PR #2, opened 2025-05-12
Add Salesforce-standard docs including license

bcantoni:


LICENSE.txt
anyoung-tableau, line 1: We should also update the `license` in package.json to `"Apache-2.0"`. Example from the Heroku MCP server: https://github.com/heroku/heroku-mcp-server/blob/main/package.json#L69

package.json
```diff
@@ -11,7 +11,7 @@
   },
   "keywords": [],
   "author": "",
-  "license": "ISC",
+  "license": "Apache-2.0",
```
anyoung-tableau, line 14: Looks like the license shows up in the package.lock file too, which will now need to be regenerated with `npm install`. I have some forthcoming changes that will regenerate the lock file so you are welcome to do it, or I can take care of it in my next PR.

bcantoni: I'll let you pick it up in the next update

PR #4, opened 2025-05-13
@W-18465174: Add unit test framework and some basic unit tests

anyoung-tableau:
These changes add [vitest](https://vitest.dev/) to the project, add a few basic unit tests for some of the utility classes, and makes a few minor refactoring changes to improve testability.

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -156,19 +156,6 @@ describe('Config', () => {
     });
   });
 
-  it('should throw error when no authentication method can be determined', () => {
-    process.env = {
-      ...process.env,
-      SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
-      SITE_NAME: 'test-site',
-    };
-
-    expect(() => new Config()).toThrow(
-      'No authentication method could be determined. Ensure the environment variables are set.',
-    );
-  });
-
   it('should set default log level to debug when not specified', () => {
     process.env = {
       ...process.env,


src/config.ts
```diff
@@ -32,23 +31,16 @@ class Config {
       DISABLE_LOG_MASKING: disableLogMasking,
     } = process.env;
 
-    this.defaultLogLevel = isLoggingLevel(defaultLogLevel) ? defaultLogLevel : 'debug';
```
anyoung-tableau, line 35: There was a circular dependency here between `config.ts` and `log.ts` that was causing some problems when running the tests. These changes resolve that.

src/config.ts
```diff
@@ -113,4 +105,15 @@ class Config {
...
 
-export const config = new Config();
+let config: Config;
+export const getConfig = (): Config => {
+  if (!config) {
+    config = new Config();
+  }
+
+  return config;
+};
```
anyoung-tableau, line 115: I converted the singleton instance of `Config` to be lazy. This made the `Config` class more readily unit-testable.

src/logging/log.ts
```diff
@@ -20,19 +20,26 @@ const loggingLevels = [
...
   if (currentLogLevel === level) {
     return;
   }
 
   currentLogLevel = level;
-  log.notice(`Logging level set to: ${level}`);
+
+  if (!silent) {
+    log.notice(`Logging level set to: ${level}`);
+  }
```
anyoung-tableau, line 42: I added the `silent` option so tests that execute `setLogLevel` don't take a run-time dependency on the MCP server instance.

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "types": ["vitest/globals"]
```
anyoung-tableau, line 13: https://vitest.dev/config/#globals

vitest.config.ts
```diff
@@ -0,0 +1,27 @@
...
+          summary: false,
+        },
+      ],
+      'junit',
+    ],
+    outputFile: 'junit/results.xml',
+    coverage: {
+      provider: 'v8',
+      include: ['src'],
+      exclude: ['src/sdks/**/*', ...coverageConfigDefaults.exclude],
```
anyoung-tableau, line 22: I'm excluding the `src/skds` folder from code coverage for now. I feel testing it is outside of what the scope of the MCP server tests should be and will require a ton of mocking anyway. Ideally the TypeScript REST API client would get moved out to its own package with its own tests, but for now it can live here.

src/config.ts
stephendeoca: Is the behavior of defaultLogLevel the same as currentLogLevel?

anyoung-tableau: `defaultLogLevel` is the value of the `DEFAULT_LOG_LEVEL` environment variable and is a constant. It determines what the initial value of `currentLogLevel` will be if the client never requests to change it.

src/config.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    process.env = {
+      ...process.env,
+      SERVER: 'test-server',
+      DATASOURCE_LUID: 'test-luid',
+      SITE_NAME: 'test-site',
+      USERNAME: 'test-user',
+      CONNECTED_APP_CLIENT_ID: 'test-client-id',
+      CONNECTED_APP_SECRET_ID: 'test-secret-id',
+      CONNECTED_APP_SECRET_VALUE: 'test-secret-value',
+      JWT_SCOPES: 'tableau:books:read',
```
stephendeoca, line 125: what happens if JWT_SCOPES is undefined or empty string? I would expect it to be fine right?

src/config.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    const config = new Config();
+    expect(config.authConfig).toEqual({
+      type: 'username-password',
+      username: 'test-user',
+      password: 'test-password',
+      siteName: 'test-site',
+    });
+  });
+
+  it('should throw error when no authentication method can be determined', () => {
```
stephendeoca: This appears to be a duplicate test case of line 66

src/config.test.ts
anyoung-tableau: Yeah: https://github.com/tableau/tableau-mcp/blob/6cba526efc0f4b3841cf5426898a80fb1fcc4f3d/src/config.ts#L91

src/config.test.ts
anyoung-tableau: Good eye. I blame Cursor. :P

PR #7, opened 2025-05-14
@W-18522213: Log failing requests/responses as well

anyoung-tableau:
Previously only successful requests and responses would get logged. These changes again use the Axios interceptors to capture failing requests and log them out too.

For example, here is an example log from a failing request to the metadata API:

```js
{
  "method": "notifications/message",
  "params": {
    "level": "info",
    "logger": "rest-api",
    "message": {
      "timestamp": "2025-05-14T23:47:46.743Z",
      "currentLogLevel": "debug",
      "message": {
        "type": "response",
        "requestId": "03c9c900-755a-4b77-81de-88074486c870",
        "url": "https:/devplat.tableautest.com/api/metadata/graphql",
        "status": 403,
        "headers": {
          "date": "Wed, 14 May 2025 23:47:46 GMT",
          "content-type": "application/json",
          "transfer-encoding": "chunked",
          "connection": "keep-alive",
          "server": "TableauMicroservicesContainer",
          "vary": "accept-encoding",
          "x-tableau": "Tableau Server",
          "p3p": "CP=\"NON\"",
          "x-ua-compatible": "IE=Edge",
          "x-content-type-options": "nosniff",
          "x-xss-protection": "1; mode=block",
          "referrer-policy": "no-referrer-when-downgrade",
          "content-security-policy-report-only": "<redacted due to length>'"
        },
        "data": {
          "timestamp": "2025-05-14T23:47:46.682+00:00",
          "status": 403,
          "error": "Forbidden",
          "path": "/relationship-service-war/graphql"
        }
      }
    }
  }
}
```

### DIFF
--- a/src/sdks/tableau/restApi.ts
+++ b/src/sdks/tableau/restApi.ts
@@ -1,5 +1,3 @@
-import { ZodiosClass } from '@zodios/core';
-
 import { AuthConfig } from './authConfig.js';
 import {
   AxiosInterceptor,


src/sdks/tableau/restApi.ts
```diff
@@ -23,14 +28,14 @@ export default class RestApi {
 
   private static _version = '3.24';
 
-  private _requestInterceptor?: RequestInterceptor;
-  private _responseInterceptor?: ResponseInterceptor;
+  private _requestInterceptor?: [RequestInterceptor, ErrorInterceptor?];
```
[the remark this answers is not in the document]
anyoung-tableau: Good idea if the tuple had more than 2 entries.

src/restApiInstance.ts
```diff
@@ -1,8 +1,13 @@
+import { isAxiosError } from '../node_modules/axios/index.js';
```
stephendeoca, line 1: Is this one of those cursor screw ups where it imports from node_modules instead of the dependency?

anyoung-tableau: No, we get Axios indirectly through our dependency on Zodios. Zodios doesn't re-export `isAxiosError` so I imported it through node_modules. Not sure if there's a better way to do this without implementing my own type guard.

PR #8, opened 2025-05-16
@W-18532773: Add instructions for VS Code debugging & log tool requests

anyoung-tableau:
These changes add instructions for how to launch the MCP Inspector and attach VS Code so the server can be debugged with breakpoints!

I also made a driveby change so tool requests and their arguments also get logged. This way requests made by other MCP clients like Cursor, will show up in the logs, and can then be "replayed" in MCP Inspector while VS Code is attached to the server process, making debugging problems way easier!

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,13 +141,14 @@ To set up local debugging with breakpoints:
    - Open the Command Palette (F1 or Cmd/Ctrl + Shift + P).
    - Type `Preferences: Open User Settings (JSON)`.
    - This should open your user's `settings.json` file.
-   - Copy the environment variables from `.vscode/settings.example.json`, append them to your user's
-     `settings.json` file, and update their values accordingly:
+   - Copy the environment variables from `.vscode/settings.example.json`, append them to the JSON
+     blob in your user's `settings.json` file, and update their values accordingly:
 
-   ```
-   "tableau.mcp.SERVER": "https://my-tableau-server.com",
-   ...
-   ```
+     ```
+     "tableau.mcp.SERVER": "https://my-tableau-server.com",
+     "tableau.mcp.DATASOURCE_LUID": "...",
+     ...
+     ```
 
 2. Set breakpoints in your TypeScript files.
 3. Locate and click the `Run and Debug` button in the Activity Bar.


src/tools/tool.ts
```diff
@@ -24,33 +26,43 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
+  }
+
+  getToolCallback = async <T>({
+    args,
+    callback,
+  }: {
+    args: unknown;
+    callback: (requestId: string) => Promise<T>;
+  }): Promise<CallToolResult> => {
+    this.logInvocation(args);
```
anyoung-tableau, line 41: This diff is bad. I moved the `getToolCallback` function up as a method on `Tool`, renamed it to `logAndExecute`, changed the signature to accept an object, and added this line.

src/tools/tool.ts
```diff
@@ -24,33 +26,43 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
 
-export async function getToolCallback<T>(
-  callback: (requestId: string) => Promise<T>,
-): Promise<CallToolResult> {
-  const result = await getResult(callback);
+  logInvocation(args: unknown): void {
+    log.debug(getToolLogMessage(this.name, args));
+  }
+
+  getToolCallback = async <T>({
```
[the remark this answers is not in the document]
anyoung-tableau: Not sure why I'm replying to a bot, but good idea on renaming the method. I'll leave the modifier off since `public` is default and should be common knowledge.

README.md
```diff
@@ -110,3 +111,58 @@ the following commands:
 | ----------------- | ---------------------------------------------------------------------------- |
 | `npm run inspect` | Start the [MCP Inspector](https://github.com/modelcontextprotocol/inspector) |
 | `npm run start`   | Start the standalone MCP server                                              |
+
+## Debugging
```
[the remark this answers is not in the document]
anyoung-tableau: We mention earlier in the readme that Node 22.15.0 is the recommended version.

package.json
```diff
@@ -15,7 +15,9 @@
     "tableau-mcp-server": "./build/index.js"
   },
   "scripts": {
-    "build": "esbuild src/index.ts --bundle --packages=external --platform=node --format=esm --outdir=build --sourcemap && npm run exec-perms",
+    ":build": "rm -rf build && esbuild src/index.ts --bundle --packages=external --platform=node --format=esm --outdir=build --sourcemap",
```
stephendeoca, line 18: is it necessary to delete the build folder each time?

anyoung-tableau: Maybe! I added this because Cursor was having trouble running the `build` command before launching the debugger, complaining about the `build/index.js` file being locked and esbuild couldn't write to it. The Heroku MCP server is doing this too: https://github.com/heroku/heroku-mcp-server/blob/main/package.json#L72

PR #11, opened 2025-05-19
@W-18465168: Add read metadata tool

stephendeoca:
Adds a tool which gets metadata from a datasource using VizQL Data Service's read metadata endpoint. This is different from the listFields tool since listFields uses the Metadata API

### DIFF
--- a/src/tools/readMetadata.ts
+++ b/src/tools/readMetadata.ts
@@ -2,23 +2,30 @@ import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 
 import { getConfig } from '../config.js';
 import { getNewRestApiInstanceAsync } from '../restApiInstance.js';
-import { getToolCallback, Tool } from './tool.js';
+import { Tool } from './tool.js';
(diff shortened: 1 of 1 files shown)


src/tools/readMetadata.ts
```diff
@@ -0,0 +1,31 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+
+import { getConfig } from '../config.js';
+import { getNewRestApiInstanceAsync } from '../restApiInstance.js';
+import { Tool } from './tool.js';
+
+export const readMetadataTool = new Tool({
```
anyoung-tableau, line 7: Could you please open a story to create unit tests for this new tool? Once https://github.com/tableau/tableau-mcp/pull/9 merges it should be straightforward.

stephendeoca: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002EhFHWYA3/view

PR #12, opened 2025-05-19
@W-18562156: Add support for enabling/disabling tools

anyoung-tableau:
By default, the Tableau MCP Server exposes all available tools to the connected agent. The set of available tools can now be customized using two new environment variables: `INCLUDE_TOOLS` and `EXCLUDE_TOOLS`. Using these variables allows you to precisely control the capabilities granted to the AI agent connecting to the server, enhancing security and alignment with intended use cases.

### DIFF
--- a/src/tools/toolName.ts
+++ b/src/tools/toolName.ts
@@ -1,4 +1,4 @@
-export const toolNames = ['query-datasource', 'list-fields'] as const;
+export const toolNames = ['query-datasource', 'list-fields', 'read-metadata'] as const;
 export type ToolName = (typeof toolNames)[number];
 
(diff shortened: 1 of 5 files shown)


src/tools/toolName.ts
```diff
@@ -0,0 +1,6 @@
+export const toolNames = ['query-datasource', 'list-fields'] as const;
```
stephendeoca: I just merged the read-metadata tool. Can you add that to this list.

anyoung-tableau: Yeah, I'll need to otherwise we'll get compile errors. :)

PR #14, opened 2025-05-20
@W-18465168: Improve queryDatasource tool

stephendeoca:
Adds a description to help the MCP Client build basic queries that contain fields and functions. There is still more work to be done such as support for filters and calculations.

### DIFF
--- a/src/tools/queryDatasource/querySchemas.ts
+++ b/src/tools/queryDatasource/querySchemas.ts
@@ -39,20 +39,3 @@ const Field = z.object({
 export const DatasourceQuery = z.object({
   fields: z.array(Field),
 });
-
-// const FilterBase = z.object({
-//   field: z.string(),
-// });
-
-// filterType: z.enum([
-//   'QUANTITATIVE_DATE',
-//   'QUANTITATIVE_NUMERICAL',
-//   'SET',
-//   'MATCH',
-//   'DATE',
-//   'TOP',
-// ])
-// const Filter = z.union([
-//   FilterBase,
-//   FilterBase.and(z.object({ filterType: 'QUANTITATIVE_DATE', value: z.string() })),
-// ]);


src/tools/queryDatasource/queryDatasource.test.ts
```diff
@@ -52,7 +52,9 @@ describe('queryDatasourceTool', () => {
 
   it('should create a tool instance with correct properties', () => {
     expect(queryDatasourceTool.name).toBe('query-datasource');
-    expect(queryDatasourceTool.description).toContain('Run a Tableau VizQL query');
+    expect(queryDatasourceTool.description).toContain(
+      'Requests relevant data from a datasource using a query.',
```
stephendeoca: Now that the description was moved to queryDescription.ts, this expectation can be changed to see if it matches the value we export. Lmk if I should change this.

anyoung-tableau: Nah this is fine. Feel free to make it even more future proof with `expect.any(String)`.

src/tools/queryDatasource/querySchemas.ts
```diff
@@ -0,0 +1,58 @@
...
+//   'QUANTITATIVE_NUMERICAL',
+//   'SET',
+//   'MATCH',
+//   'DATE',
+//   'TOP',
+// ])
+// const Filter = z.union([
+//   FilterBase,
+//   FilterBase.and(z.object({ filterType: 'QUANTITATIVE_DATE', value: z.string() })),
+// ]);
```
anyoung-tableau: Please remove

src/tools/queryDatasource/querySchemas.ts
```diff
@@ -0,0 +1,58 @@
+import { z } from 'zod';
+
+export const FunctionEnum = [
```
anyoung-tableau: This doesn't need to be exported. I'd also suggest renaming it to something like `functions` since it's just an array of strings, not an enum.

src/tools/queryDatasource/querySchemas.ts
```diff
@@ -0,0 +1,58 @@
...
+  'WEEK',
+  'DAY',
+  'TRUNC_YEAR',
+  'TRUNC_QUARTER',
+  'TRUNC_MONTH',
+  'TRUNC_WEEK',
+  'TRUNC_DAY',
+] as const;
+
+const Function = z.enum(FunctionEnum);
```
anyoung-tableau: There's quite a bit of duplication between these schemas and those from the REST API. It's fine, but also means if the schemas change we will likely have to remember to make changes in multiple spots. Do we need these?

stephendeoca: The duplication was to reduce the values used in the query we build. For example the OpenApi spec includes 'AGG', 'NONE', and 'UNSPECIFIED' in the Function values which are still present in `vizqlDataServiceApi.ts` but these are not used in the query so I removed them in `querySchemas.ts` so that the client does not accidentally use those.

anyoung-tableau: Gotcha, yeah you can use the Zod utility methods—similar to the TypeScript utility types—to transform schemas into others. For this one, you'd use `exclude` to remove AGG, NONE, and UNSPECIFIED from the enum.

https://zod.dev/api?id=exclude

You are welcome to "wont-fix" this comment but have a look if you feel it would be straightforward to take this route.

PR #16, opened 2025-05-21
Adding list datasource tool

renoyjohnm:
Adding list datasource tool and removing the dependency on DATASOURCE_LUID env variable

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -47,7 +47,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
     };
 
@@ -60,7 +59,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       PAT_NAME: 'test-pat-name',
       PAT_VALUE: 'test-pat-value',
@@ -79,7 +77,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       PAT_NAME: 'test-pat-name',
       PAT_VALUE: 'test-pat-value',
@@ -106,7 +103,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       JWT: 'test-jwt',
     };
@@ -123,7 +119,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       PAT_NAME: 'test-pat-name',
       PAT_VALUE: 'test-pat-value',
@@ -149,7 +144,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       USERNAME: 'test-user',
       CONNECTED_APP_CLIENT_ID: 'test-client-id',
@@ -174,7 +168,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       PAT_NAME: 'test-pat-name',
       PAT_VALUE: 'test-pat-value',
@@ -204,7 +197,6 @@ describe('Config', () => {
     process.env = {
       ...process.env,
       SERVER: 'test-server',
-      DATASOURCE_LUID: 'test-luid',
       SITE_NAME: 'test-site',
       USERNAME: 'test-user',
       PASSWORD: 'test-password',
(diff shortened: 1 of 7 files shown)


src/tools/toolName.ts
```diff
@@ -1,4 +1,4 @@
-export const toolNames = ['query-datasource', 'list-fields', 'read-metadata'] as const;
+export const toolNames = ['list-datasources', 'query-datasource', 'list-fields', 'read-metadata'] as const;
```
anyoung-tableau: Please keep these alphabetized.

src/tools/listDatasources.ts
```diff
@@ -0,0 +1,103 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+
+import { getConfig } from '../config.js';
+import { getNewRestApiInstanceAsync } from '../restApiInstance.js';
+import { Tool } from './tool.js';
+import { z } from 'zod';
+
+export const listDatasourcesTool = new Tool({
+    name: 'list-datasources',
+    description: `
```
anyoung-tableau: Through trial and error, and from speaking with the folks working on the Heroku MCP server, a large tool description can actually do more harm than good. They get loaded into the LLM context window, taking up many or even all of the tokens that the model can use when generating its response, increasing the chance it's going to hallucinate. We need to capture a balance where the descriptions are concise but also provide enough context to the client to provide a syntactically accurate query.

src/server.ts
```diff
@@ -35,7 +36,7 @@ class Server extends McpServer {
 export const server = new Server();
 
 const { includeTools, excludeTools } = getConfig();
-const tools = [queryDatasourceTool, listFieldsTool, readMetadataTool].filter((tool) => {
+const tools = [listDatasourcesTool, queryDatasourceTool, listFieldsTool, readMetadataTool].filter((tool) => {
```
anyoung-tableau: This changed a little earlier today, you'll need to pull and resolve conflicts.

src/tools/listDatasources.ts
anyoung-tableau: For the purposes of the tool, I'd bet searching purely by data source name would be sufficient. We don't need to expose all the complex filtering logic to the host.

src/config.ts
anyoung-tableau, line 1: You can fix all the line errors with `npx eslint --fix`. It's also a good idea to install the Prettier and ESLint extensions in VS Code, configure either as the default formatter, and enable Format on Save in the IDE settings

src/tools/listDatasources.ts
renoyjohnm: I feel like we shouldn't restrict the filter logic provided by LLM instead we could add some logic to validate the filter query created by LLM prior to the API invocation, thoughts? I'm thinking of a scenarios like where the user query is about listing all the datasources that were created/refreshed after a given date

src/sdks/tableau/methods/datasourcesMethods.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Credentials } from '../types/credentials.js';
+import AuthenticatedMethods from './authenticatedMethods.js';
+
+export default class DatasourcesMethods extends AuthenticatedMethods<typeof datasourcesApis> {
+  constructor(baseUrl: string, creds: Credentials) {
+    super(new Zodios(baseUrl, datasourcesApis), creds);
+  }
+
+  /**
+   * Returns a list of published data sources on the specified site, filtered by project name.
```
anyoung-tableau:
```suggestion
   * Returns a list of published data sources on the specified site.
```

src/sdks/tableau/restApi.ts
```diff
@@ -25,7 +26,7 @@ export default class RestApi {
 
   private _metadataMethods?: MetadataMethods;
   private _vizqlDataServiceMethods?: VizqlDataServiceMethods;
-
+  private _datasourcesMethods?: DatasourcesMethods;
```
anyoung-tableau: Hard to tell from the previous set of 2, but we should keep these alphabetized.

src/sdks/tableau/restApi.ts
```diff
@@ -72,6 +77,15 @@ export default class RestApi {
     return this._vizqlDataServiceMethods;
   }
 
+  get datasourcesMethods(): DatasourcesMethods {
```
anyoung-tableau: Same, please keep the `*Methods` property getters alphabetized.

src/tools/listDatasources.ts
anyoung-tableau: Right, but depends on how good the LLM is at crafting the query. My concern is encountering issues because we're trying to support *all* the different filter methods when in reality, most users will only need "filter by name". 

I wonder if we were to strengthen the schema of the `filter` property to validate with a regex that validates the filtering format, would the host ensure the LLM generates valid filters. Example: https://github.com/colinhacks/zod/discussions/979#discussioncomment-2275023

I am also looking into allowing server admins to override tool descriptions, similar to how the GitHub MCP server works: https://github.com/github/github-mcp-server?tab=readme-ov-file#i18n--overriding-descriptions.

If a server admin needs the ability to provide an advanced filter, they can override the tool description to help the LLM.

src/config.test.ts
```diff
@@ -12,7 +12,6 @@ describe('Config', () => {
     process.env = {
       ...originalEnv,
       SERVER: undefined,
-      DATASOURCE_LUID: undefined,
```
anyoung-tableau, line 15: There are more files in the project that reference `DATASOURCE_LUID`. Please make sure to clean them up as well.

renoyjohnm: my bad, I missed those, hopefully all cleared in the latest commit

src/tools/listDatasources.ts
renoyjohnm: The way I understand is typically if our filter validator is able to detect and articulate well the error in the LLM provided filter query, I'd expect the LLM to autocorrect and retry with the correction or at the very least provide the error to the user to revisit the ask/question from user to be refined further

>I am also looking into allowing server admins to override tool descriptions, similar to how the GitHub MCP server works: https://github.com/github/github-mcp-server?tab=readme-ov-file#i18n--overriding-descriptions.

Is it limited to when deploying MCP server as a standalone central server?

anyoung-tableau: I don't think it would necessarily need to be limited to standalone servers.

PR #21, opened 2025-05-23
 W-18465120: Adding validators for datasource filter query

renoyjohnm:
Adding validators for datasources filter query along with more detailed tool description. Also, logging api request query params

### DIFF
--- a/src/tools/listDatasources/listDatasources.ts
+++ b/src/tools/listDatasources/listDatasources.ts
@@ -56,7 +56,7 @@ Retrieves a list of published data sources from a specified Tableau site using t
 - \`lte\`: less than or equal
 
 **Filter Expression Notes**
--  Filter expressions can’t contain ampersand (&) or comma (,) characters even if those characters are encoded.
+- Filter expressions can’t contain ampersand (&) or comma (,) characters even if those characters are encoded.
 - Operators are delimited with colons (:). For example: \`filter=name:eq:Project Views\`
 - Field names, operator names, and values are case-sensitive.
 - To filter on multiple fields, combine expressions using a comma:  \`filter=lastLogin:gte:2016-01-01T00:00:00Z,siteRole:eq:Publisher\`


src/tools/listDatasources/datasourcesFilterUtils.test.ts
```diff
@@ -0,0 +1,93 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  FilterFieldSchema,
+  FilterOperatorSchema,
+  isOperatorAllowed,
+  parseAndValidateFilterString,
+} from './datasourcesFilterUtils.js';
+
+// --- parseAndValidateFilterString ---
```
anyoung-tableau: nit: remove comment; adds no value

src/tools/listDatasources/datasourcesFilterUtils.test.ts
```diff
@@ -0,0 +1,93 @@
+import { describe, expect, it } from 'vitest';
```
anyoung-tableau: nit: you can you remove the imports because we've enabled globals. https://vitest.dev/config/#globals

src/tools/listDatasources/listDatasources.ts
anyoung-tableau, line 1: Don't forget to delete the old tool and its tests

src/tools/listDatasources/listDatasources.ts
```diff
@@ -0,0 +1,108 @@
...
+  annotations: {
+    title: 'List Datasources',
+    readOnlyHint: true,
+    openWorldHint: false,
+  },
+  callback: async ({ filter }): Promise<CallToolResult> => {
+    const config = getConfig();
+    const validatedFilter = filter ? parseAndValidateFilterString(filter) : undefined;
+    return await listDatasourcesTool.logAndExecute({
+      args: { filter: validatedFilter },
```
anyoung-tableau: This should be `args: { filter }`. We want to log the arguments provided by the MCP client.

src/tools/listDatasources/datasourcesFilterUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/**
+ * Parses and validates a Tableau-style filter string
+ * @param filterString e.g. 'name:eq:Project Views,type:eq:Workbook'
+ * @returns validated filter string
+ * @throws ZodError or custom error for invalid operators
+ */
+export function parseAndValidateFilterString(filterString: string): string {
+  const expressions = filterString.split(',').filter(Boolean);
```
anyoung-tableau: I'm speculating, but if the client happened to add a space after a comma, we're not doing it any favors. Maybe change to `filterString.split(',').map((f) => f.trim()).filter(Boolean)`

src/tools/listDatasources/datasourcesFilterUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+      throw new Error(`Invalid filter expression format: "${expr}"`);
+    }
+
+    const value = valueParts.join(':');
+
+    const field = FilterFieldSchema.parse(fieldRaw);
+    const operator = FilterOperatorSchema.parse(operatorRaw);
+
+    if (!isOperatorAllowed(field, operator)) {
+      throw new Error(`Operator '${operator}' is not allowed for field '${field}'`);
```
anyoung-tableau: I think the error message could include the allowed operators to encourage the LLM to self-correct.

src/tools/listDatasources/datasourcesFilterUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  for (const expr of expressions) {
+    const [fieldRaw, operatorRaw, ...valueParts] = expr.split(':');
+    if (!fieldRaw || !operatorRaw || valueParts.length === 0) {
+      throw new Error(`Invalid filter expression format: "${expr}"`);
+    }
+
+    const value = valueParts.join(':');
+
+    const field = FilterFieldSchema.parse(fieldRaw);
```
anyoung-tableau, line 123: How friendly are the ZodErrors that get thrown here if parsing fails? Should we use `safeParse` instead and throw our own friendlier error?

src/tools/listDatasources/datasourcesFilterUtils.ts
```diff
@@ -0,0 +1,140 @@
...
+  'updatedAt',
+]);
+
+export type FilterField = z.infer<typeof FilterFieldSchema>;
+
+export const FilterOperatorSchema = z.enum(['eq', 'in', 'gt', 'gte', 'lt', 'lte']);
+
+export type FilterOperator = z.infer<typeof FilterOperatorSchema>;
+
+export const allowedOperatorsByField: Record<FilterField, FilterOperator[]> = {
```
anyoung-tableau: One thing I've been trying to do is only directly exporting things imported by the server. If it's only used by tests, export a `exportedForTesting` object that includes each thing imported only by the tests.

src/restApiInstance.ts
```diff
@@ -105,6 +105,7 @@ function logRequest(request: RequestInterceptorConfig, requestId: string): void
     ...(shouldLogWhenLevelIsAtLeast('debug') && {
       headers: maskedRequest.headers,
       data: maskedRequest.data,
+      params: maskedRequest.params,
```
anyoung-tableau, line 108: What do you think? Should we log the `params` always, or only when log level is `debug`?

src/tools/listDatasources/datasourcesFilterUtils.ts
```diff
@@ -0,0 +1,140 @@
+import { z } from 'zod';
+
+// === Field and Operator Definitions ===
+
+export const FilterFieldSchema = z.enum([
```
anyoung-tableau: Can we provide a comment to a code location or documentation that describes all this logic?

src/tools/listDatasources/listDatasources.ts
```diff
@@ -0,0 +1,108 @@
...
+- \`gte\`: greater than or equal
+- \`has\`: includes the specified string (substring search, only for Query Jobs)
+- \`in\`: any of [list] (for searching tags)
+- \`lt\`: less than
+- \`lte\`: less than or equal
+
+**Filter Expression Notes**
+- Operators are delimited with colons (:). For example: \`filter=name:eq:Project Views\`
+- Field names, operator names, and values are case-sensitive.
+- Values should be URL-encoded. For example, to search for the workbook named "Project Views", use: \`filter=name:eq:Project+Views\`
```
anyoung-tableau: This doesn't seem right. Think we'd end up with double-encoded strings in the request if the client already encoded the value.

src/tools/listDatasources/listDatasources.ts
```diff
@@ -0,0 +1,108 @@
...
+- \`cieq\`: case-insensitive equals
+- \`eq\`: equals
+- \`gt\`: greater than
+- \`gte\`: greater than or equal
+- \`has\`: includes the specified string (substring search, only for Query Jobs)
+- \`in\`: any of [list] (for searching tags)
+- \`lt\`: less than
+- \`lte\`: less than or equal
+
+**Filter Expression Notes**
```
anyoung-tableau, line 58: The docs also mention:

> Note: Filter expressions can’t contain ampersand (&) or comma (,) characters because these characters are used as field expression separators. This is true even if those characters are encoded.

https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm#datasources:~:text=Filter%20expressions%20can%E2%80%99t%20contain%20ampersand

src/tools/listDatasources/listDatasources.ts
renoyjohnm: my bad, thought it was deleted as part of the merge from the older branch, good catch

src/tools/listDatasources/datasourcesFilterUtils.ts
renoyjohnm: sounds good, always safe to catch such sceanarios

src/tools/listDatasources/datasourcesFilterUtils.ts
renoyjohnm: I think we should stick with the original ZodErrors since the LLMs might have context on it from all the publicly available web sources over the error that we provide, which would better help with correcting the errors, thoughts?

src/restApiInstance.ts
renoyjohnm: I prefer to stick with debug level just to avoid too much logging

src/tools/listDatasources/listDatasources.ts
renoyjohnm: yes, I'll remove it from the description

src/tools/listDatasources/datasourcesFilterUtils.ts
anyoung-tableau: Sounds reasonable.

src/tools/listDatasources/listDatasources.ts
```diff
@@ -0,0 +1,108 @@
...
+-  Filter expressions can’t contain ampersand (&) or comma (,) characters even if those characters are encoded.
```
anyoung-tableau:
```suggestion
- Filter expressions can’t contain ampersand (&) or comma (,) characters even if those characters are encoded.
```

PR #24, opened 2025-05-27
@W-18577048: Return VizQL Data Service errors to the client

anyoung-tableau:
Previously, if a request to VizQL Data Service returned an error the client would receive something unhelpful like `Request failed with status code 400`.

With these changes, the full error object (augmented with some additional context) will be returned to the client which are [well-documented](https://help.tableau.com/current/api/vizql-data-service/en-us/docs/vds_error_codes.html).

Example:
```json
{
  "errorCode": "400803",
  "message": "Unknown Field: Foobar.",
  "datetime": "2025-05-27T22:30:48.7122762Z",
  "condition": "Validation failed",
  "details": "The incoming request isn't valid per the validation rules."
}
```

### DIFF
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -17,6 +17,12 @@ export type ToolParams<Args extends ZodRawShape | undefined = undefined> = {
   callback: ToolCallback<Args>;
 };
 
+type LogAndExecuteParams<T, E> = {
+  args: unknown;
+  callback: (requestId: string) => Promise<Result<T, E>>;
+  getErrorText?: (error: E) => string;
+};
+
 export class Tool<Args extends ZodRawShape | undefined = undefined> {
   name: ToolName;
   description: string;
@@ -37,28 +43,19 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
   }
 
   // Overload for E = undefined (getErrorText omitted)
-  async logAndExecute<T>(params: {
-    args: unknown;
-    callback: (requestId: string) => Promise<Result<T, undefined>>;
-  }): Promise<CallToolResult>;
+  async logAndExecute<T>(
+    params: Omit<LogAndExecuteParams<T, undefined>, 'getErrorText'>,
+  ): Promise<CallToolResult>;
 
   // Overload for E != undefined (getErrorText required)
-  async logAndExecute<T, E>(params: {
-    args: unknown;
-    callback: (requestId: string) => Promise<Result<T, E>>;
-    getErrorText: (error: E) => string;
-  }): Promise<CallToolResult>;
+  async logAndExecute<T, E>(params: Required<LogAndExecuteParams<T, E>>): Promise<CallToolResult>;
 
   // Implementation
-  async logAndExecute<T, E = undefined>({
+  async logAndExecute<T, E>({
     args,
     callback,
     getErrorText,
-  }: {
-    args: unknown;
-    callback: (requestId: string) => Promise<Result<T, E>>;
-    getErrorText?: (error: E) => string;
-  }): Promise<CallToolResult> {
+  }: LogAndExecuteParams<T, E>): Promise<CallToolResult> {
     const requestId = randomUUID();
 
     this.logInvocation(args);


src/tools/tool.ts
```diff
@@ -36,53 +42,66 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
     callback,
-  }: {
-    args: unknown;
-    callback: (requestId: string) => Promise<T>;
-  }): Promise<CallToolResult> => {
-    this.logInvocation(args);
-
-    const result = await getResult(callback);
+    getErrorText,
+  }: LogAndExecuteParams<T, E>): Promise<CallToolResult> {
```
anyoung-tableau, line 58: The `callback` passed to `logAndExecute` must now return a [Result](https://github.com/vultix/ts-results?tab=readme-ov-file#result-example). If a callback only returns an `Ok` variant then they will get the same (bad) error handling we had previously. However, if a callback can also return an `Err` variant, then a `getErrorText` callback must also be supplied, which provides for flexible error handling. See the changes to `queryDatasource.ts` in this PR for an example.

src/sdks/tableau/methods/vizqlDataServiceMethods.ts
```diff
@@ -20,8 +22,16 @@ export default class VizqlDataServiceMethods extends AuthenticatedMethods<
 
   queryDatasource = async (
     queryRequest: z.infer<typeof QueryRequest>,
-  ): Promise<z.infer<typeof QueryOutput>> => {
-    return await this._apiClient.queryDatasource(queryRequest, { ...this.authHeader });
+  ): Promise<Result<z.infer<typeof QueryOutput>, z.infer<typeof TableauError>>> => {
+    try {
+      return Ok(await this._apiClient.queryDatasource(queryRequest, { ...this.authHeader }));
```
renoyjohnm, line 27: just for my understanding, are error responses going to be thrown always? Is it doing so based on the below definition:

`errors: [
      {
        status: 'default',
        schema: TableauError,
      },
    ],`

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -42,8 +44,12 @@ export const queryDatasourceTool = new Tool({
           config.authConfig,
           requestId,
         );
+
         return await restApi.vizqlDataServiceMethods.queryDatasource(queryRequest);
       },
+      getErrorText: (error: z.infer<typeof TableauError>) => {
+        return JSON.stringify(handleQueryDatasourceError(error));
```
renoyjohnm, line 51: What level of detail does the error response from VDS provide? For ex: with errors with error code `404934`, will the error response from the server include details on which requested field was invalid

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -174,6 +174,12 @@ const vizqlDataServiceApi = makeApi([
       },
     ],
     response: QueryOutput,
+    errors: [
+      {
+        status: 'default',
+        schema: TableauError,
+      },
+    ],
```
anyoung-tableau, line 182: https://www.zodios.org/docs/client/error

src/tools/queryDatasource/queryDatasourceErrorHandler.ts
```diff
@@ -0,0 +1,121 @@
...
+type TableauErrorDetail = TableauError & Partial<{ condition: string; details: string }>;
+
+/**
+ * Adds additional context to the VizQl Data Service errors.
+ *
+ * @param {TableauError} error - The error object
+ * @returns {*}  {TableauErrorDetail} - The error object with additional context
+ * @see https://help.tableau.com/current/api/vizql-data-service/en-us/docs/vds_error_codes.html
+ */
+export function handleQueryDatasourceError(error: TableauError): TableauErrorDetail {
```
anyoung-tableau, line 17: I debated adding this concept at all. Might be okay without it but also figured the additional context on the error object from the public docs couldn't hurt.

src/tools/tool.ts
anyoung-tableau: This uses function overloads to achieve this: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

src/sdks/tableau/methods/vizqlDataServiceMethods.ts
anyoung-tableau: Not always. Adding the `errors` property to the API definition structure tells Zodios that if a request to the API *does* fail, then then response body will have the shape of a `TableauError`. This can be dependent on the HTTP status code, but because we used `'default'` it will apply to all error status codes.

The `isErrorFromAlias` ensures the response matches the `TableauError` schema and acts as a type guard so we get nice autocomplete. If the response does not match the schema, then we'll throw.

References:
* https://www.zodios.org/docs/api/api-definition#errors
* https://www.zodios.org/docs/api/helpers#makeerrors
* https://www.zodios.org/docs/client/error#iserrorfromalias

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: For this example, yeah. For the other errors, it's really dependent on the error code.

* https://gitlab.tableausoftware.com/headless-bi/headless-bi-service/-/blob/main/src/main/java/com/tableausoftware/microservice/headlessbi/exceptions/errormapping/ErrorMessageProcessor.java#L179
* https://gitlab.tableausoftware.com/headless-bi/headless-bi-service/-/blob/main/src/main/java/com/tableausoftware/microservice/headlessbi/facade/HeadlessBiFacade.java#L161
* https://gitlab.tableausoftware.com/search?group_id=54203&nav_source=navbar&project_id=18908&repository_ref=main&search=TableauErrorFactory.create&search_code=true

renoyjohnm: Will such error details from HBI reach the LLM/Agent in this case? I feel like there is value in such errors being surfaced to the LLM/Agents for better error resolution over a generic error message

anyoung-tableau: Yeah, that's the `message` property on the `TableauError` object. `handleQueryDatasourceError` just *adds* the information from the table [here](https://help.tableau.com/current/api/vizql-data-service/en-us/docs/vds_error_codes.html) to the object on the `condition` and `details` properties

PR #28, opened 2025-05-29
@W-18576883: Add filters to HBI query

stephendeoca:
This MR adds filters to the Query schema. The filter schema consists of the 6 filter types (SET, TOP, MATCH, DATA, QUANTITATIVE_NUMBER, QUANTITATIVE_DATE). The schemas were designed to be as strict as possible to improve client accuracy. The details on constructing filters can be found here:
- [Create a Query VizQL Data Service Docs](https://help.tableau.com/current/api/vizql-data-service/en-us/docs/vds_create_queries.html)
- [VizQL Open API Spec](https://github.com/tableau/VizQL-Data-Service/blob/main/VizQLDataServiceOpenAPISchema.json)
- [QueryDatasource VizQL Data Service API reference](https://help.tableau.com/current/api/vizql-data-service/en-us/reference/index.html#operation/QueryDatasource)

### DIFF
--- a/src/sdks/tableau/apis/vizqlDataServiceApi.test.ts
+++ b/src/sdks/tableau/apis/vizqlDataServiceApi.test.ts
@@ -1,5 +1,3 @@
-import { describe, expect, it } from 'vitest';
-
 import { Field, Filter } from './vizqlDataServiceApi.js';
 
 describe('Field schema', () => {


src/tools/queryDatasource/queryDescription.ts
```diff
@@ -1 +1 @@
-export const queryDatasourceToolDescription = `Run a Tableau VizQL query.`;
+export const queryDatasourceToolDescription = `Run a Tableau VizQL query. You must follow the Query schema from vizqlDataServiceApi.ts!`;
```
anyoung-tableau: This should be reverted. The agent won't have any idea what vizqlDataServiceApi.ts is unless you have the project open in Cursor.

src/sdks/tableau/apis/vizqlDataServiceApi.test.ts
```diff
@@ -0,0 +1,725 @@
+import { describe, expect, it } from 'vitest';
```
anyoung-tableau: nit: these imports can be removed because we use vitest globals. https://vitest.dev/config/#globals

src/sdks/tableau/apis/vizqlDataServiceApi.ts
anyoung-tableau, line 1: Not something we have to solve *right now*, but we are going to need to think about versioning. That is, if VDS changes the schema in, say, 2025.4, but the MCP server user is running against Tableau Server 2025.1 they could encounter some problems. I'm keeping an eye on [openapi-mcp-generator](https://github.com/harsha-iiiv/openapi-mcp-generator) in case we can leverage some of its concepts to automatically generate the schemas based off the OpenAPI spec. How nifty would it be to not have to manually add schemas ourselves for every new version of the spec and to just have a pre-processing codegen step automatically generate the schemas for us.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -103,34 +100,139 @@ const FieldBase = z.object({
...
+});
+
+export const Filter = z.union([
+  SetFilter.strict(),
+  TopNFilter.strict(),
+  ...MatchFilter.options,
+  ...QuantitativeNumericalFilter.options,
+  ...QuantitativeDateFilter.options,
+  ...RelativeDateFilter.options,
+]);
```
anyoung-tableau, line 233: This feels like we should be using a discriminated union here with `filterType` as the discriminator. I remember you were exploring this, did it not work out in practice?

```typescript
export const Filter = z.discriminatedUnion('filterType', [
  SetFilter.strict(),
  TopNFilter.strict(),
  ...MatchFilter.options,
  ...QuantitativeNumericalFilter.options,
  ...QuantitativeDateFilter.options,
  ...RelativeDateFilter.options,
]);
```

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -103,34 +100,139 @@ const FieldBase = z.object({
...
+});
+
+const SetFilter = SimpleFilterBase.extend({
+  filterType: z.literal('SET'),
+  values: z.union([z.array(z.string()), z.array(z.number()), z.array(z.boolean())]),
+  exclude: z.boolean().optional().default(false),
+});
+
+const RelativeDateFilterBase = SimpleFilterBase.extend({
+  filterType: z.literal('RELATIVE_DATE'),
```
anyoung-tableau: This should be `'DATE'`

src/sdks/tableau/apis/vizqlDataServiceApi.test.ts
```diff
@@ -0,0 +1,725 @@
...
+    };
+    expect(() => Filter.parse(data)).not.toThrow();
+  });
+
+  it('rejects a MATCH filter with none of startsWith, endsWith, or contains', () => {
+    const data = {
+      filterType: 'MATCH',
+      field: { fieldCaption: 'Product Name' },
+    };
+    expect(() => Filter.parse(data)).toThrow();
```
anyoung-tableau, line 209: Here's a good example of a Zod error that might be overwhelming to the LLM and where a user-friendly error might be better.

This is the error Zod throws: https://gist.github.com/anyoung-tableau/541096fd0454236f747688ad702f9690

src/sdks/tableau/apis/vizqlDataServiceApi.ts
stephendeoca: I'm all for versioning, although I will say that we can't expect to generate reliable schemas off the OpenAPI spec. Even if it properly generates the OpenAPI schema, the generated schema will not always doesn't match what is documented. For example the QUANTITATIVE_NUMBER filter has the min and max properties available to all it's filter types (RANGE, MIN, MAX, ONLY_NULL, ONLY_NON_NULL). But if you read the documentation you can only use the min property on a MIN filter type, max on a MAX filter type, both min and max on RANGE, and you cannot use either for ONLY_NULL / ONLY_NON_NULL filter type. So we'd have to modify the generated schema to limit the use of each property.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
stephendeoca: Good catch! The OpenAPI schema calls it a relative date filter but the value of filterType is "DATE"

src/sdks/tableau/apis/vizqlDataServiceApi.ts
stephendeoca: DiscriminatedUnion was not possible here since all the sub-filters whose schemas are unions (i.e. MatchFilter, QuantitativeNumericalFilter, QuantitativeDateFilter, and RelativeDateFilter) are essentially multiple schemas under the same filterType. When we use the spread operator the compiler complains saying we have multiple schemas with the "MATCH" filter type when there can only be one.

src/sdks/tableau/apis/vizqlDataServiceApi.test.ts
stephendeoca: Awesome! With query validation this means we will have better feedback for the LLM.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
anyoung-tableau: Thanks!

src/sdks/tableau/apis/vizqlDataServiceApi.ts
anyoung-tableau: Looks like they have dependent parameters in their sights for the next version of OpenAPI "Moonwalk"

https://github.com/OAI/sig-moonwalk/discussions/100 
https://github.com/OAI/sig-moonwalk/blob/main/doc/initial-proposals/examples/dependentParameters.yaml

PR #30, opened 2025-05-30
@W-18565881: Add readMetadata unit tests

stephendeoca:
Adds tests for the readMetadata tool and removes a thrown error in the listFields tool when publishedDatasources are empty

### DIFF
--- a/src/tools/listFields.test.ts
+++ b/src/tools/listFields.test.ts
@@ -22,7 +22,7 @@ const mockMetadataResponses = vi.hoisted(() => ({
       ],
     },
   },
-  error: {
+  empty: {
     data: {
       publishedDatasources: [],
     },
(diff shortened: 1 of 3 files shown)


src/tools/listFields.test.ts
```diff
@@ -66,15 +66,6 @@ describe('listFieldsTool', () => {
     expect(mocks.mockGraphql).toHaveBeenCalledWith(getGraphqlQuery('test-luid'));
   });
 
-  it('should return error when no published datasources are found', async () => {
```
stephendeoca, line 69: No longer throws error, so I removed the test. Should I instead update the test to expect a success?

src/tools/listFields.ts
```diff
@@ -40,14 +40,7 @@ export const listFieldsTool = new Tool({
...
-        const published = response.data.publishedDatasources;
-
-        if (published.length) {
-          return new Ok(response);
-        }
-
-        return new Err('No published datasources in response');
+        return new Ok(await restApi.metadataMethods.graphql(query));
       },
       getErrorText: (error: string) => {
```
stephendeoca: readMetadata does not have this getErrorText method should I add it?

I removed the getErrorText parameter from listFields to see what happens and it made no difference with the test result. I'm guessing there will be a difference when there's an actual error right?

src/tools/listFields.test.ts
anyoung-tableau: Yeah, that sound reasonable. I'd rename the mock response named `error` to `empty` instead.

src/tools/listFields.ts
anyoung-tableau: readMetadata doesn't have one because the `callback` to `logAndExecute` doesn't return the `Err` variant of the `Result`. You should remove it because it would only ever get called if that callback returned an `Err` variant, which now it no longer can.

src/tools/readMetadata.test.ts
```diff
@@ -0,0 +1,100 @@
...
+});
+
+describe('readMetadataTool', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should create a tool instance with correct properties', () => {
+    expect(readMetadataTool.name).toBe('read-metadata');
+    expect(readMetadataTool.description).toEqual(expect.any(String))
```
anyoung-tableau:
```suggestion
    expect(readMetadataTool.description).toEqual(expect.any(String));
```

anyoung-tableau: lint step should catch this (if our CI wasn't broken rn)

PR #32, opened 2025-06-02
@W-18577048: Add query-datasource request validation

anyoung-tableau:
These changes add a validation layer for the requests made to the query-datasource tool. The error messages are written to be as descriptive as possible to help the LLM self-correct.

### DIFF
--- a/src/tools/queryDatasource/validators/validateFields.ts
+++ b/src/tools/queryDatasource/validators/validateFields.ts
@@ -102,12 +102,21 @@ export function validateFields(fields: Fields): void {
   }
 }
 
-export function hasEmptyFieldCaption(field: Field | FilterField): boolean {
-  return 'fieldCaption' in field && !field.fieldCaption;
+function hasEmptyFieldCaption(field: Field): boolean {
+  return !field.fieldCaption;
 }
 
 export function hasFunctionAndCalculation(field: Field | FilterField): boolean {
   return !!('function' in field && field.function && 'calculation' in field && field.calculation);
+}
+
+export function hasFieldCaptionAndCalculation(field: Field | FilterField): boolean {
+  return !!(
+    'fieldCaption' in field &&
+    field.fieldCaption &&
+    'calculation' in field &&
+    field.calculation
+  );
 }
 
 function hasNegativeMaxDecimalPlace(field: Field): boolean {
--- a/src/tools/queryDatasource/validators/validateFilters.ts
+++ b/src/tools/queryDatasource/validators/validateFilters.ts
@@ -1,11 +1,17 @@
 import { Err, Ok, Result } from 'ts-results-es';
 
 import { FilterField, Query } from '../queryDatasourceValidator.js';
-import { hasEmptyFieldCaption, hasFunctionAndCalculation } from './validateFields.js';
+import { hasFieldCaptionAndCalculation, hasFunctionAndCalculation } from './validateFields.js';
 
 export function validateFilters(filters: Query['filters']): void {
   if (!filters) {
     return;
+  }
+
+  if (filters.some((filter) => !('field' in filter))) {
+    throw new Error(
+      `The query must not include filters with invalid fields. The following field errors occurred: The filter must include a field property.`,
+    );
   }
 
   {
@@ -191,6 +197,19 @@ function validateFilterField(field: FilterField): Result<void, string> {
   }
 
   {
+    // Field cannot have a fieldCaption and a Calculation.
+    if (hasFieldCaptionAndCalculation(field)) {
+      const fieldCaption = 'fieldCaption' in field ? `"${field.fieldCaption}" ` : '';
+      return new Err(
+        `The field ${fieldCaption} must not contain both a fieldCaption and a calculation.`.replace(
+          '  ',
+          ' ',
+        ),
+      );
+    }
+  }
+
+  {
     // A Field cannot contain both a Function and a Calculation.
     if (hasFunctionAndCalculation(field)) {
       const fieldCaption = 'fieldCaption' in field ? `"${field.fieldCaption}" ` : '';
@@ -204,4 +223,8 @@ function validateFilterField(field: FilterField): Result<void, string> {
   }
 
   return Ok.EMPTY;
+}
+
+function hasEmptyFieldCaption(field: FilterField): boolean {
+  return 'fieldCaption' in field && !field.fieldCaption;
 }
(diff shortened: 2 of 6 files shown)


src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -54,8 +56,8 @@ export const queryDatasourceTool = new Tool({
 
         return await restApi.vizqlDataServiceMethods.queryDatasource(queryRequest);
       },
-      getErrorText: (error: z.infer<typeof TableauError>) => {
-        return JSON.stringify(handleQueryDatasourceError(error));
+      getErrorText: (requestId: string, error: z.infer<typeof TableauError>) => {
```
stephendeoca, line 59: any reason for including requestId? From the inspector it's usually the last request that fails so it's not hard to find.

anyoung-tableau: For local MCP servers, the concept of requestId doesn't really have much value in general for the reason you mentioned. I think when I initially added it, I was being forward-thinking since once the server is centrally deployed, there could be simultaneous requests and correlating an error response wouldn't be so straightforward.

anyoung-tableau: Oh also, there was also this idea: https://salesforce-internal.slack.com/archives/C0643HWPSVD/p1747250262865819?thread_ts=1747242873.221479&cid=C0643HWPSVD

Putting the requestId on the VDS request's user agent header so we could correlate VDS logs with MCP server logs.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -216,7 +219,7 @@ const QuantitativeDateFilter = z.discriminatedUnion('quantitativeFilterType', [
   }).strict(),
 ]);
 
-const TopNFilter = FilterBase.extend({
+export const TopNFilter = FilterBase.extend({
   filterType: z.literal('TOP'),
   howMany: z.number(),
```
stephendeoca: I just realized we can reduce this even further to `z.number().int()` can you throw that in there for me? Can you also do the same for the rangeN property in RelativeDateFilter for NEXTN and LASTN (lines 146 and 147).

src/tools/queryDatasource/validators/validateFields.ts
```diff
@@ -0,0 +1,115 @@
+import { FilterField, Query } from '../queryDatasourceValidator.js';
+
+type Fields = Query['fields'];
+type Field = Fields[number];
+
+export function validateFields(fields: Fields): void {
+  {
```
stephendeoca, line 7: what's going on with the brackets in this file? It seems like we have extra, is this intentional?

anyoung-tableau: Intentional. They are block statements to group each validation rule. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block

src/sdks/tableau/apis/vizqlDataServiceApi.ts
anyoung-tableau: Oh nice, good idea!

src/tools/queryDatasource/queryDatasource.ts
stephendeoca: Okay, yeah wasn't thinking about that. Thanks

src/tools/queryDatasource/validators/validateFields.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (fields.length === 0) {
+      throw new Error(
+        'The query must include at least one field. The fields property cannot be an empty array.',
+      );
+    }
+  }
+
+  {
+    // Field caption must be a non-empty string.
+    if (fields.some(hasEmptyFieldCaption)) {
```
stephendeoca, line 18: validateFields will not throw any errors a field does not have a `fieldCaption` when it should. Although the schemas and types require the `fieldCaption` property, it's entirely possible for the client to pass in a query that does not follow the Field schema.

If you add this test it will fail when it should pass.
```
it('should throw if field does not have a fieldCaption', () => {
    // @ts-expect-error - This is a test for the type validator
    expect(() => validateFields([{ calculation: 'SUM([Sales])' }])).toThrow(
      'The query must not include any fields that do not have a fieldCaption.',
    );
  });
```

Can you add logic to throw an error if there does not exist a `fieldCaption` property?

src/tools/queryDatasource/validators/validateFilters.ts
```diff
@@ -0,0 +1,207 @@
...
+
+import { FilterField, Query } from '../queryDatasourceValidator.js';
+import { hasEmptyFieldCaption, hasFunctionAndCalculation } from './validateFields.js';
+
+export function validateFilters(filters: Query['filters']): void {
+  if (!filters) {
+    return;
+  }
+
+  {
```
stephendeoca, line 17: Perhaps throw an error if there are any filters that do not have the field property. Also we should check for the filterType property if it's not there we should throw

src/tools/queryDatasource/validators/validateFilters.ts
```diff
@@ -0,0 +1,207 @@
...
+    // Field caption must be a non-empty string.
+    if (hasEmptyFieldCaption(field)) {
+      return new Err(`The fieldCaption property must be a non-empty string.`);
+    }
+  }
+
+  {
+    // A Field cannot contain both a Function and a Calculation.
+    if (hasFunctionAndCalculation(field)) {
```
stephendeoca, line 214: This logic is a bit different for filterFields, if a filter field has a calculation then it cannot have a fieldCaption.

Here's an example of a case that should pass, but fails since we don't throw when we should
```
it('should throw if filter has a fieldCaption and a calculation', () => {
    expect(() =>
      validateFilters([
        {
          field: { fieldCaption: 'Category', calculation: 'SUM([Sales])' },
          filterType: 'TOP',
          howMany: 10,
          direction: 'TOP',
          fieldToMeasure: { fieldCaption: 'Sales', function: 'SUM' },
        },
      ]),
    ).toThrow(
      'The query includes a filter with an invalid field, fields belonging to filters cannot contain both a fieldCaption and a calculation.',
    );
  });
```

src/tools/queryDatasource/validators/validateFields.ts
anyoung-tableau: Yeah, that seems reasonable. It would be great if we had assurance that the client was always going to provide queries that follow the schema but I don't think we can make that assumption. That said, I also don't feel we should write completely exhaustive tests for any scenario that the client could throw our way. We still have two more additional levels of error handling (HBI and VDS) that will act as safety nets.

PR #41, opened 2025-06-09
Joe/metadata

anyoung-tableau:

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] I have updated the version in the package.json file

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/listFields.ts
+++ b/src/tools/listFields.ts
@@ -75,7 +75,7 @@ export const listFieldsTool = new Tool({
   description: `
     Fetches rich field metadata (name, description, inherited description, dataType, dataCategory, role, etc.) for the specified datasource via Tableau's Metadata API.
     This Returns a list of field dicts or an error message. In general this tool should be used for getting the metadata to ground the use of the query_datasource tool.
-    Note that not all fields,such as Hierarchy fields, can be used with the queryDatasource tool.
+    Note that not all fields, such as Hierarchy fields, can be used with the queryDatasource tool.
     `,
   paramsSchema: {
     datasourceLuid: z.string().nonempty(),


src/tools/queryDatasource/queryDescription.ts
```diff
@@ -1 +1,310 @@
-export const queryDatasourceToolDescription = `Run a Tableau VizQL query.`;
+export const queryDatasourceToolDescription = `# Query Tableau Data Source Tool
+
+Executes VizQL queries against Tableau data sources to answer business questions from published data. This tool allows you to retrieve aggregated and filtered data with proper sorting and grouping.
+
+## Prerequisites
+Before using this tool, you should:
+1. Understand available fields and their types
+3. Understand the data structure and field relationships
```
renoyjohnm, line 8: 2 is missing?

src/tools/listFields.ts
```diff
@@ -8,19 +8,75 @@ import { Tool } from './tool.js';
...
   }`;
 
 export const listFieldsTool = new Tool({
   name: 'list-fields',
-  description:
-    "Fetches field metadata (name, description) for the specified datasource via Tableau's Metadata API, reusing the shared get_datasource_query function. Returns a list of field dicts or an error message.",
+  description: `
+    Fetches rich field metadata (name, description, inherited description, dataType, dataCategory, role, etc.) for the specified datasource via Tableau's Metadata API.
+    This Returns a list of field dicts or an error message. In general this tool should be used for getting the metadata to ground the use of the query_datasource tool.
+    Note that not all fields,such as Hierarchy fields, can be used with the queryDatasource tool.
```
anyoung-tableau:
```suggestion
    Note that not all fields, such as Hierarchy fields, can be used with the queryDatasource tool.
```

src/tools/readMetadata.ts
```diff
@@ -9,8 +9,17 @@ import { validateDatasourceLuid } from './validateDatasourceLuid.js';
...
+  description: `This tool wraps the read-metadata endpoint exposed by Tableau VizQL Data Service. It returns basic, high-level metadata for a specified data source.
+    It strictly provides the following:
+    {
+      "fieldName": "string",
+      "fieldCaption": "string",
+      "dataType": "INTEGER",
+      "defaultAggregation": "SUM",
+      "logicalTableId": "string"
+    }
+    This tool is useful for getting a quick overview of the data source, but it does not provide the rich metadata that the list-fields tool provides.
```
renoyjohnm, line 21: `but it does not provide the rich metadata that the list-fields tool provides.` Not sure if this provides value if list-fields tool is disabled by the end-user

src/tools/queryDatasource/queryDescription.ts
anyoung-tableau, line 1: You should merge the main branch into this branch. This file shouldn't be showing up in the diff.

src/tools/listFields.ts
anyoung-tableau: In general, I don't think we should reference other tools by name since there's no guarantee that one will be available by that name, or if there is one, it might not even be the one from the Tableau MCP server. If we are going to mention the tool by name, we should make sure we're being consistent (its name is `query-datasource`) and ideally use the `ToolName` type to ensure we update it in the event we ever rename it.

```ts
const queryDatasource: ToolName = 'query-datasource';

... `can be used with the ${queryDatasource} tool.`
```

src/tools/queryDatasource/queryDescription.ts
joeconstantino: fixed and will push in a subsequent PR

src/tools/readMetadata.ts
joeconstantino: it provides a small subset of the metadata that can be provided by the list-fields tool. We want MCP hosts to prefer the list fields tool if it's available. Let's try this and adjust later if needed.

src/tools/queryDatasource/queryDescription.ts
joeconstantino: done

PR #44, opened 2025-06-13
Remove singletons and make server stateless

anyoung-tableau:
## Description

These changes remove the global, singleton instances of each tool, and the `Server` and `Config` classes.

## Motivation and Context

These changes are a necessary step before we can properly support client-server communication over HTTP using the Streamable HTTP transport layer. With HTTP communication, _multiple_ clients can connect simultaneously, and having singletons put us at risk at bleeding state across requests. With these changes, each request gets its own isolated instances.

This is based off of this example: https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#without-session-management-stateless


src/server.ts
```diff
@@ -31,45 +31,51 @@ class Server extends McpServer {
...
   registerRequestHandlers = (): void => {
     this.server.setRequestHandler(SetLevelRequestSchema, async (request) => {
-      setLogLevel(request.params.level);
+      setLogLevel(this, request.params.level);
       return {};
     });
   };
-}
 
-export const server = new Server();
```
anyoung-tableau, line 47: Here is the original singleton instance of `Server`.

src/tools/listDatasources/listDatasources.ts
```diff
@@ -4,126 +4,135 @@ import { z } from 'zod';
 
 import { getConfig } from '../../config.js';
 import { getNewRestApiInstanceAsync } from '../../restApiInstance.js';
+import { Server } from '../../server.js';
```
anyoung-tableau, line 7: It looks like there are a lot of changes to each tool, but I've only introduced a factory method that returns the tool.

PR #45, opened 2025-06-13
Add support for Streamable HTTP transport layer

anyoung-tableau:
## Description

The changes add support for the [Streamable HTTP transport type](https://modelcontextprotocol.io/docs/concepts/transports#streamable-http).

It is modeled off the stateless server example described here: https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#without-session-management-stateless

The server is a basic HTTP server using Express and includes 2 basic middlewares: one for rate-limiting and another for rejecting requests who content size is too big.

New configuration environment variables include:

* `TRANSPORT`: Parameterizes transport type. Value can be `stdio` (default) or `http` for Streamable HTTP.
* `HTTP_PORT_ENV_VAR_NAME`: The environment variable name to use for the HTTP server port. The default is `PORT` because that is the environment variable Heroku uses to bind the port to a Node.js web application. See [Deploying Node.js Apps on Heroku](https://devcenter.heroku.com/articles/deploying-nodejs#specify-a-start-script).
* `SSL_KEY`: The path to the SSL key file to use for the HTTP server. If not provided server will run over HTTP.
* `SSL_CERT`: The path to the SSL certificate file to use for the HTTP server. If not provided server will run over HTTP.
* `CORS_ORIGIN_CONFIG`: The origin or origins to allow CORS requests from. See README for more info.

## Motivation and Context

This unblocks customers who want to host the MCP server remotely. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## How Has This Been Tested?

The Claude free tier unfortunately does not support remote MCP servers, so all testing was done in Cursor and MCP Inspector. Also, the [e-bikes demo AI assistant](https://ebikes-demo-a4370287451d.herokuapp.com/McKenzie/ai-assistant) is its own MCP client communicating with a deployment of the Tableau MCP server to Heroku. See https://github.com/tableau/e-bikes-demo/pull/16

## Checklist

- [x] I have updated the version in the package.json file either manually or by using
      `npm run version:minor` for a minor version bump, for example.


vitest.config.ts
```diff
@@ -19,7 +19,7 @@ export default defineConfig({
     coverage: {
       provider: 'v8',
       include: ['src'],
-      exclude: ['src/sdks/**/*', ...coverageConfigDefaults.exclude],
+      exclude: ['src/sdks/**/*', 'src/server/**/*', ...coverageConfigDefaults.exclude],
```
anyoung-tableau, line 22: I'm excluding the `server` directory from code coverage since testing it is difficult has low ROI anyway.

src/server/express.ts
```diff
@@ -0,0 +1,132 @@
...
+    }),
+  );
+
+  const middleware = [
+    rateLimitMiddleware({ windowMs: 60000, maxRequestsInWindow: 100 }), // 100 requests per minute
+    requestSizeLimit({ maxSize: 10 * 1024 * 1024 }), // 10MB max
+  ];
+
+  const path = `/${basePath}`;
+  app.post(path, ...middleware, createMcpServer);
```
anyoung-tableau: This is the important line. MCP clients make their requests to this endpoint which creates a new stateless MCP server and transport with each request. This ensures complete isolation. A single instance would cause request ID collisions when multiple clients connect concurrently.

eslint.config.mjs
```diff
@@ -31,6 +31,7 @@ export default [
       'simple-import-sort': simpleImportSort,
     },
     rules: {
+      'no-console': 'error',
```
anyoung-tableau, line 34: I added this eslint rule to prevent accidental uses of `console` when the `log` helper functions should be used instead. Accidentally using `console.log` is problematic for stdio transport if the message is not a proper JSON RPC message.

src/server/express.ts
```diff
@@ -0,0 +1,132 @@
...
+        'Accept',
+        'MCP-Protocol-Version',
+      ],
+      exposedHeaders: ['mcp-session-id', 'x-session-id'],
+    }),
+  );
+
+  const middleware = [
+    rateLimitMiddleware({ windowMs: 60000, maxRequestsInWindow: 100 }), // 100 requests per minute
+    requestSizeLimit({ maxSize: 10 * 1024 * 1024 }), // 10MB max
```
renoyjohnm: should we consider making these numbers configurable via Env variables or other ways?

anyoung-tableau: Yeah, maybe, or even disabled entirely. I'm sure some customers will want to rely on their own load balancers or proxies in front of this. One idea I've also been wanted to do was moving all the non-sensitive configuration out of environment variables and into a JSON file, which would then become the standard method for config regardless of how the MCP server is started.

anyoung-tableau: I thought about this over the weekend and I'm going to rip out the middleware. We shouldn't be in this business and if MCP server admins need it, they can put a proxy or load balancer in front of it.

PR #62, opened 2025-07-16
Adding validations for SET and MATCH filter type values

renoyjohnm:
## Description
Adding validations for `SET` and `MATCH` filter type values

## Motivation and Context
Agents could invoke the `queryDatasource` tool with invalid `SET` and `MATCH` filter type values causing it to the return empty data result. This PR addresses to validate if the used `SET` and `MATCH` filter type values were valid, by making a extra query to fetch the possible field values, validate if the Agent provided is one of them and provide a descriptive error message that includes a fuzzy matched sampling of the field values if the Agent provided field value is not present in the target field  

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues
NA

## Checklist

## Contributor Agreement

      its Contribution Checklist.


package.json
```diff
@@ -56,7 +56,9 @@
     "shx": "^0.4.0",
     "typescript": "^5.8.3",
     "typescript-eslint": "^8.31.1",
-    "vitest": "^3.1.3"
+    "vitest": "^3.1.3",
+    "fast-levenshtein": "^3.0.0",
```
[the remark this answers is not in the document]
anyoung-tableau: +1 -- `fast-levenshtein` should be under `dependencies`, not `devDependencies`

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+      filter.field.fieldCaption,
+  );
+
+  if (filtersToValidate.length === 0) {
+    return Ok.EMPTY;
+  }
+
+  // Validate each filter
+  for (const filter of filtersToValidate) {
+    const fieldCaption = (filter.field as any).fieldCaption;
```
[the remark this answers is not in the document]
anyoung-tableau: +1, this is asking for a runtime error if we ever make modifications to the conditions used to create `filtersToValidate`. Even though we as devs have confidence that `fieldCaption` exists, TS has no idea.

```js
    if (!('fieldCaption' in filter.field)) {
      continue;
    }
```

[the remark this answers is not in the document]
anyoung-tableau: Either that or use a type assertion on `filtersToValidate` with something like this:

```js
const filtersToValidate = query.filters.filter(
  (filter) =>
    (filter.filterType === 'SET' || filter.filterType === 'MATCH') &&
    'fieldCaption' in filter.field &&
    filter.field.fieldCaption,
) as Array<z.infer<typeof MatchFilter> | z.infer<typeof SetFilter>>;
```

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+          datasource,
+        );
+        if (result.isErr()) {
+          validationErrors.push(result.error);
+        }
+      }
+    } catch (error) {
+      // If validation query fails, we'll let the original query proceed
+      // This ensures we don't block valid queries due to validation issues
+      console.warn(`Filter validation failed for field ${fieldCaption}:`, error);
```
anyoung-tableau: This should be `log.warning`

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  return Ok.EMPTY;
+}
+
+/**
+ * Validates a SET filter by checking if all values exist in the target field
+ */
+async function validateSetFilter(
+  filter: Filter & { filterType: 'SET' },
```
anyoung-tableau: Type should be `z.infer<typeof SetFilter>`

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  return Ok.EMPTY;
+}
+
+/**
+ * Validates a MATCH filter by checking if the pattern matches any values in the target field
+ */
+async function validateMatchFilter(
+  filter: Filter & { filterType: 'MATCH' },
```
anyoung-tableau: Type should be `z.infer<typeof MatchFilter>`

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+      },
+    ],
+  };
+
+  const queryRequest: z.infer<typeof QueryRequest> = {
+    datasource,
+    query: distinctValuesQuery,
+    options: {
+      returnFormat: 'OBJECTS',
+      debug: false,
```
anyoung-tableau: Was setting `debug` to false intentional? https://github.com/tableau/tableau-mcp/pull/27

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+        distance: levenshtein.get(patternString.toLowerCase(), value.toLowerCase()),
+      }))
+      .filter((match) => match.distance <= 5) // Allow up to 5 character differences
+      .sort((a, b) => a.distance - b.distance)
+      .slice(0, 5)
+      .map((match) => match.value);
+
+    const suggestions = similarValues.length > 0 ? similarValues : getRandomSample(fieldValues, 5);
+
+    const patterns = [];
```
anyoung-tableau:
```suggestion
    const patterns: Array<string> = [];
```

src/tools/queryDatasource/queryDatasource.test.ts
```diff
@@ -185,6 +185,311 @@ describe('queryDatasourceTool', () => {
...
+        {
+          signal: new AbortController().signal,
+          requestId: 'test-request-id',
+          sendNotification: vi.fn(),
+          sendRequest: vi.fn(),
+        },
+      );
+
+      expect(result.isError).toBe(false);
+      if (!result.isError) {
```
anyoung-tableau, line 376: Same

src/tools/queryDatasource/queryDatasource.test.ts
```diff
@@ -185,6 +185,311 @@ describe('queryDatasourceTool', () => {
...
+        {
+          signal: new AbortController().signal,
+          requestId: 'test-request-id',
+          sendNotification: vi.fn(),
+          sendRequest: vi.fn(),
+        },
+      );
+
+      expect(result.isError).toBe(true);
+      if (result.isError) {
```
anyoung-tableau: Remove condition, here and in below tests too

src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -0,0 +1,439 @@
...
+        {
+          field: { fieldCaption: 'Region' },
+          filterType: 'SET',
+          values: ['East', 'West'],
+        },
+      ],
+    };
+
+    // Mock successful query returning existing values
+    (mockVizqlDataServiceMethods.queryDatasource as any).mockResolvedValue(
```
anyoung-tableau: Let's try to avoid using `as any`. How about a better type like `as unknown as ReturnType<typeof vi.fn>`?

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
+import * as levenshtein from 'fast-levenshtein';
```
anyoung-tableau: I am seeing a runtime error in MCP inspector:

```
Filter validation failed for field Segment: TypeError: levenshtein.get is not a function at ...
```

I changed the import to this and it resolved:

```ts
import pkg from 'fast-levenshtein';

const { get } = pkg;
```

anyoung-tableau: Alternatively, this would also work:

```ts
import levenstein from 'fast-levenshtein';

...

levenstein.get(...);
```

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,292 @@
...
+    return array;
+  }
+
+  const shuffled = [...array];
+  for (let i = shuffled.length - 1; i > 0; i--) {
+    const j = Math.floor(Math.random() * (i + 1));
+    [shuffled[i], shuffled[j]] = [shuffled[j], shuffled[i]];
+  }
+
+  return shuffled.slice(0, count);
```
anyoung-tableau: There's no test coverage for a scenario that exercises this code. Would you mind adding a test?

src/tools/queryDatasource/queryDatasource.test.ts
renoyjohnm: `npx tsc` [fails with Type guard errors](https://github.com/tableau/tableau-mcp/actions/runs/16333870316/job/46142075468?pr=62) on removing such conditions

src/tools/queryDatasource/validators/validateFilterValues.ts
renoyjohnm: Added a few tests to cover this scenario

src/tools/queryDatasource/validators/validateFilterValues.ts
renoyjohnm: no, copy-paste error

src/tools/queryDatasource/queryDatasource.test.ts
anyoung-tableau: Ah yeah. If you want, instead of using `expect`, you could use `invariant` which provides a hint to TS that the condition is `true` at compile time. Feel free to ignore this comment.

package.json
```diff
@@ -33,7 +33,9 @@
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.12.1",
+    "@types/fast-levenshtein": "^0.0.1",
```
anyoung-tableau: `@types/fast-levenshtein` should still be under `devDependencies`

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -91,7 +93,7 @@ export async function validateFilterValues(
  * Validates a SET filter by checking if all values exist in the target field
  */
 async function validateSetFilter(
-  filter: Filter & { filterType: 'SET' },
+  filter: z.infer<typeof SetFilter> & { filterType: 'SET' },
```
anyoung-tableau: Remove the ` & { filterType: 'SET' }`. That's already the case for a `SetFilter`, yeah?

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -159,7 +163,7 @@ async function validateSetFilter(
  * Validates a MATCH filter by checking if the pattern matches any values in the target field
  */
 async function validateMatchFilter(
-  filter: Filter & { filterType: 'MATCH' },
+  filter: z.infer<typeof MatchFilter> & { filterType: 'MATCH' },
```
anyoung-tableau: Same

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -41,15 +43,15 @@ export async function validateFilterValues(
       (filter.filterType === 'SET' || filter.filterType === 'MATCH') &&
       'fieldCaption' in filter.field &&
       filter.field.fieldCaption,
-  );
+  ) as Array<z.infer<typeof MatchFilter> | z.infer<typeof SetFilter>>;
```
anyoung-tableau: Feel free to add types for both of these so we don't have to use `z.infer` all the time.

`export type SetFilter = z.infer<typeof SetFilter>;`

src/tools/queryDatasource/validators/validateFilterValues.ts
renoyjohnm: ah yes, my bad

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  // Validate each filter
+  for (const filter of filtersToValidate) {
+    const fieldCaption = filter.field.fieldCaption;
+
+    try {
+      if (filter.filterType === 'SET') {
+        const result = await validateSetFilter(
+          filter,
+          fieldCaption,
```
stephendeoca: nitpick: fieldCaption doesn't need to be passed via parameter here if it's already accessible from the filter object that we pass in.

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,301 @@
...
+      returnFormat: 'OBJECTS',
+      debug: true,
+      disaggregate: false,
+    },
+  };
+
+  const result = await vizqlDataServiceMethods.queryDatasource(queryRequest);
+
+  if (result.isErr()) {
+    // If we can't query the field, let the original query proceed
```
stephendeoca, line 117: Is this comment a leftover from when you originally designed the tool to validateFilterValues before making the query?

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,301 @@
...
+    return Ok.EMPTY;
+  }
+
+  const data = result.value.data || [];
+  const existingValues = new Set(
+    (data as Record<string, unknown>[]).map((row) =>
+      String(row.DistinctValues || row[fieldCaption] || ''),
```
stephendeoca, line 124: is there a reason we check row.DistinctValues and row[fieldCaption] won't they yield the same results?

renoyjohnm: I added it as a fallback if the response does not include "DistinctValues". If `row.DistinctValues` return values `row[fieldCaption]` won't be evaluated

src/tools/queryDatasource/validators/validateFilterValues.ts
renoyjohnm: yes, removed it

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,301 @@
...
+      .map((existingValue) => ({
+        value: existingValue,
+        distance: levenshtein.get(invalidValue, existingValue),
+      }))
+      .filter((match) => match.distance <= maxDistance)
+      .sort((a, b) => a.distance - b.distance)
+      .slice(0, Math.ceil(maxSuggestions / invalidValues.length))
+      .map((match) => match.value);
+
+    matches.forEach((match) => suggestions.add(match));
```
stephendeoca: Not really a change request, just an idea for a way to improve this algorithm. Instead of adding (maxSuggestions / invalidValues.length) # of entries to the suggestions array as you progress thru the for loop, you can build up a hash map of invalidValue -> an array of objects of length maxSuggestions where each object has the suggested value and the distance where the array is sorted on based on the distance. Then after iterating over all invalidValues, you pick the closest suggestion out of all invalid values and .slice(0, Math.ceil(maxSuggestions / invalidValues.length)) them so that your suggestions array can't manage only consist really close suggestions for a single invalid value. Then if suggestions.size < maxSuggestions after all that, you can just include more suggestions for same invalid value instead of filling with random samples. Let me know if I am not clear

The current implementation might not be efficient for larger maxDistances since there might be closer suggestions for invalidValues at the end of the invalidValues array. The current implementation doesn't need to be changed, it still gets the job done and the improvements might be overkill for our use case.

Here's an example of the current implementation vs. improvement:
```
invalidValues = [''zzzzz", "xxxxx", "wwwww", "qqqqq","yyyyy", "cance", "blast"] // len = 7
existingValues = ["aaaaa", "bbbbb", "cccccc", "dddddd","eeeeee","chance","cancel", "blood", "blade", "rhyme", "ffffff","gggggg", "hhhhhh", "iiiiii"] // len = 14
maxDistance = 5
maxSuggestions = 5
```

Current implementation will find a max of 2 suggestions per invalidValue for a max of 5 suggestions
```
suggestions = ["aaaaa","bbbbb","aaaaa","bbbbb","aaaaa"]
````

Suggested Implemetation
```
// create hash map of maxSuggestion size 5
["zzzzz"] -> [ { distance: 5, value: "aaaaa" }, { distance: 5, value: "bbbbb" }] // only 2 suggestions found
["xxxxx"] -> [ { distance: 5, value: "aaaaa" }, { distance: 5, value: "bbbbb" } ] // only 2 suggestions found
...
["yyyyy"] -> [ { distance: 4, value: "rhyme" }, { distance: 5, value: "aaaaa" },  { distance: 5, value: "bbbbb" }] // only 3 suggestions found
["cance"] -> [ { distance: 1, value: "chance" }, { distance: 1, value: "cancel" }, ... ]
["blast"] -> [ { distance: 2, value: "blade" }, { distance: 3, value: "blood" }, ... ]

// pick the closest distance suggestions up to 2 max per invalidValue
suggestions = ["chance", "cancel", "blade", "blood", "rhyme"
```

File: src/tools/queryDatasource/validators/validateFilterValues.ts [hunk 21]
```diff
@@ -0,0 +1,325 @@
...
+              const pattern = filter.endsWith!.toLowerCase();
+              const len = pattern.length;
+              const dynamicDistance = Math.min(2, Math.floor(len / 2));
+              const endSlice = lowerValue.slice(-len);
+              return levenshtein.get(pattern, endSlice) <= dynamicDistance;
+            })()
+          : true;
+
+        const fuzzyContainsMatch =
+          filter.contains && lowerValue.length >= filter.contains.length
```
stephendeoca: this doesn't work for the case where we only have a `contains` value in our match filter and the `contains` value is longer than the fieldValue. This will basically allow any fieldValue that's shorter than the `contains` value string to become a suggestion.

Example: Will have 500 suggestions
```
filter = {
  contains: "really long string"
};

fieldValues = ["1", "2", "3", ..., "500"]
```

Instead remove the `&& lowerValue.length >= filter.contains.length` condition, and inside the function (lines 231-236) for the case where the `contains` string is longer, you can check if the levenshtein distance of the entire contains string and entire lowerValue string are within the dynamic distance. If the lowerValue string is longer than the contains string you remain doing the sliding window case.

renoyjohnm: good catch

src/config.ts
```diff
@@ -19,6 +19,7 @@ export class Config {
   includeTools: Array<ToolName>;
   excludeTools: Array<ToolName>;
   maxResultLimit: number | null;
+  disableDatasourceQueryFilterValidation: boolean;
```
stephendeoca: Thoughts on renaming this boolean to `disabledQueryDatasourceFilterValidation` / `DISABLE_QUERY_DATASOURCE_FILTER_VALIDATION` so the name matches the tool?

src/config.test.ts
```diff
@@ -195,6 +196,33 @@ describe('Config', () => {
     expect(config.maxResultLimit).toBe(100);
   });
 
+  it('should set disableDatasourceQueryFilterValidation to false by default', () => {
```
stephendeoca: nit: test strings still say disableDatasourceQueryFilterValidation instead of disableQueryDatasourceFilterValidation

src/config.test.ts
```diff
@@ -195,6 +196,33 @@ describe('Config', () => {
     expect(config.maxResultLimit).toBe(100);
   });
 
+  it('should set disableQueryDatasourceFilterValidation to false by default', () => {
+    process.env = {
+      ...process.env,
+      SERVER: 'https://test-server.com',
+      SITE_NAME: 'test-site',
+      PAT_NAME: 'test-pat-name',
+      PAT_VALUE: 'test-pat-value',
```
anyoung-tableau: These blocks of default variables can be replaced with `...defaultEnvVars,` now

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,351 @@
...
+): Promise<Result<void, FilterValidationError>> {
+  const fieldCaption = filter.field.fieldCaption;
+  const filterValues = filter.values.map((v) => String(v));
+
+  // Query to get distinct values from the field
+  const distinctValuesQuery: z.infer<typeof QueryType> = {
+    fields: [
+      {
+        fieldCaption: fieldCaption,
+        fieldAlias: 'DistinctValues',
```
stephendeoca, line 99: When we add fieldAlias = `DistinctValues`, does this make it so that our values our distinct in the query or is this just for readability? I'm not entirely sure what fieldAliases do for us.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -59,6 +60,22 @@ export const getQueryDatasourceTool = (server: Server): Tool<typeof paramsSchema
...
+                  server,
+                  query,
+                  restApi.vizqlDataServiceMethods,
+                  datasource,
+                );
+
+                if (filterValidationResult.isErr()) {
+                  const errors = filterValidationResult.error;
+                  const errorMessage = errors.map((error) => error.message).join('\n\n');
+                  throw new Error(errorMessage);
```
anyoung-tableau: I wanted to make sure that it was intentional that we're throwing any error returned in the result from `validateFilterValues` rather than returning a new `Err` object instead. If an error is thrown, that will get caught inside `Tool.logAndExecute()` and returned to the MCP client with a JSON-RPC response including `isError: true`. This control flow was mostly reserved for unexpectedly thrown errors.

If returned inside a new Err object, then `Tool.logAndExecute()` will return a similar JSON-RPC response, but not before passing the error to the `getErrorText` function first if it was provided to the tool, which it has for query-datasource, but was only designed to handle `TableauError`.

src/tools/queryDatasource/validators/validateFilterValues.ts
renoyjohnm: I don't think adding fieldAlias = `DistinctValues` ensures that distinct values are returned from VDS, mainly for readability

src/tools/queryDatasource/queryDatasource.ts
renoyjohnm: Thank you for explaining the differences. It wasn't intentional but since the error thrown here is custom and not `TableauError`, throwing `Error` would be appropriate here, right?

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,351 @@
...
+              } else {
+                return levenshtein.get(pattern, lowerValue) <= dynamicDistance;
+              }
+            })()
+          : true;
+        return fuzzyStartMatch && fuzzyEndMatch && fuzzyContainsMatch;
+      }),
+    );
+
+    const suggestions = Array.from(similarValues).slice(0, 5);
```
stephendeoca: an optimization can be made in this code, the second we find our 5th match we should stop fuzzy matching all the remaining values since they'll never make it in our suggestions array anyways.

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,351 @@
...
+    invalidValue: string;
+    suggestionForInvalidValue: string;
+    distance: number;
+  }> = [];
+
+  for (const existingValue of existingValues) {
+    const matches = invalidValues
+      .map((invalidValue) => ({
+        value: invalidValue,
+        distance: levenshtein.get(invalidValue, existingValue),
```
stephendeoca: Oh I just realized we aren't calculating distance based on the lower value, so invalidValue = "something" will have a distance of 9 when compared with existing value = "SOMETHING".

Can you change this to
```
distance: levenshtein.get(invalidValue.toLowerCase(), existingValue.toLowerCase())
```

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: Not necessarily. While it will work since all thrown errors will get caught and handled, that `catch` was designed to catch only unexpectedly thrown errors (like from an API call). Since this is an explicitly handled error scenario, I'd suggest returning an `Err` and updating `getErrorText` to handle both it and a `TableauError`.

A discriminated union like this would work:

```ts
type QueryDatasourceError = {
  type: 'filter-validation',
  message: string,
} | {
  type: 'tableau-error',
  error: TableauError,
}

function getErrorText(error: QueryDatasourceError) {
  switch (error.type) {
    case 'filter-validation': {
      return JSON.stringify({ requestId, message: error.message });
    }
    case 'tableau-error': {
        return JSON.stringify({ requestId, ...handleQueryDatasourceError(error) });
    }
  }
}
```

and update the tool's implementation to return the `TableauError` on the `error` property on the `{ type: 'tableau-error'; error: TableauError }` object.

src/tools/queryDatasource/validators/validateFilterValues.ts
stephendeoca: I see you break when similarValues.size >= 5, so it's not necessary to be calling slice here.

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    const suggestions = Array.from(similarValues).slice(0, 5);
+
+    const patternDescriptions: Array<string> = [];
+    if (filter.startsWith) patternDescriptions.push(`starts with "${filter.startsWith}"`);
+    if (filter.endsWith) patternDescriptions.push(`ends with "${filter.endsWith}"`);
+    if (filter.contains) patternDescriptions.push(`contains "${filter.contains}"`);
+    const similarValuesString =
+      similarValues.size > 0
+        ? `Similar values in this field: ${Array.from(similarValues).join(', ')}.`
```
stephendeoca, line 263: No need to use Array.from(similarValues) here if you already assigned this value to the suggestions array

src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -0,0 +1,685 @@
...
+
+    const result = await validateFilterValues(
+      new Server(),
+      query,
+      mockVizqlDataServiceMethods,
+      mockDatasource,
+    );
+
+    expect(result.isErr()).toBe(true);
+    if (result.isErr()) {
```
stephendeoca, line 132: Since you already expect result.isErr() to be true in the statement above, you won't need to do this if statement. The compiler might make you use the `!` operator to assert the property exists.

src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -0,0 +1,685 @@
...
```
stephendeoca, line 181: if statement not necessary.

src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -0,0 +1,685 @@
...
```
stephendeoca, line 366: No need for this if statement

src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -0,0 +1,685 @@
...
```
stephendeoca, line 460: no need for this if statement

src/tools/queryDatasource/validators/validateFilterValues.test.ts
renoyjohnm: This was discussed here, https://github.com/tableau/tableau-mcp/pull/62#discussion_r2211971565

stephendeoca (approved):

src/tools/queryDatasource/queryDatasource.ts
renoyjohnm: Thank you for the suggestions, let me know if the latest changes based on it looks good

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -72,16 +83,38 @@ export const getQueryDatasourceTool = (server: Server): Tool<typeof paramsSchema
...
                 }
               }
 
-              return await restApi.vizqlDataServiceMethods.queryDatasource(queryRequest);
+              const result = await restApi.vizqlDataServiceMethods.queryDatasource(queryRequest);
+              if (result.isErr()) {
+                return new Err({
+                  type: 'tableau-error',
+                  error: result.error,
+                } as QueryDatasourceError);
```
anyoung-tableau: Instead of using a type assertion (`as QueryDatasourceError`), you can be explicit about the generic `E` type on `Tool.logAndExecute()`. That way if you *tried* to return an `Err` that wraps an object that is not a `QueryDatasourceError` you'll still get a compile error even without the type assertion. Check out https://github.com/tableau/tableau-mcp/pull/73/

Issue #64, opened 2025-07-17
Feature: Add Pulse Metric Value Insight Bundle tool


b-carver:
## Is your feature request related to a problem? Please describe.

I want to be able to get my Pulse Metric Value Insight Bundle

## Describe the solution you'd like

Create a tool that calls the Pulse REST API to get my Pulse Metric Value Insight Bundle

PR #65, opened 2025-07-17
Bcarver/Add Pulse method to see Pulse Metric Value Insight Bundle tool

b-carver:
## Description

This adds additional support of the [Pulse Rest APIs](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_pulse.htm) for the following scenario:

- Generate Pulse Metric Value Insight Bundle

## Motivation and Context

See https://github.com/tableau/tableau-mcp/issues/64

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [X] Documentation update

## How Has This Been Tested?

I performed testing in Cursor of invoking the the new tools in different ways to get information. This also included having it use the Pulse Metric and Pulse Metric Definitions returned from the existing tools.  I also asked cursor to save the result to an HTML file and visualized it in the browser.

## Related Issues

- This will only generate insights in English for the initial release.  A future change can allow specifying language and locale.  Additional changes could also be made for specifying time_zone and output_format.
- Additional items in the request schema are actually optional and can be updated in a future change.

## Checklist

- [X] I have updated the version in the package.json file by using `npm run version`.

## Contributor Agreement

      its Contribution Checklist.


src/sdks/tableau/types/pulse.ts
```diff
@@ -117,6 +123,225 @@ export const pulseMetricSubscriptionSchema = z.object({
   metric_id: z.string(),
 });
 
+// prettier-ignore
```
anyoung-tableau: Please remove these `// prettier-ignore` comments. If you have the Prettier extension installed and Format on Save enabled in VS Code settings, it should auto replace the `"` with `'` characters.

README.md
```diff
@@ -32,6 +32,7 @@ The following MCP tools are currently implemented:
 | list-pulse-metrics-from-metric-definition-id      | List Pulse Metrics from Metric Definition ID ([Pulse API][pulse])                              |
 | list-pulse-metrics-from-metric-ids                | List Pulse Metrics from Metric IDs ([Pulse API][pulse])                                        |
 | list-pulse-metric-subscriptions                   | List Pulse Metric Subscriptions for Current User ([Pulse API][pulse])                          |
+| generate-pulse-metric-value-insight-bundle        | Generate Pulse Metric Value Insight Bundle ([Pulse API][pulse])                                |
```
anyoung-tableau, line 35: Hey Brian, thanks for these changes. Could you please share more about the motivation for adding this tool? Where would it fit into a customer's agentic workflow?

b-carver: This is generating insights for your Pulse Metric.  You might ask a question like "Can you generate insight for my first Pulse Subscription?".  This would then query the necessary tools and return a result like this:
```
Here is the insight bundle for your first Pulse Metric subscription:
Metric Name: Total customers making orders at Superstore
Current Value (Q2 2025): 291
Previous Value (Q1 2025): 225
Change: Up 29.3% (an increase of 66 compared to the prior quarter)
```

You can then ask further questions, like how does this change compare to my other Pulse subscriptions.  You can also ask it to visualize the HTML that was returned, or even look at having it create a file to share with others.

src/sdks/tableau/types/pulse.ts
b-carver: At one point it was trying to change the casing on these definitions which would make the values incorrect, but it's not anymore.  Removing and updating.

README.md
anyoung-tableau: Oh, I see, thanks. This is just the `retrievePulseInsight` tool scheduled for 260 then. Do you think there is any value in adding a parameter to the tool that would cause the agent to intelligently choose between the ban, basic, detail, or springboard endpoints instead of just ban?

b-carver: I was starting with the common one and looking to get feedback.  It's not too much to add the rest though, so I've made the changes.  It'll default to 'ban' if they don't specify, and did some additional tests with the changes.

anyoung-tableau: Nice, thanks!

PR #68, opened 2025-07-18
Suggested changes for filter validation methods

stephendeoca:

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/queryDatasource/validators/validateFilterValues.ts
+++ b/src/tools/queryDatasource/validators/validateFilterValues.ts
@@ -135,6 +135,7 @@ async function validateSetFilter(
       Array.from(existingValues),
       3, // max edit distance
       5, // max suggestions
+      true, // include random suggestions
     );
 
     const message = `Filter validation failed for field "${fieldCaption}". The following values were not found: ${invalidValues.join(', ')}. Did you mean: ${suggestedValues.join(', ')}? Please evaluate whether you included the wrong filter value or if you are trying to filter on the wrong field entirely.`;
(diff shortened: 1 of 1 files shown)


src/tools/queryDatasource/validators/validateFilterValues.test.ts
```diff
@@ -720,4 +720,77 @@ describe('validateFilterValues', () => {
       expect(result.error[0].message).toContain('Similar values in this field:');
     }
   });
+
```
stephendeoca: currently all 3 tests added will fail with the getFuzzyMatch implementation from https://github.com/tableau/tableau-mcp/pull/62/files, but pass with the new implementation changes in this MR.

src/tools/queryDatasource/validators/validateFilterValues.ts
```diff
@@ -283,31 +284,86 @@ function getRandomSample<T>(array: T[], count: number): T[] {
...
+  }
+
+  // If we still don't have enough suggestions, fill with random samples
+  if (toReturn.length < maxSuggestions && includeRandomSuggestions) {
+    const remaining = maxSuggestions - toReturn.length;
+    const randomSamples = getRandomSample(
+      existingValues.filter((v) => !toReturn.includes(v)),
+      remaining,
+    );
+    randomSamples.forEach((sample) => toReturn.push(sample));
```
renoyjohnm: `ensureMinimumSuggestions` method could be used here by adding to its method args, `includeRandomSuggestion` default to true?

renoyjohnm (approved):

src/tools/queryDatasource/validators/validateFilterValues.ts
stephendeoca: I don't think includeRandomSuggestion should default to true. For an algorithm called getFuzzyMatch I would only expect it to give me valid matches if I call it without configuring any options. It wouldn't make sense for it to possibly give me random matches when I'm looking for close matches, unless I specifically asked for that as a fall back.

PR #71, opened 2025-07-23
Add Tableau OAuth support

anyoung-tableau:
These changes introduce two new concepts:

1. The ability for server admins to protect their remotely-deployed servers that use the Streamable HTTP transport type with Tableau OAuth. This means to even use the MCP server, users must sign in to the Tableau Server or Cloud.
2. The ability for server admins to configure their MCP server to no longer require a hard-coded PAT or CA info, but to use the access token issued by Tableau during OAuth dance to authenticate to the Tableau REST APIs. This also means the data returned by these APIs are scoped to the MCP client's user, NOT the user who owns the PAT configured by the MCP server admin.

To clarify, for MCP server admins who protect their server with Tableau OAuth, bullet 2 above is NOT a requirement. They can still provide a PAT or Connected Apps config to use for authenticating to the Tableau REST APIs.

In this implementation, the MCP server is functioning as the authorization server as well. This need not be the case though; the authorization server piece could be deployed as its own distinct service. If necessary, we can revisit the design based on customer feedback.

The OAuth 2.1 authorization flow is implemented as per the [MCP spec version 2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization):

 This provider handles:
 * - Step 1: Initial 401 response with WWW-Authenticate
 * - Step 2: Resource metadata discovery
 * - Step 3: Authorization server metadata
 * - Step 4: Dynamic client registration
 * - Step 5: Authorization with PKCE
 * - Step 6: OAuth callback
 * - Step 7: Token exchange
 * - Step 8: Authenticated requests

Screenshot 2025-07-24 113435

Screenshot 2025-07-24 113606

Screenshot 2025-07-24 113756


src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
+import type { AuthInfo } from '@modelcontextprotocol/sdk/server/auth/types.js';
+import { createHash, randomBytes } from 'crypto';
+import express from 'express';
+import { jwtVerify, SignJWT } from 'jose';
+import { Err, Ok, Result } from 'ts-results-es';
+
+import { isAxiosError } from '../../../node_modules/axios/index.js';
```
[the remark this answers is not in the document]
anyoung-tableau: We have an indirect dependency on Axios via Zodios.

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
...
+
+      return Ok({
+        token,
+        clientId: 'mcp-client',
+        scopes: ['read'],
+        expiresAt: payload.exp,
+        extra: authInfo,
+      });
+    } catch {
+      // TODO: Auto-refresh logic would go here
```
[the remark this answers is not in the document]
anyoung-tableau: This is coming in a followup PR

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
...
+          if (!tokenData || tokenData.expiresAt < Date.now()) {
+            this.refreshTokens.delete(refreshToken);
+            res.status(400).json({
+              error: 'invalid_grant',
+              error_description: 'Invalid or expired refresh token',
+            });
+            return;
+          }
+
+          // TODO: Refresh tokens if needed
```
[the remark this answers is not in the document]
anyoung-tableau: This is coming in a followup PR

src/config.ts
```diff
@@ -35,26 +51,56 @@ export class Config {
...
+        maxValue: ONE_YEAR_IN_MS,
+      }),
+    };
+
+    if (this.oauth.enabled) {
+      invariant(this.oauth.issuer, 'The environment variable OAUTH_ISSUER is not set');
+      invariant(this.oauth.redirectUri, 'The environment variable OAUTH_REDIRECT_URI is not set');
+      invariant(this.oauth.jwtSecret, 'The environment variable OAUTH_JWT_SECRET is not set');
+    } else if (this.auth === 'oauth') {
+      throw new Error('When auth is "oauth", OAUTH_ISSUER must be set');
```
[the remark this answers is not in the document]
anyoung-tableau: Wrong.

src/config.ts
```diff
@@ -35,26 +51,61 @@ export class Config {
...
+      maxValue: 65535,
+    });
     this.corsOriginConfig = getCorsOriginConfig(corsOriginConfig?.trim() ?? '');
     this.datasourceCredentials = datasourceCredentials ?? '';
     this.defaultLogLevel = defaultLogLevel ?? 'debug';
     this.disableLogMasking = disableLogMasking === 'true';
     this.disableQueryDatasourceFilterValidation = disableQueryDatasourceFilterValidation === 'true';
+    this.oauth = {
+      enabled: !!oauthIssuer,
+      issuer: oauthIssuer ?? '',
```
renoyjohnm, line 157: Should we validate if this a valid URL or not?

src/config.ts
```diff
@@ -35,26 +51,61 @@ export class Config {
...
+      minValue: 1,
+      maxValue: 65535,
+    });
     this.corsOriginConfig = getCorsOriginConfig(corsOriginConfig?.trim() ?? '');
     this.datasourceCredentials = datasourceCredentials ?? '';
     this.defaultLogLevel = defaultLogLevel ?? 'debug';
     this.disableLogMasking = disableLogMasking === 'true';
     this.disableQueryDatasourceFilterValidation = disableQueryDatasourceFilterValidation === 'true';
+    this.oauth = {
+      enabled: !!oauthIssuer,
```
renoyjohnm: Any particular reason for the `enabled` value to be associated with `oauthIssuer` over `this.auth === 'oauth'`? Also, should we gate setting the values for `this.oauth` iff `this.auth === 'oauth'`

src/restApiInstance.ts
```diff
@@ -34,7 +37,21 @@ const getNewRestApiInstanceAsync = async (
...
-  await restApi.signIn(config.authConfig);
+  if (config.auth === 'pat') {
+    await restApi.signIn({
+      type: 'pat',
+      patName: config.patName,
+      patValue: config.patValue,
+      siteName: config.siteName,
+    });
+  } else {
+    if (!authInfo?.accessToken || !authInfo?.userId) {
```
renoyjohnm, line 73: Why is `userId` a requirement here?

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
...
+import {
+  AuthenticatedRequest,
+  AuthorizationCode,
+  PendingAuthorization,
+  RefreshTokenData,
+  UserAndTokens,
+} from './types.js';
+
+const TABLEAU_CLIENT_ID = '{E93A0E88-C2F8-4431-B805-11E9957FB03F}';
+const DEVICE_ID = '8FA5479C-56EE-407F-A040-F14FD7E80157';
```
renoyjohnm: I see these values being used on the Tableau OAuth redirect URLs, are these random string UUID strings? If so, should we generate it random every time on being used or are we sticking with these values for monitoring purposes?

anyoung-tableau: They are just random but I'm not sure if generating the guid at runtime is necessary. I know there exists a limit (I think, 10) on refresh tokens under the same device however I never encountered it in my testing so maybe I wasn't getting a new refresh token with each authorization. Let me try generating a fresh guid at runtime and see if there's any downside to that i.e. blowing up the list of Connected devices on your Account Settings page (perhaps there's a limit there too). I'll also see what Desktop is doing.

Image

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
...
+            error_description: tokensResult.error,
+          });
+          return;
+        }
+
+        const { accessToken, refreshToken, expiresInSeconds } = tokensResult.value;
+
+        const restApi = new RestApi(this.config.server);
+        restApi.setCredentials(accessToken, 'unknown user id');
+        const session = await restApi.serverMethods.getCurrentServerSession();
```
anyoung-tableau: Should we return a 401 if the user signs into a site other than the one specified in the `SITE_NAME` env var? I think it might be possible if they manually remove the `target_site` query parameter from the `oauth2/v1/auth` request.

src/config.ts
anyoung-tableau: A server admin can choose to protect their MCP server with Tableau OAuth but still keep `AUTH` set to `pat`. See: https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1753225317161369?thread_ts=1752019991.620519&cid=C08QYBH8AE5

src/restApiInstance.ts
anyoung-tableau: Only because `PulseMethods.listPulseMetricSubscriptionsForCurrentUser` requires it. We could update the `list-pulse-metric-subscriptions` to look it up before calling `listPulseMetricSubscriptionsForCurrentUser`

src/config.ts
anyoung-tableau: I don't really feel strongly that we should since I'd expect anyone setting up OAuth to... y'know... not do that, however trying that out I see that MCP clients don't make it terribly clear what's wrong.

src/server/oauth/provider.ts
anyoung-tableau: I tested this out and yes, the user can grab the original OAuth URL (which might look something like [this](https://prod-apsoutheast-c.online.tableau.com/#/signin?forceSitePick=true&externalRedirect=/oauth2/v1/auth%3Fclient_id%3D%257BE93A0E88-C2F8-4431-B805-11E9957FB03F%257D%26state%3D0d1eb7118eb0d7520d0ae985c80ff423c02bd9a344b87e4c5f2533078fb57a0c%253Ae8eceeb7f97bf932bec4d6e4c4bc3dcb74d63128a511ea87f996ac118f044a38%26code_challenge_method%3DS256%26code_challenge%3DRDCwKaqdAZxMF9ntQYgBFlmc8tuO2o9Rrs4N0xJY-oM%26redirect_uri%3Dhttp%253A%252F%252F127.0.0.1%253A3927%252FCallback%26device_name%3Dtableau-mcp%26device_id%3D8FA5479C-56EE-407F-A040-F14FD7E80157%26redirected%3Dtrue&target_site=andyapsec&client_type=tableau-mcp)) and remove the `target_site` query parameter.

If the user signs in to a site on the same pod as the `SERVER` env var this call should still succeed, but if it's not, it will fail. Since this requires the user to "hack" the OAuth URL maybe we don't care. However, the `SITE_NAME` env var isn't actually required when `AUTH` is `oauth` since the site the user signed into is the one used by the REST API. The original concern applies in this scenario though since the user is prompted for their site or shown a site picker when they sign in. At a minimum we should catch this failure and respond accordingly instead of with a 500.

src/server/oauth/provider.ts
renoyjohnm: As long as these values aren't any secrets or wouldn't cause any security risks, I'm fine with hardcoded values

src/config.ts
renoyjohnm: In that case, any security risks associated with Tableau Tokens being part of the JWT if it is not being used for authentication, should we gate adding the Tableau Tokens to the JWT behind `auth === 'oauth'`?

renoyjohnm (approved):

src/config.ts
anyoung-tableau: We're already doing that. :) Tableau tokens are only part of the JWT if auth == oauth

https://github.com/tableau/tableau-mcp/blob/anyoung/oauth/src/server/oauth/provider.ts#L651

renoyjohnm: My bad, was looking elsewhere for this

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,716 @@
...
+    try {
+      const result = await getTokenResult(this.config.server, {
+        grant_type: 'authorization_code',
+        code,
+        redirect_uri: redirectUri,
+        client_id: clientId,
+        code_verifier: codeVerifier,
+      });
+
+      return Ok(result);
```
anyoung-tableau: The Tableau access token is as powerful as a (site-scoped) PAT and not scoped down to the least privileges that MCP should be afforded. Unfortunately, Tableau OAuth does not support scopes or consent UX.

src/config.ts
```diff
@@ -35,26 +53,71 @@ export class Config {
...
+
+      if (this.transport === 'stdio') {
+        throw new Error('TRANSPORT must be "http" when OAUTH_ISSUER is set');
+      }
+    } else if (this.auth === 'oauth') {
+      throw new Error('When auth is "oauth", OAUTH_ISSUER must be set');
+    }
+
+    if (this.auth !== 'oauth') {
+      invariant(this.siteName, 'The environment variable SITE_NAME is not set');
```
anyoung-tableau: This is wrong. `siteName` can be empty for the default site on Server.

PR #84, opened 2025-08-16
Relax graphql schemas

stephendeoca:
## Description

Reviewed documentation for the GraphQl schemas and added null to all properties that are documented as nullable. A property which is nullable is different from the property being optional.

## Motivation and Context

Calling the listField tool to get the metadata for a datasource would possibly error due to properties that are sometimes null not matching the GraphQL response schema we define.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Ran `npm run test`.

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/sdks/tableau/apis/metadataApi.ts
```diff
@@ -5,43 +5,47 @@ const graphqlResponse = z.object({
   data: z.object({
     publishedDatasources: z.array(
       z.object({
-        name: z.string(),
+        name: z.string().or(z.null()),
```
renoyjohnm, line 8: Can the name be null?

src/sdks/tableau/apis/metadataApi.ts
```diff
@@ -5,43 +5,47 @@ const graphqlResponse = z.object({
...
         description: z.string().or(z.null()),
         owner: z.object({
-          name: z.string(),
+          name: z.string().or(z.null()),
         }),
         fields: z.array(
           z.object({
-            name: z.string(),
-            isHidden: z.boolean(),
+            name: z.string().or(z.null()),
```
renoyjohnm: Same, should we expect null value here?

src/sdks/tableau/apis/metadataApi.ts
stephendeoca: According to our documentation the name property isn't declared as non nullable. If it were declared as non-nullable, that means the value can not be null. So it could possibly be null.

Even if name was null this shouldn't be an issue since we only care about the luid (which we already have) and the fields.

src/sdks/tableau/apis/metadataApi.ts
stephendeoca: The property isn't documented as non-nullable meaning it could technically be null. But since our tools use field names for querying I'll want to remove this so that we error if it's null. I can take care of this specific case later when we merge the two metadata tools.

PR #92, opened 2025-08-26
Use encrypted JWE instead of non-encrypted JWT for MCP access token

anyoung-tableau:


src/server/oauth/provider.ts
```diff
@@ -60,11 +69,41 @@ export class OAuthProvider {
   private readonly jwtIssuer: string;
 
   constructor() {
-    this.jwtSecret = new TextEncoder().encode(this.config.oauth.jwtSecret);
     this.jwtAudience = 'tableau-mcp-server';
     this.jwtIssuer = this.config.oauth.issuer;
   }
 
+  get jwePrivateKey(): KeyObject {
```
onsubmit: These can be private getters

PR #94, opened 2025-08-28
Generate JWT from admin-provided endpoint

anyoung-tableau:
These changes allow server administrators to specify a `JWT_PROVIDER_URL` environment variable such that when `AUTH` is `jwt`, before the MCP server authenticates to the Tableau REST API, it will make a POST request to the endpoint provided in `JWT_PROVIDER_URL`. This endpoint must return the JSON web token to then be used to authenticate to the REST API.

POST request body:

```js
{
  username: "user@tableau.com", // The value of JWT_SUB_CLAIM
  scopes: ["tableau:example:scope"], // The list of scopes the JWT should have
  source: "tableau-mcp",
  resource: 'mcp-tool-name', // The name of the tool being called e.g. query-datasource
  server: "https://tableau.example.com", // The value of SERVER
  siteName: "siteName", // The value of SITE_NAME
}
```

Expected response:

```json
{
  "jwt": "eyJhbGciOiJI..."
}
```

### DIFF
--- a/src/server/express.ts
+++ b/src/server/express.ts
@@ -6,9 +6,10 @@ import fs, { existsSync } from 'fs';
 import http from 'http';
 import https from 'https';
 
-import { Config } from '../config.js';
+import { Config, getConfig } from '../config.js';
 import { setLogLevel } from '../logging/log.js';
 import { Server } from '../server.js';
+import { getJwt } from '../utils/getJwt.js';
 import { validateProtocolVersion } from './middleware.js';
 import { OAuthProvider } from './oauth/provider.js';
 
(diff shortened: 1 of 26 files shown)


src/server/express.ts
```diff
@@ -133,3 +135,35 @@ async function methodNotAllowed(_req: Request, res: Response): Promise<void> {
...
+    res.status(400).json({
+      error: 'username, scopes, source, resource, server, and siteName are required',
+    });
+    return;
+  }
+
+  const additionalPayload: Record<string, unknown> = {};
+  if (resource === 'query-datasource') {
+    additionalPayload.region = 'West';
+  }
```
anyoung-tableau, line 153: This is a simple example of how a customer could customize the user attributes based on any of the input params, including the username, Tableau server URL, Tableau site, and MCP tool being called.

src/server/express.ts
```diff
@@ -53,6 +54,7 @@ export async function startExpressServer({
   app.post(path, ...middleware, createMcpServer);
   app.get(path, ...middleware, methodNotAllowed);
   app.delete(path, ...middleware, methodNotAllowed);
+  app.post('/jwt', generateJwt);
```
anyoung-tableau, line 57: This new endpoint is just for demonstration purposes. Will remove before merging.

PR #98, opened 2025-09-04
Draft: Create a combined single metadata tool

stephendeoca:

the pull request._

# Pull Request Template

## Description

Combines the `list-fields` tool and `read-metadata` tool into the `get-datasource-metadata` tool.

## Motivation and Context

The list fields tool uses the Metadata API which needs to be enabled on server, while the read metadata tool is enabled in all scenarios and uses the VDS API but provides less context. LLMs constructing queries might lose accuracy trying to decide which of the 2 tools to use. Instead we combine both of them into one, which provides only the read-metadata results in the event the list-fields results are unable to be fetched.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Manually using an LLM

## Related Issues

N/A

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+): any {
+  // create a dataDictionary that combines field data from
+  // readMetadata (VizQL Data Service API) and listFields (GraphQL Metadata API) results
+  // to optimize for LLM accuracy and reduce tokens in response.
+  const dataDictionary = {
+    fields: Array<any>(),
+  };
+
+  if (!readMetadataResult.data) {
+    // TODO: what should we do here?
```
stephendeoca: if we don't have any read metadata fields the idea is to return an empty data dictionary. While there's still the list fields result, we don't do anything with it. It should also be empty, unless somehow the vds endpoint is failing for the environment and the metadata endpoint is unaffected.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+      toPush.defaultAggregation = field.defaultAggregation;
+    }
+
+    dataDictionary.fields.push(toPush);
+  }
+
+  if (
+    !listFieldsResult.data.publishedDatasources ||
+    listFieldsResult.data.publishedDatasources.length === 0 ||
+    !listFieldsResult.data.publishedDatasources[0].fields
```
stephendeoca: if we dont have any data from list fields we just return the data dictionary which is only the field info we took from read-metadata

src/tools/getDatasourceMetadata/getDatasourceMetadata.ts
```diff
@@ -75,18 +76,18 @@ const paramsSchema = {
...
     server,
-    name: 'list-fields',
+    name: 'get-datasource-metadata',
     description: `
-    Fetches rich field metadata (name, description, inherited description, dataType, dataCategory, role, etc.) for the specified datasource via Tableau's Metadata API.
-    This Returns a list of field dicts or an error message. In general this tool should be used for getting the metadata to ground the use of the query_datasource tool.
-    Note that not all fields, such as Hierarchy fields, can be used with the queryDatasource tool.
+    This tool retrieves field metadata for a specified datasource by taking the basic, high level, metadata results from Tableau's VizQL Data Service and enriches them with additional context provided by Tableau's Metadata API.
+    The fields provided by this tool will contain properties such as name and dataType, but may also expose richer context such as descriptions, dataCategories, roles, etc.
+    This tool should be used for getting the metadata to ground the use of the query_datasource tool.
```
anyoung-tableau: Heh, the name of the tool is `query-datasource` which we got wrong both in the old and new code. However, I don't think we should reference tools by name since there's no guarantee they'll be available in the client. Instead, I think we should use phrasing like "to ground the use of a tool that queries Tableau published data sources" and leave it to the LLM to figure out which tool applies. Maybe the user disabled our `query-datasource` tool and has some other MCP server whose job it is to query a data source.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
+import { GraphQLResponse } from '../../sdks/tableau/apis/metadataApi.js';
+import { MetadataResponse } from '../../sdks/tableau/apis/vizqlDataServiceApi.js';
+
+export function buildDataDictionary(
+  readMetadataResult: MetadataResponse,
+  listFieldsResult: GraphQLResponse,
+): any {
```
anyoung-tableau: Death to `any`s. https://github.com/tableau/tableau-mcp/pull/99/files

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
+import { GraphQLResponse } from '../../sdks/tableau/apis/metadataApi.js';
+import { MetadataResponse } from '../../sdks/tableau/apis/vizqlDataServiceApi.js';
+
+export function buildDataDictionary(
```
anyoung-tableau: nit: a dictionary implies key-value pairs, but we're just building an array wrapped in an object. Consider renaming to something like `getFields` or `combineFields`.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+  // create a dataDictionary that combines field data from
+  // readMetadata (VizQL Data Service API) and listFields (GraphQL Metadata API) results
+  // to optimize for LLM accuracy and reduce tokens in response.
+  const dataDictionary = {
+    fields: Array<any>(),
+  };
+
+  if (!readMetadataResult.data) {
+    // TODO: what should we do here?
+    return dataDictionary;
```
anyoung-tableau, line 17: Can we safely map `listFieldsResult` to the shape of the data dictionary even if the `readMetadataResult` is empty? If not, while it would great to always return an object of the same shape from a tool, it's not necessary.

src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts
```diff
@@ -0,0 +1,463 @@
...
+        readMetadata: mocks.mockReadMetadata,
+      },
+      metadataMethods: {
+        graphql: mocks.mockGraphql,
+      },
+    }),
+  ),
+}));
+
+describe('getDatasourceMetadataTool', () => {
```
anyoung-tableau, line 161: These are great tests. Nice!

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
anyoung-tableau: > unless somehow the vds endpoint is failing for the environment and the metadata endpoint is unaffected.

Can we safely map `listFieldsResult` to the shape of the data dictionary even if the `readMetadataResult` is empty? If not, while it would great to always return an object of the same shape from a tool, it's not necessary. There may be valuable info in the `listFieldsResult` which we should still return in some shape or form.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+    const matchingField = listFieldsResult.data.publishedDatasources[0].fields.find(
+      (f: any) => f.name === field.name,
+    );
+    if (!matchingField) {
+      continue;
+    }
+    if (matchingField.description) {
+      field.description = matchingField.description;
+    }
+    if (matchingField.descriptionInherited && matchingField.descriptionInherited.length > 0) {
```
anyoung-tableau:
```suggestion
    if (matchingField.descriptionInherited?.length) {
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (matchingField.dataCategory) {
+      field.dataCategory = matchingField.dataCategory;
+    }
+    if (matchingField.role) {
+      field.role = matchingField.role;
+    }
+    if (matchingField.defaultFormat) {
+      field.defaultFormat = matchingField.defaultFormat;
+    }
+    if (matchingField.formula && matchingField.formula.length > 0) {
```
anyoung-tableau:
```suggestion
    if (matchingField.formula?.length) {
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+      field.formula = matchingField.formula;
+      if (matchingField.isAutoGenerated != undefined) {
```
anyoung-tableau:
```suggestion
      if (matchingField.isAutoGenerated !== undefined) {
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+        field.isAutoGenerated = matchingField.isAutoGenerated;
+      }
+      if (matchingField.hasUserReference != undefined) {
```
anyoung-tableau:
```suggestion
      if (matchingField.hasUserReference !== undefined) {
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+        field.hasUserReference = matchingField.hasUserReference;
+      }
+    }
+    if (matchingField.binSize != undefined) {
```
anyoung-tableau:
```suggestion
    if (matchingField.binSize !== undefined) {
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
anyoung-tableau: The schema for the `listFieldsResult` doesn't allow `data.publishedDatasources` or `fields` to be undefined so this can be simplified to:

```ts
if (!listFieldsResult.data.publishedDatasources[0]?.fields.length) {
```

src/tools/getDatasourceMetadata/getDatasourceMetadata.ts
joeconstantino: fair point. Let's make it generic

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
stephendeoca: Yes, we can map the `listFieldsResult` to the shape of the data dictionary. But it's a corner case where somehow `readMetadata` (VDS API) was returning empty but `listFields` (Metadata API) was returning something. In most cases if the VDS API result is empty then the Metadata API would be empty or the fields it has are unqueryable (such as hidden fields). We can return that just so that customers aren't confused with the results in this specific corner case.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
stephendeoca: I intentionally left this as a != instead of !== since the Metadata API reference docs does not assert that this is non nullable, so technically the value could be null or undefined, but we only care to include the property if it's specifically true or false.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
stephendeoca: I agree with you, I'll also rename the return type to be CombinedFieldsResult

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
anyoung-tableau: Gotcha, thanks, please leave a comment explaining that.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -0,0 +1,125 @@
...
+    fields: [],
+  };
+
+  if (!readMetadataResult.data) {
+    if (listFieldsResult.data.publishedDatasources[0]?.fields.length) {
+      // fallback to returning listFields results if we don't have any fields from readMetadata.
+      for (const field of listFieldsResult.data.publishedDatasources[0].fields) {
+        const toPush: Field = { name: field.name };
+        if (field.dataType?.length) {
+          toPush.dataType = field.dataType as Field['dataType'];
```
stephendeoca: listFields' dataTypes are string | null | undefined. If I change `dataType` of our Field (line 10) to be just string, I wont need to typecast here. Lmk I should do this. Same idea with defaultAggregation.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
stephendeoca: Does that mean the schema validation step will throw an error if data.publishedDatasources / fields are undefined?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
anyoung-tableau: Yeah, changing the type to `string` is fine. Since we don't have any assertions about the possible value of listFields' dataTypes, `string` is the best we can do. The typecast (called a type "assertion" in TypeScript) is dangerous and could be misleading.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -1,26 +1,68 @@
...
+        }
+        if (field.aggregation?.length) {
+          toPush.defaultAggregation = field.aggregation as FieldMetadata['defaultAggregation'];
+        }
+        populateFieldWithAdditionalProperties(field as Field, toPush);
```
anyoung-tableau: Consider removing the `as Field` type assertion and updating the `Field` type to allow for nulls that come from the listFields result. Type assertions can sometimes be harmless, but other times can cause really subtle runtime bugs because we are lying to ourselves about what the types actually are.

```ts
type Field = Partial<{
  name: string;
  dataType: string | null;
  defaultAggregation: string | null;
  description: string | null;
  descriptionInherited: Array<{
    attribute: string;
    value: string | null;
  } | null> | null;
  dataCategory: string | null;
  role: string | null;
  defaultFormat: string | null;
  formula: string | null;
  isAutoGenerated: boolean | null;
  hasUserReference: boolean | null;
  binSize: number | null;
}>;
```

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -29,53 +71,55 @@ export function buildDataDictionary(
...
+  if (sourceField.role) {
+    targetField.role = sourceField.role;
+  }
+  if (sourceField.defaultFormat) {
+    targetField.defaultFormat = sourceField.defaultFormat;
+  }
+  if (sourceField.formula?.length) {
+    targetField.formula = sourceField.formula;
+    // Possibly null or undefined, only populate if true or false.
+    if (sourceField.isAutoGenerated != undefined) {
```
anyoung-tableau, line 134: If you really wanted to avoid the wonky JavaScript null/undefined loose equality business:

```ts
if ([true, false].includes(sourceField.isAutoGenerated))
```

stephendeoca: I tried to go with this but the .includes method throws a typescript error since `isAutoGenerated` can be `boolean | null | undefined` according to the new definition for field.

PR #103, opened 2025-09-12
Bump MCP inspector and Claude DXT dependencies

anyoung-tableau:
Some housekeeping changes to pick up some bug fixes in MCP inspector (most notably that HTTP URL prefill in the UI works now) and to use the new MCP Bundle (MCPB) package since they renamed and deprecated the DXT package.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Build Docker image
         run: npm run build:docker
 
-      - name: Build Claude Desktop Extension
+      - name: Build Claude MCP Bundle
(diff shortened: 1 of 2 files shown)


src/scripts/createClaudeMcpBundleManifest.ts
```diff
@@ -258,6 +258,7 @@ const manifest = {
     url: 'https://github.com/tableau/tableau-mcp',
   },
   homepage: packageJson.homepage,
+  documentation: 'https://tableau.github.io/tableau-mcp/',
```
anyoung-tableau, line 261: Driveby addition here

PR #109, opened 2025-09-16
Stop returning error results for schema validation errors

anyoung-tableau:
The Tableau REST API client used by the MCP server uses Zodios for type safety and validating the responses of these APIs. Currently, when an API call succeeds but its response does not match the schema we've defined, we return an error result from the tool.

I've witnessed, at least with Claude Desktop, that the tool result will appear in red and the LLM notes to the user that the tool encountered an error, but because Zodios by default includes the full API in the error message it continues as if the tool hadn't failed at all.

These changes:

1. Identify successful API responses with schema validation errors.
2. Convert the Zod issues into a user-friendly error message using [zod-validation-error](https://www.npmjs.com/package/zod-validation-error)
3. Return a *successful* tool result instead that contains the full API response but with a data quality warning using the error message from 2.

### DIFF
--- a/src/tools/tool.ts
+++ b/src/tools/tool.ts
@@ -1,7 +1,9 @@
 import { ToolCallback } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { CallToolResult, RequestId, ToolAnnotations } from '@modelcontextprotocol/sdk/types.js';
+import { ZodiosError } from '@zodios/core';
 import { Result } from 'ts-results-es';
 import { z, ZodRawShape, ZodTypeAny } from 'zod';
+import { fromError, isZodErrorLike } from 'zod-validation-error';
 
 import { getToolLogMessage, log } from '../logging/log.js';
 import { Server } from '../server.js';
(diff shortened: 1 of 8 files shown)


src/tools/tool.ts
```diff
@@ -133,6 +179,28 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
 
 function getErrorResult(requestId: RequestId, error: unknown): CallToolResult {
+  if (error instanceof ZodiosError && isZodErrorLike(error.cause)) {
+    // Schema validation errors on otherwise successful API calls will not return an "error" result to the MCP client.
+    // We instead return the full response from the API with a data quality warning message
+    // that mentions why the schema validation failed.
+    // This should make it so users don't get "stuck" when our schemas are too strict or wrong.
+    // The only con is that the full response from the API might be larger than normal
+    // since a successful schema validation "trims" the response down to the shape of the schema.
+    const validationError = fromError(error.cause);
```
anyoung-tableau, line 189: `zod-validation-error` also automatically limits the number of issues included in the error message. For example, say the API returned an array with 100 objects, where each object had some validation error. If we returned the Zod issue message to the client, it would be enormous and exhaustively include a message for all 100 objects. With `zod-validation-error`, it defaults to truncating to 10 objects.

src/tools/tool.ts
```diff
@@ -12,24 +14,64 @@ type ArgsValidator<Args extends ZodRawShape | undefined = undefined> = Args exte
   ? (args: z.objectOutputType<Args, ZodTypeAny>) => void
   : never;
 
+/**
+ * The parameters for creating a tool instance
+ *
+ * @typeParam Args - The schema of the tool's parameters
+ */
 export type ToolParams<Args extends ZodRawShape | undefined = undefined> = {
+  // The MCP server instance
```
anyoung-tableau, line 23: Added some driveby comments here

src/tools/tool.ts
```diff
@@ -12,24 +14,64 @@ type ArgsValidator<Args extends ZodRawShape | undefined = undefined> = Args exte
...
 type LogAndExecuteParams<T, E, Args extends ZodRawShape | undefined = undefined> = {
+  // The request ID of the tool call
   requestId: RequestId;
+
+  // The arguments of the tool call
   args: Args extends ZodRawShape ? z.objectOutputType<Args, ZodTypeAny> : undefined;
-  callback: () => Promise<Result<T, E>>;
+
+  // A function that contains the business logic of the tool to be logged and executed
+  callback: () => Promise<Result<T, E | ZodiosError>>;
```
anyoung-tableau, line 60: This signature change is subtle but important. This means the error returned by the callback could potentially be a ZodiosError, which would be thrown if a schema validation happens, but could be something else entirely. The `logAndExecute` method can then check to see if the callback did return a ZodiosError and is a schema validation error (aka ZodError).

stephendeoca:
Can you add a unit test for the getDatasourceMetadata tool where there's a schema validation error in list fields but read-metadata is good, and another test for when read-metadata has a schema error while list-fields is good.

PR #111, opened 2025-09-18
Return a friendly error message when Pulse or VDS is disabled

anyoung-tableau:
The tools that invoke Pulse and VizQL Data Service APIs can potentially fail because their underlying services can be disabled or unavailable. Currently, the user experience is poor because the tools return error results with unhelpful 4xx errors with no hints on how to resolve these errors.

These changes introduce friendly error messages for these scenarios which include instructions on how to resolve them.

## Pulse tool called against Tableau Server
image

## Pulse tool called against Tableau Cloud but Pulse is disabled
image

## Query Datasource called against Tableau Server with VDS disabled
image

## Get Datasource Metadata called against Tableau Server with VDS disabled
image

### DIFF
--- a/src/sdks/tableau/methods/vizqlDataServiceMethods.ts
+++ b/src/sdks/tableau/methods/vizqlDataServiceMethods.ts
@@ -38,11 +38,15 @@ export default class VizqlDataServiceMethods extends AuthenticatedMethods<
    */
   queryDatasource = async (
     queryRequest: z.infer<typeof QueryRequest>,
-  ): Promise<Result<QueryOutput, TableauError | ZodiosError>> => {
+  ): Promise<Result<QueryOutput, 'feature-disabled' | TableauError | ZodiosError>> => {
     try {
       return Ok(await this._apiClient.queryDatasource(queryRequest, { ...this.authHeader }));
     } catch (error) {
       if (isErrorFromAlias(this._apiClient.api, 'queryDatasource', error)) {
+        if (error.response.status === 404) {
+          return Err('feature-disabled');
+        }
+
         return Err(error.response.data);
       }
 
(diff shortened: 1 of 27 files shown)


src/tools/getDatasourceMetadata/getDatasourceMetadata.ts
```diff
@@ -97,44 +106,55 @@ export const getGetDatasourceMetadataTool = (server: Server): Tool<typeof params
...
+                },
+              });
 
-                if (config.disableMetadataApiRequests) {
-                  // Exit early since requests to the Tableau Metadata API are disabled.
-                  return simplifyReadMetadataResult(readMetadataResult);
-                }
+              if (readMetadataResult.isErr()) {
+                return Err({ type: 'feature-disabled' });
+              }
```
anyoung-tableau, line 131: The diff here is kinda bad but this is really only the new lines. All the other changes are to return an `Ok` result.

src/tools/pulse/listMetricDefinitionsFromDefinitionIds/listPulseMetricDefinitionsFromDefinitionIds.test.ts
```diff
@@ -171,33 +172,63 @@ describe('listPulseMetricDefinitionsFromDefinitionIdsTool', () => {
...
-        "inclusive": true,
-        "exact": true,
-        "message": "String must contain exactly 36 character(s)",
-        "path": [
-          "metricDefinitionIds",
-          0
-        ]
-      }
-    ]`,
+  it('should return an error for is metricDefinitionId is too small', async () => {
```
anyoung-tableau, line 176: I moved this test up into the `describe`.

src/sdks/tableau/methods/vizqlDataServiceMethods.ts
```diff
@@ -64,7 +68,18 @@ export default class VizqlDataServiceMethods extends AuthenticatedMethods<
    */
   readMetadata = async (
     readMetadataRequest: z.infer<typeof ReadMetadataRequest>,
-  ): Promise<MetadataResponse> => {
-    return await this._apiClient.readMetadata(readMetadataRequest, { ...this.authHeader });
+  ): Promise<Result<MetadataResponse, 'feature-disabled'>> => {
+    try {
+      return Ok(await this._apiClient.readMetadata(readMetadataRequest, { ...this.authHeader }));
```
stephendeoca, line 73: Is there a need to wrap the response of the apiClient as an `Ok()` result? The list-fields tool has similar failures but doesn't need to be wrapped.

stephendeoca: What's the motivation behind using Ok and Err?

anyoung-tableau: Yeah, we essentially have a tri-state thing going on.

1. An `Ok` result means 👍
2. An `Err` result, which can ONLY have the value `'feature-disabled'`, means "the call failed because the feature is disabled". We handle any `Err` result in the tool's `getErrorText` function to determine what string gets returned back to the client.
3. An exception was thrown, which means something else went wrong and we'll fall back to generic error handling.

The `graphql` method on the Metadata API doesn't need to use `Result<T, E>` because there's only 2 states: 👍 and exception (which we ignore anyway).

PR #118, opened 2025-09-23
Add search tool

stephendeoca:

the pull request._

# Pull Request Template

## Description

Adds the `search-content` tool

## Motivation and Context

Allows for search on all types of content, not just datasources or workbooks.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit tests and manual testing

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/sdks/tableau/methods/contentExplorationMethods.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export default class ContentExplorationMethods extends AuthenticatedMethods<
+  typeof contentExplorationApis
+> {
+  constructor(baseUrl: string, creds: Credentials) {
+    super(new Zodios(baseUrl, contentExplorationApis), creds);
+  }
+
+  /**
+   * Searches across all supported content types for objects relevant to the search expression specified in the querystring of the request URI.
```
anyoung-tableau, line 23: I've been trying to make a note of the required scopes for each REST API method but since this one doesn't support scopes, maybe leave a line like:

```
* Required scopes: Not available.
```

src/sdks/tableau/methods/contentExplorationMethods.ts
```diff
@@ -0,0 +1,51 @@
...
+    filter,
+  }: {
+    terms?: string;
+    page?: number;
+    limit?: number;
+    orderBy?: string;
+    filter?: string;
+  }): Promise<SearchContentResponse> => {
+    const response = await this._apiClient.searchContent({
+      queries: { terms, page, limit, order_by: orderBy, filter },
```
anyoung-tableau: Double check the MCP log notifications, but I am pretty sure that when an arg is undefined, rather than omitting it from the request URL, it will get included e.g. `/search?terms=undefined`. This may not matter for some of the args, but I can see it being problematic for `terms`. Not sure if there's a good way to control this behavior other than doing something like:

```ts
queries: {
  terms: terms ?? '',
  ...
}
```

src/sdks/tableau/types/contentExploration.ts
```diff
@@ -0,0 +1,83 @@
+import { z } from 'zod';
+
+const OrderingMethods = z.enum([
```
anyoung-tableau: nit: these are schemas, not types, so we should use camelCase. Their inferred types should be PascalCase.

src/sdks/tableau/types/contentExploration.ts
```diff
@@ -0,0 +1,83 @@
...
+]);
+
+export type SearchContentFilter = z.infer<typeof SearchContentFilterSchema>;
+
+const SearchContentItemSchema = z.object({
+  uri: z.string(),
+  content: z.record(z.string(), z.unknown()),
+});
+
+export const SearchContentResponseSchema = z.object({
```
anyoung-tableau: If all the properties are optional, consider using Zod's `partial` method. https://v3.zod.dev/?id=partial

src/sdks/tableau/restApi.ts
```diff
@@ -38,6 +39,7 @@ export default class RestApi {
   private _vizqlDataServiceMethods?: VizqlDataServiceMethods;
   private _viewsMethods?: ViewsMethods;
   private _workbooksMethods?: WorkbooksMethods;
+  private _contentExplorationMethods?: ContentExplorationMethods;
```
anyoung-tableau: nit: I've been trying to keep these alphabetized

src/tools/contentExploration/searchContent.ts
```diff
@@ -0,0 +1,104 @@
...
+      const filterString = filter ? buildFilterString(filter) : undefined;
+      if (
+        orderByString?.includes('downstreamWorkbookCount') &&
+        !filterString?.includes('type:eq:table') &&
+        !filterString?.includes('type:eq:database')
+      ) {
+        throw new Error(
+          "When 'orderBy' includes 'downstreamWorkbookCount', the filter must include of content type of 'table' or 'database'",
+        );
+      }
```
anyoung-tableau: Consider using the `argsValidator` concept to validate the arguments. Check out the QueryDatasourceTool for an example.

src/tools/contentExploration/searchContentUtils.ts
```diff
@@ -0,0 +1,163 @@
...
+        return []; // skip duplicate methods
+      }
+      methodsUsed.add(ordering.method);
+      return ordering.method + (ordering.sortDirection ? `:${ordering.sortDirection}` : '');
+    })
+    .join(',');
+}
+
+export function buildFilterString(filter: SearchContentFilter): string {
+  const filterExpressions = new Array<string>();
```
anyoung-tableau: nit: using the `Array` constructor is pretty uncommon in favor of the literal notation instead.

```ts
const filterExpressions: Array<string> = [];
```

src/tools/contentExploration/searchContentUtils.ts
```diff
@@ -0,0 +1,163 @@
...
+    } else if (filter.modifiedTime.endDate) {
+      filterExpressions.push(`modifiedTime:lte:${filter.modifiedTime.endDate}`);
+    }
+  }
+
+  return filterExpressions.join(',');
+}
+
+export function reduceSearchContentResponse(response: SearchContentResponse): Array<object> {
+  const searchResults = new Array<object>();
```
anyoung-tableau: Can we use a narrower type than `object` like `Record<string, unknown>`?

src/tools/contentExploration/searchContentUtils.ts
```diff
@@ -0,0 +1,163 @@
...
+  if (response.items) {
+    for (const item of response.items) {
+      searchResults.push(getReducedSearchItemContent(item.content));
+    }
+  }
+  return searchResults;
+}
+
+function getReducedSearchItemContent(content: Record<string, any>): Record<string, unknown> {
```
anyoung-tableau, line 85: This function is pretty verbose. Can it instead leverage `Object.entries(content)` and loop over the keys, and only add keys whose values are populated?

src/tools/toolName.ts
```diff
@@ -31,6 +38,7 @@ export const toolGroups = {
     'list-pulse-metric-subscriptions',
     'generate-pulse-metric-value-insight-bundle',
   ],
+  contentExploration: ['search-content'],
```
anyoung-tableau: I was thinking that tool names and tool group names should be kebab-case for consistency. Thoughts?

src/tools/contentExploration/searchContent.ts
```diff
@@ -0,0 +1,104 @@
...
+} from './searchContentUtils.js';
+
+const paramsSchema = {
+  terms: z.string().trim().nonempty().optional(),
+  limit: z.number().int().min(1).max(2000).default(2000).optional(),
+  orderBy: OrderBySchema.optional(),
+  filter: SearchContentFilterSchema.optional(),
+};
+
+export const getSearchContentTool = (server: Server): Tool<typeof paramsSchema> => {
```
anyoung-tableau, line 26: Please be sure to add a docs page for the new tool.

src/tools/contentExploration/searchContent.ts
```diff
@@ -0,0 +1,104 @@
...
+            await useRestApi({
+              config,
+              requestId,
+              server,
+              jwtScopes: [],
+              callback: async (restApi) => {
+                const response = await restApi.contentExplorationMethods.searchContent({
+                  terms,
+                  page: 0,
+                  limit: limit ?? 2000, // TODO: determine default limit
```
anyoung-tableau: Note: `limit` should also be no larger than `config.maxResultLimit`

src/tools/toolName.ts
stephendeoca: I'm not a bit fan of kebab case in javascript since accessing properties is limited to using strings `toolGroups["content-exploration"]` but for consistency sake I think that's alright.

src/sdks/tableau/methods/contentExplorationMethods.ts
stephendeoca: Yes, `terms` will be undefined in the query parameter. It's actually not problematic for `terms` to be undefined, it will search as if you didn't include it. Using the empty string as a fallback is actually problematic since it's a query param, so it's not possible to assign a query param a value of empty string and we'll get a 500 error.

src/sdks/tableau/types/contentExploration.ts
stephendeoca: Oh, I think we have some inconsistencies in that case. If you look at `vizqlDataServiceApi.ts` there's a bunch of schemas using PascalCase.

stephendeoca: That might be a special case, I'll fix this.

src/sdks/tableau/methods/contentExplorationMethods.ts
anyoung-tableau: Hah, that sounds like a bug on the API.

src/sdks/tableau/types/contentExploration.ts
anyoung-tableau: Yeah, we haven't been very good at naming consistency in general.

src/tools/toolName.ts
anyoung-tableau: It's that these are the values that get used in the `INCLUDE_TOOLS` env var since you can provide tool names OR tool group names.

src/sdks/tableau/methods/contentExplorationMethods.ts
anyoung-tableau: What happens if we don't provide `terms` in the `queries` object at all when it is undefined? Does Zodios still build the URL with `terms=undefined`? I wish it would just omit the query param entirely.

stephendeoca: the query params are all marked optional in the API schema, so if you don't provide it at all it will omit it from the URL. I tested and confirmed this.

anyoung-tableau: Awesome, thanks for checking that. Let's go with that and we should probably fix the other places we have query params defined in the Zodios api definitions.

src/tools/contentExploration/searchContent.ts
stephendeoca: I actually just discovered that the downstreamWorkbookCount value only appears in table and database types but you can still use this as a sort method for other content types so no need to validate there.

src/tools/contentExploration/searchContentUtils.ts
stephendeoca: I agree. There's actually so many cases for the responses, Joe and I discovered the items have a lot of different properties based on the content type so it's not as straight forward to reduce it as we originally thought. It's most likely we're going to need to have special reductions for each content type in a follow up story. So I'll keep this for now.

But overall what we are doing here is a pretty common operation with our tools. So I think we should come up with a class that reduces the response of any tool for you. you just got to provide the names of the properties and if there are aliases or conversions you want to apply to each property it will do that for you.

src/sdks/tableau/methods/contentExplorationMethods.ts
```diff
@@ -0,0 +1,66 @@
...
+    const queries: Record<string, unknown> = {};
+    if (terms) {
+      queries.terms = terms;
+    }
+    if (page != undefined && page >= 0) {
+      queries.page = page;
+    }
+    if (limit != undefined && limit >= 0) {
+      queries.limit = limit;
+    }
```
anyoung-tableau: Business logic doesn't really belong here. If a consumer wants to pass a negative value for the `limit` they are welcome to. Removing the biz logic would also let you make the construction of `queries` less verbose and future proof with something like:

```ts
searchContent = async (input: {
    terms?: string;
    page?: number;
    limit?: number;
    orderBy?: string;
    filter?: string;
  }): Promise<SearchContentResponse> => {

  Object.entries(input).forEach(([k, v]) => {
   if (v === undefined) {
     delete input[k];
    }
  });

  const response = await this._apiClient.searchContent({ queries: input, ...this.authHeader });
```

docs/docs/tools/content-exploration/search-content.md
```diff
@@ -0,0 +1,161 @@
...
+- `hitsSmallSpanTotal`: Number of times a content item was viewed in the last month
+- `hitsMediumSpanTotal`: Number of times a content item was viewed in the last 3 months
+- `hitsLargeSpanTotal`: Number of times a content item was viewed in the last year
+- `downstreamWorkbookCount`: Number of workbooks in a given project. This value is only available when the content type filter includes 'database' or 'table'
+
+For each sort method, you can specify a sort direction: `"asc"` for ascending or `"desc"` for descending (default: `"asc"`).
+
+The `orderBy` parameter is an array of objects containing the sorting method and direction. The first element determines primary sorting, with subsequent elements used as tiebreakers.
+
+If the `orderBy` parameter is ommitted, the search will sort items by their "relevance score" in descending order, which is Tableau's internal algorithm for providing the most relevant results.
```
anyoung-tableau:
```suggestion
If the `orderBy` parameter is omitted, the search will sort items by their "relevance score" in descending order, which is Tableau's internal algorithm for providing the most relevant results.
```

src/tools/contentExploration/searchContent.test.ts
```diff
@@ -0,0 +1,324 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
```
anyoung-tableau, line 1: Please add a basic e2e test. A followup story is fine too.

stephendeoca: Thanks for reminding me, I'll add an e2e tests in a follow up story.

anyoung-tableau (approved):

PR #119, opened 2025-09-24
Support schemas and tool descriptions that depend on Tableau Server version

anyoung-tableau:


src/getTableauServerVersion.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!tableauServerVersions) {
+    tableauServerVersions = new ExpiringMap<string, ProductVersion>({
+      expirationTimeMs: getConfig().tableauServerVersionCheckIntervalInHours * 60 * 60 * 1000,
+    });
+  }
+
+  const serverVersion = tableauServerVersions.get(server);
+  if (serverVersion) {
+    return serverVersion;
+  }
```
anyoung-tableau, line 37: We cache the version based off the server host for 1 hour (by default). The server host is currently static (based off the `SERVER` environment variable) but once OAuth support lands the MCP server could be pointed at Tableau Cloud and the user could sign into *any* pod. This approach should work for both scenarios.

src/server.ts
```diff
@@ -52,10 +52,15 @@ export class Server extends McpServer {
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- remove this eslint comment once a tool exists whose behavior depends on the Tableau server version
+  private _getToolsToRegister = async (tableauServer: string): Promise<Array<Tool<any>>> => {
     const { includeTools, excludeTools } = getConfig();
 
-    const tools = toolFactories.map((tool) => tool(this));
+    // TODO: Once a tool exists whose behavior depends on the Tableau server version,
+    // we should get the product version here.
+    // const productVersion = await getTableauServerVersion(tableauServer);
+
+    const tools = toolFactories.map((tool) => tool(this /*,  productVersion */));
```
anyoung-tableau: I am realizing now this won't work quite as I expected for long-running Node processes using stdio transport. That is because we only do this once as opposed to with every request over http transport. Let me see if I can adjust accordingly.

src/restApiInstance.test.ts
```diff
@@ -10,16 +10,9 @@ import {
   useRestApi,
 } from './restApiInstance.js';
 import { AuthConfig } from './sdks/tableau/authConfig.js';
-import RestApi from './sdks/tableau/restApi.js';
+import { RestApi } from './sdks/tableau/restApi.js';
 import { Server } from './server.js';
 
-vi.mock('./sdks/tableau/restApi.js', () => ({
```
anyoung-tableau, line 16: Moved this to testSetup.ts so any tool that calls `getTableauServerVersion` won't have to mock out the serverMethods.

src/config.ts
```diff
@@ -76,12 +77,24 @@ export class Config {
...
+    const defaultTableauServerVersionCheckIntervalInHours = 1;
+    const tableauServerVersionCheckIntervalInHoursNumber = tableauServerVersionCheckIntervalInHours
+      ? parseFloat(tableauServerVersionCheckIntervalInHours)
+      : defaultTableauServerVersionCheckIntervalInHours;
+
+    this.tableauServerVersionCheckIntervalInHours =
+      isNaN(tableauServerVersionCheckIntervalInHoursNumber) ||
+      tableauServerVersionCheckIntervalInHoursNumber <= 0
+        ? defaultTableauServerVersionCheckIntervalInHours
+        : tableauServerVersionCheckIntervalInHoursNumber;
```
anyoung-tableau: This will get replaced with the `parseNumber` helper function from the OAuth branch. Ref: https://github.com/tableau/tableau-mcp/pull/142/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R162

package.json
```diff
@@ -20,8 +20,10 @@
   },
   "scripts": {
     ":build": "npx rimraf ./build && esbuild src/index.ts --bundle --packages=external --platform=node --format=esm --minify --outdir=build --sourcemap",
+    ":build:dev": "npx rimraf ./build && esbuild src/index.ts --bundle --packages=external --platform=node --format=esm --outdir=build --sourcemap",
```
anyoung-tableau, line 23: Added a `:build:dev` script which doesn't minify the output. I noticed the debugger in Cursor/VSCode commonly choked on variable previews even with sourcemaps enabled. 🤷‍♂️

src/config.test.ts
```diff
@@ -393,6 +394,28 @@ describe('Config', () => {
...
+    it('should set tableauServerVersionCheckIntervalInHours to the specified value when specified', () => {
+      process.env = {
+        ...process.env,
+        ...defaultEnvVars,
+        TABLEAU_SERVER_VERSION_CHECK_INTERVAL_IN_HOURS: '2',
+      };
+
+      const config = new Config();
+      expect(config.tableauServerVersionCheckIntervalInHours).toBe(2);
+    });
```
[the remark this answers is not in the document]
anyoung-tableau: Oops, thanks

src/getTableauServerVersion.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @returns The version of the Tableau Server or Cloud pod.
+ */
+export const getTableauServerVersion = async (server: string): Promise<ProductVersion> => {
+  if (!server) {
+    // TODO: Once OAuth is available and enabled, the SERVER environment variable may be empty.
+    // This implies the server host name comes from the Tableau Cloud session of the user,
+    // which we include in the MCP access token.
+    // Get the server host name from `req.auth.extra.server` set by the auth middleware.
+    // Remove this condition once this TODO is resolved.
+    throw new Error('server cannot be empty');
```
stephendeoca, line 25: Now that you merged in the OAuth changes is this to be updated? I know were still waiting on support for redirect URIs but the logic here can be added without it.

In this code's current state the SERVER env var is always set, so the map will really only contain one mapping right?

src/utils/isTableauVersionAtLeast.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (
+    year > minYear ||
+    (year === minYear && major > minMajor) ||
+    (year === minYear && major === minMajor && minor >= minMinor)
+  ) {
+    // Tableau version is newer than the minimum required version
+    return true;
+  }
+
+  // Tableau version is older than the minimum required version
```
stephendeoca, line 36: As per our last talk, I asked "What happens when we fail to get product version or it gets a result that doesn't match the release branch format (X.Y.Z)?". I think we said that the logic should fallback to the latest version. That only seems to happen for main.

What about changing the logic to check if the versionValue is in X.Y.Z number format, and if it is we check if it's less than the version and return false. Else we return true and fallback to latest for all other scenarios?

anyoung-tableau: Yeah, that sounds reasonable.

src/getTableauServerVersion.ts
anyoung-tableau: Synced up offline. All calling sites will have a valid server to pass. It may come from the `SERVER` env var or it may come from the auth token. Once we actually *have* a calling site we can either remove just this comment, or the entire condition if we feel confident `server` will always be non-empty in practice.

PR #123, opened 2025-09-30
Adjust docs build path location

bcantoni:
After the update to Docusaurus 3.9.0 (#122), the built files are actually down one level, so that needs to be adjusted for publishing to GH pages.


bcantoni:
As it stands, the docs page is giving a 404 right now: https://tableau.github.io/tableau-mcp/

anyoung-tableau (approved):

PR #131, opened 2025-10-17
Constrain tool results by project/datasource/workbook

anyoung-tableau:
These changes allow server administrators to limit the scope of its tools to a set of data sources, workbooks, or projects.

Enabling tool scoping can cause:

1. Tools to return an error if they are called with arguments that are not within the allowed scope,
   and
2. Tools to respond with results that have been filtered to only include content from the allowed
   scope.

## Examples use-cases

- Only allow clients to query a single data source with the Query Data Source tool. A client attempting to query any other data source will result in an error.
- Filter the results of the List Workbooks tool to only include workbooks that exist in a single project. Workbooks from other projects will not be included in the results.


src/tools/queryDatasource/datasourceCredentials.test.ts
```diff
@@ -5,33 +5,35 @@ import {
...
+}));
 
+describe('getDatasourceCredentials', () => {
   beforeEach(() => {
     resetDatasourceCredentials();
-    process.env = {
-      ...originalEnv,
-      DATASOURCE_CREDENTIALS: undefined,
-    };
-  });
```
anyoung-tableau, line 17: I refactored these tests to mock `getConfig` instead of relying on setting and resetting environment variables.

src/tools/queryDatasource/queryDatasource.test.ts
```diff
@@ -55,19 +58,23 @@ vi.mock('../../restApiInstance.js', () => ({
   ),
 }));
 
-describe('queryDatasourceTool', () => {
```
anyoung-tableau, line 58: I refactored these tests to mock `getConfig` instead of relying on setting and resetting environment variables.

src/tools/workbooks/getWorkbook.ts
```diff
@@ -25,21 +32,36 @@ export const getGetWorkbookTool = (server: Server): Tool<typeof paramsSchema> =>
...
               config,
               requestId,
               server,
               jwtScopes: ['tableau:content:read'],
               callback: async (restApi) => {
-                const workbook = await restApi.workbooksMethods.getWorkbook({
-                  workbookId,
-                  siteId: restApi.siteId,
-                });
+                // Notice that we already have the workbook if it had been allowed by a project scope.
```
anyoung-tableau, line 58: Leaving a note here since this does something a bit special. Checking whether the workbook is allowed could potentially have called `restApi.workbooksMethods.getWorkbook()`, the same method as below. No need to call it twice. See https://github.com/tableau/tableau-mcp/pull/131/files#r2458957992

src/tools/tool.ts
```diff
@@ -65,6 +79,9 @@ type LogAndExecuteParams<T, E, Args extends ZodRawShape | undefined = undefined>
   // A function that can transform an error result of the callback into a string.
   // Required if the callback can return an error result.
   getErrorText?: (error: E) => string;
+
+  // A function that constrains the success result of the tool
+  constrainSuccessResult: (result: T) => ConstrainedResult<T> | Promise<ConstrainedResult<T>>;
```
anyoung-tableau, line 84: I decided not to make this function optional, so we don't forget to implement it for any new tools

src/tools/resourceAccessChecker.ts
```diff
@@ -0,0 +1,365 @@
...
+          message: [
+            'The set of allowed workbooks that can be queried is limited by the server configuration.',
+            `An error occurred while checking if the workbook with LUID ${workbookId} is in an allowed project:`,
+            getExceptionMessage(error),
+          ].join(' '),
+        };
+      }
+    }
+
+    return { allowed: true, content: workbook };
```
anyoung-tableau, line 256: Here's that "special" thing I mentioned in comment: https://github.com/tableau/tableau-mcp/pull/131/files#r2458160645

This case doesn't apply for the other content types which is why you don't see it for data sources or views.

docs/docs/configuration/mcp-config/tool-scoping.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Tool Scoping
+
+The Tableau MCP server can be configured to limit the scope of its tools to a set of data sources,
+workbooks, or projects.
+
+Enabling tool scoping can cause:
+
+1. Tools to return an error if they are called with arguments that are not within the allowed scope,
+   and
```
stephendeoca: is this "and" intentional or did you mean to add more to this sentence?

anyoung-tableau: It was intentional but I can remove it. The 2nd bullet is the continuation of the sentence.

docs/docs/configuration/mcp-config/tool-scoping.md
```diff
@@ -0,0 +1,91 @@
...
+
+A comma-separated list of data source IDs by which to constrain tool arguments and results. Only
+data sources or Pulse metrics and definitions derived from those data sources can be queried or will
+be included in the results of the tools.
+
+- When set, cannot be empty.
+- Data source IDs can be determined using the
+  [Query Data Sources](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_data_sources.htm#query_data_sources)
+  REST API or the [List Data Sources](../../tools/data-qna/list-datasources.md) tool (assuming tool
+  scoping is disabled).
```
stephendeoca, line 68: Perhaps note that limiting a datasource does not limit workbooks or views (assuming that's how that works)

src/config.ts
```diff
@@ -181,6 +214,22 @@ function getCorsOriginConfig(corsOriginConfig: string): CorsOptions['origin'] {
...
 }
 
+// Creates a set from a comma-separated string of values.
+// Returns null if the value is undefined.
+function createSetFromCommaSeparatedString(value: string | undefined): Set<string> | null {
+  if (value === undefined) {
+    return null;
+  }
+
+  return new Set(
```
stephendeoca, line 224: Should we perform validation of these strings?

Like check that the string is either all numbers or follows the LUID format?

I'm not sure if there's an exact format for LUIDs, but let's assume that there is one. If we had a set of rules such as 32 characters equally separated by 4 dashes. Then we can have an error message in case some one configuring the MCP server accidentally copy pastes 31 characters and then they know why their tools aren't letting them query a datasource or something.

src/sdks/tableau/apis/datasourcesApi.ts
```diff
@@ -33,5 +33,15 @@ const listDatasourcesRestEndpoint = makeEndpoint({
   }),
 });
 
-const datasourcesApi = makeApi([listDatasourcesRestEndpoint]);
+const queryDatasourceEndpoint = makeEndpoint({
```
stephendeoca, line 36: Damn, another query datasource endpoint?

src/sdks/tableau/types/pulse.test.ts
```diff
@@ -20,8 +22,9 @@ describe('PulseMetricDefinition schema', () => {
   it('rejects a PulseMetricDefinition with invalid metrics', () => {
     const data = createValidPulseMetricDefinition({
       // is_default should be boolean
+      // @ts-expect-error - is_default should be boolean
```
stephendeoca, line 24: you seem to repeat the same comment here may want to delete lines 24 and 46

src/config.ts
anyoung-tableau: I don't feel strongly that we need to hold their hand.

src/sdks/tableau/types/pulse.test.ts
anyoung-tableau: Sure

docs/docs/configuration/mcp-config/tool-scoping.md
anyoung-tableau: Good idea

src/tools/contentExploration/searchContentUtils.test.ts
```diff
@@ -450,4 +453,119 @@ describe('searchContentUtils', () => {
...
+          datasourceIds: null,
+          workbookIds: null,
+        },
+      });
+
+      invariant(result.type === 'success');
+      expect(result.result).toEqual([items[0]]);
+    });
+
+    it('should return success result when some items were filtered out by allowed projects in the bounded context', () => {
```
stephendeoca: seems like this test case is a duplicate of the one above it, if you're looking to add more scenarios perhaps a test where all bounds are set (project, datasource ids, workbook ids). For the boundedContext you could include ids of a datasource and workboook that both belong to the projectId '123456' and the ids of a datasource and workbook that do not belong to the projectId '123456'. I recall the bounds use AND logic and not OR logic, then the only results we should have are the datsources and workbooks which belong to project '123456'.

Then in another test you can add the same bounds as before but remove the project ID from the bounded context. Then expect the result to include all datasources and workbooks in the bounded context right? since datasources bounds have no impact on workbook bounds?

src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts
```diff
@@ -701,6 +720,28 @@ describe('getDatasourceMetadataTool', () => {
     expect(result.content[0].text).toBe(getVizqlDataServiceDisabledError());
     expect(mocks.mockGraphql).not.toHaveBeenCalled();
   });
+
+  it('should return data source not allowed error when datasource is not allowed', async () => {
```
stephendeoca, line 724: If you want to add more coverage perhaps 2 tests with the following boundedContext.
projectIds: '12345'
datsourceIds: 'luid-1', 'luid-2'
workbookIds: 'luid-3'

The datasource with 'luid-1' belong to project '12345' while datasource with 'luid-2' does not belong to project '12345'

Test 1: Get metadata for 'luid-1', datasource should be allowed

Test 2: Get metadata for 'luid-2', datasource should not be allowed

Mainly to confirm that we use AND logic instead of OR logic for bounds and that workbookId's do not affect datasource queries

src/tools/listDatasources/listDatasources.test.ts
```diff
@@ -72,6 +52,59 @@ describe('listDatasourcesTool', () => {
...
+      const result = constrainDatasources({
+        datasources: mockDatasources.datasources,
+        boundedContext: { projectIds: null, datasourceIds: null, workbookIds: null },
+      });
+
+      invariant(result.type === 'success');
+      expect(result.result).toBe(mockDatasources.datasources);
+    });
+
+    it('should return success result when some datasources were filtered out by the bounded context', () => {
```
stephendeoca: what about a test where we only set project id bounds and from the datasources list we confirm there aren't any datsources included in the result that do not belong to the project

src/tools/pulse/constrainPulseDefinitions.ts
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      type: 'empty',
+      message:
+        'No Pulse Metric Definitions were found. Either none exist or you do not have permission to view them.',
+    };
+  }
+
+  const { datasourceIds } = boundedContext;
+
+  if (datasourceIds) {
```
stephendeoca, line 22: We don't do any project filtering here?

src/tools/resourceAccessChecker.test.ts
```diff
@@ -0,0 +1,585 @@
...
+          await resourceAccessChecker.isDatasourceAllowed({
+            datasourceLuid: mockDatasource.id,
+            restApiArgs,
+          }),
+        ).toEqual({ allowed: true });
+
+        expect(mocks.mockQueryDatasource).not.toHaveBeenCalled();
+      });
+
+      it('should return allowed when the datasource exists in a project that is allowed by the projects in the bounded context', async () => {
```
stephendeoca, line 69: what about a test where we have a projectId bound but not a datasourceId bound?

src/tools/resourceAccessChecker.test.ts
```diff
@@ -0,0 +1,585 @@
...
+
+        // Check again to exercise the cache.
+        expect(
+          await resourceAccessChecker.isWorkbookAllowed({
+            workbookId: mockWorkbook.id,
+            restApiArgs,
+          }),
+        ).toEqual({ allowed: true });
+
+        expect(mocks.mockGetWorkbook).not.toHaveBeenCalled();
```
stephendeoca, line 253: Shouldn't mockGetWorkbook get called once from the initial request before we cached it?

src/tools/resourceAccessChecker.test.ts
```diff
@@ -0,0 +1,585 @@
...
+            workbookId: mockWorkbook.id,
+            restApiArgs,
+          }),
+        ).toEqual({
+          allowed: false,
+          message: [
+            'The set of allowed workbooks that can be queried is limited by the server configuration.',
+            `Querying the workbook with LUID ${mockWorkbook.id} is not allowed.`,
+          ].join(' '),
+        });
```
stephendeoca, line 330: Perhaps do a second call to check if the failure result was cached.

stephendeoca (approved):

src/tools/contentExploration/searchContentUtils.test.ts
anyoung-tableau: This was just a copy/paste error duplication

src/tools/resourceAccessChecker.test.ts
anyoung-tableau: I didn't do a 2nd call on this one because both checks are O(1) time regardless of whether it was a cached result. Non-cached checks if the WorkbookID set contains the workbook ID and cached checks if the `_cachedWorkbookIds` map contains the workbook ID as a key. Caching the result doesn't actually buy us any time, but it'd be more complex to NOT cache the result in this scenario.

anyoung-tableau: Actually, I'm realizing now I'm exercising the cache in the "allowed" version of the test. I don't have a good reason for why I did it there and not here. I'll add it.

src/tools/resourceAccessChecker.test.ts
anyoung-tableau: No, we'd only call mockGetWorkbook if there was a project filter to look up the project ID for the workbook. Since there's only a workbook filter we just check the set of allowed workbook ids in that filter.

src/tools/resourceAccessChecker.test.ts
anyoung-tableau: Sure. Not sure why I didn't include this combination. I may have though I was already over-testing, but this is a good test for this suite.

src/tools/pulse/constrainPulseDefinitions.ts
anyoung-tableau: Joe didn't ask for it. :)

src/tools/listDatasources/listDatasources.test.ts
anyoung-tableau: Sure

src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts
anyoung-tableau: Good test, but I feel pretty good about the coverage already.

PR #138, opened 2025-10-29
Update query datasource tool for 2025.3

stephendeoca:

the pull request._

# Pull Request Template

## Description

Updates Query Datasource tool to support 2025.3 release of VDS:
- Support for creating bins in queries
- Support for overriding parameter values in queries
- Validation of queries using metadata

**Breaking Changes**:
The `DISABLE_QUERY_DATASOURCE_FILTER_VALIDATION` environment variable was deprecated and replaced with 
`DISABLE_QUERY_DATASOURCE_VALIDATION_REQUESTS` to better reflect what the variable does.

## Motivation and Context

VDS released more features for querying as part of the 2025.3 release we want to update our tools to leverage these features to provide more robust and customizable queries.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit tests and manual validation

## Related Issues

N/A

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/scripts/createClaudeMcpBundleManifest.ts
+++ b/src/scripts/createClaudeMcpBundleManifest.ts
@@ -213,7 +213,7 @@ const envVars = {
     type: 'boolean',
     title: 'Disable Metadata API Requests',
     description:
-      'Disable requests to the Tableau Metadata API in the get-datasource-metadata tool. Deprecated in Tableau MCP v1.13.0 and replaced by DISABLE_QUERY_DATASOURCE_VALIDATION_REQUESTS.',
+      'Disable requests to the Tableau Metadata API in the get-datasource-metadata tool.',
     required: false,
     sensitive: false,
   },


docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -160,7 +160,9 @@ If a tool has a `limit` parameter and returns an array of items, the maximum len
 
 <hr />
 
-## `DISABLE_QUERY_DATASOURCE_FILTER_VALIDATION`
+## `DISABLE_QUERY_DATASOURCE_VALIDATION_REQUESTS`
```
anyoung-tableau, line 185: We should be wary of breaking changes. We should either revert the name change or do something like:

1. Add a "🚨 Breaking Changes" section to any releases that contain a breaking change since the last release.
2. Add a "Upgrading" section to the docs that mention to check the GitHub releases for any that have breaking changes.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -36,13 +35,17 @@ export type QueryDatasourceError =
...
       message: string;
     }
+  | {
+      type: 'metadata-validation';
+      message: string;
+    }
   | {
       type: 'filter-validation';
       message: string;
     }
```
anyoung-tableau, line 46: It doesn't look like we make any distinction between metadata and filter validation errors in the final tool call result. Consider combining these into a single `validation` type.

stephendeoca: Having a single validation type sounds like the way to go, although there is a difference between the two errors.

We validate SET and MATCH filters by calling the query-datasource endpoint getting data values on the fields that are used in the filter. Meanwhile metadata validation calls the read-metadata endpoint which uses the metadata to determine if the query has any problems (for example: metadata says field is STRING type, but SUM is being applied to the field in the query).

stephendeoca: let's call the new type `query-validation`

src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+        if (matchingParameter.max != undefined && parameter.value > matchingParameter.max) {
+          validationErrors.push({
+            parameter: parameter.parameterCaption,
+            message: `Parameter '${parameter.parameterCaption}' was provided a value that is greater than the maximum value for the parameter.`,
+          });
+          continue;
+        }
+        // TODO: account for floating point precision issues.
+        if (matchingParameter.step != undefined && parameter.value % matchingParameter.step !== 0) {
```
stephendeoca: this logic needs to be updated. we want to know if parameter value holds form `N*(step size) + min value` where N is an integer. The current calculation doesn't properly reflect that, instead it checks if the parameter value is `N * (step size)`.

Another issue, all values here can be floats which means there's a chance that the % operator returns a result which is 0.00000001 when it's supposed to be 0 in the calculations, we have to enforce a max level of floating point precision and check if its close enough.

Ex: `0.1 + 0.2 % 0.3` actually returns `0.0000000000000004` since `(0.1 + 0.2) = 0.30000000000000004`.

The cutoff point should be the 1 + max number of decimals we can specify in a step size parameter. What I mean is, if the smallest step size is 0.001 in tableau then we accept the parameter value if modulus result that is less than `0.001`

src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
```diff
@@ -0,0 +1,310 @@
...
+        continue;
+    }
+  }
+}
+
+function parameterValueMatchesDataType(
+  value: QueryParameter['value'],
+  dataType: DataType,
+): boolean {
+  // TODO: Determine when parameter values are nullable.
```
stephendeoca: sometimes parameter values are null, i'm not sure when this can occur but for now we return true.

src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
anyoung-tableau: https://www.npmjs.com/package/decimal.js-light

src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
anyoung-tableau: I'd be comfortable leaving this as-is and wait for feedback.

src/scripts/createClaudeMcpBundleManifest.ts
```diff
@@ -212,7 +213,7 @@ const envVars = {
     type: 'boolean',
     title: 'Disable Metadata API Requests',
     description:
-      'Disable requests to the Tableau Metadata API in the get-datasource-metadata tool.',
+      'Disable requests to the Tableau Metadata API in the get-datasource-metadata tool. Deprecated in Tableau MCP v1.13.0 and replaced by DISABLE_QUERY_DATASOURCE_VALIDATION_REQUESTS.',
```
anyoung-tableau: This isn't true, right?

PR #143, opened 2025-11-03
A few housekeeping things

anyoung-tableau:
This PR is mostly just a collection of driveby stuff I noticed while working in the OAuth branch that never made their way to the main branch yet. Figured I'd collect them in a single PR so they don't distract from the real OAuth changes.


config.http.json
```diff
@@ -2,7 +2,7 @@
   "mcpServers": {
     "tableau": {
       "type": "streamable-http",
-      "url": "http://localhost:3927/tableau-mcp"
+      "url": "http://127.0.0.1:3927/tableau-mcp"
```
anyoung-tableau, line 5: This is somewhat OAuth related since the OAuth issuer will be `http://127.0.0.1:3927/tableau-mcp` and MCP inspector complains if there's a mismatch.

eslint.config.mjs
```diff
@@ -14,8 +14,9 @@ export default [
   {
     files: ['**/*.{js,mjs,cjs,ts}'],
     rules: {
-      'no-console': 'error',
+      'no-console': ['error', { allow: ['warn', 'error'] }],
```
anyoung-tableau, line 17: console.warn and console.error are fine to use when the transport is stdio.

eslint.config.mjs
```diff
@@ -14,8 +14,9 @@ export default [
...
       'no-duplicate-imports': ['error', { includeExports: true }],
+      quotes: ['error', 'single', { avoidEscape: true }],
```
anyoung-tableau, line 19: This enforces consistent quote character style.

PR #145, opened 2025-11-04
Add OAuth 2.1 implementation

anyoung-tableau:
This is the authorization implementation using OAuth 2.1.

A detailed writeup of this implementation can be found here: https://docs.google.com/document/d/1NLPNt9WflU9ZeFNG8zVbBDFtl9Nn_rd2amwNrNQtZRs/edit?tab=t.0#heading=h.z44zwmyorvuk

A summary of the flow is described here: https://docs.google.com/document/d/1NLPNt9WflU9ZeFNG8zVbBDFtl9Nn_rd2amwNrNQtZRs/edit?tab=t.0#heading=h.mkfieihap2g2

The sequence diagram is here: https://github.com/tableau/tableau-mcp/blob/anyoung/oauth/docs/docs/configuration/mcp-config/oauth.md#sequence-diagram

Note: No scope support yet. See: https://docs.google.com/document/d/1fVbgiupEuHoI8_5RhnE85nd8B0DMAKotkn-8PB6qSzw/edit?tab=t.0#heading=h.navwdwatt0cf


src/server/oauth/token.ts
```diff
@@ -0,0 +1,340 @@
...
+async function createClientCredentialsAccessToken(
+  clientCredentials: ClientCredentials,
+  publicKey: KeyObject,
+): Promise<string> {
+  const config = getConfig();
+  const payload = JSON.stringify({
+    sub: clientCredentials.clientId,
+    tableauServer: clientCredentials.server,
+    iat: Math.floor(Date.now() / 1000),
+    exp: Date.now() + config.oauth.accessTokenTimeoutMs,
```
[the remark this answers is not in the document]
anyoung-tableau: Good catch. I fixed this for authz code exchange, but not client cred access token.

src/server/oauth/token.ts
```diff
@@ -0,0 +1,337 @@
...
+              error: 'invalid_grant',
+              error_description: 'Invalid code verifier',
+            });
+            return;
+          }
+
+          // Generate tokens
+          const refreshTokenId = randomBytes(32).toString('hex');
+          const accessToken = await createAccessToken(authCode, publicKey);
+          refreshTokens.set(refreshTokenId, {
```
anyoung-tableau, line 86: Storing the refresh tokens in memory is not a long-term solution. Please see: https://docs.google.com/document/d/1NLPNt9WflU9ZeFNG8zVbBDFtl9Nn_rd2amwNrNQtZRs/edit?tab=t.0#heading=h.ro7jv4g2g9jq

src/server/oauth/callback.ts
```diff
@@ -0,0 +1,187 @@
...
+            error: 'invalid_request',
+            error_description: `Invalid origin host: ${originHost}. Expected: ${new URL(config.server).hostname}`,
+          });
+          return;
+        }
+      }
+
+      const server = originHostUrl.toString();
+      const restApi = new RestApi(server);
+      restApi.setCredentials(accessToken, 'unknown user id');
```
anyoung-tableau, line 96: The `'unknown user id'` is used because we don't actually know the user id yet until after we make this call. It doesn't matter though since the user id is normally populated by the result of the Sign In request and is not used as part of the access token

src/server/oauth/authorize.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      return defaultDeviceName;
+    } else if (url.protocol === 'cursor:') {
+      return 'tableau-mcp (Cursor)';
+    } else {
+      return `tableau-mcp (${url.protocol.slice(0, -1)})`;
+    }
+  } catch {
+    return defaultDeviceName;
+  }
```
anyoung-tableau, line 124: This method of determining the device name isn't great, but there doesn't seem to be any good way to figure it out. The OAuth dance occurs *before* the MCP init request which contains information about the client, including its name and version.

src/server/oauth/token.ts
```diff
@@ -0,0 +1,341 @@
...
+              publicKey,
+            );
+          } else {
+            const {
+              accessToken: newTableauAccessToken,
+              refreshToken: newTableauRefreshToken,
+              expiresInSeconds,
+            } = tokensResult.value;
+
+            accessToken = await createAccessToken(
```
anyoung-tableau, line 171: We need to rotate the refresh token here.

See https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#token-theft

src/server/oauth/token.ts
```diff
@@ -0,0 +1,355 @@
...
+          // Rotate the refresh token and extend its expiration time
+          refreshTokens.delete(refreshToken);
+          const refreshTokenId = randomBytes(32).toString('hex');
+
+          refreshTokens.set(refreshTokenId, {
+            user: tokenData.user,
+            server: tokenData.server,
+            clientId: tokenData.clientId,
+            tokens: tokenData.tokens,
+            expiresAt: Math.floor((Date.now() + config.oauth.refreshTokenTimeoutMs) / 1000),
```
anyoung-tableau, line 195: We may need to adjust whether/how we extend the expiration date of the refresh token. I went with the simplest approach for now, which is to extend indefinitely, but there are alternatives e.g.

* Don't extend
* Extend up to a max time since original login

src/sdks/tableau/restApi.ts
```diff
@@ -181,6 +181,24 @@ export default class RestApi {
     this._creds = undefined;
   };
 
+  setCredentials = (accessToken: string, userId: string): void => {
+    const parts = accessToken.split('|');
+    if (parts.length < 3) {
```
stephendeoca, line 186: shouldn't this be parts.length !== 3 if it must have 3 parts. Or does it need at least 3 parts where the 3rd part is always the siteId?

anyoung-tableau: We need to parse out the site id from the access token which is always going to be the 3rd part. We wouldn't want to use a strict check of `!== 3` because if Tableau ever added a 4th part for some other thing, then we'd break.

src/server/oauth/callback.ts
```diff
@@ -0,0 +1,187 @@
...
+      // Parse state to get auth key and Tableau state
+      const [authKey, tableauState] = state?.split(':') ?? [];
+      const pendingAuth = pendingAuthorizations.get(authKey);
+
+      if (!pendingAuth || pendingAuth.tableauState !== tableauState) {
+        res.status(400).json({
+          error: 'invalid_request',
+          error_description: 'Invalid state parameter',
+        });
+        return;
```
stephendeoca, line 59: if we have a pendingAuth but the tableau state doesn't match do we care to delete it's entry from the pendingAuth map?

anyoung-tableau: So, the purpose of the `state` parameter is to prevent XSRF (cross site request forgery) attacks. The only time the state wouldn't match is if:

1. Tableau OAuth didn't reflect it back properly, but this would be a major bug.
2. Some attacker had tricked the user into authorizing their own malicious authorization request and intercepted their authorization code.

Since the attacker *can't* know the actual state value, any value they provide for the `state` parameter would get rejected by this condition and we wouldn't issue them the access token.

Now for the user, if we were to have removed its entry from the map when the attacker tried and failed to exchange the user's authz code for an access token, then when the user DOES provide the correct `state` param value, we'd fail to issue the user the access token they are expecting with no indication why the flow broke for them. In any case, by default authz codes only live 10 minutes and if a server admin wanted, they could shorten their lifetime as much as they see fit.

src/server/oauth/authorize.ts
```diff
@@ -0,0 +1,122 @@
...
+      tableauState,
+      tableauClientId,
+    });
+
+    // Clean up expired authorizations
+    setLongTimeout(() => pendingAuthorizations.delete(authKey), config.oauth.authzCodeTimeoutMs);
+
+    // Redirect to Tableau OAuth
+    const server = config.server || TABLEAU_CLOUD_SERVER_URL;
+    const tableauCodeChallenge = generateCodeChallenge(codeChallenge);
```
stephendeoca: The code challenge we receive from the MCP Client gets used as the verifier for the authorization request to Tableau OAuth. In the OAuth request, we just hash the MCP client code challenge to make the Tableau code challenge. Which means if the MCP client's code challenge is known and the authorization code is known, then the tableau access token can be obtained. Would it be better to use a different verifier for the Tableau code challenge?

anyoung-tableau: Nice catch! I've updated here to generate a new verifier instead of reusing the challenge as the verifier.

src/server/oauth/provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  private readonly pendingAuthorizations = new Map<string, PendingAuthorization>();
+  private readonly authorizationCodes = new Map<string, AuthorizationCode>();
+  private readonly refreshTokens = new Map<string, RefreshTokenData>();
+
+  private readonly privateKey: KeyObject;
+  private readonly publicKey: KeyObject;
+
+  constructor() {
+    this.privateKey = this.getPrivateKey();
+    this.publicKey = createPublicKey(this.privateKey);
```
stephendeoca, line 37: can you remind me the functionality of the public + private key? I think you have some notes on this somewhere so maybe I just need to find that.

anyoung-tableau: I talk about this a bit in the [Access Token Encryption](https://docs.google.com/document/d/1NLPNt9WflU9ZeFNG8zVbBDFtl9Nn_rd2amwNrNQtZRs/edit?tab=t.0#heading=h.xg9merpyj0c9) section.

PR #148, opened 2025-11-04
Update tools to use OAuth info

anyoung-tableau:
These changes update the tool definitions to use the Tableau access token held within the MCP access token to authenticate to the REST APIs


src/index.ts
```diff
@@ -34,6 +35,30 @@ async function startServer(): Promise<void> {
...
+          input: process.stdin,
+          output: process.stdout,
+        });
+
+        const input = (
+          await rl.question('⚠️ To accept these risks, please type "I accept" and press Enter.\n> ')
+        )
+          .toLocaleLowerCase()
+          .replaceAll('"', '')
+          .trim();
```
anyoung-tableau: I might need to change this approach. Admins may not have terminal access to the server they are deploying to and can't type anything. Instead, I might rename the `DISABLE_OAUTH` environment variable to `DANGEROUSLY_DISABLE_OAUTH`.

Issue #153, opened 2025-11-11
Optional Stateful Session Management


aldofebriii:
### Feature Request: Optional Stateful Session Management
---
#### Is your feature request related to a problem? Please describe.

Currently, the `tableau-mcp server` operates in a stateless manner. This presents a challenge when integrating with MCP clients that are designed to manage concurrent connections with multiple MCP servers. For these clients to reliably track context and state for each distinct server, they rely on session management, as outlined in the [MCP Specification on Session Management](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#session-management).

---

#### Describe the solution you'd like

I propose the introduction of optional stateful session management, which can be enabled via an environment variable. This approach would provide flexibility for all users while maintaining backward compatibility.

My suggestion is to add a `STATEFUL` environment variable.
*   If `STATEFUL=true` is set in the environment configuration, the server will initialize the transport with session management enabled. Upon a new client connection, it would generate and assign a unique session ID (e.g., using UUIDv7) to manage the state for that session.
*   If the `STATEFUL` variable is omitted or set to `false`, the server will continue to operate in its default stateless mode.

An example configuration in the `.env` file would be:
```
STATEFUL=true
```

I have prototyped this functionality in my local environment and found that it successfully addresses the integration problem.

Thank you for considering this enhancement.

PR #155, opened 2025-11-11
Add the session management into the MCP Server Express. Add the GET a…

aldofebriii:

the pull request._

# Pull Request Template

#### feat: Implement Stateful Session Management

This commit introduces session management to the MCP Server, enabling it to support multiple concurrent client connections. The server can now operate in either a stateful or stateless mode, configurable via the optional `STATEFUL` environment variable.

**Key Changes:**

*   **Stateful Session Management:** When `STATEFUL=true`, the server now:
    *   Assigns a unique session ID (using UUIDv7) to each new client.
    *   Implements `GET` and `DELETE` endpoint handlers for clients to manage their sessions, aligning with the MCP specification.
    *   Cleans up transport connections gracefully using the `onclose` event handler.
*   **Server Instantiation:** The MCP Server is now instantiated only once at startup, rather than on every incoming request. This ensures a persistent state and improves performance.
*   **Stateless Mode:** In the default stateless mode, the server maintains its original behavior, closing the transport immediately after each response.
*   **Improved Logging:** The server's startup message now indicates whether it is running in stateful or stateless mode.
*   **New Dependency:** Added the `uuid` library to generate unique session identifiers.

This enhancement allows the server to interoperate correctly with MCP clients that require persistent sessions, paving the way for more complex, multi-client interactions.
## Type of Change

- [] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
I only run the npm test and made sure the test is passed but I try to connect to the mcp server with stateless and stateful mode and the result is connected (with sessionId and without sessionId)

## Related Issues
https://github.com/tableau/tableau-mcp/issues/153

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/server/express.ts
+++ b/src/server/express.ts
@@ -1,10 +1,11 @@
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
-import { LoggingLevel } from '@modelcontextprotocol/sdk/types.js';
+import { isInitializeRequest, LoggingLevel } from '@modelcontextprotocol/sdk/types.js';
 import cors from 'cors';
 import express, { Request, Response } from 'express';
 import fs, { existsSync } from 'fs';
 import http from 'http';
 import https from 'https';
+import { v7 as uuidv7 } from 'uuid';
 
 import { Config } from '../config.js';
 import { setLogLevel } from '../logging/log.js';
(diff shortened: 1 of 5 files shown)


src/server/express.ts
```diff
@@ -1,10 +1,11 @@
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
-import { LoggingLevel } from '@modelcontextprotocol/sdk/types.js';
+import { isInitializeRequest, LoggingLevel } from '@modelcontextprotocol/sdk/types.js';
```
anyoung-tableau, line 2: Hey @aldofebriii, thanks for the PR. I have what I think is the same thing in this branch: https://github.com/tableau/tableau-mcp/compare/main...anyoung/client-info

Its intent was to store the MCP client info in the session object so we can better track *which* clients are using the Tableau MCP server, but it looks like it will solve your issue as well.

Happy to push my PR through and we can then close this one.

aldofebriii: Oh yea, thank you 
my bad for not seeing the other branch.

PR #160, opened 2025-11-17
Improve content exploration datasource scoping

anyoung-tableau:
These changes improve datasource tool scoping for the search-content tool. They do a few things:

1. Combine the search results where the same datasource is found, once with `type=unifieddatasource` and again with `type=datasource`. See internal thread [here](https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1763144190913459?thread_ts=1763141863.121709&cid=C08QYBH8AE5).
2. Always use the `luid` property for the datasource luid. Previously, unified datasources used the `datasourceLuid` property.

image

### DIFF
--- a/src/tools/contentExploration/searchContentUtils.test.ts
+++ b/src/tools/contentExploration/searchContentUtils.test.ts
@@ -456,6 +456,40 @@ describe('searchContentUtils', () => {
         luid: 'test-luid',
       });
     });
+
+    it('should combine unifieddatasource entries with correspdonding datasource entries', () => {
+      const response = {
+        total: 1,
+        items: [
+          {
+            uri: 'test-uri',
+            content: {
+              type: 'unifieddatasource',
+              title: 'Test Unified Datasource',
+              datasourceLuid: 'test-datasource-luid',
+              luid: 'test-unifieddatasource-luid',
+              ownerId: 123,
+            },
+          },
+          {
+            uri: 'test-uri-2',
+            content: {
+              type: 'datasource',
+              title: 'Test Datasource',
+              luid: 'test-datasource-luid',
+            },
+          },
+        ],
+      };
+      const result = reduceSearchContentResponse(response);
+      expect(result).toHaveLength(1);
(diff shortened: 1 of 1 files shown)


src/tools/contentExploration/searchContentUtils.ts
```diff
@@ -286,6 +285,18 @@ export function constrainSearchContent({
...
 
+  // Remove duplicate datasources with luid matching a unifieddatasource's datasourceLuid
+  items = items.filter((item) => {
+    if (item.type === 'datasource') {
+      return !items.some(
+        (otherItem) =>
+          otherItem.type === 'unifieddatasource' && otherItem.datasourceLuid === item.luid,
+      );
+    }
+
```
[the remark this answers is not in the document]
anyoung-tableau: Good idea

src/tools/contentExploration/searchContentUtils.ts
```diff
@@ -79,12 +79,38 @@ export function buildFilterString(filter: SearchContentFilter): string {
...
-  const searchResults: Array<ReducedSearchContentResponse> = [];
+  let searchResults: Array<ReducedSearchContentResponse> = [];
   if (response.items) {
     for (const item of response.items) {
       searchResults.push(getReducedSearchItemContent(item.content));
     }
   }
+
+  // Remove duplicate datasources with luid matching a unifieddatasource's datasourceLuid
+  const unifiedDatasourceLuids = new Set(
```
jarhun88, line 90: do we have a test case for this data dedup and normalization?

anyoung-tableau: Yeah, it happens indirectly but I updated the test case to expect the one entry to be removed. Since the expected entry only has the `luid` property if the actual had the `datasourceLuid` property, the test would fail.

anyoung-tableau: I did add an explicit test. Thanks for the comment. :)

PR #161, opened 2025-11-18
Wire up getTableauServerVersion and isTableauVersionAtLeast, make Provider class support async

anyoung-tableau:


src/getTableauServerVersion.ts
```diff
@@ -15,13 +15,8 @@ let tableauServerVersions: ExpiringMap<string, ProductVersion> | undefined;
...
  */
-export const getTableauServerVersion = async (server: string): Promise<ProductVersion> => {
+export const getTableauServerVersion = async (server?: string): Promise<ProductVersion> => {
   if (!server) {
-    // TODO: Once OAuth is available and enabled, the SERVER environment variable may be empty.
-    // This implies the server host name comes from the Tableau Cloud session of the user,
-    // which we include in the MCP access token.
-    // Get the server host name from `req.auth.extra.server` set by the auth middleware.
-    // Remove this condition once this TODO is resolved.
     throw new Error('server cannot be empty');
```
anyoung-tableau, line 20: I decided to leave this condition here. In practice `server` should never be empty/undefined, but a downstream error would be significantly more difficult to identify and debug

src/sdks/tableau/methods/serverMethods.ts
```diff
@@ -29,6 +30,20 @@ export default class ServerMethods extends AuthenticatedMethods<typeof serverApi
   getServerInfo = async (): Promise<ServerInfo> => {
```
anyoung-tableau, line 30: I accidentally added this as an authenticated method initially. It does not require an access token.

src/server/express.ts
```diff
@@ -98,7 +100,7 @@ export async function startExpressServer({
         server.close();
       });
 
-      server.registerTools();
+      await server.registerTools(getTableauAuthInfo(req.auth));
```
anyoung-tableau, line 103: This is how we communicate which server the user is connected to. The server host is included inside the access token.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -37,23 +41,31 @@ export type QueryDatasourceError =
...
   const queryDatasourceTool = new Tool({
     server,
     name: 'query-datasource',
-    description: queryDatasourceToolDescription,
+    description: new Provider(async () => {
+      const productVersion = await getTableauServerVersion(config.server || authInfo?.server);
+      if (isTableauVersionAtLeast({ productVersion, minVersion: '2025.3.0' })) {
+        return queryDatasourceToolDescription;
+      }
+
```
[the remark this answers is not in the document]
anyoung-tableau: Yes I know, it's just an example.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -37,23 +41,31 @@ export type QueryDatasourceError =
...
+      return queryDatasourceToolDescription;
```
anyoung-tableau, line 67: Here's the example where we customize the description based off the Tableau version

stephendeoca: This is just an example, but technically we're returning the same thing for now which we'll add in later.

PR #162, opened 2025-11-19
Pulse pagination

jarhun88:

the pull request._

# Pull Request Template

## Description
Pulse apis utilize a different form of pagination than the standard tableau apis, the major difference being the use of page_token.

Adding support for pulse api pagination will make it much easier to implement the pulse discover tool. 

Screenshot 2025-11-19 at 1 22 25 AM
Screenshot 2025-11-19 at 1 24 53 AM

## Motivation and Context

- unlocks pulse pagination for pulse apis (this pr only implemented for listDefinitionTool)
- fixed some hidden failing validations in the definition schema 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- unit tests on pulse pagination
- manually tested scenarios

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
+++ b/docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
@@ -40,9 +40,8 @@ pagination.
 
 Example: `25`
 
-**Note:** This parameter is optional. If not specified, the API will use its default page size. You
-may want to provide a larger value if you know in advance that you have more than the 10 (default
-page size) definitions to retrieve.
+**Note:** This parameter is optional. If not specified, the API will use its default page size of 10 definitions. You
+may want to provide a larger value if you know in advance that you have more than the 10 definitions to retrieve.
 
 ## Example result
 


src/sdks/tableau/types/pulse.ts
```diff
@@ -34,7 +34,7 @@ const pulseBasicSpecificationSchema = z.object({
 
 const pulseSpecificationSchema = z.object({
   datasource: pulseDatasourceSchema,
-  basic_specification: pulseBasicSpecificationSchema,
+  basic_specification: pulseBasicSpecificationSchema.optional(),
```
jarhun88, line 37: this schema in pulse is a oneof basic_specification or viz_state_specification. Hence there was some validation errors when we received a definition that was viz_state_specification. 

The simple solution was to have them both optional. read calls will always return oneof the 2 fields.

src/sdks/tableau/types/pulse.ts
```diff
@@ -101,9 +101,13 @@ export const comparisonSchema = z.object({
 
 export const datasourceGoalsSchema = z.array(
   z.object({
-    basic_specification: pulseBasicSpecificationSchema,
+    basic_specification: pulseBasicSpecificationSchema.optional(),
```
jarhun88, line 104: a dsgoal schema is a oneof of these 4 fields. Hence adding them all as optional. Fixes broken mcp validations that was occuring

src/sdks/tableau/types/pulse.ts
```diff
@@ -101,9 +101,13 @@ export const comparisonSchema = z.object({
...
+    threshold_basic_specification: pulseBasicSpecificationSchema.optional(),
+    threshold_viz_state_specification: z.object({ viz_state_string: z.string() }).optional(),
     viz_state_specification: z.object({ viz_state_string: z.string() }).optional(),
     minimum_granularity: z.string(),
+    benchmark_sentiment_type: z.string(),
```
jarhun88, line 109: adding missing fields

src/utils/paginate.ts
```diff
@@ -46,3 +48,54 @@ export async function paginate<T>({ pageConfig, getDataFn }: PaginateArgs<T>): P
...
+  config: PulsePaginateConfig;
+  getDataFn: (
+    pageToken?: string,
+  ) => Promise<PulseResult<{ pagination: PulsePagination; definitions: Array<T> }>>;
+};
+
+export async function pulsePaginate<T>({
+  config,
+  getDataFn,
+}: PulsePaginateArgs<T>): Promise<PulseResult<Array<T>>> {
```
jarhun88, line 72: The current implementation of pulse methods expect the callback within logAndExecute() to return a PulseResult. 

The simplest minimally invasive approach was to have pulsePaginate return PulseResult, this required having to 

1. check the result of getDataFn() for err, and returning early if so.
2. unwrap non error data and append paginated data
3. wrap and return final data in an Ok()

src/sdks/tableau/types/pulse.ts
anyoung-tableau: Just a thought because I am curious what your opinion is on whether the type on `benchmark_sentiment_type` should be a string or an enum. I like it being a string so we won't break if Pulse were to add a new enum entry. However, if it were to be used as a tool input argument, an LLM could technically provide a bad value for the enum and we wouldn't prevent it. Thoughts?

From the OpenAPI spec:
```
benchmark_sentiment_type:
  type: "string"
  enum:
  - "BENCHMARK_SENTIMENT_TYPE_UNSPECIFIED"
  - "BENCHMARK_SENTIMENT_TYPE_NONE"
  - "BENCHMARK_SENTIMENT_TYPE_ABOVE_THRESHOLD_IS_UNFAVORABLE"
  - "BENCHMARK_SENTIMENT_TYPE_BELOW_THRESHOLD_IS_UNFAVORABLE"
```

src/utils/paginate.ts
```diff
@@ -46,3 +48,54 @@ export async function paginate<T>({ pageConfig, getDataFn }: PaginateArgs<T>): P
...
+  })
+  .optional();
+
+type PulsePaginateConfig = z.infer<typeof pulsePaginateConfigSchema>;
+
+type PulsePaginateArgs<T> = {
+  config: PulsePaginateConfig;
+  getDataFn: (
+    pageToken?: string,
+  ) => Promise<PulseResult<{ pagination: PulsePagination; definitions: Array<T> }>>;
```
anyoung-tableau: Since this pagination could apply to more than just Pulse *definitions*, considering renaming `definitions` to something more generic, like `items` or `data` like the other `paginate` function. Line 86 too.

src/utils/paginate.ts
anyoung-tableau: Good idea and I think that's fine.

src/sdks/tableau/apis/pulseApi.ts
```diff
@@ -26,9 +26,18 @@ const listAllPulseMetricDefinitionsRestEndpoint = makeEndpoint({
...
+      type: 'Query',
+      schema: z.optional(z.string()),
+      description: 'Token for retrieving the next page of results. Omit for the first page.',
+    },
   ],
   response: z.object({
     definitions: z.array(pulseMetricDefinitionSchema),
+    next_page_token: z.string().optional(),
+    offset: z.coerce.number(),
+    total_available: z.coerce.number(),
```
anyoung-tableau, line 46: There's also a `page_size` param I think. The other tools that support pagination also include a `pageSize` argument. Was it intentionally omitted? I think we should probably include it so folks who want 1000 results don't have to call the API 20 times with a default page size of 50 (I'm assuming it's 50 but have no idea)

src/tools/pulse/listAllMetricDefinitions/listAllPulseMetricDefinitions.ts
```diff
@@ -3,15 +3,20 @@ import { z } from 'zod';
...
 import { Server } from '../../../server.js';
 import { getTableauAuthInfo } from '../../../server/oauth/getTableauAuthInfo.js';
+import { pulsePaginate } from '../../../utils/paginate.js';
 import { Tool } from '../../tool.js';
 import { constrainPulseDefinitions } from '../constrainPulseDefinitions.js';
 import { getPulseDisabledError } from '../getPulseDisabledError.js';
 
 const paramsSchema = {
   view: z.optional(z.enum(pulseMetricDefinitionViewEnum)),
+  limit: z.coerce.number().gt(0).optional(),
```
anyoung-tableau, line 20: Please update the docs for this tool similar to: https://github.com/tableau/tableau-mcp/blob/main/docs/docs/tools/data-qna/list-datasources.md?plain=1#L26-L45

src/sdks/tableau/apis/pulseApi.ts
jarhun88: oh hmm ill look into that

src/sdks/tableau/types/pulse.ts
jarhun88: I suppose if we plan to support pulse write operations, an enum will be safer overall (though we dont rn, and its not clear to me if thats on the roadmap), perhaps something we could revisit down the line?

anyoung-tableau: Yeah, definitely fine to revisit later.

src/tools/pulse/listAllMetricDefinitions/listAllPulseMetricDefinitions.ts
jarhun88: done

docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
```diff
@@ -24,6 +24,25 @@ Possible values:
...
+
+### `pageSize`
+
+The number of results per page. Controls how many definitions are fetched in each API request during
+pagination.
+
+Example: `25`
+
+**Note:** This parameter is optional. If not specified, the API will use its default page size.
+Larger page sizes can reduce the number of API calls but may increase response time per request.
```
jarhun88: i guess this is not really correct with smart pagination its actually more optimal not to pass a page size lol

src/sdks/tableau/apis/pulseApi.ts
jarhun88: ok ive added the ability to pass a specific page size, and also added smart pagesize if its not passed

src/utils/paginate.ts
```diff
@@ -46,3 +48,68 @@ export async function paginate<T>({ pageConfig, getDataFn }: PaginateArgs<T>): P
...
+  const total_available = pagination.total_available;
+
+  let next_page_token = pagination.next_page_token;
+
+  // If pageSize is not provided, set it to the minimum of the total available data and the remaining limit
+  if (!pageSize && total_available) {
+    pageSize = Math.min(
+      total_available - resultArray.length,
+      limit ? limit - resultArray.length : Number.MAX_SAFE_INTEGER,
+    );
```
anyoung-tableau, line 92: Nice idea on only doing smart paging if the user didn't request an explicit page size.

anyoung-tableau, line 92: Consider adding a test that exercises the smart pagination. Looks like we have a test gap on it.

docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
anyoung-tableau: Yeah, with the introduction of smart pagination the only real benefit this argument seems to have is to provide a way to get all the results in a single API call, instead of relying on the smart pagination to get a small amount in the first request, and then a large amount in the 2nd request. Maybe we just mention something similar to what we say for the list-datasources tool: `You may want to provide a larger value if you know in advance that you have more than 100 definitions to retrieve.` (Assuming 100 is the default value).

src/utils/paginate.ts
jarhun88: added 2 tests

docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
jarhun88: i have added so

docs/docs/tools/pulse/list-all-pulse-metric-definitions.md
```diff
@@ -24,6 +24,26 @@ Possible values:
...
+### `pageSize`
+
+The number of results per page. Controls how many definitions are fetched in each API request during
+pagination.
+
+Example: `25`
+
+**Note:** This parameter is optional. If not specified, the API will use its default page size. You
+may want to provide a larger value if you know in advance that you have more than the 10 (default
+page size) definitions to retrieve.
```
anyoung-tableau:
```suggestion
**Note:** This parameter is optional. If not specified, the API will use its default page size of 10 definitions. You
may want to provide a larger value if you know in advance that you have more than the 10 definitions to retrieve.
```

anyoung-tableau (approved):

src/utils/paginate.ts
anyoung-tableau: Thanks!

PR #163, opened 2025-11-19
Fix OAuth in Claude

anyoung-tableau:
After debugging some more into what I had initially thought was a [bug in Claude](https://github.com/anthropics/claude-code/issues/5826), turns out it was just a small bug on us.

These changes:

1. Add `client_secret_post` as an advertised auth method. Previously we were only advertising `client_secret_basic` but Claude uses `client_secret_post` and posts client credentials on the token request body, not in the Authorization header. We did *support* this method but we weren't *advertising* that fact.

3. We had a bug where the `WWW-Authenticate` header always used `http` for the `resource_metadata` URL. Every other client I've used seems to "upgrade" it to https for me but Claude didn't and I hadn't noticed that until now.

### DIFF
--- a/src/server/express.ts
+++ b/src/server/express.ts
@@ -44,7 +44,7 @@ export async function startExpressServer({
     }),
   );
 
-  if (config.trustProxyConfig) {
+  if (config.trustProxyConfig !== null) {
     // https://expressjs.com/en/guide/behind-proxies.html
     app.set('trust proxy', config.trustProxyConfig);
   }


tests/oauth/oauth.test.ts
```diff
@@ -94,7 +94,7 @@ describe('OAuth', () => {
       grant_types_supported: ['authorization_code', 'refresh_token', 'client_credentials'],
       code_challenge_methods_supported: ['S256'],
       scopes_supported: [],
-      token_endpoint_auth_methods_supported: ['client_secret_basic'],
+      token_endpoint_auth_methods_supported: ['client_secret_basic', 'client_secret_post'],
```
anyoung-tableau, line 97: Here we are now advertising we support `client_secret_post` instead of only `client_secret_basic` even though we actually *did* support both on the token endpoint. 🫠

PR #165, opened 2025-11-20
Anyoung/query datasource 258

anyoung-tableau:

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
+++ b/src/tools/queryDatasource/validators/validateQueryAgainstDatasourceMetadata.ts
@@ -291,7 +291,6 @@ function parameterValueMatchesDataType(
   value: QueryParameter['value'],
   dataType: DataType,
 ): boolean {
-  // TODO: Determine when parameter values are nullable.
   if (value === null) {
     return true;
   }


src/utils/isTableauVersionAtLeast.ts
```diff
@@ -1,6 +1,27 @@
...
+  server,
+  mappings,
+}: {
+  server: string | undefined;
+  mappings: { [minVersion: TableauVersion]: T } & { default: T };
+}): Promise<T> {
+  const productVersion = await getTableauServerVersion(server);
+  for (const [minVersion, value] of Object.entries(mappings) as Array<[TableauVersion, T]>) {
+    if (isTableauVersionAtLeast({ productVersion, minVersion })) {
+      return value;
```
stephendeoca, line 17: is iteration done in order of latest / highest release? What happens if our server version is 2026.1 but we have mappings for 2026.1 and 2025.3. If we don't iterate in order of latest / highest release, if we check if 2026.1 is at least 2025.3 it will be true and return the 2025.3 mapping result instead of the 2026.1 mapping.

stephendeoca (approved):

src/utils/isTableauVersionAtLeast.ts
anyoung-tableau: It goes by the order of the keys provided in the `mappings` object so yeah you'd want to list 2026.1 before 2025.3.

anyoung-tableau: We could define a way to compare two product versions and sort the keys but meh

PR #167, opened 2025-11-21
Pulse discover

jarhun88:

the pull request._

# Pull Request Template

## Motivation and Context

Today, the tableau mcp server mostly provides direct access to existing APIs on the tableau platform (e.g /search, /query-datasource, /get-workbooks, etc.). However, in the case of the Pulse discover experience, we have invested a lot of time and resources into building an intelligent experience. The goal of integrating the Pulse /discover experience into the Tableau MCP server is to:

Externalize the intelligence that we have implemented natively on the tableau platform
Provide a clearer value proposition for how the MCP server improves as customers leverage the capabilities on our premium SKUs

The intelligence in the discover service is that it has been instructed how to compare and analyze a group of metric insights that belong to a metric group. In addition, it provides relevant follow-up questions based on additional deterministic insights that can be gleaned from the metric group in scope. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- vibe tested
- sanity tool unit tests

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
+++ b/src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
@@ -2,6 +2,7 @@ import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 import { Err, Ok } from 'ts-results-es';
 
 import { Server } from '../../../server.js';
+import { Provider } from '../../../utils/provider.js';
 import { exportedForTesting as resourceAccessCheckerExportedForTesting } from '../../resourceAccessChecker.js';
 import { getGeneratePulseInsightBriefTool } from './generatePulseInsightBriefTool.js';
 
@@ -297,16 +298,7 @@ describe('getGeneratePulseInsightBriefTool', () => {
     });
     mocks.mockGeneratePulseInsightBrief.mockResolvedValue(new Ok(mockBriefResponse));
 
-    const tool = getGeneratePulseInsightBriefTool(new Server());
-    const result = await tool.callback(
-      { briefRequest: twoMetricRequest },
-      {
-        signal: new AbortController().signal,
-        requestId: 'test-request-id',
-        sendNotification: vi.fn(),
-        sendRequest: vi.fn(),
-      },
-    );
+    const result = await getToolResult(twoMetricRequest);
 
     // Should succeed
     expect(result.isError).toBe(false);
@@ -339,10 +331,13 @@ describe('getGeneratePulseInsightBriefTool', () => {
     );
   });
 
-  async function getToolResult(): Promise<CallToolResult> {
+  async function getToolResult(
+    overrideBriefRequest?: typeof briefRequest,
+  ): Promise<CallToolResult> {
     const tool = getGeneratePulseInsightBriefTool(new Server());
-    return await tool.callback(
-      { briefRequest },
+    const callback = await Provider.from(tool.callback);
+    return await callback(
+      { briefRequest: overrideBriefRequest ?? briefRequest },
       {
         signal: new AbortController().signal,
         requestId: 'test-request-id',


docs/docs/tools/pulse/generate-pulse-insight-brief.md
```diff
@@ -0,0 +1,279 @@
...
+---
+
+# Generate Pulse Insight Brief
+
+Generates AI-powered conversational insights for Pulse metrics based on natural language questions.
+This tool enables interactive, multi-turn conversations about your metrics data.
+
+## What is an Insight Brief?
+
+An **insight brief** is an AI-generated response to questions about Pulse metrics that provides:
```
anyoung-tableau, line 12: Out of curiosity, where did this info about insight briefs come from? Documentation?

src/sdks/tableau/methods/pulseMethods.ts
```diff
@@ -145,6 +147,18 @@ export default class PulseMethods extends AuthenticatedMethods<typeof pulseApis>
...
 
+  generatePulseInsightBrief = async (
+    briefRequest: z.infer<typeof pulseInsightBriefRequestSchema>,
+  ): Promise<PulseResult<PulseInsightBriefResponse>> => {
+    return await guardAgainstPulseDisabled(async () => {
+      const response = await this._apiClient.generatePulseInsightBrief(
+        briefRequest,
+        this.authHeader,
+      );
+      return response ?? {};
```
anyoung-tableau: nit: I don't think `response` can be null or undefined. Can remove the `??`

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      { content: 'What factors contributed to the increase?' },
+      { content: 'How does this compare to last year?' },
+    ],
+    group_context: briefRequest.messages[0].metric_group_context,
+    not_enough_information: false,
+  };
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    resetResourceAccessCheckerSingleton();
```
anyoung-tableau, line 129: I think this line may be unnecessary. I think the Pulse tools don't care about tool scoping and the resource access checker.

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    mocks.mockGetConfig.mockReturnValue({
+      disableMetadataApiRequests: false,
+      boundedContext: {
+        projectIds: null,
+        datasourceIds: null,
+        workbookIds: null,
+      },
+    });
```
anyoung-tableau, line 136: Same, I feel like having the mock config may be unnecessary

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export type GeneratePulseInsightBriefError =
+  | {
+      type: 'feature-disabled';
+      reason: PulseDisabledError;
+    }
+  | {
+      type: 'datasource-not-allowed';
+      message: string;
+    };
```
anyoung-tableau, line 28: The `datasource-not-allowed` result is impossible, I think. That only comes from the tool scoping feature which we aren't doing in this tool.

anyoung-tableau: I guess that begs the question though... if a datasource filter is defined, should we first check the input metric uses that datasource?

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
```diff
@@ -0,0 +1,242 @@
...
+            callback: async (restApi) =>
+              await restApi.pulseMethods.generatePulseInsightBrief(briefRequest),
+          });
+
+          if (result.isErr()) {
+            return new Err({
+              type: 'feature-disabled',
+              reason: result.error,
+            });
+          }
```
anyoung-tableau, line 243: If we end up not needing tool scoping, I think this `if` block can be removed and `getErrorText` can just be `getPulseDisabledError`, similar to https://github.com/tableau/tableau-mcp/blob/11347a20d6affc139f5932d1a9c90ddeb2a7417a/src/tools/pulse/listMetricsFromMetricIds/listPulseMetricsFromMetricIds.ts#L61

docs/docs/tools/pulse/generate-pulse-insight-brief.md
jarhun88: tbh i actually got cursor to generate it based on the pulse discover feature... and then i tuned the response a little bit until it seemed reasonable to me

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: from our convo at standup, i will be implementing the tool scoping for ds and so will keep this

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: i think we can resolve this based on our discussion outcome?

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
jarhun88: im not 100% sure what this is for but it doesnt seem necessary for the tests to pass so i removed it?

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
jarhun88: keeping this cuz it seems i need it for the scoping

docs/docs/tools/pulse/generate-pulse-insight-brief.md
anyoung-tableau: It did feel a little AI-ish. It seems reasonable to me too and I don't think it's PR-blocking, but consider starting a thread with Joe and Dan Platt with a link to the docs after they deploy for any feedback.

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
anyoung-tableau: With tool scoping you will likely need to add back `resetResourceAccessCheckerSingleton()` in case you ever added a 2nd test that tested scoping.

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
```diff
@@ -201,6 +201,29 @@ An insight brief is an AI-generated response to questions about Pulse metrics. I
         authInfo,
         args: { briefRequest },
         callback: async () => {
+          // Check datasource access for all metrics in all messages
+          const { datasourceIds } = config.boundedContext;
+          if (datasourceIds) {
+            for (const message of briefRequest.messages) {
+              if (message.metric_group_context) {
+                for (const metricContext of message.metric_group_context) {
```
anyoung-tableau: I see the tool description suggests that all metrics in the group context should be from the same datasource. However, if they aren't and there is at least one metric that *is* part of the bounded context, should we allow it rather than throwing an error because there are some metrics with a disallowed datasource?

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
jarhun88: hmm tests still pass without it, tho its not clear for me what this does

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: I was considering that, however i copied the logic based off of the insightGenerateBundle tool which has the same fail all logic. https://github.com/tableau/tableau-mcp/blob/main/src/tools/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts#L171

docs/docs/tools/pulse/generate-pulse-insight-brief.md
jarhun88: what is meant by "after they deploy?" deploy what?

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
anyoung-tableau: The resource access checker singleton caches the results of whether a datasource is allowed or not. Resetting it clears the cache

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
anyoung-tableau: I think the difference is that that tool only accepts a single metric instead of a group of them.

anyoung-tableau: I would vote for only throwing an error if _all_ metrics use a disallowed datasource. Might need Joe's input.

docs/docs/tools/pulse/generate-pulse-insight-brief.md
anyoung-tableau: After the docs deploy. They autodeploy to the docs site with every PR merge.

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: oh yea you are right

jarhun88: so just filter out the "bad" ones

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
```diff
@@ -201,25 +201,15 @@ An insight brief is an AI-generated response to questions about Pulse metrics. I
...
-                      type: 'datasource-not-allowed',
-                      message: [
-                        'The set of allowed metric insights that can be queried is limited by the server configuration.',
-                        'Generating the Pulse Insight Brief is not allowed because one or more metrics are derived',
-                        `from the data source with LUID ${datasourceLuid}, which is not in the allowed set of data sources.`,
-                      ].join(' '),
-                    });
-                  }
-                }
+                message.metric_group_context = message.metric_group_context.filter(
```
anyoung-tableau, line 209: If all metrics get filtered out of the group what is the experience like? Should we detect this and return a user/LLM-friendly error? Or just remove the entire message from the array of messages? What if the array of messages becomes empty? :)

src/sdks/tableau/methods/pulseMethods.ts
```diff
@@ -145,6 +147,18 @@ export default class PulseMethods extends AuthenticatedMethods<typeof pulseApis>
     });
   };
 
+  generatePulseInsightBrief = async (
```
anyoung-tableau, line 158: nit: please document this method similar to the other ones in this file

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: ah yea good point, mm i think it kind of makes sense to error if the request becomes completely empty after filter. Perhaps similar to the error we had before

jarhun88: ok how about this logic. 

So each request essentially requires the entire conversational history (structured as a list of messages) to be passed as a param. 

Each message should contain the list of metrics that we are discovering on, which ideally should be all the same ds, but not necessarily guaranteed.

Typically once a new discover conversation is started, youre going to stick with the same metric context throughout the entire conversation, thus if you validate the metric ds on the first request, you probably dont have to do it again.

But to be safe Im going to have it check thru all metrics for each message. If any message ends up with 0 valid metrics, i will throw an error.

How does that sound?

anyoung-tableau: I like it, thanks for wiring this up!

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
```diff
@@ -0,0 +1,266 @@
...
+        args: { briefRequest },
+        callback: async () => {
+          // Filter out metrics that are not in the allowed datasource set
+          const { datasourceIds } = config.boundedContext;
+          if (datasourceIds) {
+            for (const message of briefRequest.messages) {
+              if (message.metric_group_context) {
+                message.metric_group_context = message.metric_group_context.filter(
+                  (metricContext) =>
+                    datasourceIds.has(metricContext.metric.definition.datasource.id),
```
anyoung-tableau, line 211: Non-blocking, but I noticed that Pulse tool scoping checks the `datasourceIds` from the `boundedContext` directly, instead of using `resourceAccessChecker.isDatasourceAllowed()` which _also_ checks if the datasource exists in an allowed project if a project filter is defined. I don't think this was intentional.

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.test.ts
jarhun88: hmm ok i have added it back

src/sdks/tableau/methods/pulseMethods.ts
jarhun88: done

src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
jarhun88: ah i copied this logic from generatePulseMetricValueInsightBundleTool, are you suggesting that we should be using resourceAccessChecker.isDatasourceAllowed() to leverage the additional project filter ?

anyoung-tableau: Actually, I think I was wrong about it not being intentional. The [PRD](https://docs.google.com/document/d/1sEw1qnoCOBTxykTZQWwAYcwcgZoO5CvgMejpxMAro-o/edit?tab=t.0#heading=h.6ppnte3pl4i) does say the project filter has no impact on the Pulse tools. That said, I don't see any good reason why we _couldn't_ use `resourceAccessChecker.isDatasourceAllowed()` to honor the project filters too. But again, this is non-blocking and doesn't need to be considered for this PR.

jarhun88: ok i will not consider it for now

PR #177, opened 2025-12-09
Upload NodeJS single executable applications to each release

anyoung-tableau:
1. NodeJS [single executable applications](https://nodejs.org/api/single-executable-applications.html) (SEA) currently only support [CommonJS](https://nodejs.org/api/modules.html#modules-commonjs-modules) module system so this PR changes the `type` field of `package.json` to `"CommonJS"` so Node.js will treat the build output as a CommonJS module, and the `esbuild` format to `cjs`  so the build output uses CommonJS syntax.
2. With each release, a GitHub action will generate a Linux and Windows SEA and upload them to the release assets.
3. Every week, another GitHub action will delete all assets (including the MCPB) on releases older than 10 releases ago.

With these assets, 1st and 3rd parties can easily get up and going. Just download and extract the asset, create a `.env` file, and run the executable. This has importance for unblocking internal feature work too.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,7 +1,7 @@
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { InitializeRequest, SetLevelRequestSchema } from '@modelcontextprotocol/sdk/types.js';
 
-import pkg from '../package.json' with { type: 'json' };
+import pkg from '../package.json';
 import { getConfig } from './config.js';
 import { setLogLevel } from './logging/log.js';
 import { TableauAuthInfo } from './server/oauth/schemas.js';
--- a/src/scripts/createClaudeMcpBundleManifest.ts
+++ b/src/scripts/createClaudeMcpBundleManifest.ts
@@ -7,10 +7,11 @@ import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
 import { z } from 'zod';
 
-import packageJson from '../../package.json' with { type: 'json' };
+import packageJson from '../../package.json';
 import { ProcessEnvEx } from '../../types/process-env.js';
 import { toolNames } from '../tools/toolName.js';
 
+// @ts-expect-error - import.meta is not allowed in CommonJS output, but this file is built using esbuild as ESM
 const __filename = fileURLToPath(import.meta.url);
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -53,9 +53,7 @@ async function startServer(): Promise<void> {
...
 }
 
-try {
-  await startServer();
-} catch (error) {
+startServer().catch((error) => {
   writeToStderr(`Fatal error when starting the server: ${getExceptionMessage(error)}`);
   process.exit(1);
-}
+});
```
anyoung-tableau, line 59: No top-level await in CommonJS

src/config.ts
```diff
@@ -305,8 +305,10 @@ export class Config {
 }
 
 function validateServer(server: string): void {
-  if (!server.startsWith('https://')) {
-    throw new Error(`The environment variable SERVER must start with "https://": ${server}`);
+  if (!['https://', 'http://'].find((prefix) => server.startsWith(prefix))) {
```
anyoung-tableau, line 308: Driveby fix here so we can run Tableau MCP against an http Tableau Server

src/server.ts
```diff
@@ -1,7 +1,7 @@
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { InitializeRequest, SetLevelRequestSchema } from '@modelcontextprotocol/sdk/types.js';
 
-import pkg from '../package.json' with { type: 'json' };
```
anyoung-tableau, line 4: Import attributes (the `with` statement) are an ESM-only feature but aren't required.

.github/workflows/cleanup-releases.yml
```diff
@@ -0,0 +1,39 @@
...
+    - cron: '0 18 * * 1' # Weekly on Monday 18:00 UTC
+  workflow_dispatch:
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v8
+        with:
+          script: |
```
anyoung-tableau, line 13: I haven't fully tested these GitHub actions so there might be a follow up PR once they execute.

stephendeoca (approved):

Issue #176, opened 2025-12-08
REST API request timeouts should be configurable


anyoung-tableau:
Currently, for the requests TMCP makes to the Tableau REST APIs, there is no timeout policy. According to the spec: [implementations SHOULD establish timeouts for all sent requests](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#timeouts:~:text=Implementations%20SHOULD%20establish%20timeouts%20for%20all%20sent%20requests).

There is a [community PR](https://github.com/tableau/tableau-mcp/pull/139) to add support for configurable timeouts for these requests, but it has grown stale and only makes the request timeouts configurable at the server level, not the request level. MCPs clients should be able to [control the timeout](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L127) of the MCP requests themselves.

PR #181, opened 2025-12-11
@W-20600683: Add configurable max timeout for REST APIs, and handle request cancellation notifications

anyoung-tableau:
Addresses https://github.com/tableau/tableau-mcp/issues/176

Before these changes, we had two issues:

1. All requests to Tableau APIs could potentially never timeout.
2. We were doing nothing when an MCP client sends a [cancel notification](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation), instead of cancelling in any-progress requests.

With these changes:
1. Requests will timeout based off the value of the `MAX_REQUEST_TIMEOUT_MS` env var.
2. Cancel notifications from MCP clients are now honored, cancelling any in-progress API calls. See [protocol.ts#L367](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/protocol.ts#L367) which triggers the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for the request whose signal is plumbed down into Axios. (Note that the SDK has changed slightly and looks a tiny bit different compared to the version we are using, but the concept is the same).

I tested this by adding a 60 second delay in the execution of one of the REST APIs and set the MCP Inspector timeout to 10 seconds. After 10 seconds, MCP Inspector sends the cancel notification to the server, the MCP SDK handles it and triggers the AbortController that's been passed in the AxiosConfig for each Zodios instance.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.13.8",
+  "version": "1.13.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.13.8",
+      "version": "1.13.9",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.13.8",
+  "version": "1.13.9",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/restApiInstance.ts
```diff
@@ -35,12 +35,27 @@ const getNewRestApiInstanceAsync = async (
...
+    log.info(
+      server,
+      {
+        type: 'request-cancelled',
+        requestId,
+        reason: signal.reason,
+      },
+      { logger: server.name, requestId },
+    );
+  };
```
anyoung-tableau: This isn't necessary, but it does send a signal back to the client that says "yep, we got your request to cancel any pending requests and we cancelled them!"

eslint.config.mjs
```diff
@@ -16,6 +16,7 @@ export default [
     rules: {
       'no-console': ['error', { allow: ['warn', 'error'] }],
       'no-duplicate-imports': ['error', { includeExports: true }],
+      'no-trailing-spaces': 'error',
```
anyoung-tableau, line 19: Driveby fix here since we had a few places with trailing spaces which my IDE removed when autoformatting.

src/restApiInstance.ts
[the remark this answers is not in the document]
anyoung-tableau: Hmm, yeah, you're probably right.

src/restApiInstance.ts
```diff
@@ -35,12 +35,31 @@ const getNewRestApiInstanceAsync = async (
...
+        {
+          type: 'request-cancelled',
+          requestId,
+          reason: signal.reason,
+        },
+        { logger: server.name, requestId },
+      );
+    },
+    { once: true },
+  );
```
anyoung-tableau, line 55: This isn't necessary, but it does send a signal back to the client that says "yep, we got your request to cancel any pending requests and we cancelled them!"

src/server/oauth/callback.ts
```diff
@@ -92,7 +92,10 @@ export function callback(
       }
 
       const server = originHostUrl.toString();
-      const restApi = new RestApi(server);
+      const restApi = new RestApi(server, {
+        maxRequestTimeoutMs: config.maxRequestTimeoutMs,
```
stephendeoca, line 96: We don't use an AbortController for OAuth?

anyoung-tableau: Good question, but no, there wouldn't be any way for an MCP client to ask for a request cancellation. They're still doing the OAuth dance so they haven't yet established a connection to the MCP server.

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -173,6 +173,15 @@ be available.
 
 <hr />
 
+## `MAX_REQUEST_TIMEOUT_MS`
+
+The maximum timeout for requests to the Tableau Server REST API.
+
+- Default: `600000` (10 minutes)
```
jarhun88, line 180: just curious where did 10min come from? seems like a pretty generous amount of time

anyoung-tableau: I chose a generous value intentionally since some REST API calls can take a long time and I didn't want to accidentally make it too small. If admins want a shorter value they can shorten it.

PR #182, opened 2025-12-12
Add PowerShell script to download, install, and manage Tableau MCP single application executable

anyoung-tableau:
These changes add a PowerShell script to help admins install Tableau MCP without needing to clone the repo or even install NodeJS.

image

image

image

image

image

image

### DIFF
--- a/scripts/Manage-Server.ps1
+++ b/scripts/Manage-Server.ps1
@@ -439,12 +439,14 @@ function Get-ServerStatus {
     Write-Host "Body: $($body | ConvertTo-Json -Compress)" -ForegroundColor Magenta
     Write-Host ""
 
+    [System.Net.ServicePointManager]::ServerCertificateValidationCallback = { $true }
     $response = Invoke-WebRequest -Uri $uri `
       -Method Post `
       -Body ($body | ConvertTo-Json -Compress) `
       -ContentType "application/json" `
       -TimeoutSec 5 `
       -UseBasicParsing
+    [System.Net.ServicePointManager]::ServerCertificateValidationCallback = $null
 
   }
   catch {


src/server/express.ts
```diff
@@ -52,7 +52,7 @@ export async function startExpressServer({
     app.set('trust proxy', config.trustProxyConfig);
   }
 
-  const middleware: Array<RequestHandler> = [];
+  const middleware: Array<RequestHandler> = [validatePingRequest];
```
anyoung-tableau: `ping` requests shouldn't require OAuth and the health check uses the ping method.

anyoung-tableau: I'll also be deleting the SEA assets from the 1.13.4 release since they won't include these changes and the health check won't work.

docs/docs/getting-started.md
```diff
@@ -34,15 +34,69 @@ This standard config works in most MCP clients:
 
 ## Node.js Single Executable Applications
 
-Node.js [Single Executable Applications](https://nodejs.org/api/single-executable-applications.html) (SEA) are a way to package a Node.js application into a single executable file.
+Node.js [Single Executable Applications](https://nodejs.org/api/single-executable-applications.html)
```
anyoung-tableau, line 37: image

scripts/Manage-Server.ps1
```diff
@@ -0,0 +1,491 @@
...
+  3. Start the MCP server.
+  4. Check the status of the MCP server.
+  5. Upgrade the MCP server.
+  6. Stop the MCP server.
+
+  .INPUTS
+  None. You can't pipe objects to Manage-Server.ps1.
+
+  .OUTPUTS
+  None. Manage-Server.ps1 doesn't generate any output.
```
stephendeoca, line 21: We don't do any console logging once the server is up?

scripts/Manage-Server.ps1
```diff
@@ -0,0 +1,491 @@
...
+"@
+      }
+    }
+    @{
+      label  = "OAuth (Tableau Server 2025.3+ only)"
+      action = {
+        $oauthIssuer = Read-Host "OAuth Issuer"
+        $oauthRedirectUri = Read-Host "OAuth Redirect URI ([ENTER] to use default)"
+        $oauthJwePrivateKey = Read-Host "OAuth JWE Private Key ([ENTER] to provide path instead)"
+        $oauthJwePrivateKeyPath = Read-Host "OAuth JWE Private Key Path"
```
stephendeoca: If someone provides a private key, the they'll be prompted for the private key path after right? So they should press enter in that case right?

scripts/Manage-Server.ps1
```diff
@@ -0,0 +1,491 @@
...
+        $oauthJwePrivateKeyPassphrase = Read-Host "OAuth JWE Private Key Passphrase ([ENTER] to leave blank)"
+        $oauthAuthzCodeTimeoutMs = Read-Host "OAuth Authz Code Timeout MS ([ENTER] to use default)"
+        $oauthAccessTokenTimeoutMs = Read-Host "OAuth Access Token Timeout MS ([ENTER] to use default)"
+        $oauthRefreshTokenTimeoutMs = Read-Host "OAuth Refresh Token Timeout MS ([ENTER] to use default)"
```
stephendeoca, line 194: Do we supply a link to docs if the user wants to better understand the default values for these variable or is unsure of what they do?

scripts/Manage-Server.ps1
anyoung-tableau: `OUTPUTS` means return values, not console logs.

scripts/Manage-Server.ps1
anyoung-tableau: Yeah

scripts/Manage-Server.ps1
anyoung-tableau: If someone is using this script I'm assuming they got to it from the docs themselves and they're only a click away from the docs that describe these environment variables. I don't think we need to supply a link.

scripts/Manage-Server.ps1
anyoung-tableau: https://learn.microsoft.com/en-us/powershell/scripting/developer/help/comment-based-help-keywords?view=powershell-7.5#outputs

scripts/Manage-Server.ps1
stephendeoca: consider adding to the comment "([ENTER] if you already provided a OAuth JWE Private Key)"

stephendeoca (approved):

scripts/Manage-Server.ps1
anyoung-tableau: I'll do one better and bypass when the key is already provided

PR #184, opened 2025-12-16
@W-20600734: fix openai json schema error

jarhun88:

the pull request._

# Pull Request Template

## Description

This error was happening when using mcp with chatgpt

`Error creating connector
Invalid schema for function 'generate-pulse-insight-brief': In context=('properties', 'briefRequest', 'properties', 'messages', 'items', 'properties', 'metric_group_context', 'items', 'properties', 'metric', 'properties', 'candidates'), array schema missing items.`

This is the findings from investigation:

Why Strict Clients Require items
JSON Schema technically allows { "type": "array" } without items. But:
LLMs need to know what type of elements to generate
Without items, the schema is ambiguous
Strict clients (OpenAI, Anthropic, Tableau) enforce this for reliability

The Fix
Replace z.any() with something that produces a real type. In this case I updated it to be what is expected by the API. https://sourcegraph.prod.tableautools.com/ntbue/ntbue-protobuf-definitions@1f51f2147c81a1fc8ae8e437d225a4e566fe015e/-/blob/protos/ntbueinsightsservice/tableau/pulse/insightsservice/v1/insightsservice_messages.proto?L48

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
+++ b/src/tools/pulse/generateInsightBrief/generatePulseInsightBriefTool.ts
@@ -104,6 +104,7 @@ An insight brief is an AI-generated response to questions about Pulse metrics. I
               metric: {
                 definition: { /* metric definition */ },
                 specification: { /* metric specification */ },
+                candidates: [ /* optional array of candidate definitions */ ],
               },
             }
           ],


src/sdks/tableau/types/pulse.ts
```diff
@@ -308,7 +316,7 @@ export const metricGroupContextSchema = z.array(
           metric_goals: pulseGoalsSchema.optional(),
         })
         .optional(),
-      candidates: z.array(z.any()).optional(),
+      candidates: z.array(pulseCorrelationCandidateDefinitionSchema),
```
anyoung-tableau, line 319: Now that we've marked `candidates` as required, we should also update the example request on the docs page for the tool, even if an empty array is acceptable.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.13.7",
+  "version": "1.13.8",
```
anyoung-tableau, line 4: While we're touching this tool, would you mind also adding it to the list of tools on https://github.com/tableau/tableau-mcp/blob/main/docs/docs/intro.md ?

jarhun88: done

src/sdks/tableau/types/pulse.ts
jarhun88: hmm i made a 1 line update to the desc, is this suitable?

PR #190, opened 2025-12-24
@W-20481777 Setup Generic Telemetry interface

jarhun88:

the pull request._

# Pull Request Template

## Description

Creating a generic interface for plugging in the salesforce telemetry agent OR a custom telemetry provider. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/telemetry/types.ts
+++ b/src/telemetry/types.ts
@@ -27,19 +27,27 @@ export interface TelemetryProvider {
 }
 
 /**
- * Attributes that can be attached to telemetry data.
+ * Schema for telemetry attributes.
  * Values can be strings, numbers, booleans, or undefined.
  */
-export interface TelemetryAttributes {
-  [key: string]: string | number | boolean | undefined;
-}
+export const telemetryAttributesSchema = z.record(
+  z.string(),
+  z.union([z.string(), z.number(), z.boolean(), z.undefined()]),
+);
+export type TelemetryAttributes = z.infer<typeof telemetryAttributesSchema>;
 
 /**
- * Configuration for noop provider (no telemetry)
+ * Valid telemetry provider names
  */
-interface NoopTelemetryConfig {
-  provider: 'noop';
-}
+export const telemetryProviderSchema = z.enum(['noop', 'custom']);
+export type TelemetryProviderType = z.infer<typeof telemetryProviderSchema>;
+
+/**
+ * Schema for noop telemetry config (no telemetry)
+ */
+export const noopTelemetryConfigSchema = z.object({
+  provider: z.literal('noop'),
+});
 
 /**
  * Schema for custom telemetry provider config.
@@ -52,34 +60,36 @@ export const providerConfigSchema = z
   .passthrough();
 
 /**
- * Configuration for custom provider with required providerConfig
+ * Schema for custom telemetry config
+ *
+ * @example
+ * ```json
+ * {
+ *   "provider": "custom",
+ *   "providerConfig": {
+ *     "module": "./my-otel-provider.js"
+ *   }
+ * }
+ * ```
  */
-interface CustomTelemetryConfig {
-  provider: 'custom';
-  /**
-   * Configuration for the custom provider.
-   *
-   * @example
-   * ```json
-   * {
-   *   "module": "./my-otel-provider.js"
-   * }
-   * ```
-   */
-  providerConfig: z.infer<typeof providerConfigSchema>;
-}
-
-export type TelemetryConfig = NoopTelemetryConfig | CustomTelemetryConfig;
+export const customTelemetryConfigSchema = z.object({
+  provider: z.literal('custom'),
+  providerConfig: providerConfigSchema,
+});
 
 /**
- * Valid telemetry provider names
+ * Combined telemetry config schema (discriminated union)
  */
-const telemetryProviders = ['noop', 'custom'] as const;
-type TelemetryProviderType = (typeof telemetryProviders)[number];
+export const telemetryConfigSchema = z.discriminatedUnion('provider', [
+  noopTelemetryConfigSchema,
+  customTelemetryConfigSchema,
+]);
+
+export type TelemetryConfig = z.infer<typeof telemetryConfigSchema>;
 
 /**
  * Type guard for telemetry provider names
  */
 export function isTelemetryProvider(provider: unknown): provider is TelemetryProviderType {
-  return telemetryProviders.some((p) => p === provider);
+  return telemetryProviderSchema.safeParse(provider).success;
 }


src/telemetry/init.ts
```diff
@@ -0,0 +1,156 @@
...
+    provider.initialize();
+    return provider;
+  }
+
+  let provider: TelemetryProvider;
+
+  try {
+    // Select provider based on configuration
+    switch (config.telemetry.provider) {
+      case 'moncloud':
```
jarhun88: i think having separate cases for moncloud vs custom will become even more apparent when i need to write custom metrics for the tools using telemetry.addAttributes(), as each underlying telemetry provider has different apis for doing that

src/telemetry/types.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Values can be strings, numbers, booleans, or undefined.
+ */
+export interface TelemetryAttributes {
+  [key: string]: string | number | boolean | undefined;
+}
+
+/**
+ * Configuration for telemetry providers
+ */
+export interface TelemetryConfig {
```
anyoung-tableau: Consider enforcing the requirement at the type level that `providerConfig` must be provided when `provider === "custom"` with something like:

```ts
export type TelemetryConfig = {
  enabled: boolean;
} & (
  | {
      provider: "noop" | "moncloud";
    }
  | {
      provider: "custom";
      providerConfig: Record<string, unknown>;
    }
);
```

Image

src/config.ts
```diff
@@ -223,6 +228,12 @@ export class Config {
         : null,
     };
 
+    this.telemetry = {
+      enabled: telemetryEnabled === 'true',
+      provider: (telemetryProvider as 'noop' | 'moncloud' | 'custom') || 'noop',
```
anyoung-tableau: Consider using a type guard for the telemtryProvider, similar to how `this.auth` is set below. Currently the runtime value could be anything.

Something like:

```ts
const telemetryProviders = ["noop", "moncloud", "custom"] as const;
type TelemetryProvider = (typeof telemetryProviders)[number];

export function isTelemetryProvider(value: unknown): value is TelemetryProvider {
    return !!(telemetryProviders.find((t) => t === value));
}

...
provider: isTelemetryProvider(telemetryProvider) ? telemetryProvider : 'noop',
```

src/telemetry/init.ts
```diff
@@ -0,0 +1,156 @@
...
+        break;
+
+      case 'custom':
+        // Load custom provider from user's filesystem
+        provider = loadCustomProvider(config.telemetry.providerConfig);
+        break;
+
+      case 'noop':
+      default:
+        if (config.telemetry.provider !== 'noop') {
```
anyoung-tableau: The type of `config.telemetry.provider` and the return type on this method prevent this `default` case from ever happening. If you remove the `default` case and, say, add a new provider type without adding the new `case` for the new provider, TypeScript will yell at you because the method could potentially now return `undefined`

src/telemetry/init.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    if (!ProviderClass) {
+      throw new Error(
+        `Module ${modulePath} must export a default class or named export "TelemetryProvider" ` +
+          'that implements the TelemetryProvider interface',
+      );
+    }
+
+    // Instantiate the provider with the full config
+    return new ProviderClass(config);
```
anyoung-tableau: Before returning the instance, let's verify the class implements the interface. I did something similar here: https://github.com/tableau/tableau-mcp/blob/anyoung/session-store/src/server/storage/storeFactory.ts#L157

src/telemetry/moncloud.ts
```diff
@@ -0,0 +1,49 @@
...
+ * - SFDC_TRACES_ENDPOINT: Traces endpoint URL
+ * - SFDC_EVENTS_ENDPOINT: Events endpoint URL
+ *
+ * See: https://git.soma.salesforce.com/monitoring/salesforce-apmagent
+ */
+
+import { TelemetryAttributes, TelemetryProvider } from './types.js';
+
+export class MonCloudTelemetryProvider implements TelemetryProvider {
+  private trace: any;
```
anyoung-tableau: What sets `trace`? And can it have a better type?

src/telemetry/types.ts
jarhun88: ive updated based on your input

src/telemetry/init.ts
```diff
@@ -0,0 +1,163 @@
...
+import { getConfig } from '../config.js';
+import { MonCloudTelemetryProvider } from './moncloud.js';
+import { NoOpTelemetryProvider } from './noop.js';
+import { TelemetryProvider } from './types.js';
+
+// Use global to share provider across bundles (tracing.js and index.js)
+declare global {
+  // eslint-disable-next-line no-var
+  var __telemetryProvider: TelemetryProvider | undefined;
+}
```
anyoung-tableau, line 45: Any reason we need to modify the global scope? Could `__telemetryProvider` be a non-exported singleton here with a get and set function?

```ts
let telemetryProvider: TelemetryProvider | undefined = undefined;

export function getTelemetryProvider(): TelemetryProvider {
  if (!telemetryProvider) {
    telemetryProvider = new NoOpTelemetryProvider();
  }

  return telemetryProvider;
}

export function setTelemetryProvider(provider: TelemetryProvider): void {
  telemetryProvider = provider;
}

File: src/config.ts (hunk 3 above)
Comment (jarhun88) [reply src/config.ts]: ive added the update

File: src/telemetry/moncloud.ts (hunk 6 above)
Comment (jarhun88) [reply src/telemetry/moncloud.ts]: ive decided to remove trace, but i took your advice to consideration and remove usage of any

File: src/tools/tool.ts [hunk 8]
```diff
@@ -173,6 +174,14 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
 
     this.logInvocation({ requestId, args, username });
 
+    // Record custom metric for this tool call
+    const telemetry = getTelemetryProvider();
+    telemetry.recordMetric('mcp.tool.calls', 1, {
+      'mcp.tool.name': this.name,
+      'mcp.request.id': String(requestId),
+      ...(username && { 'mcp.user.name': username }),
```
anyoung-tableau: Out of curiosity, is this format for the attribute names a moncloud or OpenTelemetry thing? We already know the metric is for `mcp.tool.calls`, do the attribute names need the `mcp.*` prefix too?

src/tools/tool.ts
```diff
@@ -173,6 +174,14 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
 
     this.logInvocation({ requestId, args, username });
 
+    // Record custom metric for this tool call
+    const telemetry = getTelemetryProvider();
+    telemetry.recordMetric('mcp.tool.calls', 1, {
+      'mcp.tool.name': this.name,
+      'mcp.request.id': String(requestId),
```
anyoung-tableau: nit: using the `String` ctor is not very common anymore. Consider using `requestId.toString()` or `` `${requestId}` ``

src/telemetry/init.ts
jarhun88: ah yea makes sense, ive removed the default

src/config.ts
```diff
@@ -223,6 +228,23 @@ export class Config {
...
 
+    const parsedProvider = (telemetryProvider as 'noop' | 'moncloud' | 'custom') || 'noop';
+    if (parsedProvider === 'custom') {
+      if (!telemetryProviderConfig) {
+        throw new Error('TELEMETRY_PROVIDER_CONFIG is required when TELEMETRY_PROVIDER is "custom"');
+      }
+      this.telemetry = {
+        enabled: telemetryEnabled === 'true',
+        provider: 'custom',
+        providerConfig: JSON.parse(telemetryProviderConfig),
```
anyoung-tableau: JSON.parse doesn't provide any type safety here. Consider defining a zod schema and using it to parse the input. You can even replace your interfaces and infer the type using zod!

Something like:

```ts
const telemetryConfigBaseSchema = z.object({
  enabled: z.boolean(),
});

const noopTelemetryConfigSchema = telemetryConfigBaseSchema.extend({
  provider: z.literal('noop'),
});

const monCloudTelemetryConfigSchema = telemetryConfigBaseSchema.extend({
  provider: z.literal('moncloud'),
});

const customTelemetryConfigSchema = telemetryConfigBaseSchema.extend({
  provider: z.literal('custom'),
  providerConfig: z.record(z.string(), z.unknown()),
});

export const telemetryConfigSchema = z.discriminatedUnion('provider', [
  noopTelemetryConfigSchema,
  monCloudTelemetryConfigSchema,
  customTelemetryConfigSchema,
]);

export type TelemetryConfig = z.infer<typeof telemetryConfigSchema>;
```

```ts
  providerConfig: telemetryConfigSchema.parse(telemetryProviderConfig),
```

src/telemetry/init.ts
jarhun88: updated with your example

src/telemetry/init.ts
jarhun88: from my understanding, `node -r ./build/telemetry/tracing.ts build/index.ts` builds 2 separate bundles, hence both bundles would get their own copy of that code. Since tracing.ts is the one that actually calls initializeTelemetry(), if i want to add customMetrics to the index.js (main), I need to utilize the telemetryProvider that was initialized by tracing.ts.

As to why its done like this... well its just the way suggested by moncloud docs, but also i believe the agent needs to load before the http server else the auto instrumentation will not succeed. (whether or not we actually need auto instrumentation is a question for later down the line... for now i think its good to have our options open?)

src/config.ts
jarhun88: i made a zod for the providerConfig, but do i need one for the other telemetryConfigSchemas?

src/tools/tool.ts
jarhun88: good catch, we dont need to do this, in fact even the username is likely unnecessary. I have simplified this call

src/telemetry/moncloud.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * MonCloud telemetry provider for Salesforce's internal monitoring platform.
+ */
+
+import { TelemetryAttributes, TelemetryProvider } from './types.js';
+import { Apm } from '@salesforce/apmagent';
```
jarhun88: so im not going to add the salesforce monitoring agent here in the package.json, still is this safe to reference?

src/scripts/build.ts
```diff
@@ -32,5 +32,16 @@ const dev = process.argv.includes('--dev');
     console.log(`⚠️ ${warning.text}`);
   }
 
+  console.log('🏗️ Building telemetry/tracing.js...');
+  await mkdir('./build/telemetry', { recursive: true });
+  await build({
```
anyoung-tableau: Let's include the `minify` and `sourcemap` properties too. Also, consider logging out errors and warnings like we do for the main build.

src/telemetry/init.test.ts
```diff
@@ -0,0 +1,100 @@
+import { beforeEach, describe, expect, it, Mock, vi } from 'vitest';
```
anyoung-tableau: Because `tsconfig.json` has `"types": ["vitest/globals"]` these imports should be unnecessary. Is that the case?

src/telemetry/init.ts
anyoung-tableau: Cool, thanks, that makes sense. A few notes just for posterity since I wasn't familiar with Node's `-r`/`--require` module argument:

- The `node -r` command preloads `tracing.js` before starting the `index.js` main script.
- There is still a single Node.js process and `tracing.js` updates the global scope that `index.js` can then read later on.

src/telemetry/moncloud.ts
anyoung-tableau: No, we'll need to delete this file and include it only in the Hyperforce repo and load it as a `custom` provider.

anyoung-tableau: Either that, or you could try dynamically importing the dependency at runtime with `import('@salesforce/apmagent')` (without type safety)

jarhun88: yea i was wondering if the hyperforce repo makes more sense with custom... so maybe ill just keep this branch as reference for now and make a new branch without the moncloud option?

src/telemetry/init.test.ts
jarhun88: yes, i also just refactored the test file to align more with other tests

src/telemetry/init.ts
```diff
@@ -0,0 +1,194 @@
...
+ * @example
+ * function main() {
+ *   // Initialize telemetry first
+ *   const telemetry = initializeTelemetry();
+ *
+ *   // Start application...
+ * }
+ */
+export function initializeTelemetry(): TelemetryProvider {
+  const config = getConfig();
```
anyoung-tableau, line 78: I think we should make an effort to limit the cross-script dependencies between the telemetry script and the main script. What do you think about adding a `telemetryConfig` parameter to `initializeTelemetry()` and the main script can pass it in instead of having to import `getConfig` from `config.ts`?

src/scripts/build.ts
jarhun88: done

src/telemetry/init.ts
jarhun88: i can see some benefits with testing, but only the telemetry script (or bundle) calls initializeTelemetry, so wouldnt there still be both main and telemetry needing to use getConfig?

src/telemetry/init.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+function validateTelemetryProvider(provider: unknown): asserts provider is TelemetryProvider {
+  if (typeof provider !== 'object' || provider === null) {
+    throw new Error('Provider must be an object');
+  }
+
+  // Use NoOpTelemetryProvider as the reference implementation
+  const requiredMethods = getInstanceMethods(NoOpTelemetryProvider);
+  const missingMethods = requiredMethods.filter(
+    (method) => typeof (provider as Record<string, unknown>)[method] !== 'function',
```
anyoung-tableau: Nit: if you want to get rid of the type assertion (I generally try to avoid them), you could add a basic type guard like this:

```ts
function isRecord(obj: unknown): obj is Record<string, unknown> {
  return typeof obj === 'object' && obj !== null && !Array.isArray(obj);
}

...

  if (!isRecord(provider)) {
    throw new Error('Provider must be an object');
  }

...

const missingMethods = requiredMethods.filter((method) => typeof provider[method] !== 'function');
```

src/telemetry/init.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    // Validate the provider implements TelemetryProvider interface
+    validateTelemetryProvider(provider);
+
+    return provider;
+  } catch (error) {
+    // Provide helpful error message with common issues
+    let errorMessage = `Failed to load custom telemetry provider from "${modulePath}". `;
+
+    if ((error as any).code === 'MODULE_NOT_FOUND') {
```
anyoung-tableau: Consider improving this condition to eliminate the `as any`. The type of `error` when NodeJS includes the `code` property is `NodeJS.ErrnoException` but the TypeScript types are underwhelming. [This document](ErrnoException) talks about the problem in depth but something basic like would be sufficient:

```ts
if (error instanceof Error && 'code' in error && error.code === 'MODULE_NOT_FOUND') {
```

src/config.ts
anyoung-tableau: You don't, my thinking was to use one or the other, but not both concepts. It's fine as it is and up to you if you want to make any changes. What's nice about Zod is you can define the schemas and then infer the type from them. No type construction needed.

Also, in TypeScript it's generally favorable to use `type`s over `interface`s unless you need [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces).

src/telemetry/moncloud.ts
anyoung-tableau: Yeah, that works. Or just delete the file from this branch to avoid another PR and pull it from git history before copying it to HF repo.

src/telemetry/init.ts
anyoung-tableau: Oh sorry, nevermind. I had it in my head that main called `initializeTelemetry`

src/telemetry/init.ts
```diff
@@ -0,0 +1,192 @@
...
+ */
+function loadCustomProvider(config?: Record<string, unknown>): TelemetryProvider {
+  if (!config?.module) {
+    throw new Error(
+      'Custom telemetry provider requires "module" in providerConfig. ' +
+        'Example: TELEMETRY_PROVIDER_CONFIG=\'{"module":"./my-telemetry.js"}\'',
+    );
+  }
+
+  const modulePath = config.module as string;
```
anyoung-tableau: You could avoid the type assertion with a quick check:

```ts
if (typeof config.module !== 'string') {
  throw new Error(...);
}
```

src/telemetry/init.ts
jarhun88: sounds good, that doc is 404 btw

src/telemetry/init.ts
jarhun88: updated

src/telemetry/init.ts
jarhun88: ok sounds good thanks for bringing up the type assertion stuff

src/telemetry/moncloud.ts
jarhun88: ok ive removed moncloud, i pushed it to the branch origin/telemetry_moncloud_example

src/config.ts
jarhun88: i have determined to zoddify everything just to match the existing pattern

PR #194, opened 2026-01-06
@W-20785682: Trigger E2E tests manually for fork PRs

anyoung-tableau:
https://github.com/tableau/tableau-mcp/issues/172

Issue #172, opened 2025-12-01
E2E tests fail in PRs from forks and dependabot


anyoung-tableau:
Presumably PRs created by dependabot cannot read the secrets so the environment variables are empty at runtime.

Example:

https://github.com/tableau/tableau-mcp/actions/runs/19810326451/job/56751377665#step:13:353

PR #195, opened 2026-01-06
@W-20785682: Run E2E tests after PR merge, disable E2E tests for fork PRs

anyoung-tableau:
Addresses https://github.com/tableau/tableau-mcp/issues/172: E2E tests fail in PRs from forks and dependabot

These changes:

1. Refactor the E2E and OAuth test steps into reusable, composite workflows.
2. Triggers the E2E and OAuth tests after a PR merges to main.

I initially tried adding a PR comment workflow (https://github.com/tableau/tableau-mcp/pull/194) but it quickly grew complex and somewhat scary from a security standpoint. This approach is much simpler.

### DIFF
--- a/.github/actions/run-e2e-tests/action.yml
+++ b/.github/actions/run-e2e-tests/action.yml
@@ -6,11 +6,3 @@ runs:
     - name: E2E Tests
       shell: bash
       run: npm run test:e2e
-      env:
-        SERVER: ${{ env.SERVER }}
-        SITE_NAME: ${{ env.SITE_NAME }}
-        AUTH: ${{ env.AUTH }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -46,7 +46,8 @@ jobs:
         run: npm run coverage
 
       - name: E2E Tests
-        run: npm run test:e2e
+        if: github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
```
anyoung-tableau, line 49: E2E tests from forks would always fail because PRs from forks can't access the secrets. E2E tests will now only run after the PR merges. If we as maintainers are worried about E2E test failures, we can run them locally before approving fork PRs.

jarhun88 (approved):

Issue #192, opened 2026-01-05
Allow admins to enable OAuth but lock the site users can sign into to SITE_NAME


anyoung-tableau:
See Slack thread on the Tableau #DataDev workspace: https://salesforce-external.slack.com/archives/C07LMAVG4N6/p1765895685559239

If the user signs in to a site other than SITE_NAME, return an appropriate 400 during the OAuth callback.

TMCP doesn't know the site the user signed into until after the call to Get Current Server Session @ https://github.com/tableau/tableau-mcp/blob/main/src/server/oauth/callback.ts#L100

PR #196, opened 2026-01-07
@W-20813610: Allow admins to enable OAuth but lock the site users can sign into to SITE_NAME

anyoung-tableau:
These changes introduce a new config variable `OAUTH_LOCK_SITE` (default: true) which ensures that when users connect their agent, they sign in to the site specified in the `SITE_NAME` config variable.

This addresses https://github.com/tableau/tableau-mcp/issues/192 where users are able to sign in to some other site if they already happened to have an active Tableau session in their browser when connecting their agent.

The one major gotcha is that on Server, when site locking is enabled and the user already has an active Tableau session in their browser for some other site, then they will get an error that says `Invalid request. Did you sign into the wrong site? Expected: [SITE_NAME]`. The user will need to sign out in their browser and reattempt to connect their agent.

Here is the behavior matrix I compiled when doing my testing. I did test against Tableau Cloud but the oddities can be ignored since OAuth can't currently be used against Tableau Cloud outside of local testing.

image

### DIFF
--- a/src/server/oauth/callback.ts
+++ b/src/server/oauth/callback.ts
@@ -42,7 +42,8 @@ export function callback(
       if (error === 'invalid_request') {
         res.status(400).json({
           error: 'invalid_request',
-          error_description: `Invalid request. Did you sign into the wrong site? Expected: ${config.siteName || 'Default site'}`,
+          error_description:
+            'Invalid request. Did you sign in to the wrong site? From your browser, please sign out of your site and reconnect your agent to Tableau MCP.',
         });
       } else {
         res.status(400).json({
--- a/tests/oauth/authorizationCodeCallback.test.ts
+++ b/tests/oauth/authorizationCodeCallback.test.ts
@@ -61,7 +61,8 @@ describe('authorization code callback', () => {
     expect(response.headers['content-type']).toBe('application/json; charset=utf-8');
     expect(response.body).toEqual({
       error: 'invalid_request',
-      error_description: 'Invalid request. Did you sign into the wrong site? Expected: mcp-test',
+      error_description:
+        'Invalid request. Did you sign in to the wrong site? From your browser, please sign out of your site and reconnect your agent to Tableau MCP.',
     });
   });
 


docs/src/css/custom.css
```diff
@@ -3,3 +3,8 @@
  * bundles Infima by default. Infima is a CSS framework designed to
  * work well for content-centric websites.
  */
+
+table td,
+table th {
+  vertical-align: top;
+}
```
anyoung-tableau, line 10: I originally had added a big table to the OAuth docs (the one from the screenshot in the description) but I ended up removing it. Going to leave this change here though since by default table cells have a middle vertical alignment and they look weird.

src/server/oauth/callback.ts
```diff
@@ -39,10 +39,18 @@ export function callback(
...
     if (error) {
-      res.status(400).json({
-        error: 'access_denied',
-        error_description: 'User denied authorization',
-      });
+      if (error === 'invalid_request') {
+        res.status(400).json({
+          error: 'invalid_request',
+          error_description: `Invalid request. Did you sign into the wrong site? Expected: ${config.siteName || 'Default site'}`,
+        });
```
anyoung-tableau, line 47: This can happen when the user already has an active Tableau session in their browser for a site that is _different_ than the one specified in the SITE_NAME config variable.

src/server/oauth/authorize.ts
```diff
@@ -133,10 +133,50 @@ export function authorize(
     oauthUrl.searchParams.set('device_name', getDeviceName(redirect_uri, state ?? ''));
     oauthUrl.searchParams.set('client_type', 'tableau-mcp');
 
-    res.redirect(oauthUrl.toString());
+    if (config.oauth.lockSite) {
+      // The "redirected" parameter is used by Tableau's OAuth controller to determine whether the user will be shown the site picker.
+      // When provided, the user will not be shown the site picker.
+      oauthUrl.searchParams.set('redirected', 'true');
```
anyoung-tableau, line 139: If you are internal, you can see how the `redirected` parameter gets used here: https://github.com/sf-analyticscloud/monolith/blob/main/workgroup/src/silos/tableau-server/services/war-vizportal/src/com/tableausoftware/api/oauth/OAuthController.java#L194

jarhun88: is it just the presence of the field value? redirected: true kind of feels like the user would be redirected to the site picker. (unless im misinterpreting what redirected means)

src/server/oauth/authorize.ts
```diff
@@ -133,10 +133,50 @@ export function authorize(
...
+    }
+
+    const redirectUrl = await getOAuthRedirectUrl(oauthUrl, { lockSite: config.oauth.lockSite });
+    res.redirect(redirectUrl.toString());
   });
 }
 
+async function getOAuthRedirectUrl(
+  initialOAuthUrl: URL,
+  { lockSite }: { lockSite: boolean },
```
jarhun88, line 149: cant this param just be a bool instead of an object

src/server/oauth/authorize.ts
```diff
@@ -133,10 +133,50 @@ export function authorize(
...
+  // Tableau does not show the site picker.
+  // We can force it to by changing the path from #/signin to #/site.
+
+  try {
+    const response = await fetch(initialOAuthUrl, { redirect: 'manual' });
+    if (response.status === 302) {
+      // The response is a redirect to the Tableau OAuth login page.
+      // Force it to ultimately show the site picker by changing the path from #/signin to #/site.
+      const location = response.headers.get('location');
+      if (location?.startsWith('#/signin') || location?.startsWith('/#/signin')) {
```
jarhun88, line 168: nit: is this cleaner 

location?.includes('#/signin')

src/server/oauth/authorize.ts
anyoung-tableau: It needs to have *some* value, what the value is doesn't matter. The comments in the controller don't make it terribly clear exactly what "redirected" mean, but it doesn't mean "redirected to the site picker". When a value for "redirected" is not provided the comments say: "it means that the request is from the client side" but the end result is what we want. I think "client" refers to Tableau Desktop in this case. The PR also unfortunately has no description. https://github.com/sf-analyticscloud/monolith/pull/30596. I reached out to the author, but he's left the team.

src/server/oauth/authorize.ts
anyoung-tableau: Yeah, but I generally prefer objects for config args since it acts as a way to provide "named" arguments.

`getOAuthRedirectUrl(url, true);`

vs

`getOAuthRedirectUrl(url, { lockSite: true } );`

src/server/oauth/authorize.ts
anyoung-tableau: I wanted to avoid the possibility of a false positive when `#/signin` appears somewhere *else* in the string. The reason why there are 2 parts to the condition is because the first matches on Server and the second matches on Cloud.

PR #201, opened 2026-01-15
@W-20918624: Add build script, clean up dependencies

anyoung-tableau:
These changes came about while I was experimenting in the [anyoung/embed-view](https://github.com/tableau/tableau-mcp/tree/anyoung/embed-view) branch.

These changes:

1. Introduce a build script that uses the esbuild API so we no longer need to have the lengthy NPM script that uses the esbuild CLI.
2. Adds `tsx` to execute that build script.
3. Execute `createClaudeMcpBundleManifest.ts` with `tsx` rather than explicitly building with `esbuild` first and then executing with Node.js. `tsx` uses esbuild anyway.
4. Removed some unnecessary dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,6 @@ jobs:
 
       - name: Build Claude MCP Bundle
         run: |
-          npm run build:manifest:script
           npm run build:manifest
           npm ci --omit=dev
           npx -y @anthropic-ai/mcpb pack . tableau-mcp.mcpb
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,7 +36,6 @@ jobs:
           git checkout -b mcpb ${{ github.event.release.tag_name }}
           npm ci
           npm run build
-          npm run build:manifest:script
           npm run build:manifest
           npm ci --omit=dev
           npx -y @anthropic-ai/mcpb pack . tableau-mcp-${{ github.event.release.tag_name }}.mcpb


package.json
```diff
@@ -69,7 +64,6 @@
     "@anthropic-ai/mcpb": "^1.0.0",
     "@eslint/eslintrc": "^3.3.1",
     "@eslint/js": "^9.25.1",
-    "@modelcontextprotocol/inspector": "^0.16.6",
```
anyoung-tableau, line 72: We never really needed this dependency on the Inspector anyway since the script that launches the inspector will install it anyway if it's not already installed. Removing this dependency also fixes a few npm audit vulnerabilities.

package.json
```diff
@@ -19,17 +19,12 @@
     "tableau-mcp-server": "./build/index.js"
   },
   "scripts": {
-    "clean": "npx rimraf ./build",
```
jarhun88, line 22: we no longer need a clean script?

package.json
```diff
@@ -19,17 +19,12 @@
...
-    ":build": "esbuild src/index.ts --bundle --platform=node --format=cjs --minify --outfile=build/index.js --sourcemap --log-override:empty-import-meta=silent",
-    ":build:dev": "esbuild src/index.ts --bundle --packages=external --platform=node --format=cjs --outfile=build/index.js --sourcemap",
-    "build": "run-s clean :build exec-perms",
-    "build:dev": "run-s clean :build:dev exec-perms",
-    "build:watch": "npm run :build:dev -- --watch",
```
anyoung-tableau, line 27: Removing build:watch for now since it's not super useful imho. When transport is http you have to restart the express server anyway after each rebuild.

package.json
anyoung-tableau: It's incorporated into the build script

stephendeoca (approved):

PR #191, opened 2025-12-30
@W-20786426: queryDatasource tool description Improvement

joeconstantino:
## Description

Updates the queryDatasource tool description to include details about how to use the context property in the HBI spec. This change was motivated by qualitative feedback and testing from Alexis Guinebertiere. Documented here - https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1762282334243879
## Type of Change
- Bug fix

## How Has This Been Tested?
Successfully ran Alexis's documented failure cases in Claude and Cursor

## Checklist
- I have updated the version in the package.json file by using `npm run version`. For example,
      use `npm run version:patch` for a patch version bump.
- New and existing unit tests pass locally with my changes

## Contributor Agreement
By submitting this pull request, I confirm that:

- [x] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) for this project and followed
      its Contribution Checklist.

PR #202, opened 2026-01-16
@W-19426562: Expose VizQL Data Service rowLimit option, add MAX_RESULT_LIMITS config entry

anyoung-tableau:
These changes add a `limit` parameter to the `query-datasource` tool when the version of Tableau is at least 2026.1. When provided, it limits the maximum number of rows that can be returned by VDS.

These changes also introduce tool-specific result limiting. @stephendeoca shared some good analysis that having only a global limit doesn't make a lot of sense. A limit for one tool may be wildly different than another.

Here is the corresponding VDS MR for internal folks: https://gitlab.tableausoftware.com/headless-bi/headless-bi-service/-/merge_requests/561

### DIFF
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -226,9 +226,9 @@ This means that:
 - For a list of available tools and groups, see
   [toolName.ts](https://github.com/tableau/tableau-mcp/blob/main/src/tools/toolName.ts).
 - Only applies to tools that have a `limit` parameter and return an array of items.
-- Tool names take precedence over tool group names. That is, `datasource:10,list-datasources:20`
+- Tool names take precedence over tool group names. That is, `datasource:1000,list-datasources:20`
   means that the `list-datasources` tool can return up to 20 data sources but the `query-datasource`
-  tool can only return up to 10 results.
+  tool can only return up to 1000 results.
 - If a tool-specific limit is not set, the global limit specified by the
   [`MAX_RESULT_LIMIT`](#max_result_limit) environment variable will be used instead.
 - Each limit must be a positive number.


src/tools/queryDatasource/queryDescription.2025.3.ts
```diff
@@ -221,6 +221,37 @@ If a bin field already exists in the datasource (e.g., "Profit (bin)"), query it
 To control existing bin field behavior, use parameters if available (e.g., "Profit Bin Size").
 
 ## Filter Types and Usage
+### Filter Context Property
```
anyoung-tableau, line 224: Driveby fix here to apply the changes from https://github.com/tableau/tableau-mcp/pull/191 in the 2025.3 description too.

src/server.test.ts
```diff
@@ -27,11 +27,13 @@ describe('server', () => {
 
     const tools = toolFactories.map((toolFactory) => toolFactory(server));
     for (const tool of tools) {
-      expect(server.tool).toHaveBeenCalledWith(
+      expect(server.registerTool).toHaveBeenCalledWith(
```
anyoung-tableau, line 30: Driveby: `server.tool` is deprecated in the SDK.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -68,15 +75,24 @@ export const getQueryDatasourceTool = (
...
         }),
     ),
-    paramsSchema,
+    paramsSchema: new Provider(
+      async () =>
+        await getResultForTableauVersion<ParamsSchema>({
+          server: config.server || authInfo?.server,
+          mappings: {
+            '2026.1.0': paramsSchema,
+            default: (({ limit: _, ...rest }) => rest)(paramsSchema), // remove 'limit' for older versions
```
anyoung-tableau: This is what makes it so the `limit` parameter is not even advertised to MCP clients if the Tableau Server is old.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -30,6 +30,13 @@ import { validateQueryAgainstDatasourceMetadata } from './validators/validateQue
 const paramsSchema = {
   datasourceLuid: z.string().nonempty(),
   query: querySchema,
+  limit: z.number().min(1).optional(),
```
stephendeoca: it's likely in the future we might offer more query options, so if you want you can move limit inside an options object. Or we can save that for when there's actually more options to offer.

I also recommend updating the schema to force an integer value with `z.number().int().min(1).optional()`

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -30,6 +30,13 @@ import { validateQueryAgainstDatasourceMetadata } from './validators/validateQue
...
+};
+
+type ParamsSchema = {
```
stephendeoca: What's the motivation for this change here? can't we do `type ParamsSchema = typeof paramsSchema` or `z.infer` type?

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -101,6 +117,12 @@ export const getQueryDatasourceTool = (
             returnFormat: 'OBJECTS',
             debug: true,
             disaggregate: false,
+            rowLimit:
+              limit === undefined
+                ? undefined
+                : config.maxResultLimit
```
stephendeoca: I'm not sure if MAX_RESULT_LIMIT makes sense here. We might want an environment variable specific to query result limits instead.

MAX_RESULT_LIMIT is used by the searchContent, listDatasources, listViews, list... tools to limit search results. So its likely someone would use this environment variable to limit results to like 25 datasources or something. So queries might get limited to 25 rows by accident if someone doesn't realize it's being used here too.

anyoung-tableau: Hmm, the docs mention that MAX_RESULT_LIMIT applies to any tool that has a `limit` parameter. It was added as a global "catch-all" to prevent huge amounts of data getting returned and exhausting the context and sets a hard cap on the `limit` argument that users specify.

I think if we want to add more tool-specific granularity to the max limits, we should do it for all the tools and keep MAX_RESULT_LIMIT as the default value if no tool-specific value is provided.

Something like this. Thoughts?

```ts
TOOL_RESULT_LIMITS: {
  "query-datasource": 500,
  "list-datasources": 1000
}
```

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: `ParamsSchema` and `type paramsSchema` are slightly different. 

`typeof paramsSchema` looks like this:

```ts
{
    datasourceLuid: ZodString;
    query: z.ZodObject<typeof querySchema.shape>;
    limit: ZodOptional<ZodNumber>;
}
```

Notice that `limit` is *required* in the type, even if its *value* could be `undefined`. In `ParamsSchema`, `limit` is not required in the type so objects of type `ParamsSchema` do *not* need a `limit` key. If we used `typeof paramsSchema` then we'd be forced to advertise to clients that the tool *always* has a `limit` parameter whose value can be a number or `undefined`. This approach allows us to not even advertise the `limit` parameter at all when the Tableau Server version is too old.

Simple example:
image

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: Good call on the integer forcing. I'd like to keep `limit` outside of an `options` param because that's how all the other tools are currently doing it.

src/tools/queryDatasource/queryDatasource.ts
stephendeoca: A per tool limit could work, but that might be over engineering the solution

I think we can make an exception for a env variable specifically for limiting query results such as `MAX_QUERY_RESULT_LIMIT`. Then the rest are captured by `MAX_RESULT_LIMIT`.

I don't think it makes sense to share `MAX_RESULT_LIMIT` with the other tools, since if you want to limit list_datasource results to something simple like 10 cuz there's many duplicates that means you also limit your query_datasource result to 10 rows. So essentially the `MAX_RESULT_LIMIT` would really only make sense using it for the query_datasource tool.

anyoung-tableau: I actually feel like a per tool limit is less engineering than a one-off just for query-datasource. Doing things generally covers all our bases and makes it easy for admins to specify either a single global limit, or control the limits of each tool.

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -184,13 +184,57 @@ The maximum timeout for requests to the Tableau Server REST API.
...
+Example:
+
+```
+MAX_RESULT_LIMIT=30
+MAX_RESULT_LIMITS=query-datasource:10,list-datasources:20
+```
+
+This means that:
+
+- The `query-datasource` tool can return up to 10 results.
```
stephendeoca: Maybe make the example more realistic, since I don't think limiting query_datasource to 10 results is a typical use case. Maybe someone would limit it to 1000 rows or so.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -113,16 +98,16 @@ export const getQueryDatasourceTool = (
           }
 
           const datasource: Datasource = { datasourceLuid };
+          const maxResultLimit = config.getMaxResultLimit(queryDatasourceTool.name);
+          const rowLimit = maxResultLimit
+            ? Math.min(maxResultLimit, limit ?? Number.MAX_SAFE_INTEGER)
+            : limit;
```
stephendeoca, line 104: should this be `limit ?? Number.MAX_SAFE_INTEGER`

stephendeoca (approved):

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: Nah. An undefined limit is effectively the same as a huge limit.

docs/docs/configuration/mcp-config/env-vars.md
anyoung-tableau: Sure

Issue #204, opened 2026-01-21
MCP OAuth scope support + best‑practice authorization challenges


mattcfilbert:
## Is your feature request related to a problem? Please describe.

The MCP server currently lacks a formal, spec‑aligned way to advertise and enforce authorization scopes for MCP tools/resources. This makes it hard for clients to request least‑privilege access, and prevents the MCP server from returning correct insufficient_scope challenges when access is missing. We also need a clear, documented scope set for the beta and a rationale for why MCP scopes are required in addition to Tableau API scopes.

## Describe the solution you'd like

Implement MCP scope support end‑to‑end per the MCP authorization spec:
Add scope configuration knobs (OAUTH_SCOPES_SUPPORTED, OAUTH_REQUIRED_SCOPES, OAUTH_DISABLE_SCOPES).

Advertise scopes_supported in OAuth metadata.

Parse/validate requested scopes at /oauth/authorize and /oauth/token.

Persist scopes through auth code and refresh flows and include scopes in issued access tokens.

Enforce required scopes with WWW-Authenticate challenges and insufficient_scope errors.

Document a recommended inclusive scope set (MCP + Tableau API scopes) for the beta and explain why MCP scopes are needed even without token exchange.

## Additional context

This aligns with MCP authorization guidance and Andy Young’s advice: scopes are MCP‑defined and needed to protect MCP‑specific tools/prompts/resources (not all MCP operations call Tableau APIs). We’ll keep Server in no‑scope mode for now via OAUTH_DISABLE_SCOPES=true, and avoid token exchange until a future phase.

PR #205, opened 2026-01-21
@W-20976712: Add MCP OAuth scope support and enforcement

mattcfilbert:
## Description

Implemented MCP OAuth scope support and best‑practice authorization challenges. Added scope config knobs, advertised scopes_supported, validated requested scopes, persisted scopes across auth/refresh flows, embedded scopes in access tokens, and enforced required scopes with insufficient_scope challenges. Updated OAuth docs with a recommended beta scope set and rationale for MCP scopes.

## Motivation and Context

MCP lacked spec‑aligned scope handling and challenge responses, which blocked least‑privilege access and step‑up behavior. This change brings OAuth behavior in line with the MCP authorization spec and provides a clear scope configuration for beta.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## How Has This Been Tested?

Not run (logic is unit‑level; tests can be added in a follow‑up).

## Related Issues

MCP OAuth scope support + best‑practice authorization challenges #204 

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.15.3",
+  "version": "1.17.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.15.3",
+      "version": "1.17.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.15.3",
+  "version": "1.17.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+export type McpScope =
+  | 'tableau:content:read'
+  | 'tableau:viz_data_service:read'
+  | 'tableau:views:download'
+  | 'tableau:insight_definitions_metrics:read'
+  | 'tableau:insight_metrics:read'
+  | 'tableau:metric_subscriptions:read'
+  | 'tableau:insights:read'
+  | 'tableau:insight_brief:create';
```
anyoung-tableau, line 34: In my mind, I was thinking we wouldn't use an API scope to protect an MCP resource directly ("resource" as in OAuth resource, not MCP [resource](https://modelcontextprotocol.io/specification/2025-11-25/server/resources)). That is: an access token must have, say, the `tableau:mcp:datasource:read` MCP scope to even call the `query-datasource` tool, but that same access token would *also* need the `'tableau:viz_data_service:read'` API scope since it's going to be re-used to call into VDS.

src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Minimal default scopes suggested when no specific tool is known.
+ */
+export const DEFAULT_REQUIRED_SCOPES: McpScope[] = ['tableau:content:read'];
+
+/**
+ * Validates that a scope string is a valid MCP scope
+ */
+export function isValidScope(scope: string): scope is McpScope {
+  return DEFAULT_SCOPES_SUPPORTED.includes(scope as McpScope);
```
anyoung-tableau: You can avoid the type assertion using `DEFAULT_SCOPES_SUPPORTED.some((s) => s === scope);`

src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Determines the required scopes for a given MCP endpoint/tool
+ *
+ * This function maps MCP tools to their required scopes.
+ * This is used to provide scope guidance in WWW-Authenticate headers.
+ *
+ * @param endpoint - The MCP endpoint or tool name
+ * @returns Array of required scopes for the endpoint
+ */
+export function getRequiredScopesForTool(toolName: ToolName | string): string[] {
```
anyoung-tableau: Since we are going to need to maintain this map, let's use it as the source of truth for when tools generate a JWT with scopes.

For example, the `query-datasource` tool currently passes `jwtScopes: ['tableau:viz_data_service:read'],` to `useRestApi()`.

Instead, it could pass: `jwtScopes: getRequiredScopesForTool(queryDatasourceTool.name),`.

This way we can never accidentally forget to update this map or the tool implementation.

src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,143 @@
...
+  const toolScopeMap: Record<ToolName, McpScope[]> = {
```
anyoung-tableau: Let's pull `toolScopeMap` outside of the function definition. It doesn't need to be initialized every time the function is called.

src/server/oauth/scopes.ts
anyoung-tableau: Let's change the type on the `toolName` parameter to be `ToolName`. Calling sites can provide their tool name using `queryDatasourceTool.name` (or even `this.name` if we call this function from `tool.ts`) for example. Then we don't need the `if (toolName in toolScopeMap)` or the fallback `return DEFAULT_REQUIRED_SCOPES;` below.

anyoung-tableau: If scopes are disabled, `getRequiredScopesForTool` should return an empty array? Or were you planning to not call this function when scopes are disabled?

src/server/oauth/.well-known/oauth-authorization-server.ts
```diff
@@ -10,16 +10,16 @@ import { getConfig } from '../../../config.js';
...
+      issuer,
+      authorization_endpoint: `${issuer}/oauth/authorize`,
+      token_endpoint: `${issuer}/oauth/token`,
+      registration_endpoint: `${issuer}/oauth/register`,
       response_types_supported: ['code'],
       grant_types_supported: ['authorization_code', 'refresh_token', 'client_credentials'],
       code_challenge_methods_supported: ['S256'],
-      scopes_supported: [],
+      scopes_supported: scopesSupported,
       token_endpoint_auth_methods_supported: ['client_secret_basic', 'client_secret_post'],
```
anyoung-tableau: While we're here, would you mind making a driveby change here and add `'none'` to `token_endpoint_auth_methods_supported`?

https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1769075292144669

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -168,6 +172,78 @@ Where `Y2xpZW50SWQ6c2VjcmV0` is the base64 encoding of `clientId:secret`.
...
+- Optional, but recommended when OAuth is enabled.
+- Used to populate `scopes_supported` in the OAuth metadata.
+- Example: `tableau:content:read tableau:view:read tableau:datasource:query`
+
+<hr />
+
+## Recommended scope set (beta)
+
+For the initial release, use an inclusive scope set (MCP + Tableau API scopes) to avoid token
+exchange. This keeps consent simple and aligns with the current MCP server behavior.
```
anyoung-tableau: Internal decisions and implementation details don't need to be documented.

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -168,6 +172,78 @@ Where `Y2xpZW50SWQ6c2VjcmV0` is the base64 encoding of `clientId:secret`.
...
+## Why MCP scopes (in addition to Tableau API scopes)
+
+Tableau API scopes alone are not sufficient to protect all MCP functionality.
+
+- Not all MCP tools call Tableau APIs. Some tools can operate entirely within the MCP server
+  (for example, generating a TWB). Tableau API scopes do not describe those operations.
+- MCP also exposes non-API concepts like prompts and resources that should be gated behind MCP
+  scopes. Those do not have a natural Tableau API scope equivalent.
+- Keeping MCP scopes separate from API scopes clarifies intent and avoids over-granting: MCP scopes
+  authorize what the MCP server can do; Tableau API scopes authorize what downstream APIs can do.
```
anyoung-tableau: I don't think we need to go into these details in the docs. We should describe more of the "what", and not necessarily the "why" here.

src/server/oauth/authorize.ts
```diff
@@ -132,6 +168,9 @@ export function authorize(
     oauthUrl.searchParams.set('target_site', config.siteName);
     oauthUrl.searchParams.set('device_name', getDeviceName(redirect_uri, state ?? ''));
     oauthUrl.searchParams.set('client_type', 'tableau-mcp');
+    if (scopesToGrant.length > 0) {
+      oauthUrl.searchParams.set('scope', formatScopes(scopesToGrant));
+    }
```
anyoung-tableau: So, the `/oauth2/v1/auth` endpoint on Tableau doesn't support scopes or consent. Not sure it makes sense to provide the scopes if they're going to be ignored.

src/server/oauth/scopes.ts
anyoung-tableau: Also were you planning to wire this function up in a followup PR?

tests/e2e/workbooks/listWorkbooks.test.ts
```diff
@@ -34,11 +34,13 @@ describe('list-workbooks', () => {
     const workbooks = await callTool('list-workbooks', {
       env,
       schema: z.array(workbookSchema),
-      toolArgs: { filter: 'name:eq:Super*' },
+      toolArgs: { filter: 'name:eq:Superstore' },
```
anyoung-tableau, line 37: Oops, I might have messed this up. I duplicated the Superstore workbook yesterday. Did this cause a test failure?

OAUTH_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,315 @@
...
+## Open Questions for Auth Team
+
+1. **Scope Format**: What is the exact format of Tableau OAuth scopes? (e.g., `tableau:content:read`, `tableau.content.read`, etc.)
+
+2. **Scope Mapping**: How should we map MCP scopes to Tableau scopes? One-to-one, one-to-many, or many-to-one?
+
+3. **Token Exchange**: 
+   - What is the endpoint for token exchange?
+   - What format should the request/response be?
+   - How do we convert MCP tokens to Tableau REST API tokens?
```
anyoung-tableau: We aren't going to use token exchange for initial release

tests/e2e/workbooks/listWorkbooks.test.ts
mattcfilbert: short answer: yes

OAUTH_IMPLEMENTATION_PLAN.md
mattcfilbert: thank you for the reminder - I'm going through all of your great feedback now

src/config.ts
```diff
@@ -135,8 +138,11 @@ export class Config {
       OAUTH_AUTHORIZATION_CODE_TIMEOUT_MS: authzCodeTimeoutMs,
       OAUTH_ACCESS_TOKEN_TIMEOUT_MS: accessTokenTimeoutMs,
       OAUTH_REFRESH_TOKEN_TIMEOUT_MS: refreshTokenTimeoutMs,
-      TELEMETRY_PROVIDER: telemetryProvider,
-      TELEMETRY_PROVIDER_CONFIG: telemetryProviderConfig,
+      OAUTH_SCOPES_SUPPORTED: oauthScopesSupported,
+      OAUTH_REQUIRED_SCOPES: oauthRequiredScopes,
+      OAUTH_DISABLE_SCOPES: oauthDisableScopes,
+      TELEMETRY_PROVIDER: _telemetryProvider,
+      TELEMETRY_PROVIDER_CONFIG: _telemetryProviderConfig,
```
anyoung-tableau: Was the rename of the `telemetryProvider` and `telemetryProviderConfig` variables intentional?

src/restApiInstance.ts
```diff
@@ -16,19 +16,12 @@ import {
...
-type JwtScopes =
-  | 'tableau:viz_data_service:read'
-  | 'tableau:content:read'
-  | 'tableau:insight_definitions_metrics:read'
-  | 'tableau:insight_metrics:read'
-  | 'tableau:metric_subscriptions:read'
-  | 'tableau:insights:read'
-  | 'tableau:views:download'
-  | 'tableau:insight_brief:create';
+type JwtScopes = TableauApiScope;
```
anyoung-tableau: Let's kill the `JwtScopes` type entirely.

src/config.ts
mattcfilbert: Accident after linting + merge. Will revert

package.json
```diff
@@ -49,6 +49,7 @@
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.25.2",
     "@zodios/core": "^10.9.6",
+    "axios": "^1.13.4",
```
anyoung-tableau: Was adding axios explicitly intentional? We already get it for free with Zodios

mattcfilbert: hit a local build hiccup that was seeking axis. no longer seem to have the issue so I'm removing it

src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Determines the required scopes for a given MCP endpoint/tool
+ *
+ * This function maps MCP tools to their required scopes.
+ * This is used to provide scope guidance in WWW-Authenticate headers.
+ *
+ * @param endpoint - The MCP endpoint or tool name
+ * @returns Array of required scopes for the endpoint
+ */
+export function getRequiredScopesForTool(toolName: ToolName): McpScope[] {
```
anyoung-tableau, line 196: I think there's a smell here that is exposing a small gap. We're not calling this anywhere, right?

1. TMCP advertises the comprehensive set of MCP and API scopes needed in the `scopes_supported` key of `/.well-known/oauth-authorization-server`.
2. The scopes included in the `WWW-Authenticate` challenge will match that of `scopes_supported` (unless overridden by the admin using `OAUTH_REQUIRED_SCOPES`).
3. MCP clients will request and be issued an access token that include the scopes provided in the `WWW-Authenticate` challenge.
4. Our authMiddleware ensures the access token for every request has ALL the scopes needed by ALL the tools collectively.

Bullet 4 doesn't seem right. We should instead only be checking that the access token has the scopes to serve the _current_ request. For example, if the request is for calling the `list-datasources` tool, we should only be checking if the access token has the MCP scopes and the API scopes needed by the `list-datasources` tool. If it's missing scopes needed by other tools, that doesn't matter.

OAUTH_IMPLEMENTATION_PLAN.md
bcantoni, line 1: @mattcfilbert I would omit this plan file

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -168,6 +172,37 @@ Where `Y2xpZW50SWQ6c2VjcmV0` is the base64 encoding of `clientId:secret`.
 
 <hr />
 
+### `OAUTH_SCOPES_SUPPORTED`
```
anyoung-tableau: Hmm... these environment variables put the responsibility on the MCP server admin to know in advance what scopes are needed at runtime. This is a pretty big ask and I imagine some/most admins don't care. I think we should remove both of these environment variables and populate `scopes_supported` in `/.well-known/oauth-authorization-server` and `scopes` in the `WWW-Authenticate` challenge with the comprehensive set of MCP and API scopes.

anyoung-tableau: We should discuss the API scopes a bit more too. They are meaningless to include on the access token when it is issued by TMCP's authz server so I think there's value in not including them by default.

src/server/oauth/scopes.ts
mattcfilbert: Thanks @anyoung-tableau. I have some clarifying Q's:

If we remove OAUTH_SCOPES_SUPPORTED/OAUTH_REQUIRED_SCOPES, should we always advertise the union of MCP scopes only (and maybe keep API scopes out entirely)?

For per‑request checks, should we enforce only MCP scopes for the tool, and ignore API scopes on the token?

Where do you recommend the per‑request scope check to live (auth middleware vs tool dispatch)?

For non‑tool endpoints, should we require any scopes or none?

OAUTH_IMPLEMENTATION_PLAN.md
mattcfilbert: Fair enough - thanks, Brian

src/server/oauth/scopes.ts
anyoung-tableau: > If we remove OAUTH_SCOPES_SUPPORTED/OAUTH_REQUIRED_SCOPES, should we always advertise the union of MCP scopes only (and maybe keep API scopes out entirely)?

Yeah. We can leave the API scopes out for self-hosted but we'll need them for 1st-party hosted since the access token issued by the Tableau Authz server will be reused to call Tableau APIs.

> For per‑request checks, should we enforce only MCP scopes for the tool, and ignore API scopes on the token?

For first-party hosted I think it wouldn't be a bad idea to check the API scopes too. That way we can respond with a scope step-up challenge rather than a hard 401 fail.

> Where do you recommend the per‑request scope check to live (auth middleware vs tool dispatch)?

Not sure yet. I'm leaning toward auth middleware but that means we'll have to parse the tool name from the request body.

> For non‑tool endpoints, should we require any scopes or none?

We should require MCP scope(s) only.

mattcfilbert: applied. thank you!

docs/docs/configuration/mcp-config/oauth.md
mattcfilbert: looking forward to chatting tomorrow more on scoping, but I believe we're on the same page and I made updates based on your feedback so I'll resolve this thread

.gitignore
```diff
@@ -17,4 +17,5 @@ tableau-mcp.mcpb
 *.pem
 *.tgz
 *.tar.gz
-*.zip
\ No newline at end of file
+*.zip
+OAUTH_IMPLEMENTATION_PLAN.md
```
jarhun88: should this be in ignore?

OAUTH_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,301 @@
+# Tableau MCP OAuth Integration Implementation Plan
```
jarhun88: is it typical for us to publicize an implementation plan to public?

.gitignore
mattcfilbert: good feedback, I will be 10% less lazy and move my doc

OAUTH_IMPLEMENTATION_PLAN.md
mattcfilbert: narp, moving now

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -241,6 +256,16 @@ References:
 
 <hr />
 
+### `ADVERTISE_API_SCOPES`
+
+Include Tableau API scopes in OAuth metadata and scope challenges.
+
+- Default: `false` (self-hosted)
+- When `true` (Falcon), `scopes_supported` includes MCP + API scopes and scope challenges may
```
anyoung-tableau: This `(Falcon)` distinction does not need to be advertised in the docs.

package-lock.json
```diff
@@ -2521,9 +2521,9 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.12.1",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.12.1.tgz",
-      "integrity": "sha512-Kn4kbSXpkFHCGE6rBFNwIv0GQs4AvDT80jlveJDKFxjbTYMUeB4QtsdPCv6H8Cm19Je7IU6VFtRl2zWZI0rudQ==",
+      "version": "1.13.4",
```
anyoung-tableau: Please revert this file. `axios` is already bumped to 1.13.5 in the main branch

src/server/oauth/.well-known/oauth-authorization-server.ts
```diff
@@ -10,17 +11,17 @@ import { getConfig } from '../../../config.js';
  */
 export function oauthAuthorizationServer(app: express.Application): void {
   app.get('/.well-known/oauth-authorization-server', (_req, res) => {
-    const origin = getConfig().oauth.issuer;
+    const { issuer, advertiseApiScopes } = getConfig().oauth;
```
anyoung-tableau: We need to update the URLs in `oauth-authorization-server.ts` to use the `oauth2` path. Easiest approach is to update the MCP authz server paths to match.

src/config.ts
anyoung-tableau: We need some conditional logic here. These constraints only apply to the embedded authz server.

src/server/oauth/provider.ts
anyoung-tableau, line 49: We need conditional logic to not setup the routes except for the `.well-known` endpoints when the issuer is the Tableau authz server

anyoung-tableau: We either need some conditional logic here not create the key pair when the issuer is the Tableau authz server or consider an approach where we have a base abstract `OAuthProvider` class and then two concrete classes that extend it, one for the embedded authz provider, and another for the Tableau authz provider.

src/server/oauth/authMiddleware.ts
```diff
@@ -80,11 +95,116 @@ export function authMiddleware(privateKey: KeyObject): RequestHandler {
...
+      const missingScopes = [...missingMcpScopes, ...missingApiScopes];
+
+      if (missingScopes.length > 0) {
+        const baseUrl = `${req.protocol}://${req.get('host')}`;
+        const requiredScopesForChallenge = [
+          ...requiredMcpScopes,
+          ...(shouldCheckApiScopes ? requiredApiScopes : []),
+        ];
+        const scopeParam = `scope="${formatScopes(requiredScopesForChallenge)}"`;
+        const wwwAuthenticate = `Bearer realm="MCP", error="insufficient_scope", error_description="Missing required scopes", resource_metadata="${baseUrl}/.well-known/oauth-protected-resource", ${scopeParam}`;
```
anyoung-tableau, line 130: Were you able to find a client to test the step-up authz flow? Everything looks good here and we're doing the [recommended approach](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors) by including *all* the scopes, not just the missing scopes, but both MCP Inspector and MCPJam don't seem to implement the [step up flow](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#step-up-authorization-flow).

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -241,6 +270,26 @@ References:
 
 <hr />
 
+### `OAUTH_EMBEDDED_AUTHZ_SERVER`
+
+When `true` (default), the MCP server runs the embedded OAuth authorization server (authorize, token,
```
anyoung-tableau: > When `true` (default), the MCP server runs the embedded OAuth authorization server

This is a bit misleading. This env var isn't *responsible* for running the embedded authz server, it's just a hint to TMCP that `OAUTH_ISSUER` is provided so that the embedded authz server is being used.

We should aim to remove this env var and determine at run-time if the embedded authz server is in use, but for now let's at least change the messaging to be more reflective of its real purpose. Alternatively, let's just remove it from the docs entirely until the Tableau authz server is ready for public consumption.

src/server/oauth/scopes.ts
```diff
@@ -0,0 +1,221 @@
...
+  return supportedApiScopes;
+}
+
+export function getSupportedScopes({ includeApiScopes }: { includeApiScopes: boolean }): string[] {
+  return includeApiScopes ? [...supportedMcpScopes, ...supportedApiScopes] : supportedMcpScopes;
+}
+
+export function isTableauApiScope(scope: string): scope is TableauApiScope {
+  return supportedApiScopes.some((supported) => supported === scope);
+}
```
anyoung-tableau: This function is unused

src/utils/axios.ts
```diff
@@ -1,8 +1,4 @@
-import axios, {
-  AxiosRequestConfig,
-  AxiosResponse,
-  isAxiosError,
-} from '../../node_modules/axios/index.js';
+import axios, { AxiosRequestConfig, AxiosResponse, isAxiosError } from 'axios';
```
anyoung-tableau: I may be off-base here, but I was intentionally not importing from `'axios'` because we don't have a direct dependency on it, it's indirect through Zodios. Maybe this is fine, but having this extra file is kinda meaningless now. I would suggest reverting this file and reserving these changes for their own PR, updating all places to import directly from `'axios'` instead of from `utils/axios.js`

tests/oauth/fixtures/jwe-private-key.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
anyoung-tableau: Remove this file please. Even though it's a test private key it will undoubtedly get flagged in security scans

tests/oauth/testEnv.ts
```diff
@@ -5,7 +5,12 @@ import { resetEnv as resetBaseEnv, setEnv as setBaseEnv } from '../testEnv.js';
 export function setEnv(): void {
   setBaseEnv();
   dotenv.config({ path: 'tests/oauth/.env.oauth', override: true });
+  process.env.OAUTH_ISSUER = process.env.OAUTH_ISSUER ?? 'http://127.0.0.1:3927';
```
anyoung-tableau: These changes shouldn't be necessary. The values should come from `tests/oauth/.env.oauth` that are read on line 7.

src/server/oauth/provider.ts
anyoung-tableau: It's not the end of the world if we still create the routes for the embedded authz server even when the issuer is the Tableau authz server but not creating them at all would be ideal. Please create a story for this as a followup item.

anyoung-tableau:
Thanks @mattcfilbert !! Please run `npm version:minor` to bump the version and merge!

anyoung-tableau (approved):

PR #209, opened 2026-01-23
@W-20742343 Product telemetry

jarhun88:

the pull request._

# Pull Request Template

## Description

- product telemetry to get UIP data that we can observe on how the server is used. For now we are aiming to track the following fields

        * Timestamp (ISO 8601)
        * RequestID (for correlation)
        * Session ID
        * Requesting site
        * requesting pod
        * Parameters (sanitized - remove PII, auth tokens)
        * Success/failure status
        * Error code
        * For tools/call: specific tool name (e.g., local-tableau:query-datasource)

- here is the product telemetry tableau system that were utilizing https://docs.google.com/document/d/11zY6e2uXLlwetLo7NtrebmDa61TVa5WkYPIi7ngg63Q/edit?tab=t.0#heading=h.mskhvfu51hln

- here is the associated juno schema http://juno.tsi.lan/manage/analytics.tableau.tableau_mcp.tableau_mcp.tool_call

- created a new env variable to disable the product telemetry. (default is enabled)
- passing through sessionID through all tools

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.14.5",
+  "version": "1.14.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.14.5",
+      "version": "1.14.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.25.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.14.5",
+  "version": "1.14.6",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,107 @@
...
+   * Service name. Default: 'tableau-mcp'
+   */
+  serviceName?: string;
+}
+
+/**
+ * A simplified telemetry forwarder that sends events directly to Tableau's
+ * telemetry JSON endpoint (e.g., qa.telemetry.tableausoftware.com).
+ */
+export class DirectTelemetryForwarder {
```
jarhun88: i basically copied this class from the prod telemetry team typescript telemetry forwarder https://gitlab.tableausoftware.com/Product-Analytics/typescript-telemetry-forwarder/-/merge_requests/37

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
       if (result.isOk()) {
         const constrainedResult = await constrainSuccessResult(result.value);
 
         if (constrainedResult.type !== 'success') {
-          return {
-            isError: constrainedResult.type === 'error',
+          // Constrained result is either 'empty' or 'error'
+          const isError = constrainedResult.type === 'error';
+          success = !isError;
+          errorCode = ''; // Constrained errors don't have HTTP status
```
jarhun88: im not really sure what if any api error this should be...

anyoung-tableau: Take a look at the places where `constrainedResult.type` is `error`. It's currently only in `listPulseMetricSubscriptions` and should only happen when calling the `listPulseMetricsFromMetricIds` REST API. 

One idea would be to extend the `ConstrainedResult` type so that when its `type !== 'success'`, it must also include an `httpStatus` property too. Then you could use it here.

src/tools/tool.ts
```diff
@@ -6,14 +6,41 @@ import { Result } from 'ts-results-es';
...
+  const axiosError = error instanceof ZodiosError ? error.cause : error;
+
+  if (isAxiosError(axiosError) && axiosError.response?.status) {
+    return String(axiosError.response.status);
+  }
+
+  return '';
+}
+
+const PRODUCT_TELEMETRY_ENDPOINT = 'https://prod.telemetry.tableausoftware.com';
```
anyoung-tableau: This should be dependent on whether TMCP is running in prod vs preprod. Preprod should use `https://qa.telemetry.tableausoftware.com/`. Easiest would probably be to make `PRODUCT_TELEMETRY_ENDPOINT ` an env var that defaults to prod url.

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
     });
 
-    if (args) {
-      try {
-        (await Provider.from(this.argsValidator))?.(args);
-      } catch (error) {
-        return getErrorResult(requestId, error);
-      }
-    }
+    const config = getConfig();
```
anyoung-tableau: I was trying to keep this class independent on `config`. Consider adding `siteName` as a parameter to `logAndExecute`.

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
+      errorCode = getHttpStatus(error);
+      toolResult = getErrorResult(requestId, error);
+      return toolResult;
+    } finally {
+      // Single telemetry call - always executed
+      productTelemetry.send('tool_call', {
+        tool_name: this.name,
+        request_id: requestId.toString(),
+        session_id: sessionId ?? '',
+        site_name: config.siteName,
```
anyoung-tableau: `config.siteName` is a little tricky.

For Falcon-hosted TMCP, `siteName` will be empty so we will need to read it from the access token. The access token won't have the site name, but it will have the site luid.

For customer hosted TMCP, if they are using Tableau Server and have enabled OAuth, they will be using the authz server baked into TMCP. The access token also will only contain the site luid. Also, an empty valid for `siteName` on Server with OAuth disabled implies they are targeting the "Default" site.

In both cases we should be able to call [Get Current Server Session](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_server.htm#get-current-server-session) to get the site name if we absolutely need it vs just the luid.

src/tools/tool.ts
anyoung-tableau: Alternatively, since it's only used as part of building up the telemetry payload, consider an approach where the new parameter to `logExecute` is something like `telemetryBase` which would be an object containing the shared values of the payload, independent of result of the tool.

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
+        podname: process.env.SERVER || '',
```
anyoung-tableau: Please don't read env vars outside of config.ts. Please parameterize it, or address as part of this comment: https://github.com/tableau/tableau-mcp/pull/209/changes#r2771358686

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
+        success,
+        error_code: errorCode,
+        parameters: JSON.stringify(args ?? {}),
```
anyoung-tableau: I'm concerned about logging the args without any sanitization or anonymizing. Please check with Headless BI team ([#tab-dev-headless-bi](https://salesforce.enterprise.slack.com/archives/C0643HWPSVD)) to understand what they do before logging the HBI query.

src/tools/tool.ts
```diff
@@ -71,6 +98,9 @@ type LogAndExecuteParams<T, E, Args extends ZodRawShape | undefined = undefined>
   // The request ID of the tool call
   requestId: RequestId;
 
+  // The session ID from the transport, if available
+  sessionId: string;
```
anyoung-tableau: Consider making the type `string | undefined` to simplify the calling sites. (No more `sessionId ?? ''`)

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * Build and send a telemetry event.
+   *
+   * @param eventType - The event type/name
+   * @param serviceName - The service name emitting the event
+   * @param properties - Key-value properties for the event
+   */
+  public send(eventType: string, properties: PropertiesType): void {
+    if (process.env.PRODUCT_TELEMETRY_ENABLED === 'false') {
```
anyoung-tableau: Let's avoid accessing env vars and add an `enabled` param to the ctor which can be provided by `config.productTelemetryEnabled`

src/tools/tool.ts
jarhun88: hmm ok so preprod envs should overwrite the default value? and then in cases of self hosting it defaults to prod?

src/tools/tool.ts
jarhun88: sure thing

src/config.ts
```diff
@@ -261,7 +261,7 @@ export class Config {
       };
     } else {
       this.telemetry = {
-        provider: parsedProvider,
+        provider: 'noop',
```
anyoung-tableau, line 268: The config should be the only place env vars are read and validated. Please add a `productTelemetryEnabled` boolean field

src/tools/tool.ts
```diff
@@ -6,14 +6,41 @@ import { Result } from 'ts-results-es';
...
+
+const PRODUCT_TELEMETRY_ENDPOINT = 'https://prod.telemetry.tableausoftware.com';
+let productTelemetry: DirectTelemetryForwarder | null = null;
+
+function getProductTelemetry(): DirectTelemetryForwarder {
+  if (!productTelemetry) {
+    productTelemetry = new DirectTelemetryForwarder(PRODUCT_TELEMETRY_ENDPOINT);
+  }
+  return productTelemetry;
+}
```
anyoung-tableau: imho these should be in telemetryForwarder.ts and the DirectTelemetryForwarder class should not be exported. That maintains the "singleton" access pattern we're going for here.

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -340,6 +340,17 @@ interface TelemetryProvider {
...
+<hr />
+
+## `PRODUCT_TELEMETRY_ENABLED`
+
+Enables product telemetry for tool usage tracking.
+
+- Default: `true`
+- When `true`, the server will send telemetry events to Tableau's telemetry endpoint for each tool
+  call, including tool name, request ID, session ID, and site name.
+- Set to `false` to disable product telemetry.
```
anyoung-tableau: Let's also link to https://help.tableau.com/current/server/en-us/usage_data_basic_product_data.htm

src/telemetry/product_telemetry/telemetryForwarder.test.ts
```diff
@@ -0,0 +1,95 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
```
anyoung-tableau: These imports shouldn't be necessary.

src/tools/tool.ts
anyoung-tableau: Yep yep

src/tools/tool.ts
anyoung-tableau: If the tests need the ctor, you can export it inside an object called `exportedForTesting`. We do this in other places.

src/tools/tool.ts
```diff
@@ -181,61 +213,79 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
...
+    } finally {
+      // Single telemetry call - always executed
+      productTelemetry.send('tool_call', {
+        tool_name: this.name,
+        request_id: requestId.toString(),
+        session_id: sessionId ?? '',
+        site_name: config.siteName,
+        podname: process.env.SERVER || '',
+        success,
+        error_code: errorCode,
```
anyoung-tableau, line 277: We should maybe talk about what we want `error_code` to mean. I'm assuming this is totally left to us and not some Tableau-enforced decree...

Currently, it's mostly parroting the http status code of any failing REST API responses, and not necessarily the http status code of the response of the MCP server.

This is fine to help get these changes in quickly, but I think we should formalize error codes and define specific values for specific errors so we can more easily track issues with better granularity and can answer questions like:

* Which REST API call failed?
* Did argument validation fail? Why?
* Did tool scoping return an empty result?
* Was there some completely unhandled error? What was the stack trace?

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Build and send a telemetry event.
+   *
+   * @param eventType - The event type/name
+   * @param serviceName - The service name emitting the event
+   * @param properties - Key-value properties for the event
+   */
+  public send(eventType: string, properties: PropertiesType): void {
```
anyoung-tableau: nit: I should probably enable an eslint rule for this, but `public` modifier is unnecessary.

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const req = new Request(this.endpoint, init);
+    fetch(req)
+      .then(async (res) => {
+        const body = await res.text();
+        if (!res.ok) {
+          console.error(`[Telemetry] Failed: ${res.status} ${res.statusText}`, body);
+        }
+      })
+      .catch((error) => console.error('[Telemetry] Network error:', error));
```
anyoung-tableau: Let's not mix `async/await` syntax with `.then()/.catch()` syntax. async/await is preferable.

src/tools/tool.test.ts
```diff
@@ -1,12 +1,21 @@
...
 import { z } from 'zod';
 
 import { log } from '../logging/log.js';
 import { Server } from '../server.js';
 import invariant from '../utils/invariant.js';
 import { Tool } from './tool.js';
 
+// Mock for product telemetry - tracks calls to send()
+const mockTelemetrySend = vi.fn();
+vi.mock('../telemetry/product_telemetry/telemetryForwarder.js', () => ({
```
anyoung-tableau: nit: no underscores in folder names please.

Consider `telemetry/product/`

src/tools/tool.ts
```diff
@@ -6,14 +6,41 @@ import { Result } from 'ts-results-es';
...
+import { isAxiosError } from '../utils/axios.js';
 import { getExceptionMessage } from '../utils/getExceptionMessage.js';
 import { Provider, TypeOrProvider } from '../utils/provider.js';
 import { ToolName } from './toolName.js';
 
+/**
+ * Extracts HTTP status code from an error if available (e.g., "401", "404", "500")
+ * Returns empty string if no HTTP status can be determined
+ */
+function getHttpStatus(error: ZodiosError | Error | unknown): string {
```
anyoung-tableau: A type defined as `ZodiosError | Error | unknown` is the same as `unknown`.

Image

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const init: RequestInit = {
+      method: this.httpMethod,
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+      cache: 'default',
+      mode: 'cors',
+      credentials: 'omit',
```
anyoung-tableau: Since this code is running in Node.JS and not the browser, providing `mode: 'cors'` shouldn't be necessary. Same with `credentials`

src/tools/tool.ts
jarhun88: great idea, i went with the telemetryBase, lmk if this looks better

src/tools/tool.ts
jarhun88: ok ive set it, lmk how this looks

src/config.ts
jarhun88: ? did you mean to write this comment somewhere else? not sure what im supposed to do here

anyoung-tableau: No, this was the only place in config.ts that had changes to leave a comment on. Up above where we read all the environment variables, we should also be reading the new one you added.

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -73,21 +79,26 @@ export class DirectTelemetryForwarder {
...
         'Content-Type': 'application/json',
       },
-      cache: 'default',
-      mode: 'cors',
-      credentials: 'omit',
       body: JSON.stringify([event]),
     };
 
+    // eslint-disable-next-line no-console
+    console.log('[Telemetry] Sending event:', JSON.stringify(event, null, 2));
```
anyoung-tableau: :) The reason why we have the `no-console` eslint rule is because when the MCP server is configured to communicate over standard ouput, writing to the console will pipe a message to the MCP client. But since this is just a string and not in JSON-RPC format, the client will choke on it.

jarhun88: ah yea i accidentlly pushed this, but its only for local validation, im not planning on merging this

src/telemetry/productTelemetry/telemetryForwarder.ts
jarhun88: done

src/config.ts
jarhun88: ah understood and i think ive done so now

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -53,7 +59,7 @@ export class DirectTelemetryForwarder {
    * @param serviceName - The service name emitting the event
    * @param properties - Key-value properties for the event
    */
-  public send(eventType: string, properties: PropertiesType): void {
+  send(eventType: string, properties: PropertiesType): void {
```
anyoung-tableau: Consider changing the type of `eventType` to a string literal to avoid possible typos.

`type EventType = 'tool_call';`

In the future we can use a union of literals e.g. `'tool_call' | 'some_other_event'`

src/tools/tool.ts
jarhun88: ill speak to joe if siteLuid is fine. I feel it should be

src/tools/tool.ts
jarhun88: something for me to bring up in standup tmr...

src/tools/tool.ts
jarhun88: updated

src/telemetry/productTelemetry/telemetryForwarder.ts
anyoung-tableau: Generally, when using the singleton access pattern, you don't `export` the class so nobody other than this module can create a new instance. Since the tests use it, check out the `exportedForTesting` pattern we use elsewhere.

jarhun88: that makes sense thanks

src/tools/tool.ts
jarhun88: what if theyre using PAT?

src/tools/tool.ts
jarhun88: ive simplified it

src/tools/tool.ts
jarhun88: shoot i didnt get a chance to bring it up since we went so far over time. I'll bring it up next chance we get, but how do we feel about just submitting this as is for now to not block, and update in the future?

src/tools/tool.ts
jarhun88: so this is what the hbi team does for sanitation https://gitlab.tableausoftware.com/headless-bi/headless-bi-service/-/blob/main/src/main/java/com/tableausoftware/microservice/headlessbi/util/HBILogger.java?ref_type=heads

however i spoke to joe and he said that if all these fields are anonymized, the params field would be useless to track. So i think ill just remove this

src/tools/tool.ts
jarhun88: makes sense for error, but i wonder if empty is technically a 200 since api just couldnt return any results or they were filtered out by boundedcontext

jarhun88: ok i 

- added an error field to the error constrainedResult
- refactored PulseDisabledError into a new class PulseDisabledError which stores the http status, 
- moevd getHttpStatus to its own file, and added a check for the PulseDsabledError to extract the errorocde specifically
- updated pulse  tools to pass error object
- udpated tool.ts to call getHttpStatus from constrained result

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,123 @@
...
+      method: 'PUT',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify([event]),
+    };
+
+    const req = new Request(this.endpoint, init);
+    sendTelemetryRequest(req);
```
anyoung-tableau, line 81: Consider leaving a comment here that describes that the lack of `await` is intentional as to not block execution while waiting for the telemetry request to finish. I say this because I have been wanting to enable the [no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/) eslint rule which will flag this line. We'd want to ignore the eslint error, and not fix it by adding an `await`.

src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!res.ok) {
+      console.error(`[Telemetry] Failed: ${res.status} ${res.statusText}`, body);
+    }
+  } catch (error) {
+    console.error('[Telemetry] Network error:', error);
+  }
+}
+
+const getDefaultPod = (): string => {
+  return process.env.SERVER || '';
```
anyoung-tableau: Please pass the podname into the constructor so we're not reading env vars outside of the `config`

src/tools/tool.ts
anyoung-tableau: Yeah, that's more than fine.

src/utils/getSiteLuidFromAccessToken.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Extracts the site LUID from a Tableau access token.
+ * The access token format is: part0|part1|siteLuid|...
+ *
+ * @param accessToken - The Tableau access token
+ * @returns The site LUID, or empty string if it cannot be extracted
+ */
+export function getSiteLuidFromAccessToken(accessToken: string | undefined): string {
```
anyoung-tableau, line 8: Please update [RestApi.setCredentials](https://github.com/tableau/tableau-mcp/blob/main/src/sdks/tableau/restApi.ts#L234) to use this function

src/telemetry/productTelemetry/telemetryForwarder.ts
jarhun88: done

src/utils/getSiteLuidFromAccessToken.ts
jarhun88: done

anyoung-tableau (approved):

PR #212, opened 2026-01-28
@W-20797528: Implement tag-based scoping

anyoung-tableau:
These changes extend tool scoping to include the concept of [tags](https://help.tableau.com/current/pro/desktop/en-us/tags.htm).

When tag-based tool scoping is enabled:

1. Tools that query a data source, workbook, or view can only be called if the asset has one of the allowed tags.
2. Tools can only return information for data sources, workbooks, and views when those assets have one of the allowed tags.

Examples:

```
INCLUDE_TAGS=sales
```

1. Call `query-datasource` tool using a data source ID for a data source that does not have the `sales` tag: error.
2. Call `get-view-data` tool using a view ID for a view that does not have the `sales` tag: error.
3. Call `list-datasources` tool: only return data sources with the `sales` tag.
4. Call `search-content` tool: only return content with the `sales` tag.
5. Call `get-workbook` using a workbook ID for a workbook that has the `sales` tag: return the workbook but its list of views contains only the views that also have the `sales` tag. Adding a tag to a workbook automatically adds the same tag to all of its views, however that tag *can* be removed from each view individually.

### DIFF
--- a/src/sdks/tableau/types/view.ts
+++ b/src/sdks/tableau/types/view.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 
-import { tagsSchema } from './tags';
+import { tagsSchema } from './tags.js';
 
 export const viewSchema = z.object({
   id: z.string(),


src/tools/resourceAccessChecker.ts
```diff
@@ -271,33 +339,32 @@ class ResourceAccessChecker {
...
-
-        if (!this.allowedWorkbookIds.has(viewWorkbookId)) {
+        view = await getView();
+
+        if (!this.allowedWorkbookIds.has(view.workbook?.id ?? '')) {
           return {
             allowed: false,
             message: [
-              'The set of allowed workbooks that can be queried is limited by the server configuration.',
+              'The set of allowed views that can be queried is limited by the server configuration.',
```
anyoung-tableau, line 367: These driveby changes are to fix what I felt was a bug. All the other asset types reference themselves, not the type of scoping applied.

src/tools/resourceAccessChecker.test.ts
```diff
@@ -1,5 +1,6 @@
 import { getConfig } from '../config.js';
 import { Server } from '../server.js';
+import { getCombinationsOfBoundedContextInputs } from '../utils/getCombinationsOfBoundedContextInputs.js';
```
anyoung-tableau, line 3: I refactored this entire file. Before, we had 1 test for each combination of bounded context inputs, now we leverage [test.each](https://vitest.dev/api/#test-each) to run the same test with different variables, once for each combination of bounded context inputs.

src/sdks/tableau/types/view.ts
```diff
@@ -1,5 +1,7 @@
 import { z } from 'zod';
 
+import { tagsSchema } from './tags';
```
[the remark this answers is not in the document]
anyoung-tableau: Weird, not sure why the IDE didn't add the extension like it normally does.

src/scripts/createClaudeMcpBundleManifest.ts
```diff
@@ -249,6 +249,14 @@ const envVars = {
     required: false,
     sensitive: false,
   },
+  INCLUDE_TAGS: {
```
jarhun88, line 252: what is this file?

src/tools/workbooks/getWorkbook.ts
```diff
@@ -103,3 +99,32 @@ export const getGetWorkbookTool = (server: Server): Tool<typeof paramsSchema> =>
...
+  if (!workbook.views || !tags) {
+    return {
+      type: 'success',
+      result: workbook,
+    };
+  }
+
+  workbook.views.view = workbook.views.view.filter((view) =>
+    view.tags?.tag?.some((tag) => tags.has(tag.label)),
+  );
```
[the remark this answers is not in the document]
anyoung-tableau: It's fine as it is.

src/scripts/createClaudeMcpBundleManifest.ts
anyoung-tableau: It's used to generate the manifest json file for the Claude Desktop Extension. See https://tableau.github.io/tableau-mcp/docs/getting-started#claude-desktop-extension

src/tools/resourceAccessChecker.ts
```diff
@@ -316,25 +383,9 @@ class ResourceAccessChecker {
...
-              });
-
-              return view.project?.id ?? '';
-            },
-          }));
-
-        if (!this.allowedProjectIds.has(viewProjectId)) {
+        view = view ?? (await getView());
+
+        if (!this.allowedProjectIds.has(view.project?.id ?? '')) {
```
jarhun88, line 388: curious if the view project id is '' will it just auto fail or succeed anyways?

src/tools/listDatasources/listDatasources.ts
```diff
@@ -157,6 +157,12 @@ export function constrainDatasources({
     datasources = datasources.filter((datasource) => datasourceIds.has(datasource.id));
   }
 
+  if (tags) {
```
jarhun88, line 160: i notice this pattern repeats a good few times, wondering if theres value in making some function that does the filtering for any boundedContext object?

src/tools/resourceAccessChecker.ts
anyoung-tableau: In practice, I don't think `view.project?.id` can ever be undefined since I don't think a view can exist outside of a project. But even if the ID was `''`, `allowedProjectIds` cannot contain an empty string.

src/tools/listDatasources/listDatasources.ts
anyoung-tableau: Hmm, we repeat it for `listDatasources`, `listWorkbooks`, and `listViews`, but each one is unique to the content type. Not sure if there's value in abstracting that out into a helper method since it's different for each content type.

src/tools/resourceAccessChecker.ts
jarhun88: This is probably a nit so ff to ignore but would it then be cleaner to do smth like this to avoid the magic string?

if (!view.project?.id || !this.allowedProjectIds.has(view.project?.id)

anyoung-tableau: This isn't new code (notice how `viewProjectId` was initialized before) so I don't feel strongly any change is needed.

PR #214, opened 2026-02-04
@W-19426562: Don't provide rowLimit option to VDS before 2026.1

anyoung-tableau:
Customer report here: https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1770197628514829

Fixes https://github.com/tableau/tableau-mcp/issues/215

### DIFF
--- a/src/tools/queryDatasource/queryDatasource.test.ts
+++ b/src/tools/queryDatasource/queryDatasource.test.ts
@@ -70,6 +70,7 @@ describe('queryDatasourceTool', () => {
     resetDatasourceCredentials();
     resetResourceAccessCheckerSingleton();
     mocks.mockGetConfig.mockReturnValue({
+      server: 'https://10ax.online.tableau.com',
       datasourceCredentials: undefined,
       boundedContext: {
         projectIds: null,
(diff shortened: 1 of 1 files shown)


src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -103,12 +103,22 @@ export const getQueryDatasourceTool = (
...
+                disaggregate: false,
+                rowLimit, // rowLimit can only be provided in 2026.1.0 and later
+              } as const,
+              default: {
+                returnFormat: 'OBJECTS',
+                debug: true,
+                disaggregate: false,
+              } as const,
+            },
+          });
```
anyoung-tableau, line 121: Unfortunately, I've learned that checking the Tableau product version can potentially be insufficient since HBI deploys separately. For example, I was testing this against prod-uswest-c. It is running 2026.1 but HBI was still complaining about the `rowLimit` option. HBI team agreed to looking into potentially adding a new endpoint that would server the OpenAPI spec that we could check to see what request format it supports.

See thread: https://salesforce-internal.slack.com/archives/C0643HWPSVD/p1770231781612089

PR #216, opened 2026-02-09
@W-21200034: Add scaffolding to override some config entries at runtime

anyoung-tableau:
These changes lay the groundwork for [@W-21200059: Add Get/Update MCP site settings REST APIs](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002UNWjMYAX/view).

Once the "Get MCP Site Settings" REST API is available, we can call it to get the actual site settings. Currently, the `ENABLE_MCP_SITE_SETTINGS` feature is disabled but can be enabled for testing. When enabled, the method that retrieves the MCP site settings returns a placeholder empty object (implying no overrides).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.14.7",
+  "version": "1.14.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.14.7",
+      "version": "1.14.8",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.25.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.14.7",
+  "version": "1.14.8",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/sdks/tableau/getSiteIdFromAccessToken.ts
```diff
@@ -0,0 +1,9 @@
+export function getSiteIdFromAccessToken(accessToken: string): string {
+  const parts = accessToken.split('|');
+  if (parts.length < 3) {
+    throw new Error('Could not determine site ID. Access token must have 3 parts.');
+  }
+
+  const siteId = parts[2];
+  return siteId;
+}
```
anyoung-tableau: Resolve this with the new function from https://github.com/tableau/tableau-mcp/pull/209

src/tools/resourceAccessChecker.ts
```diff
@@ -38,48 +31,78 @@ class ResourceAccessChecker {
...
+  }): Promise<Set<string> | null> {
+    return (
+      this._testOverrides.projectIds ??
+      (
+        await getConfigWithOverrides({
+          restApiArgs,
+        })
+      ).boundedContext.projectIds
+    );
   }
```
anyoung-tableau, line 61: None of the scoping logic has changed. The changes in this file are so we retrieve the MCP site settings to see if a tool scoping override exists. Remember, the response from the REST API is cached for the site for n=10 minutes so once we call it once, all subsequence responses come from cache for the next 10 minutes.

src/config.test.ts
```diff
@@ -189,68 +163,28 @@ describe('Config', () => {
...
-      const config = new Config();
-      expect(config.includeTools).toEqual(['query-datasource']);
-    });
-
-    it('should filter out invalid tool names from EXCLUDE_TOOLS', () => {
-      vi.stubEnv('EXCLUDE_TOOLS', 'query-datasource,order-hamburgers');
+  it('should set mcpSiteSettingsCheckIntervalInMinutes to default when not specified', () => {
+    const config = new Config();
+    expect(config.mcpSiteSettingsCheckIntervalInMinutes).toBe(10);
+  });
```
anyoung-tableau, line 169: Diff is kinda bad in this file. I moved the tests for overridable config to their own file and added tests for `MCP_SITE_SETTINGS_CHECK_INTERVAL_IN_MINUTES` and `ENABLE_MCP_SITE_SETTINGS`.

src/config.ts
```diff
@@ -58,16 +47,13 @@ export class Config {
   datasourceCredentials: string;
   defaultLogLevel: string;
   disableLogMasking: boolean;
-  includeTools: Array<ToolName>;
-  excludeTools: Array<ToolName>;
```
anyoung-tableau, line 62: I had initially left all overridable config items in the `Config` class, leaving the responsibility of checking the site settings for an updated value on the developer. I felt this would be prone to bugs so I moved all the overridable items to `OverridableConfig` class instead.

src/overridableConfig.test.ts
```diff
@@ -0,0 +1,332 @@
+import { exportedForTesting } from './overridableConfig.js';
+import { stubDefaultEnvVars } from './testShared.js';
+
+describe('OverridableConfig', () => {
```
anyoung-tableau, line 4: All these tests are copy/paste from `config.test.ts`

src/overridableConfig.ts
```diff
@@ -0,0 +1,182 @@
...
+  return overridableVariables.some((v) => v === variable);
+}
+
+function filterEnvVarsToOverridable(
+  environmentVariables: Record<string, string | undefined>,
+): Record<OverridableVariable, string | undefined> {
+  return Object.fromEntries(
+    Object.entries(environmentVariables).filter(([key]) => isOverridableVariable(key)),
+  ) as Record<OverridableVariable, string | undefined>;
+}
```
anyoung-tableau, line 29: This is the magic that ensures only the environment variables that we have specified as overridable get overridden.

src/restApiInstance.ts
```diff
@@ -28,46 +28,68 @@ type JwtScopes =
   | 'tableau:metric_subscriptions:read'
   | 'tableau:insights:read'
   | 'tableau:views:download'
-  | 'tableau:insight_brief:create';
+  | 'tableau:insight_brief:create'
+  | 'tableau:mcp_site_settings:read';
```
anyoung-tableau, line 32: This is currently a "fake" scope but will be the one we use when adding the new Get MCP Site Settings REST API

src/sdks/tableau/types/mcpSiteSettings.ts
```diff
@@ -0,0 +1,4 @@
+import { z } from 'zod';
+
+export const mcpSiteSettingsSchema = z.record(z.string(), z.string());
+export type McpSiteSettings = z.infer<typeof mcpSiteSettingsSchema>;
```
anyoung-tableau, line 4: The idea is that the "Get MCP Site Settings" REST API will return a JSON object in this format, representing key-value pairs of environment variable names and their values

```
{
  "EXCLUDE_TOOLS": "query-datasource",
  "INCLUDE_DATASOURCE_IDS": "2d935df8-fe7e-4fd8-bb14-35eb4ba31d4"
}
```

src/restApiInstance.ts
```diff
@@ -28,46 +28,68 @@ type JwtScopes =
...
+  signal: AbortSignal;
+  authInfo?: TableauAuthInfo;
+} & (
+  | {
+      requestId: RequestId;
+      disableLogging?: false;
+    }
+  | {
+      disableLogging: true;
+    }
```
anyoung-tableau, line 46: If it's not clear, the `requestId` is only used in logging. This type makes it so you can't provide it when `disableLogging` is `true`. TypeScript rules and I'll die on this hill. :)

src/tools/pulse/listMetricsFromMetricIds/listPulseMetricsFromMetricIds.ts
```diff
@@ -44,26 +45,38 @@ Retrieves a list of published Pulse Metrics from a list of metric IDs using the
...
+        constrainSuccessResult: async (metrics) => {
+          const configWithOverrides = await getConfigWithOverrides({
+            restApiArgs,
+          });
+
+          return constrainPulseMetrics({
+            metrics,
+            boundedContext: configWithOverrides.boundedContext,
+          });
+        },
```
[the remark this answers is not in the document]
anyoung-tableau: You are confused Copilot

src/restApiInstance.ts
jarhun88: by fake scope do you mean its just a placeholder? or is it in fact correct but the api just doesnt exist yet?  might adding a comment would help explain

src/restApiInstance.ts
```diff
@@ -28,46 +28,68 @@ type JwtScopes =
...
-        { logger: server.name, requestId },
-      );
-    },
-    { once: true },
-  );
+  const { config, server, jwtScopes, signal, authInfo, disableLogging } = args;
+
+  if (!disableLogging) {
+    const { requestId } = args;
+    signal.addEventListener(
```
jarhun88, line 58: what might trigger the abort? cancelled apis?

src/restApiInstance.ts
anyoung-tableau: Yeah, this scope doesn't exist in Tableau yet. It's a placeholder for now until it becomes available as part of the changes to add the REST APIs.

src/restApiInstance.ts
anyoung-tableau: Yeah, MCP clients can [request pending requests get cancelled](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation). The MCP SDK handles those requests by triggering the abort signal.

You can see this in action in MCP Inspector if you configure it to have a very short timeout, like 1 ms.

src/restApiInstance.ts
anyoung-tableau: Dunno if we need a comment since we'll just have to remove it in a week or 2 when the REST APIs land. Jill is working on that right now.

src/restApiInstance.ts
anyoung-tableau: https://github.com/modelcontextprotocol/typescript-sdk/blob/42b335953a2e3c744276c3e2a41c2f330baa8741/packages/core/src/shared/protocol.ts#L440

https://github.com/modelcontextprotocol/typescript-sdk/blob/42b335953a2e3c744276c3e2a41c2f330baa8741/packages/core/src/shared/protocol.ts#L636

jarhun88: oh interesting, so the typescript lib will automatically receive the cancelled event that the mcp would generate and throw the abort event.

src/restApiInstance.ts
jarhun88: ok if its that soon then yea doesnt matter

src/config.ts
jarhun88: can you expand abit on this? I was just contemplating on how devs now will need to be aware that there are 2 sets of configs, which have different values. Not bad per say, but i am curious towards what bugginess this extraction would reduce?

jarhun88 (approved):

src/restApiInstance.ts
anyoung-tableau: Yep, and we plumb that `AbortSignal` instance all the way down into the instance of Axios that is responsible for making the REST API requests. That means if a client ever sends a cancellation notification, any pending REST API requests will also get cancelled.

src/config.ts
anyoung-tableau: If we kept the overridable values in `Config` then we would either have to:

1. Call the "Get MCP Site Settings" REST API every time we need to get the config to determine whether overrides exist, even if the only config entries that apply are non-overridable. This is wasteful.
2. OR, we force the developer to make the decision in-advance on whether a config entry they need to access is potentially overridable and call the "Get MCP Site Settings" REST API first to determine the override values. This is prone to bugs because a developer could forget and then an override wouldn't take effect when it should have.

By moving the overridable entries into their own class, we get bullet 1 above, but ONLY for the config entries that are overridable. We don't make wasteful REST API calls when accessing entries that can't be overridden. We also remove the bug possibility of bullet 2 since the prescribed way of getting the overridable config entries is through `getConfigWithOverrides` which calls the REST API for the developer.

PR #219, opened 2026-02-11
@W-20742343 adding new is_hyperforce column to product telemetry

jarhun88:

the pull request._

# Pull Request Template

## Description

adding new product telemetry field is_hyperforce which tracks whether or not event data is coming from within hyperforce environment or externally.

Default will be false. I dont think theres a need to document this env var publically as all it does is set the telemetry column.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.14.6",
+  "version": "1.14.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.14.6",
+      "version": "1.14.7",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.25.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.14.6",
+  "version": "1.14.7",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/config.ts
```diff
@@ -150,6 +151,7 @@ export class Config {
       TELEMETRY_PROVIDER_CONFIG: telemetryProviderConfig,
       PRODUCT_TELEMETRY_ENDPOINT: productTelemetryEndpoint,
       PRODUCT_TELEMETRY_ENABLED: productTelemetryEnabled,
+      IS_HYPERFORCE: isHyperforce,
```
anyoung-tableau, line 154: Just checking if leaving this out of `process-env.d.ts` was intentional since it won't be documented to customers? I think there is some value in adding it since it:

1. gives a hint to the IDE so we get autocomplete on `process.env` since we extend the global interface.
2. is good practice because it forces us to update `createClaudeMcpBundleManifest` to add the new env var so we don't accidentally omit an env var we might *want* to let users configure in Claude Desktop when installing the extension.

It's fine as is since this is an "internal" env var, but we could get bullet 1 above by adding another interface just for these internal env vars.

Image

src/tools/contentExploration/searchContent.ts
```diff
@@ -103,6 +103,7 @@ This tool searches across all supported content types for objects relevant to th
           endpoint: config.productTelemetryEndpoint,
           siteLuid: getSiteLuidFromAccessToken(getTableauAuthInfo(authInfo)?.accessToken),
           podName: config.server,
+          isHyperforce: config.isHyperforce,
           enabled: config.productTelemetryEnabled,
         },
```
anyoung-tableau: I foresee us ultimately converting this object creation into a helper function since it's the same for all tools. If you feel the same, consider including that in this PR  (later is fine too)

src/config.ts
jarhun88: it was intentional cuz i wasnt sure the value, but sure i can add it, only takes a few seconds

src/tools/contentExploration/searchContent.ts
jarhun88: should be relatively easy to do

PR #223, opened 2026-02-13
@W-21264036: Provide common context to tools

anyoung-tableau:
Over time, I've noticed that each tool has a lot of the same boilerplate which has grown what I'd considered bloated. These changes are an attempt to help reel it in a bit and provide these common items to each tool automatically. These are things like the environment `config`, the `tableauAuthInfo`, and the instance of the `server`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.15.2",
+  "version": "1.15.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.15.2",
+      "version": "1.15.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.15.2",
+  "version": "1.15.3",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/telemetry/productTelemetry/telemetryForwarder.ts
```diff
@@ -125,28 +120,6 @@ export function getProductTelemetry(
...
-  authInfo: AuthInfo | undefined,
-): ProductTelemetryBase {
-  return {
-    endpoint: config.productTelemetryEndpoint,
-    siteLuid: getSiteLuidFromAccessToken(getTableauAuthInfo(authInfo)?.accessToken),
-    podName: config.server,
-    isHyperforce: config.isHyperforce,
-    enabled: config.productTelemetryEnabled,
-  };
-}
```
anyoung-tableau, line 148: @jarhun88 I rolled this into `Tool.logAndExecute()` since it now has access to `config` and the `productTelemetryBase` no longer needs to be explicitly provided.

src/server.ts
```diff
@@ -53,22 +61,41 @@ export class Server extends McpServer {
...
+          ...extra,
+          config,
+          server: this,
+          tableauAuthInfo,
+          getConfigWithOverrides: async () =>
+            getConfigWithOverrides({ restApiArgs: tableauRequestHandlerExtra }),
+        };
+
+        return tableauToolCallback(args, tableauRequestHandlerExtra);
+      };
```
anyoung-tableau, line 89: Here's the magic that extends the normal `extra` object the SDK provides and attaches the additional shared stuff like `config` and `server` to it.

src/tools/toolContext.ts
```diff
@@ -0,0 +1,30 @@
...
+export type TableauToolContext = {
+  config: Config;
+  server: Server;
+  tableauAuthInfo: TableauAuthInfo | undefined;
+  getConfigWithOverrides: () => Promise<OverridableConfig>;
+};
+
+// An extension of the RequestHandlerExtra type that includes the TableauToolContext
+export type TableauRequestHandlerExtra = TableauToolContext &
+  RequestHandlerExtra<ServerRequest, ServerNotification>;
```
anyoung-tableau, line 25: The MCP WG refactored the `RequestHandlerExtra` type [last week](https://github.com/modelcontextprotocol/typescript-sdk/pull/1467) (unreleased as of today) which is a minor breaking change but it shouldn't be too bad. Take a peek at the delta outlined in the table in the [migration skill](https://github.com/modelcontextprotocol/typescript-sdk/pull/1467/changes#diff-9610a7bb06e139c019b8f74a7a0f6da1d9a1994c807839bb8bcbe64ebe8af043) from that PR.

src/telemetry/productTelemetry/telemetryForwarder.ts
yyk16293: Cool

jarhun88 (approved):

PR #227, opened 2026-02-20
@W-20680233 Add new "warnings" field in queryDatasource response schema to capture ContextFilterMissing Validation Errors

jarhun88:

the pull request._

# Pull Request Template

## Description

Adds a non-blocking warning system to queryDatasource that detects when TOP/BOTTOM filters are combined with dimension filters missing context: true. In Tableau's Order of Operations, dimension filters without context run alongside TOP rather than before it, causing the TOP filter to rank against the full dataset instead of the filtered subset — leading to unexpected results like null values or fewer rows than expected.

Changes

New: src/tools/queryDatasource/validators/validateContextFilters.ts
- validateContextFilters(query) inspects the query's filters and returns warnings when dimension filters (SET, DATE, MATCH, or non-aggregated QUANTITATIVE) are used alongside TOP/BOTTOM without context: true
Correctly distinguishes dimension filters from aggregated measure filters (QUANTITATIVE with a function on the field), since measure filters execute after TOP in Tableau's Order of Operations and don't need context
Returns structured ContextFilterWarning objects with the warning type, severity, human-readable message, and list of affected filter names

Modified: src/tools/queryDatasource/queryDatasource.ts
- Runs validateContextFilters before the API call to avoid unnecessary processing if warnings are needed
Attaches warnings to the successful response payload when present, preserving backward compatibility (no warnings field when there are none)

New: src/tools/queryDatasource/validators/validateContextFilters.test.ts
- 19 unit tests covering: no-warning cases (no filters, single filter, existing context, aggregated measure filters), and warning cases (SET, DATE, MATCH, dimension QUANTITATIVE_NUMERICAL, QUANTITATIVE_DATE, BOTTOM direction, multiple affected filters, context: false, calculated fields, mixed dimension/measure filters)

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tableau/mcp-server",
-  "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
+  "description": "Helping agents see and understand data.",
   "version": "1.17.3",
   "repository": {
     "type": "git",


src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: Consider changing the generic type param `QueryOutput` to `QueryOutput & { warnings?: ContextFilterWarning[] }`. Doing this will give the `constrainSuccessResult` function the correct type that includes the `warnings` key.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -184,6 +187,13 @@ export const getQueryDatasourceTool = (
                 result.value.data.length = rowLimit;
               }
 
+              if (contextWarnings.length > 0) {
+                return new Ok({
+                  ...result.value,
+                  warnings: contextWarnings,
```
anyoung-tableau: Hmm... if HBI ever decides to add a `warning` key to its output, we're going to overwrite with ours. Let's try to proactively avoid that bug and also set precedent for adding a warning to any tool response.

Thoughts on either of these?:

```ts
return new Ok({
  ...result.value,
  mcp: {
    warnings: contextWarnings,
  },
});
```

or

```ts
return new Ok({
  result: result.value,
  warnings: contextWarnings,
});
```

anyoung-tableau: I would vote for the 2nd one.

jarhun88: i think i need to go with 1, because if i do 2, this tools response object is going to look different than all others. Which im not sure if that inconsistency is going to be a problem in the future. 

{
  "result": { "data": [...] },
}

vs 

{
  "data": [...]
}

src/tools/queryDatasource/queryDatasource.ts
jarhun88: i updated the type

src/tools/queryDatasource/queryDatasource.ts
anyoung-tableau: Approach 1 is fine, but tool response shape really doesn't matter. The nice thing about MCP is that you can ship "breaking" changes to APIs because agents will just run with it. They will automatically adapt.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "An MCP server for Tableau, providing a suite of tools that will make it easier for developers to build AI applications that integrate with Tableau.",
-  "version": "1.15.3",
+  "version": "1.15.5",
```
anyoung-tableau: Please pull, merge, and bump the version.

jarhun88: done

src/tools/queryDatasource/queryDatasource.ts
jarhun88: that is pretty impressive that it can do that i will not lie

PR #234, opened 2026-02-27
Add passthrough auth mode for gateway/proxy deployments

karimtabet:
## Summary

- Add `AUTH=passthrough` mode that trusts pre-authenticated credentials from HTTP headers (`Authorization: Bearer <token>` + `X-Tableau-User-Id: <user-luid>`)
- Enable enterprise gateway/proxy deployments where users authenticate via Kerberos, OIDC, or other mechanisms and the proxy forwards a valid Tableau REST API token
- Reuse the existing `restApi.setCredentials()` code path (same as OAuth) — no new dependencies or Tableau API changes needed

## Motivation

There is currently no way for a trusted upstream proxy to pass per-user credentials to the MCP server via HTTP headers. This is the standard pattern in enterprise gateway deployments where:

- A reverse proxy authenticates users (via Kerberos, mTLS, OIDC, etc.)
- The proxy acquires a Tableau-compatible token on behalf of the user
- The proxy forwards the token to the MCP server in an HTTP header

This is particularly important for enterprises running Tableau Server versions that don't support OAuth (< 2025.3), where Kerberos SPNEGO authentication produces a `workgroup_session_id` that works as a standard REST API bearer token.

| Deployment | Auth Mode | How Users Authenticate |
|---|---|---|
| Single user, CLI | `pat` | Static PAT in env vars |
| Multi-user, Tableau 2025.3+ | `oauth` | Browser redirect |
| Multi-user, Connected Apps | `direct-trust` | JWT via Connected App |
| **Multi-user, enterprise gateway** | **`passthrough`** | **Proxy-provided token** |

## Changes

- **`src/config.ts`**: Add `'passthrough'` to valid auth types, auto-detect transport as `'http'`, exempt from `OAUTH_ISSUER` requirement, validate transport must be HTTP
- **`src/server/express.ts`**: Add `getPassthroughAuthInfo()` to extract credentials from request headers, add `X-Tableau-User-Id` to CORS allowed headers, wire up passthrough validation middleware
- **`src/restApiInstance.ts`**: Check for pre-authenticated credentials (`accessToken` + `userId`) before the auth-type switch, skip sign-out for passthrough sessions
- **`src/server/middleware.ts`**: Add `validatePassthroughHeaders()` middleware returning 401 JSON-RPC error on missing headers

## Test plan

- [x] 7 new config tests: passthrough auth creation, transport auto-detection, stdio rejection, OAuth/PAT requirements relaxed, SERVER still required
- [x] 4 new middleware tests: 401 on missing Authorization, missing X-Tableau-User-Id, both missing; passes through when both present
- [x] 1 new restApiInstance test: verifies `setCredentials` called (not `signIn`), `signOut` skipped
- [x] Full test suite passes: 896 tests across 52 files, 0 failures
- [x] TypeScript compilation clean

## Security considerations

- `AUTH=passthrough` trusts the `Authorization` header completely — same trust model as a reverse proxy terminating TLS or Kerberos
- The MCP server should NOT be exposed directly to end users in passthrough mode; it must sit behind a trusted proxy/gateway
- This follows the standard pattern used by many enterprise services (Kubernetes API server, Envoy ext_authz, etc.)


src/server/express.ts
```diff
@@ -10,14 +10,37 @@ import { Config } from '../config.js';
...
 import { AuthenticatedRequest } from './oauth/types.js';
 
 const SESSION_ID_HEADER = 'mcp-session-id';
 
+function getPassthroughAuthInfo(req: Request, config: Config): TableauAuthInfo | undefined {
+  if (config.auth !== 'passthrough') {
+    return undefined;
+  }
+
+  const authHeader = req.headers['authorization'];
```
anyoung-tableau: @karimtabet would you be willing to pass the access token on the `X-Tableau-Auth` header instead of the `Authorization` header? The scenario we want to support is one where we *don't* introduce a new `passthrough` value for `config.auth`, but for requests that have the `X-Tableau-Auth` we ignore the normal auth mechanism and let the value of `X-Tableau-Auth` pass through. That way clients that don't provide the `X-Tableau-Auth` header will use PAT/Direct Trust/OAuth/etc and clients that do provide the header will use the provided access token. Thoughts?

anyoung-tableau: Hey @karimtabet , please let me know your thoughts, otherwise I will move forward with the `X-Tableau-Auth` header approach seen in #241

bbartels: I let karim know to have a look, your comments seem sensible

karimtabet: Hey @anyoung-tableau, sorry for the delay! I've updated the PR to use the X-Tableau-Auth header approach as you suggested — removed the dedicated passthrough auth type and replaced it with an ENABLE_PASSTHROUGH_AUTH flag that works alongside existing auth modes. Let me know if this aligns with what you had in mind, happy to adjust.

anyoung-tableau: Thanks @karimtabet, last thing I wanted to check... are you already utilizing or planning to utilize these changes soon in some way? I'd like to move forward, but only if there's an active use case, otherwise it's idle code we need to maintain.

bbartels: Yep, we want to run the tableau in a multi-tenant way, where no interactive flow is necessary in order to auth with tableau. This is useful in scenarios where you have headless agents that you want to spin up at scale without user interaction. In our case we do something fairly complicated to fetch PAT's in a headless manner (fetch session token via kerberos, then use session token to create a pat). But this would be the last missing piece for us to run this as desired. The alternative is we would fork and create our own artefacts, but we operate in an airgapped environment so thats a lot trickier than it should be.

bbartels: We use this type of flow for all other mcps we deploy already, tableau being one of the ones that don't currently support passthrough flows

anyoung-tableau: Cool, thanks guys. Would you be willing to close this PR in favor of us moving forward with #241 instead?

karimtabet: Fine by me.

PR #236, opened 2026-03-02
@W-21415611: Only advertise client secret auth methods when secret pairs are defined

### LINKED ISSUE #232: Update .well-known/oauth-authorization-server's data according to current env vars

Cherry studio 1.7.21 can't connect to a Tableau MCP Server via streamable http and OAuth if Tableau MCP doesn't have client ID/secret pairs in the env vars because it tries to connect using **client_secret_basic** authentication.

When trying to connect to Cherry Studio 1.7.21 via streamable http I had the OAuth page opening in my browser, OAuth worked but after the redirection to Cherry studio, I had this error message : 
Image

Removing `"client_secret_basic"` in `src/server/oauth/.well-known/oauth-authorization-server.ts` and only leaving `"client_secret_post"` solved the problem.

It would be great if this field could have some tiny bit of logic checking if client secret related env vars are set so that it could add or not the corresponding auth method to the supported list.

anyoung-tableau:
Fixes #232

### DIFF
--- a/tests/oauth/oauth.test.ts
+++ b/tests/oauth/oauth.test.ts
@@ -25,6 +25,7 @@ describe('OAuth', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+    vi.unstubAllEnvs();
     _server = undefined;
   });
 


src/server/oauth/token.ts
```diff
@@ -40,20 +40,24 @@ export function token(
...
-        error: 'invalid_client',
-        error_description: clientCredentialsResult.error,
+    let clientCredentialClientId = '';
+    if (config.oauth.clientIdSecretPairs) {
+      const clientCredentialsResult = verifyClientCredentials({
+        clientId: result.data.clientId,
+        clientSecret: result.data.clientSecret,
+        clientIdSecretPairs: config.oauth.clientIdSecretPairs,
+        authorizationHeader: req.headers.authorization,
       });
```
anyoung-tableau, line 50: Driveby fix I've been meaning to make here. Client creds should always be validated for confidential clients, not just when the `grant_type` is `client_credentials`. See https://www.rfc-editor.org/rfc/rfc6749#section-3.2.1

tests/oauth/oauth.test.ts
```diff
@@ -117,6 +117,37 @@ describe('OAuth', () => {
     });
   });
 
+  it('should provide a authorization server metadata endpoint for the OAuth 2.1 flow without client credentials', async () => {
+    vi.stubEnv('OAUTH_CLIENT_ID_SECRET_PAIRS', '');
```
mattcfilbert, line 122: Worth confirming that vi.stubEnv doesn’t leak into other tests here? Or since we have "fileParallelism: false" it’s all good? 

I’ve seen stubs stick around without a reset and am potentially over-paranoid

anyoung-tableau: Oh, it does bleed actually. It doesn't cause any problems, but thanks, good catch.

PR #229, opened 2026-02-21
fix: LIST parameter members schema to match VDS response format

mnoles-vc:
**IMPORTANT: Issue created first — see #228.**

## Description

Fix `get-datasource-metadata` crashing with `Cannot read properties of undefined (reading 'map')` for any datasource containing LIST parameters.

The VDS `read-metadata` endpoint returns LIST parameter `members` as an array of objects (`[{value: "Daily"}, ...]`) but the Zod schema expected an array of primitives (`["Daily", ...]`). The `.strict()` modifier on the discriminated union causes Zod validation to fail when parsing the response.

## Motivation and Context

Any published datasource with LIST parameters (e.g., dynamic dimension selectors, date granularity pickers) is completely unusable with `get-datasource-metadata`. This blocks schema discovery for datasources that use parameters.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe):

## How Has This Been Tested?

Tested against Tableau Cloud datasources before and after the fix:

| Datasource | Parameters | Before | After |
|---|---|---|---|
| Published DS (28 fields, 0 params) | None | ✅ | ✅ |
| Same DS after adding 1 LIST param (28 fields, 1 param) | `Choose Date Granularity`: Daily, Monthly, Quarterly, Yearly | ❌ Crashes | ✅ 28 fields, 1 parameter returned correctly |
| Published DS with 2 LIST params | `Choose Dimension`, `Choose Date Granularity` | ❌ Crashes | ✅ All fields and parameters returned correctly |

## Related Issues

Fixes #228

## Checklist

- [x] I have updated the version in the package.json file by using `npm run version`. For example, use `npm run version:patch` for a patch version bump.
- [x] I have made any necessary changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have documented any breaking changes in the PR description. For example, renaming a config environment variable or changing its default value.

## Contributor Agreement

- [x] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) for this project and followed its Contribution Checklist.

## Changes

3 files, 5 lines changed:

- **`src/sdks/tableau/apis/vizqlDataServiceApi.ts`** — Update LIST parameter schema to expect `{value: x}` objects
[...]

Issue #228, opened 2026-02-20
get-datasource-metadata crashes on datasources with LIST parameters due to Zod schema mismatch


mnoles-vc:
### Environment
- `@tableau/mcp-server` v1.15.0
- Tableau Cloud

### Description

`get-datasource-metadata` throws an uncaught error for any datasource that contains LIST-type parameters. Datasources without parameters work fine.

### Steps to Reproduce

1. Publish a datasource to Tableau Cloud **without** any parameters
2. Run `get-datasource-metadata` → ✅ succeeds
3. Add a LIST parameter (e.g., "Choose Date Granularity" with values Daily/Monthly/Quarterly/Yearly) and republish
4. Run `get-datasource-metadata` again → ❌ fails with `Cannot read properties of undefined (reading 'map')`

Same datasource, only difference is the presence of a LIST parameter.

### Root Cause

The VDS `read-metadata` endpoint returns LIST parameter members as an array of **objects**:

```json
{
  "parameterType": "LIST",
  "parameterCaption": "Choose Date Granularity",
  "members": [
    {"value": "Daily"},
    {"value": "Monthly"},
    {"value": "Quarterly"}
  ]
}
```

But in `src/sdks/tableau/apis/vizqlDataServiceApi.ts`, the schema expects members as an array of **primitives**:

```typescript
const parameterValueSchema = z.union([z.number(), z.string(), z.boolean(), z.null()]);

// ...
parameterBaseSchema.extend({
  parameterType: z.literal('LIST'),
  members: z.array(parameterValueSchema),  // expects ["Daily", "Monthly"] not [{value: "Daily"}, ...]
}).strict(),
```

The `.strict()` modifier combined with the type mismatch causes Zod validation to fail when parsing the VDS response. The error propagates uncaught because `readMetadata` only has a 404-specific catch handler.

### Suggested Fix

Update the LIST parameter schema to match the actual VDS response format:

```typescript
// Before
members: z.array(parameterValueSchema),

// After
members: z.array(z.object({ value: parameterValueSchema })),
```

And in `src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts`, unwrap the values when building the simplified response:

```typescript
if (parameter.parameterType === 'LIST' && parameter.members) {
  toPush.members = parameter.members.map(m => m.value);  // unwrap {value: x} → x
}
```

### Impact

[...]

PR #240, opened 2026-03-03
@W-21420603: Fix parameterValueSchema to accept object-typed parameter members

mattcfilbert:
# Pull Request Template

## Description

VDS `read-metadata` responses can return LIST parameter `members` as objects (e.g. `{value: "..."}`) rather than plain primitives (`string | number | boolean | null`). The Zod schema was too strict, causing Zodios response validation to throw `Cannot read properties of undefined (reading 'map')` when calling `get-datasource-metadata`.

Widened `parameterValueSchema` in both `vizqlDataServiceApi.ts` and `datasourceMetadataUtils.ts` to also accept `Record<string, unknown>`.

## Motivation and Context

During HBI sign-off testing against `test-dataplane1`, the `get-datasource-metadata` tool crashed with `Cannot read properties of undefined (reading 'map')`. Root cause: the Sales Commission datasource has LIST parameters whose `members` are objects (`{value: "% quota ascending"}`) rather than plain strings. Zodios threw a validation error because the schema didn't account for this shape.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- All 903 unit tests pass locally
- Manually verified end-to-end via MCPjam + OAuth against `test-dataplane1`:
  - `list-datasources` returned datasources
  - `get-datasource-metadata` returned 15 fields and 4 parameters (previously crashed)
  - `query-datasource` returned 41 rows of real sales data

## Related Issues

@W-21420603

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
+++ b/src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
@@ -33,14 +33,12 @@ export const parameterSchema = z
     dataType: z.string().nullable(),
     value: z.union([z.number(), z.string(), z.boolean(), z.null()]),
     members: z.array(
-      z
-        .union([z.number(), z.string(), z.boolean(), z.null()])
-        .or(
-          z.object({
-            value: z.union([z.number(), z.string(), z.boolean(), z.null()]),
-            alias: z.string(),
-          }),
-        ),
+      z.union([z.number(), z.string(), z.boolean(), z.null()]).or(
+        z.object({
+          value: z.union([z.number(), z.string(), z.boolean(), z.null()]),
+          alias: z.string(),
+        }),
+      ),
     ),
     min: z.number().nullable(),
     max: z.number().nullable(),


src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -64,7 +64,13 @@ const fieldMetadataSchema = z
   .passthrough();
 
 // Parameter schemas
-const parameterValueSchema = z.union([z.number(), z.string(), z.boolean(), z.null()]);
+const parameterValueSchema = z.union([
+  z.number(),
+  z.string(),
+  z.boolean(),
+  z.null(),
+  z.record(z.unknown()),
```
anyoung-tableau: Let's chat. While this works, I think the type could be better. There was some discussion for this issue on this PR: https://github.com/tableau/tableau-mcp/pull/229/changes#r2842632159

And the [internal discussion](https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1763063062004409?thread_ts=1763060868.308809&cid=C08QYBH8AE5)

mattcfilbert: Sounds good Andy, thank you. The crux of it is making something for pre and post 2026.1 yeah?

mattcfilbert: Reverted parameterValueSchema back to primitives only. Made the members array on LIST parameters accept both the primitive format and the { value, alias } object format. 

Ought to be backward compatible with pre-2026.1 while also handling the new schema but please lmk if I missed something

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -31,8 +31,10 @@ export const parameterSchema = z
     name: z.string(),
     parameterType: z.string(),
     dataType: z.string().nullable(),
-    value: z.union([z.number(), z.string(), z.boolean(), z.null()]),
-    members: z.array(z.union([z.number(), z.string(), z.boolean(), z.null()])),
+    value: z.union([z.number(), z.string(), z.boolean(), z.null(), z.record(z.unknown())]),
+    members: z.array(
+      z.union([z.number(), z.string(), z.boolean(), z.null(), z.record(z.unknown())]),
+    ),
```
anyoung-tableau, line 42: We should update the schema here too so the types match down in `simplifyReadMetadataResult()`

mattcfilbert: thanks Andy, pushed

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -78,7 +78,9 @@ const parameterSchema = z.discriminatedUnion('parameterType', [
   parameterBaseSchema
     .extend({
       parameterType: z.literal('LIST'),
-      members: z.array(parameterValueSchema),
+      members: z.array(
+        parameterValueSchema.or(z.object({ value: parameterValueSchema, alias: z.string() })),
+      ),
```
anyoung-tableau, line 83: Thanks Matt! Please:

* Follow up on and close https://github.com/tableau/tableau-mcp/pull/229 in favor of this PR.
* Add `Fixes #228` to the PR description so it closes the issue when the PR merges.

PR #241, opened 2026-03-03
@W-21581935: Add passthrough auth

### LINKED ISSUE #237: Add passthrough auth mode for gateway/proxy deployments
See https://github.com/tableau/tableau-mcp/pull/234

anyoung-tableau:
This PR originated from #234 and the discussion on it.

Fixes #237

This PR introduces "pass through" auth which allows clients to provide the Tableau access token used by the REST APIs to be provided directly on the `X-Tableau-Auth` header or `workgroup_session_id` cookie. This allows for user-less agents, both external or internal, to use Tableau MCP without being forced to authenticate interactively or provide a client credential.

When enabled, this auth method *complements* the default auth method configured on the `AUTH` environment variable. That is, requests with the `X-Tableau-Auth` header will take precedence over whatever form of auth would have occurred otherwise. This includes OAuth. See the new docs page for more details.

### DIFF
--- a/src/server/passthroughAuthMiddleware.test.ts
+++ b/src/server/passthroughAuthMiddleware.test.ts
@@ -11,7 +11,7 @@ describe('passthroughAuthMiddleware', () => {
       toolsWithoutMcpScopes,
       [
         'This test is designed to fail the first time a tool is added that does not require API scopes.',
-        'If you see this error, and your tool indeed requires no API scopes,you must add the appropriate logic to prevent calling the tool with passthrough auth.',
+        'If you see this error, and your tool indeed requires no API scopes, you must add the appropriate logic to prevent calling the tool with passthrough auth.',
         'See: https://github.com/tableau/tableau-mcp/pull/241/changes#r2942474421',
       ].join('\n'),
     ).toHaveLength(0);


docs/docs/configuration/mcp-config/authentication/passthrough.md
```diff
@@ -0,0 +1,51 @@
...
+`X-Tableau-Auth` header, otherwise the client will be considered unauthorized and will be forced to
+authenticate using OAuth. This even includes MCP lifecycle requests like the
+[Initialization request](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization),
+even though it does not make any downstream Tableau REST API calls.
+
+:::
+
+:::danger
+
+Do not use a Personal Access Token (PAT) to generate the `X-Tableau-Auth` credential when when using
```
anyoung-tableau: `when when`

tests/oauth/embedded-authz/oauth.test.ts
```diff
@@ -213,6 +213,116 @@ describe('OAuth', () => {
     expect(data).toMatchObject({ result: { tools: expect.any(Array) } });
   });
 
+  it('should allow authenticated requests using the X-Tableau-Auth header', async () => {
```
anyoung-tableau, line 216: This isn't the best spot for these tests, but they are the only tests that use HTTP transport.

src/restApiInstance.test.ts
```diff
@@ -23,23 +22,50 @@ vi.mock('./logging/log.js', () => ({
 
 describe('restApiInstance', () => {
   const mockHost = 'https://my-tableau-server.com';
-  const mockAuthConfig: AuthConfig = {
```
anyoung-tableau, line 26: Small refactor in this test file to stub environment variables instead

src/restApiInstance.ts
```diff
@@ -101,58 +101,71 @@ const getNewRestApiInstanceAsync = async (
...
-    setSiteLuid?.(restApi.siteId);
-    setUserLuid?.(restApi.userId);
+  let signOutWhenCompleted = true;
+  if (tableauAuthInfo?.type === 'Passthrough') {
+    if (!tableauAuthInfo.raw || !tableauAuthInfo.userId) {
+      throw new Error('Auth info is required when not signing in first.');
+    }
+
+    signOutWhenCompleted = false;
+    restApi.setCredentials(tableauAuthInfo.raw, tableauAuthInfo.userId);
```
anyoung-tableau, line 111: Sorry for the bad diff here. These are really the only changes. Also the new `signOutWhenCompleted` variable.

docs/docs/configuration/mcp-config/authentication/passthrough.md
```diff
@@ -0,0 +1,61 @@
...
+With passthrough authentication enabled, authentication to the MCP server acts similarly to the
+Tableau REST APIs. The same
+[`X-Tableau-Auth` header](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_auth.htm#using_auth_token)
+used to authenticate to the Tableau REST APIs can also used to authenticate to the MCP server.
+
+When a request is made to the MCP server, the `X-Tableau-Auth` header is read.
+
+- When the header is present, the value will be "passed through" and re-used during MCP tool calls
+  when they authenticate to the Tableau REST APIs.
+- When absent, normal authentication will resume as defined by the [`AUTH`](../env-vars.md#auth)
```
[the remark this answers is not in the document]
anyoung-tableau: This is an internal cookie and was omitted intentionally.

docs/docs/configuration/mcp-config/authentication/passthrough.md
```diff
@@ -0,0 +1,61 @@
...
+:::
+
+:::danger
+
+Do not use a Personal Access Token (PAT) to generate the `X-Tableau-Auth` credential when using
+passthrough authentication since PATs cannot be used concurrently. Signing in multiple times with
+the same PAT at the same time will terminate any prior session and will result in an authentication
+error. See
+[Understand personal access tokens](https://help.tableau.com/current/server/en-us/security_personal_access_tokens.htm#understand-personal-access-tokens)
+for more details.
```
stephendeoca, line 43: Can I get a refresher on when we support PATs in MCP? Due to the concurrency limitations, I imagine we don't support it with HTTP transport types, only STDIO right?

src/server/oauth/authMiddleware.ts
```diff
@@ -23,11 +23,13 @@ import { AuthenticatedRequest } from './types.js';
  * @returns Express middleware function
  */
 export function authMiddleware(accessTokenValidator: AccessTokenValidator): RequestHandler {
-  return async (
-    req: AuthenticatedRequest,
-    res: express.Response,
-    next: express.NextFunction,
-  ): Promise<void> => {
+  return async (req: AuthenticatedRequest, res: Response, next: NextFunction): Promise<void> => {
+    if (req.auth) {
```
mattcfilbert, line 27: When passthrough sets req.auth upstream we skip scope enforcement entirely, yeah? 

Saw the comment in the passthrough middleware about relying on downstream REST APIs for scope checks. Want to be aware since this bypasses the MCP-level scope checks I added

src/server/passthroughAuthMiddleware.ts
```diff
@@ -0,0 +1,103 @@
...
+export type PassthroughAuthInfo = z.infer<typeof passthroughAuthInfoSchema>;
+
+let passthroughAuthInfoCache: ExpiringMap<string, PassthroughAuthInfo> | undefined;
+
+export async function passthroughAuthMiddleware(
+  req: AuthenticatedRequest,
+  res: Response,
+  next: NextFunction,
+): Promise<void> {
+  const tableauAccessToken: string =
```
stephendeoca: Although its not necessary, consider adding in a check for `config.enablePassthrough`. I know we wont have this middleware when it's false, but if we someone manage to register the passthrough middleware when the feature is disabled, then any request with `X_TABLEAU_AUTH_HEADER` set should be able to authenticate. (not 100% sure though)

src/server/passthroughAuthMiddleware.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  req.auth = {
+    token: 'passthrough',
+    clientId: 'passthrough',
+    // From TMCP, we have no way of verifying which scopes, if any, were used to obtain the access token.
+    // We will rely on the downstream REST APIs to enforce scopes at the API level.
+    // MCP scopes will not be enforced for passthrough auth.
+    scopes: [],
```
mattcfilbert: Since passthrough sets `req.auth` before the auth middleware runs, the `if (req.auth) { next(); return; } ` early return in authMiddleware.ts (site of my other comment) skips scope checks entirely

So these empty scopes never actually get compared against required scopes. Is that the intended behavior, or should `scopes: []` be populated with something if we ever want scope enforcement on passthrough in the future?

docs/docs/configuration/mcp-config/authentication/passthrough.md
anyoung-tableau: We only support it in that we don't actively *prevent* it for http transport since it's helpful for testing and prototyping.

src/server/oauth/authMiddleware.ts
anyoung-tableau: Yeah, passthrough auth is an escape hatch. Tableau MCP won't do scope enforcement but the underlying REST APIs will do the API scope check for us.

src/server/passthroughAuthMiddleware.ts
anyoung-tableau: Yeah, I'd call it intended but only because there's no other option currently. TMCP has no way of cracking open the access token to see a) if the session is even scoped at all and b) if it has the MCP scopes necessary to serve the MCP request.

src/server/passthroughAuthMiddleware.ts
anyoung-tableau: I don't think it's necessary, but it doesn't hurt to have the additional check. I do want to avoid parsing the environment variables unless necessary by the middleware but I think I can accomplish this by converting to a factory function.

src/server/passthroughAuthMiddleware.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  const config = getConfig();
+  const enableCaching = config.passthroughAuthUserSessionCheckIntervalInMinutes > 0;
+
+  if (enableCaching && !passthroughAuthInfoCache) {
+    passthroughAuthInfoCache = new ExpiringMap<string, PassthroughAuthInfo>({
+      defaultExpirationTimeMs: config.passthroughAuthUserSessionCheckIntervalInMinutes * 60 * 1000,
+    });
+  }
```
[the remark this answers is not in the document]
yyk16293: As we are using defaultExpirationTimeMs, we should be good here.

src/server/passthroughAuthMiddleware.ts
anyoung-tableau: As a compromise we could disallow passthrough auth requests from executing MCP tools that don't require any API scopes. Thoughts?

mattcfilbert: Word: can't enforce what we can't see. The compromise works for me.

Only thing I'd flag: as we add tools we'd want to make sure new ones either have API scopes or are explicitly handled. Otherwise we could silently open up passthrough access to a tool that has no downstream enforcement. Worth a test?

anyoung-tableau: Test added!

src/server/passthroughAuthMiddleware.ts
anyoung-tableau: Ended up going with a guard. Anything else was too heavy

PR #243, opened 2026-03-03
@W-21433654: Fix 260 filter schemas by adding support for rules to control tool behavior across different versions

stephendeoca:

the pull request._

# Pull Request Template

## Description

Instead of having a tool make calls to get the tableau server version and have different results for each release, we can use rules to control the flow of logic instead.

## Motivation and Context

When updating SET, MATCH, and DATE filters to support functions and calculations, the args validators would need to be passed version context or a new version of validator functions would have to be created which is okay as a one time change but this issue may be ran into again in the future. So instead we created rules to control the validation logic as well as behavior within a tool's callback implementation

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit tests and versioned requests

## Related Issues

N/A

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.17.6",
+  "version": "1.17.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.17.6",
+      "version": "1.17.7",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.17.6",
+  "version": "1.17.7",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -240,3 +210,20 @@ export const getQueryDatasourceTool = (
...
 };
+
+function getQueryDatasourceRules(tableauServerVersion: ProductVersion): ToolRules {
+  return getResultForTableauVersion({
+    productVersion: tableauServerVersion,
+    mappings: {
+      '2026.1.0': {},
+      '2025.3.0': {
+        dontSpecifyRowLimits: true,
+        restrictFunctionsAndCalculationsInFilters: true,
```
stephendeoca: I intentionally named the variables in such a way where `false` or `undefined` means the feature is enabled. That way if a new rule mapping is needed newer versions, folks wont have to specify any preexisting rules again. Notice how '2026.1.0' opts in to all features (supports row limits and functions / calculations in filters) yet it doesn't specify any rules.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -240,3 +210,20 @@ export const getQueryDatasourceTool = (
...
+      },
```
anyoung-tableau: Consider removing the rules for 2025.3 since they are the same as the default.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -109,22 +100,12 @@ export const getQueryDatasourceTool = (
...
-                debug: true,
-                disaggregate: false,
-              } as const,
-            },
-          });
+          const options: QueryRequest['options'] = {
+            returnFormat: 'OBJECTS',
+            debug: true,
+            disaggregate: false,
+            ...(rules.dontSpecifyRowLimits ? {} : { rowLimit }),
```
anyoung-tableau, line 117: Good change. 🙌

src/tools/tool.ts
```diff
@@ -19,6 +19,8 @@ type ArgsValidator<Args extends ZodRawShape | undefined = undefined> = Args exte
   ? (args: z.objectOutputType<Args, ZodTypeAny>) => void
   : never;
 
+export type ToolRules = Record<string, boolean | undefined>;
+
```
anyoung-tableau, line 23: Non-blocking, but consider removing this type and moving it closer to where it's being used. Now that the `rules` aren't defined at the Tool level, it feels a little out of place here.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -144,17 +144,17 @@ const filterBaseSchema = z.object({
 });
 
 const simpleFilterBaseSchema = z.object({
-  field: z.object({ fieldCaption: z.string() }).strict(),
+  field: z.union([dimensionFilterFieldSchema, calculatedFilterFieldSchema]),
```
stephendeoca, line 149: This schema is only used by Match filters since it does not support the measureFilterFieldSchema (aggregated fields) but all other filter types will support both aggregated fields and calculated fields in their schema.

src/tools/queryDatasource/descriptions/queryDescription.2026.1.ts
```diff
@@ -0,0 +1,605 @@
+export const queryDatasourceToolDescription20261 = `# Query Tableau Data Source Tool
```
stephendeoca, line 1: The only differences between this and the 2025.3 version is that I removed the following:
- "Note: SET, MATCH, and DATE filters can't have functions or calculations."
- "(only available with QUANTITATIVE and TOP filters)" from the line specifying using calculations in filters.

src/tools/tool.ts
stephendeoca: I think I will leave it here because if other tools want to add in rules they can just borrow this type in all their mappings objects.

src/tools/queryDatasource/queryDatasource.ts
```diff
@@ -61,32 +52,32 @@ export type QueryDatasourceError =
...
-    description: new Provider(
-      async () =>
-        await getResultForTableauVersion({
-          server: config.server || authInfo?.server,
-          mappings: {
-            '2025.3.0': queryDatasourceToolDescription20253,
-            default: queryDatasourceToolDescription,
-          },
-        }),
+    description: new Provider(() =>
```
stephendeoca, line 71: @anyoung-tableau I think since we have the product version we don't need to use the provider here any more. want me to refactor this into its own method called `getQueryDatasourceDescription` like I did with `getQueryDatasourceRules`?

anyoung-tableau: Only if you want to. No strong feelings either way.

PR #246, opened 2026-03-04
@W-21447604 - Fix schema validation errors causing unhelpful crash (#158)

### LINKED ISSUE #158: Schema validation errors cause unhelpful error
Internal conversation: https://salesforce-internal.slack.com/archives/C08QYBH8AE5/p1763062220113849?thread_ts=1763060868.308809&cid=C08QYBH8AE5

The logic around building up a friendly error message for Zodios errors needs another look.

https://github.com/tableau/tableau-mcp/blob/main/src/tools/tool.ts#L267

Ultimately in `zod-validation-error`, `issue.errors` is undefined and throws on this line:

https://github.com/causaly/zod-validation-error/blob/4b4f8a78ab0478c1ff81cab626b369b722b07c2f/lib/v4/MessageBuilder.ts#L58

mattcfilbert:
## Description

Wraps `fromError()` in a try-catch in `getErrorResult()` so that when `zod-validation-error` crashes on certain Zod error structures (e.g. `discriminatedUnion` with undefined `issue.errors`), we still return the raw API response with a fallback warning instead of an unhandled `Cannot read properties of undefined (reading 'map')` error.

(Also adds `.vscode/` to `.gitignore`.)

## Motivation and Context

When a Zod schema validation fails on an API response (e.g. LIST parameter members returning objects instead of primitives), the `fromError()` call from `zod-validation-error` can itself crash internally when processing certain error structures. This causes users to see an unhelpful `Cannot read properties of undefined (reading 'map')` error instead of getting the raw API data with a validation warning. With this fix, even if `fromError()` fails, users still receive the API response data along with a generic schema mismatch warning so they aren't blocked.

PS: Also addressed the outstanding comment from #247 — changed clientName?: string to clientName: string | undefined so callers can't accidentally omit it.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added 3 new unit tests in `tool.test.ts`:
- ZodiosError with a valid ZodError cause returns `isError: false` with data and a validation warning
- ZodiosError with a malformed issue (simulating the `discriminatedUnion` crash) returns `isError: false` with data and a fallback warning
- ZodiosError thrown (not returned as `Err`) is still handled gracefully

All 908 tests pass. Lint, Prettier, and TypeScript checks pass.

## Related Issues

- Closes #158
- Related to #228, #229

## Checklist

- [x] I have updated the version in the package.json file by using `npm run version`. For example, use `npm run version:patch` for a patch version bump.

## Contributor Agreement

### DIFF
--- a/tests/oauth/authorizationCodeFlow.test.ts
+++ b/tests/oauth/authorizationCodeFlow.test.ts
@@ -83,7 +83,7 @@ describe('authorization code flow', () => {
       expect(response.body).toEqual({
         error: 'invalid_request',
         error_description:
-          'Validation error: Client_id is required at "client_id"; Redirect_uri is required at "redirect_uri"; Response_type is required at "response_type"; Code_challenge is required at "code_challenge"; Code_challenge_method is required at "code_challenge_method"',
+          'Validation error: client_id is required at "client_id"; redirect_uri is required at "redirect_uri"; response_type is required at "response_type"; code_challenge is required at "code_challenge"; code_challenge_method is required at "code_challenge_method"',
       });
     });
 


src/tools/tool.ts
```diff
@@ -273,15 +273,20 @@ function getErrorResult(requestId: RequestId, error: unknown): CallToolResult {
     // This should make it so users don't get "stuck" when our schemas are too strict or wrong.
     // The only con is that the full response from the API might be larger than normal
     // since a successful schema validation "trims" the response down to the shape of the schema.
-    const validationError = fromError(error.cause);
+    let warning: string;
+    try {
+      warning = fromError(error.cause).toString();
```
anyoung-tableau: I was hoping we could fix what was actually wrong on this line rather than throwing a try/catch at it. The purpose of `zod-validation-error` is to convert Zod schema validation errors into user-friendly error messages that make it easier to know which fields in the schema didn't match. If we're doing everything right and there's a bug in the dependency, then the try/catch is fine. Would you mind digging in a bit please? Can we define the discriminated union schemas in a way so that `issue.errors` won't be undefined?

src/tools/tool.test.ts
```diff
@@ -384,4 +385,112 @@ describe('Tool', () => {
...
+      const rawApiData = { fields: [], parameters: [{ unexpected: 'data' }] };
+
+      // Construct a ZodError with a malformed issue that will cause fromError() to crash.
+      // This simulates the discriminatedUnion bug where issue.errors is undefined.
+      const brokenIssue = {
+        code: 'invalid_union',
+        unionErrors: [],
+        path: ['parameters', 0],
+        message: 'Invalid union',
+      } as unknown as ZodIssue;
```
anyoung-tableau: Rather than crafting a fake ZodIssue, let's create a zod schema and parse an object that results in the ZodIssue we want.

src/tools/tool.ts
mattcfilbert: Thanks, Andy 

Did a deeper dive. We were importing from `zod-validation-error` which defaults to the v4 code path, but we use Zod v3

The v3 and v4 error structures are different (unionErrors vs errors), so fromError() was crashing on the mismatch. Switched all imports to `zod-validation-error/v3` and removed the try-catch since it's no longer needed

src/tools/tool.test.ts
mattcfilbert: +1 

Now using a real discriminatedUnion schema now and parsing invalid data through it to get the actual ZodError

src/tools/tool.ts
```diff
@@ -2,7 +2,7 @@ import { CallToolResult, RequestId, ToolAnnotations } from '@modelcontextprotoco
 import { ZodiosError } from '@zodios/core';
 import { Result } from 'ts-results-es';
 import { z, ZodRawShape, ZodTypeAny } from 'zod';
-import { fromError, isZodErrorLike } from 'zod-validation-error';
+import { fromError, isZodErrorLike } from 'zod-validation-error/v3';
```
anyoung-tableau, line 5: Oh, nice! Glad this was an easy fix!

src/tools/tool.test.ts
```diff
@@ -384,4 +385,112 @@ describe('Tool', () => {
...
+        callback: () => Promise.resolve(Err(zodiosError)),
+        getErrorText: () => 'unused',
+        constrainSuccessResult: (result) => ({ type: 'success', result }),
+      });
+
+      expect(result.isError).toBe(false);
+      invariant(result.content[0].type === 'text');
+      const parsed = JSON.parse(result.content[0].text);
+      expect(parsed.data).toEqual(rawApiData);
+      expect(parsed.warning).toContain('Expected string, received object');
```
jarhun88, line 421: will it always be the zodError.message?

src/tools/tool.test.ts
```diff
@@ -384,4 +385,112 @@ describe('Tool', () => {
...
+      };
+
+      const schema = z.discriminatedUnion('parameterType', [
+        z.object({ parameterType: z.literal('LIST'), members: z.array(z.string()) }).strict(),
+        z.object({ parameterType: z.literal('RANGE'), min: z.number(), max: z.number() }).strict(),
+      ]);
+
+      const parseResult = schema.safeParse(rawApiData.parameters[0]);
+      expect(parseResult.success).toBe(false);
+      if (parseResult.success) return;
```
jarhun88, line 437: do we need this? isnt this a scenario thats never going to happen? or is it purely for compiler?

src/tools/tool.test.ts
```diff
@@ -384,4 +385,112 @@ describe('Tool', () => {
...
+      });
+
+      expect(result.isError).toBe(false);
+      invariant(result.content[0].type === 'text');
+      const parsed = JSON.parse(result.content[0].text);
+      expect(parsed.data).toEqual(rawApiData);
+      expect(parsed.warning).toContain('Validation error');
+    });
+
+    it('should return isError: false with data when ZodiosError is thrown (not returned)', async () => {
```
jarhun88, line 461: curious whats the scenario when its thrown vs returned?

src/tools/tool.ts
mattcfilbert: love the 3 character resolution

src/tools/tool.test.ts
mattcfilbert: yeah, we're constructing the ZodError with that message in the test so it's deterministic 

just checking that the issue message makes it through fromError() into the warning

src/tools/tool.test.ts
mattcfilbert: just for the TypeScript compiler because expect doesn't narrow the type, so without it TS won't let us access parseResult.error 

yay coding

src/tools/tool.test.ts
mattcfilbert: good Q: two code paths in logAndExecute can hit getErrorResult() 

the callback can return Err(zodiosError) or throw one

Zodios throws during response validation so the "thrown" path is the real-world case, but both funnel through the same handler

jarhun88: by handler you mean the caller aka logandExcute right

PR #247, opened 2026-03-04
@W-21448601 - Use CIMD client_name in device_name for Connected Clients (#199)

### LINKED ISSUE #199: client_name from CIMD should be used in device_name
When an MCP client supports CIMD and provides its metadata document URL as its client_id to the /authorize endpoint, this document should contain a `client_name`.

For example, from https://vscode.dev/oauth/client-metadata.json:

```json
{
  "client_name": "Visual Studio Code",
  "grant_types": [
    "authorization_code",
    "refresh_token",
    "urn:ietf:params:oauth:grant-type:device_code"
  ],
  "response_types": [
    "code"
  ],
  "token_endpoint_auth_method": "none",
  "application_type": "native",
  "client_id": "https://vscode.dev/oauth/client-metadata.json",
  "client_uri": "https://vscode.dev/product",
  "redirect_uris": [
    "http://127.0.0.1:33418/",
    "https://vscode.dev/redirect"
  ]
}
```

The value `Visual Studio Code` should be included in the value of the `device_name` provided to Tableau's OAuth auth endpoint so it appears in the user's accounts settings page Connected Clients list (instead of `Unknown agent`)

Image

The changes necessary are in https://github.com/tableau/tableau-mcp/blob/main/src/server/oauth/authorize.ts

mattcfilbert:
## Description

When an MCP client supports CIMD and provides its metadata document URL as its `client_id` to the `/authorize` endpoint, the `client_name` from the metadata document is now used in the `device_name` sent to Tableau's OAuth endpoint. This shows the actual client name (e.g. "Visual Studio Code") in the Connected Clients list instead of "Unknown agent".

The existing redirect_uri/state heuristics (for Cursor, VS Code URL Handler, etc.) are preserved as fallbacks when CIMD is not used or `client_name` is not provided.

## Motivation and Context

Clients that provide a CIMD metadata URL (like `https://vscode.dev/oauth/client-metadata.json`) include a `client_name` field, but it was being ignored. Users saw "Unknown agent" in their Connected Clients list even when the client had identified itself. This is confusing when multiple clients are connected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Added 7 unit tests for `getDeviceName` covering:
  - CIMD `client_name` is used when provided
  - `client_name` takes priority over redirect_uri heuristics
  - Cursor detected from protocol when no `client_name`
  - VS Code detected from redirect_uri + state when no `client_name`
  - Custom protocol names used as fallback
  - "Unknown agent" for http/https without matching heuristics
  - "Unknown agent" for invalid redirect URIs
- All 905 unit tests + 7 oauth tests pass
- Lint, Prettier, and TypeScript checks pass

## Related Issues

- Closes #199

## Checklist

- [x] I have updated the version in the package.json file by using `npm run version`. For example, use `npm run version:patch` for a patch version bump.

## Contributor Agreement

### DIFF
--- a/src/server/oauth/authorize.ts
+++ b/src/server/oauth/authorize.ts
@@ -354,7 +354,7 @@ async function getClientFromMetadataDoc(
   return Ok(clientMetadataResult.data);
 }
 
-export function getDeviceName(redirectUri: string, state: string, clientName?: string): string {
+function getDeviceName(redirectUri: string, state: string, clientName?: string): string {
   if (clientName) {
     return `tableau-mcp (${clientName})`;
   }
@@ -365,6 +365,8 @@ export function getDeviceName(redirectUri: string, state: string, clientName?: s
     const url = new URL(redirectUri);
     if (url.protocol === 'https:' || url.protocol === 'http:') {
       if (redirectUri === 'https://vscode.dev/redirect' && new URL(state).protocol === 'vscode:') {
+        // VS Code normally authenticates in a way that doesn't give any clues about who it is.
+        // It has a backup authentication method they call "URL Handler" that does though.
         return 'tableau-mcp (VS Code)';
       }
 
@@ -378,3 +380,5 @@ export function getDeviceName(redirectUri: string, state: string, clientName?: s
     return defaultDeviceName;
   }
 }
+
+export const exportedForTesting = { getDeviceName };
--- a/tests/oauth/clientIdMetadataDocuments.test.ts
+++ b/tests/oauth/clientIdMetadataDocuments.test.ts
@@ -142,7 +142,7 @@ describe('clientIdMetadataDocuments', () => {
     expect(location.searchParams.get('state')).toEqual(expect.any(String));
     expect(location.searchParams.get('device_id')).toEqual(expect.any(String));
     expect(location.searchParams.get('target_site')).toBe('mcp-test');
-    expect(location.searchParams.get('device_name')).toBe('tableau-mcp (Unknown agent)');
+    expect(location.searchParams.get('device_name')).toBe('tableau-mcp (Fake MCP Client)');
     expect(location.searchParams.get('client_type')).toBe('tableau-mcp');
   });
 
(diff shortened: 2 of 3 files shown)


src/server/oauth/authorize.ts
```diff
@@ -352,15 +354,17 @@ async function getClientFromMetadataDoc(
...
+  }
+
   const defaultDeviceName = 'tableau-mcp (Unknown agent)';
 
   try {
     const url = new URL(redirectUri);
     if (url.protocol === 'https:' || url.protocol === 'http:') {
       if (redirectUri === 'https://vscode.dev/redirect' && new URL(state).protocol === 'vscode:') {
-        // VS Code normally authenticates in a way that doesn't give any clues about who it is.
-        // It has a backup authentication method they call "URL Handler" that does though.
```
anyoung-tableau, line 363: Is this no longer true? I think if you click "Cancel" on the little dialog in VS Code when it's authenticating, it still offers the backup auth method.

src/server/oauth/authorize.ts
```diff
@@ -352,15 +354,17 @@ async function getClientFromMetadataDoc(
   return Ok(clientMetadataResult.data);
 }
 
-function getDeviceName(redirectUri: string, state: string): string {
+export function getDeviceName(redirectUri: string, state: string, clientName?: string): string {
```
anyoung-tableau: nit: we've generally avoided exporting private functions globally by wrapping them in an `exportedForTesting` object e.g.

```ts
export const exportedForTesting = { getDeviceName };
```

src/server/oauth/authorize.ts
mattcfilbert: oh true - good catch

src/server/oauth/authorize.ts
```diff
@@ -352,7 +354,11 @@ async function getClientFromMetadataDoc(
   return Ok(clientMetadataResult.data);
 }
 
-function getDeviceName(redirectUri: string, state: string): string {
+function getDeviceName(redirectUri: string, state: string, clientName?: string): string {
```
anyoung-tableau, line 357: nit: `clientName` shouldn't be optional, it should be required but with a type that is `string | undefined`. The distinction is subtle but important imho. This isn't a super important scenario but it's good practice to make it impossible to accidentally forget to provide what's actually a required parameter.

PR #249, opened 2026-03-04
@W-21451162 datasource descriptions now returned in getDatasourceMetadata response

joeconstantino:
## Description

Added datasourceDescription: z.string() to fieldsResultSchema so the field is always present in the response.
In combineFields (the path taken when the Metadata API is enabled): populated from the GraphQL response's publishedDatasources[0].description, falling back to an empty string if absent. The GraphQL query already fetched this field, so no query changes were needed.

In simplifyReadMetadataResult (the path taken when the Metadata API is disabled): always returns an empty string, since the VizQL Data Service's readMetadata API does not expose datasource description.

Updated tests to assert datasourceDescription is present and correctly valued across all relevant scenarios.

## Motivation and Context
datasource descriptions weren't showing up in the response from the getDatasourceMetadata tool calls. 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

all 15 tests passed when running npx vitest run src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts

## Related Issues
n/a 

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
+++ b/src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
@@ -51,6 +51,7 @@ export const parameterSchema = z
   .partial();
 
 export const fieldsResultSchema = z.object({
+  datasourceDescription: z.string(),
   fields: z.array(fieldSchema),
   parameters: z.array(parameterSchema),
 });
@@ -61,6 +62,7 @@ export type FieldsResult = z.infer<typeof fieldsResultSchema>;
 
 export function simplifyReadMetadataResult(readMetadataResult: MetadataResponse): FieldsResult {
   const simplifiedResponse: FieldsResult = {
+    datasourceDescription: '',
     fields: [],
     parameters: [],
   };
@@ -127,6 +129,7 @@ export function combineFields(
   // readMetadata (VizQL Data Service API) and listFields (GraphQL Metadata API) results
   // to optimize for LLM accuracy and reduce tokens in response.
   const combinedFields: FieldsResult = {
+    datasourceDescription: listFieldsResult.data.publishedDatasources[0]?.description ?? '',
     fields: [],
     parameters: [],
   };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.17.8",
+  "version": "1.17.9",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.17.8",
+      "version": "1.17.9",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.17.8",
+  "version": "1.17.9",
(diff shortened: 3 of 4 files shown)


src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -51,6 +51,7 @@ export const parameterSchema = z
   .partial();
 
 export const fieldsResultSchema = z.object({
+  datasourceDescription: z.string(),
```
stephendeoca, line 54: When adding this in the response schema it's important to make sure the following:
1. Is datasourceDescription guaranteed to have a string value in all responses for all supported API versions? If so, this is fine
2. If datasourceDescription is guaranteed to have a value, but can also be null you want to change the schema to `z.string().nullable()`
3. If datasourceDescription is not guaranteed to have a value (i.e. the datasourceDescription property might not always appear in a response), then you want the schema to be `z.string().optional()` or `z.string().nullish()` if it's possible to be missing and null

stephendeoca, line 54: Disregard the other comment, I mixed up the API endpoint schemas with our Tool response schemas. This here is our tool response schema. In our tool response we should make this property optional i.e. use `z.string().optional()` instead. That way if we don't have a value for the property we don't have to include it in the tool response.

It's just an unnecessary use of tokens to include a `datasourceDescription` property when it's value is empty (unless you think the agent might benefit from knowing that there's an empty description)

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -61,6 +62,7 @@ export type FieldsResult = z.infer<typeof fieldsResultSchema>;
 
 export function simplifyReadMetadataResult(readMetadataResult: MetadataResponse): FieldsResult {
   const simplifiedResponse: FieldsResult = {
+    datasourceDescription: '',
```
stephendeoca, line 65: simplifyReadMetadataResult gets called when we only have the results from VDS metadata API (no GraphQL API response), so we'll always have an empty string here. Might as well remove it.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -127,6 +129,7 @@ export function combineFields(
   // readMetadata (VizQL Data Service API) and listFields (GraphQL Metadata API) results
   // to optimize for LLM accuracy and reduce tokens in response.
   const combinedFields: FieldsResult = {
+    datasourceDescription: listFieldsResult.data.publishedDatasources[0]?.description ?? '',
```
stephendeoca, line 132: Let's only add this property if it has a non empty string value, you can do this with an if statement after combineFields is initialized:

```
if (listFieldsResult.data.publishedDatasources[0]?.description) {
  combinedFields.datasourceDescription = listFieldsResult.data.publishedDatasources[0]?.description
}
```

joeconstantino: I think i actually want to keep the field and make it an empty string even if there is no description. we're going to be encouraging tableau customers more and more to leverage datasource descriptions so an agent being able to proactively see that's it empty could help it guide the user to add a description. Also I confirmed "description" with the quotes is 2 tokens, so it has a negligible impact.

stephendeoca: That works too.

PR #251, opened 2026-03-04
@W-21453276 - Fix OAuth refresh token site context and stale token storage (#211)

### LINKED ISSUE #211: OAuth refresh token switches to default site rather than user selected site
## Describe the bug

I have an app which makes use of the refresh token in oauth to keep the access token fresh however I've noticed that when a user first logs in to the Tableau Server and selects a site (for instance siteB) their session will then be moved to siteA (the default site where contentUrl is "") after the token refresh occurs.

## Expected behavior

I would like the refresh token to have the original site ID embedded in it in order to keep future access tokens locked to that same site (in my example a refresh would keep the user in siteB)

## Versions

What version of Tableau does the issue reproduce on?

Tableau Server Version: 2025.3.1

What version of the Tableau MCP server does the issue reproduce on?

1.13.10

## Steps to reproduce

1. Authenticate via oauth and select a non-default site
2. Check the correct site is active either by asking the AI agent for the current site or decoding the access token and using the Tableau auth token against the REST API current session endpoint
3. Trigger an access token refresh using the refresh token
4. Check the site info again. Notice you should now be on the default site

mattcfilbert:
Preserve the user's selected site across token refreshes by storing `siteContentUrl` from the session and passing it as `site_namespace` instead of hardcoding empty string. Also fix rotated `RefreshTokenData` to store the new Tableau tokens on successful refresh instead of the old ones.

## Description

Two bugs fixed in the OAuth refresh token flow:

1. **Site context lost on refresh:** `exchangeRefreshToken` hardcoded `site_namespace: ''`, causing Tableau to default the user back to their primary site after a token refresh. Now we capture `contentUrl` from the session at callback time, store it in `RefreshTokenData`, and pass it through as `site_namespace` on every refresh.

2. **Stale tokens stored on rotation:** After a successful Tableau refresh token exchange, the rotated `RefreshTokenData` was storing the *old* Tableau tokens instead of the new ones. If Tableau invalidates used refresh tokens (standard OAuth practice), subsequent refreshes would fail. Now the new tokens are stored on success, with the old tokens preserved only in the failure fallback path.

## Motivation and Context

Users on multi-site Tableau Server environments were getting silently switched to their default site after a token refresh (#211). The `site_namespace` parameter exists in Tableau's token endpoint specifically to preserve site context, but we were always sending an empty string.

The stale tokens bug was discovered during investigation — it's in the same code block and would cause cascading failures on subsequent refreshes if Tableau uses single-use refresh tokens.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Added test asserting `site_namespace: 'mcp-test'` is passed to Tableau during refresh (not `''`)
- Added test verifying site context and updated tokens survive across multiple refresh rotations (second refresh uses `refreshed-refresh-token-1` from the first refresh, not the original token)
- All 908 unit tests pass
- All 62 OAuth tests pass (9 test files)
- TypeScript and ESLint clean

## Related Issues

Fixes #211

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/server/oauth/token.ts
```diff
@@ -224,8 +231,9 @@ export function token(
             user: tokenData.user,
             server: tokenData.server,
             clientId: tokenData.clientId,
-            tokens: tokenData.tokens,
+            tokens: tokensToStore,
```
anyoung-tableau, line 234: Nice, thanks for catching that one!

mattcfilbert: Found it on a double check after not wanting to accidentally ignore your feedback again lol

Issue #159, opened 2025-11-14
Publish docker image on dockerhub


bbartels:
It would be useful to have a docker image published on hub.docker.com so as not to deal with a binary directly

googlemarketingsite24:
Google marketing site @joeconstantino @jarhun88 #202 

> [/

![Image](https://github.com/user-attachments/assets/eacf2b8e-20c4-4e0f-b8ee-a65da2a7e7d9)

](url)

PR #200, opened 2026-01-14
Add GitHub Actions workflow to publish Docker images to GHCR

bbartels:
## Summary
  - Adds a GitHub Actions workflow to automatically build and publish Docker images to GitHub Container Registry when releases are published
  - Supports manual workflow dispatch for ad-hoc builds

  ## Features
  - Multi-architecture builds (linux/amd64, linux/arm64)
  - Automatic semantic version tagging (e.g., release v1.2.3 creates tags: `1.2.3`, `1.2`, `1`, `latest`)
  - GitHub Actions build caching for faster subsequent builds
  - Manual trigger option via workflow_dispatch
  - Zero configuration required - uses built-in `GITHUB_TOKEN`

  ## Admin Setup Required

  **None!** The workflow uses the built-in `GITHUB_TOKEN` and automatic `packages: write` permission.

  After the first publish, the package will appear at:
  `ghcr.io/<org>/<repo>` (e.g., `ghcr.io/tableau/tableau-mcp`)

  ### Optional: Package Visibility
  By default, packages inherit the repository's visibility. If needed, admins can adjust package settings at:
  **Settings → Packages → tableau-mcp → Package settings**

  ## Usage

  Pull the image:
  ```bash
  docker pull ghcr.io/<org>/tableau-mcp:latest

  Test Plan

  - Create a test release and verify image appears in the repo's Packages tab
  - Verify multi-arch support: docker manifest inspect ghcr.io/<org>/tableau-mcp:latest
  - Test manual workflow dispatch from Actions tab

  🤖 Generated with https://claude.com/claude-code

PR #252, opened 2026-03-04
@W-21453541: Publish container to GitHub Container Registry

anyoung-tableau:
See the discussion on #200.

Fixes: #159

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.17.10",
+  "version": "1.17.11",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.17.10",
+      "version": "1.17.11",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.17.10",
+  "version": "1.17.11",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"


docs/docs/getting-started.md
```diff
@@ -32,6 +32,14 @@ This standard config works in most MCP clients:
 
 <hr />
 
+## Docker Container
+
+See
+[Tableau MCP Docker Container](https://github.com/tableau/tableau-mcp/pkgs/container/tableau-mcp)
```
anyoung-tableau, line 38: I _think_ this will be the URL once the first container publishes. If not, I'll fix it. Here is what the page will look like (from [@bbartels](https://github.com/bbartels) fork): https://github.com/bbartels/tableau-mcp/pkgs/container/tableau-mcp

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
anyoung-tableau: Removed workflow_dispatch

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to build (e.g., v1.0.0)'
+        required: false
+        default: 'latest'
```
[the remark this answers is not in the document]
anyoung-tableau: Removed workflow_dispatch

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Extract metadata (tags, labels)
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=semver,pattern={{major}}
+            type=raw,value=latest,enable={{is_default_branch}}
```
mattcfilbert: nit: enable={{is_default_branch}} is normally used for push triggers to gate latest to main only

This workflow only fires on release: published, yeah? 

I think the tag would always get created anyway since releases are tied to the default branch. (Flagging in case there's an edge case with pre-release tags or releases cut from non-default branches where you might not want latest applied.)

mattcfilbert (approved):

.github/workflows/docker-publish.yml
anyoung-tableau: Hmm, Copilot made a [similar comment](https://github.com/tableau/tableau-mcp/pull/252/changes/71458d9fad1d6b7a7fb7f9ed25249259e34f6f28#r2886745560) and I thought I updated that line... must've gotten reverted locally before I pushed. I meant it to be

```
type=raw,value=latest,enable=${{ github.event_name == 'release' }}
```

PR #254, opened 2026-03-06
@W-21436847 send siteluid and userluid for MCP product telemetry

jarhun88:

the pull request._

# Pull Request Template

## Description

I validated that siteLuid and userLuid are available for the following scenarios

non oauth (Pat/uat/connectedappis)
embedded oauth (was already working)
server oauth with mfa

- added siteluid and userluid to the TableauToolContext which is part of TableauRequestHandlerExtra
- setting the siteLuid and userLuid values in getNewRestApiInstanceAsync. Method depends on which auth is being used.
- updated all tools to pass an extra field so that it can be modified at the rest calling level (getNewRestApiInstanceAsync)
- _getToolsToRegister client creates an extra to pass to the restApi call. Not the prettiest but it works as TableauRequestHandlerExtra is a superset of restApiArgs

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe):

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -85,7 +85,7 @@ export class Server extends McpServer {
           userLuid: undefined,
           siteLuid: undefined,
           getConfigWithOverrides: async () =>
-            getConfigWithOverrides({ restApiArgs: tableauRequestHandlerExtra }),
+            getConfigWithOverrides(tableauRequestHandlerExtra as TableauRequestHandlerExtra),
         };
 
         return tableauToolCallback(args, tableauRequestHandlerExtra);
@@ -114,13 +114,20 @@ export class Server extends McpServer {
     tableauAuthInfo?: TableauAuthInfo,
   ): Promise<Array<Tool<any>>> => {
     const config = getConfig();
-    const configOverrides = await getConfigWithOverrides({
-      restApiArgs: {
-        server: this,
-        tableauAuthInfo,
-        disableLogging: true, // MCP server is not connected yet so we can't send logging notifications
-      },
-    });
+    const extra: TableauRequestHandlerExtra = {
+      config,
+      server: this,
+      tableauAuthInfo,
+      userLuid: undefined,
+      siteLuid: undefined,
+      signal: AbortSignal.timeout(config.maxRequestTimeoutMs),
+      requestId: 0 as const,
+      // MCP server is not connected yet so logging/notifications are no-ops
+      sendNotification: async () => {},
+      sendRequest: async () => ({}) as never,
+      getConfigWithOverrides: async () => getConfigWithOverrides(extra),
+    };
+    const configOverrides = await getConfigWithOverrides(extra);
 
     const tableauServerVersion = await getTableauServerVersion(
       config.server || tableauAuthInfo?.server,
--- a/src/tools/resourceAccessChecker.ts
+++ b/src/tools/resourceAccessChecker.ts
@@ -53,11 +53,7 @@ class ResourceAccessChecker {
   }): Promise<Set<string> | null> {
     return (
       this._testOverrides.projectIds ??
-      (
-        await getConfigWithOverrides({
-          restApiArgs: { ...extra },
-        })
-      ).boundedContext.projectIds
+      (await getConfigWithOverrides(extra)).boundedContext.projectIds
     );
   }
 
@@ -68,11 +64,7 @@ class ResourceAccessChecker {
   }): Promise<Set<string> | null> {
     return (
       this._testOverrides.datasourceIds ??
-      (
-        await getConfigWithOverrides({
-          restApiArgs: { ...extra },
-        })
-      ).boundedContext.datasourceIds
+      (await getConfigWithOverrides(extra)).boundedContext.datasourceIds
     );
   }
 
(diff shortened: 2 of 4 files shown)


src/server/service-auth/serviceAuthInfoCache.ts
```diff
@@ -0,0 +1,21 @@
...
+): TableauServiceAuthInfo | undefined => serviceAuthInfoCache.get(requestId);
+
+export const setServiceAuthInfoInCache = (
+  requestId: RequestId,
+  info: TableauServiceAuthInfo,
+): void => {
+  serviceAuthInfoCache.set(requestId, info);
+};
+
+export const clearServiceAuthInfoFromCache = (requestId: RequestId): void => {
```
anyoung-tableau: Since we only ever write and read to the map once, consider a push/pop design so consumers don't have to remember to clean up the map after reading from it.

```ts
class PopMap<K, V> {
  private map = new Map<K, V>();

  push(key: K, value: V): void {
    this.map.set(key, value);
  }

  pop(key: K): V | undefined {
    const value = this.map.get(key);
    this.map.delete(key);
    return value;
  }
}
```

src/server/service-auth/schemas.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod';
+
+// Auth info resolved after signing in to Tableau Server (PAT, direct-trust, UAT).
+// Not applicable for OAuth, where identity is already available in tableauAuthInfo.
+export const tableauServiceAuthInfoSchema = z.object({
```
anyoung-tableau: Zod schemas are only needed if we need to ensure some untrusted object is in the shape of the schema. Please remove and move the `TableauServiceAuthInfo` type to `serviceAuthInfoCache.ts`

src/server/service-auth/serviceAuthInfoCache.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Request-scoped cache storing service auth info (userId, siteLuid) resolved after signIn for PAT, direct-trust, UAT.
+const serviceAuthInfoCache = new Map<RequestId, TableauServiceAuthInfo>();
+
+export const getServiceAuthInfoFromCache = (
+  requestId: RequestId,
+): TableauServiceAuthInfo | undefined => serviceAuthInfoCache.get(requestId);
+
+export const setServiceAuthInfoInCache = (
+  requestId: RequestId,
```
anyoung-tableau: Hmm... I think we need to re-think the key. [Request IDs](https://modelcontextprotocol.io/specification/2025-11-25/schema#jsonrpcrequest) are unique (like `1`, `2`, `3`), but they are not _globally_ unique. This means simultaneous requests from two different clients can have the same request id and we'll get collisions. Last week we talked about an idea that I think would work which is to augment the `extra` property so that user info could be stored/retrieved. Data on `extra` are unique to the current request.

jarhun88: hmm ok gotcha, if were ok with passing values dynamically into extra then ill go with that

src/tools/pulse/listMetricDefinitionsFromDefinitionIds/listPulseMetricDefinitionsFromDefinitionIds.ts
```diff
@@ -60,16 +60,19 @@ Retrieves a list of specific Pulse Metric Definitions using the Tableau REST API
...
+              jwtScopes: ['tableau:insight_definitions_metrics:read'],
+              callback: async (restApi) => {
+                return await restApi.pulseMethods.listPulseMetricDefinitionsFromMetricDefinitionIds(
+                  metricDefinitionIds,
+                  view,
+                );
+              },
             },
-          });
+            extra,
```
jarhun88, line 74: this was the only way i could think of modifying the extra to pass on the siteId and userId

src/server.ts
```diff
@@ -112,13 +114,20 @@ export class Server extends McpServer {
     tableauAuthInfo?: TableauAuthInfo,
   ): Promise<Array<Tool<any>>> => {
     const config = getConfig();
-    const configOverrides = await getConfigWithOverrides({
```
jarhun88, line 115: not as clean as id prefer but it works

src/server/service-auth/serviceAuthInfoCache.ts
jarhun88: no longer applicable as i switched to having extra hold the site and user fields

src/server/service-auth/serviceAuthInfoCache.ts
jarhun88: ive made changes to try to do this, it made the pr much bigger but i think it is somewhat a better flow? lmk your thoughts

src/server/oauth/schemas.ts
```diff
@@ -113,6 +113,7 @@ export const tableauBearerTokenSchema = z.object({
   scope: requiredString('scope'),
   'https://tableau.com/siteId': requiredString('https://tableau.com/siteId'),
   'https://tableau.com/targetUrl': requiredString('https://tableau.com/targetUrl'),
+  'https://tableau.com/userId': requiredString('https://tableau.com/userId'),
```
anyoung-tableau, line 116: This can't use the `requiredString` helper since it won't be available when MFA is disabled. Needs to be `z.string().optional()`

PR #255, opened 2026-03-06
@W-21474807 Group get-datasource-metadata fields by logicalTableId

joeconstantino:
## Summary
- remove `sourceTable` from get-datasource-metadata field payloads
- reshape `fields` to grouped entries: `{ logicalTableId, columns[] }` with `logicalTableId: null` supported
- update metadata utility/test contract and bump package version to `1.17.13`

## Test plan
- [x] `npm run lint -- \"src/tools/getDatasourceMetadata/**/*.ts\"`
- [x] `npm test -- getDatasourceMetadata.test.ts`
- [x] invoke `get-datasource-metadata` for datasource `3f8f54d4-1587-4b5a-b6d7-4a1b37887122` and verify grouped `fields`

## GUS
- Work item: W-21474807
- Epic: [262] [TMCP] Tool Additions and Enhancements

### DIFF
--- a/tests/eval/getDatasourceMetadata.test.ts
+++ b/tests/eval/getDatasourceMetadata.test.ts
@@ -92,7 +92,10 @@ describe('get-datasource-metadata', () => {
       'get_datasource_metadata tool execution not found',
     );
 
-    const { fieldGroups } = getCallToolResult(getDatasourceMetadataToolExecution, fieldsResultSchema);
+    const { fieldGroups } = getCallToolResult(
+      getDatasourceMetadataToolExecution,
+      fieldsResultSchema,
+    );
     const flatFields = fieldGroups.flatMap((group) => group.fields ?? []);
     expect(flatFields.length).toBeGreaterThan(0);
 


src/sdks/tableau/apis/metadataApi.ts
```diff
@@ -27,6 +27,13 @@ export const graphqlResponse = z.object({
               .nullable(),
             fullyQualifiedName: z.string(),
             __typename: z.string(),
+            upstreamTables: z
```
stephendeoca, line 30: according to the [Field docs for Metadata API](https://help.tableau.com/current/api/metadata_api/en-us/reference/field.doc.html), the `upstreamTables` property is always populated and will never be null and always be an array with non null elements. The only way I can see this property being null is if it didn't always exist in the Metadata API, so people using older versions it'd be missing for them but I doubt this property is new. If you want remove the `.nullish()` part feel free to do so. Make sure to keep the `.nullable()` you have on the `name` property though since that can still be null.

That'd mean you wont have to do as many null checks in your typescript code when you build the response. Or you can keep it as is, your code is still type safe for all null scenarios it seems.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -52,12 +58,13 @@ export const parameterSchema = z
 
 export const fieldsResultSchema = z.object({
   datasourceDescription: z.string(),
-  fields: z.array(fieldSchema),
+  fields: z.array(logicalTableGroupSchema),
```
stephendeoca: Should this `fields` property be renamed to something like `fieldGroupings` or `fieldGroups` and rename `columns` in the `logicalTableGroupSchema` to `fields`? One thing I'm thinking is that our descriptions don't make any references to `columns` which are essentially now our `fields`, while the actual `fields` property is being changed to be tables of columns / fields

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -133,23 +144,28 @@ export function combineFields(
...
+  const fields: Field[] = [];
 
   if (!readMetadataResult.data) {
     if (listFieldsResult.data.publishedDatasources[0]?.fields.length) {
       // fallback to returning listFields results if we don't have any fields from readMetadata.
       for (const field of listFieldsResult.data.publishedDatasources[0].fields) {
-        const toPush: Field = { name: field.name };
+        const toPush: Field = {
+          name: field.name,
+          logicalTableId: null,
```
stephendeoca, line 155: As you know we only hit this branch of logic if we have no field results from the VDS readMetadata endpoint but for some reason have field data from our graphQL metadata API. It's pretty much an edge case that should never happen, so I'm not too concerned with what we do here, but I noticed that you added upstream tables to metadataApi.ts did you forget to use their table names here?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -249,3 +269,28 @@ function populateFieldWithAdditionalProperties(sourceField: Field, targetField:
...
+    columns,
+  }));
+}
+
+function normalizeLogicalTableId(logicalTableId?: string | null): string | null {
+  if (logicalTableId == undefined) {
+    return null;
+  }
+
+  const normalizedLogicalTableId = logicalTableId.trim();
```
stephendeoca, line 294: So we're just trimming whitespace here? We no longer split the logical table id's into two parts on the `_`  character?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
joeconstantino: Good suggestion. I'll do that.

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
joeconstantino: no I did remove upstream table in favor of logicalTable_id because that is the same field name returned by the new HBI endpoint (getDatasourceModel) that we want to implement. I didn't consider the edge case you're referring to though. I think it's ok to ignore for now? The scenario you're saying assumes that the readmetadata call succeeds but has no results while the graphql query does have results... I don't think that's even possible?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
joeconstantino: Correct we no longer split logicalTableId on. We now preserve the full identifier because logicalTableId is now the canonical grouping key and should match what will be returned by getDatasourceModel exactly

PR #256, opened 2026-03-06
@W-21483837: Stop creating /oauth2 routes when authz server is external

anyoung-tableau:
When using the Tableau authorization server, we are still creating the embedded authZ server's routes which is unnecessary cruft we don't need in our lives. These changes fix that by a small refactor to break the OAuth provider and access token validation logic into separate classes, one for the Tableau authZ server and one for the embedded authZ server.

### DIFF
--- a/src/server/oauth/accessTokenValidator.ts
+++ b/src/server/oauth/accessTokenValidator.ts
@@ -4,15 +4,15 @@ import { compactDecrypt } from 'jose';
 import { Err, Ok, Result } from 'ts-results-es';
 import { fromError } from 'zod-validation-error/v3';
 
-import { getConfig } from '../../config';
+import { getConfig } from '../../config.js';
 import {
   mcpAccessTokenSchema,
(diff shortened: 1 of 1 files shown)


src/server/oauth/provider.ts
```diff
@@ -89,3 +105,13 @@ export class OAuthProvider {
...
+
+/**
+ * OAuth provider for the Tableau authorization server
+ *
+ */
+export class TableauOAuthProvider extends OAuthProvider {
+  get accessTokenValidator(): AccessTokenValidator {
+    return new TableauAccessTokenValidator();
+  }
+}
```
anyoung-tableau, line 117: The "fix" is here. Since `TableauOAuthProvider` doesn't override the `setupRoutes` method, it gets the default behavior from the base class, which is to only create the `./well-known` routes.

yyk16293: Very good use of abstract classes and inheritance. Code is lot more cleaner to understand.

PR #257, opened 2026-03-09
@W-21514752 - use OAUTH_RESOURCE_URI for resource_metadata in auth middleware

mattcfilbert:
## Description

Use OAUTH_RESOURCE_URI for the resource_metadata URL in auth middleware 401/403 responses instead of deriving it from the request.

- Read base URL from getConfig().oauth.resourceUri (OAUTH_RESOURCE_URI env) instead of req.protocol/req.get('host')
- Ensures 401/403 WWW-Authenticate resource_metadata URL matches the configured deployment base
- Add test asserting resource_metadata uses OAUTH_RESOURCE_URI when set

## Motivation and Context

Some OAuth clients require the resource/resource_metadata URL to match the configured MCP server base. When the server is behind a proxy or custom host, using req.protocol and req.get('host') can produce the wrong URL. Using OAUTH_RESOURCE_URI keeps the advertised URL consistent with the deployment and with the .well-known/oauth-protected-resource document, which already uses this config.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Existing OAuth test still passes (401 resource_metadata still matches default http://127.0.0.1:port when OAUTH_RESOURCE_URI is unset).
- New test stubs OAUTH_RESOURCE_URI to https://mcp.example.com and asserts the 401 WWW-Authenticate header contains that base in resource_metadata.

## Checklist

- [x] I have updated the version in the package.json file by using npm run version. For example, use npm run version:patch for a patch version bump.
- [ ] I have documented any breaking changes in the PR description. For example, renaming a config environment variable or changing its default value.

## Contributor Agreement

_Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of the pull request._

### DIFF
--- a/.cursor/rules/auth.mdc
+++ b/.cursor/rules/auth.mdc
@@ -1,0 +1,86 @@
+---
+description: Auth architecture, middleware chain, and auth-specific coding standards
+globs:
+  - "src/server/**/*.ts"
+  - "src/restApiInstance.ts"
+---
+
+# Auth Standards
+
+## Three Auth Paths
+
+### 1. Passthrough Auth
+- Header `x-tableau-auth` or cookie `workgroup_session_id`
+- Validated via `RestApi.authenticatedServerMethods.getCurrentServerSession()`
+- Optional session caching via `ExpiringMap` (configurable interval)
+- Location: `src/server/passthroughAuthMiddleware.ts`
+- Enabled when `config.enablePassthroughAuth` is true
+- MCP scopes NOT enforced (no way to verify original token scopes)
+
+### 2. Embedded OAuth (PKCE)
+- JWE-encrypted MCP access tokens signed with server's private key
+- Contains embedded Tableau access/refresh tokens
+- `EmbeddedAccessTokenValidator` validates via `jose`
+- Location: `src/server/oauth/provider.ts`, `accessTokenValidator.ts`,
+  `token.ts`, `authorize.ts`, `callback.ts`
+
+### 3. Tableau Cloud OAuth
+- Tableau Cloud JWT Bearer tokens
+- `TableauAccessTokenValidator` validates JWT claims
+- Location: `src/server/oauth/provider.ts`, `accessTokenValidator.ts`
+
+## Middleware Chain Order (express.ts)
+```
+handlePingRequest
+  → passthroughAuthMiddleware (if enabled)
+    → oauthProvider.authMiddleware
+      → validateProtocolVersion
+        → createMcpServer
+```
+
+## Auth Info Flow
+```
+req.auth
+  → getTableauAuthInfo(req.auth)
+    → TableauAuthInfo (discriminated union: 'X-Tableau-Auth' | 'Bearer' | 'Passthrough')
+      → TableauRequestHandlerExtra.tableauAuthInfo
+        → useRestApi() (handles sign-in, token setup, sign-out lifecycle)
+```
+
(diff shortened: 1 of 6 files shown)


src/server/oauth/authMiddleware.ts
```diff
@@ -105,7 +105,8 @@ export function authMiddleware(accessTokenValidator: AccessTokenValidator): Requ
       const missingScopes = [...missingMcpScopes, ...missingApiScopes];
 
       if (missingScopes.length > 0) {
-        const baseUrl = `${req.protocol}://${req.get('host')}`;
+        const { resourceUri } = getConfig().oauth;
+        const baseUrl = resourceUri.replace(/\/$/, '');
```
anyoung-tableau: These changes also mean we can get rid of the [TRUST_PROXY_CONFIG](https://tableau.github.io/tableau-mcp/docs/configuration/mcp-config/http-server#trust_proxy_config) env var. See bullet 2 on https://github.com/tableau/tableau-mcp/pull/163.

Would you mind cleaning it up too?

mattcfilbert: salute emoji

src/server/oauth/authMiddleware.ts
```diff
@@ -46,8 +46,8 @@ export function authMiddleware(accessTokenValidator: AccessTokenValidator): Requ
         return;
       }
 
-      const baseUrl = `${req.protocol}://${req.get('host')}`;
-      const { enforceScopes, advertiseApiScopes } = getConfig().oauth;
+      const { enforceScopes, advertiseApiScopes, resourceUri } = getConfig().oauth;
+      const baseUrl = resourceUri.replace(/\/$/, '');
```
anyoung-tableau: Was thinking about this some more, and while this is much better, there is still a small pathing issue.

`OAUTH_RESOURCE_URI` will be something like `https://foo.com/tableau-mcp` so now the `WWW-Authenticate` header will include `resource_metadata="https://foo.com/tableau-mcp/.well-known/oauth-protected-resource"` which is not correct. The `.well-known` routes currently do not include the `tableau-mcp/` path.

I think it makes sense to keep the current paths so we need to remove the trailing `/tableau-mcp/`.

CC @yyk16293 since I believe you are working on a related problem with this header for the Falcon service.

mattcfilbert: Thanks, Andy. Updated to use new URL(resourceUri).origin instead of the raw string; that strips any path so https://foo.com/tableau-mcp becomes https://foo.com for the resource_metadata URL. Added a test for that case too.

One thing to flag: the .well-known/oauth-protected-resource handler also uses resourceUri directly — resource: `${resourceUri}/${serverName} `

So if OAUTH_RESOURCE_URI includes the path, that resource field would double up (https://foo.com/tableau-mcp/tableau-mcp). I saw you CC'd Yogi on the related issue so figured that's being handled separately, but wanted to call it out

anyoung-tableau: Oh shoot, I had it in my head that OAUTH_RESOURCE_URI already included the path and we weren't appending it in the oauth-protected-resource handler. What you previously had would have been fine but in any case, what we have here now and in the oauth-protected-resource handler is fine. The value for OAUTH_RESOURCE_URI used in Falcon will also just be the root. Sorry for the confusion.

PR #259, opened 2026-03-11
[draft] Add user/site luid getter/setters to tool context

anyoung-tableau:


src/server.ts
```diff
@@ -82,6 +82,20 @@ export class Server extends McpServer {
           config,
           server: this,
           tableauAuthInfo,
+          _userLuid: tableauAuthInfo?.userId,
+          _siteLuid: tableauAuthInfo?.siteId,
+          get userLuid(): string {
+            return tableauRequestHandlerExtra._userLuid ?? '';
+          },
```
anyoung-tableau: Consider either converting these `get`ters to regular functions or convert the `setUser*` functions to `set`ters. I tried the latter but spun my wheels a bit trying to figure out the correct types.

src/restApiInstance.ts
```diff
@@ -123,9 +136,13 @@ const getNewRestApiInstanceAsync = async (
       scopes: jwtScopes,
       additionalPayload: getJwtAdditionalPayload(config, tableauAuthInfo),
     });
+    setSiteLuid?.(restApi.siteId);
+    setUserLuid?.(restApi.userId);
   } else {
     if (tableauAuthInfo?.type === 'Bearer') {
       restApi.setBearerToken(tableauAuthInfo.raw);
+      setSiteLuid?.(tableauAuthInfo.siteId);
+      setUserLuid?.(tableauAuthInfo.userId ?? '');
```
anyoung-tableau, line 145: These shouldn't be necessary because we set their default values in server.ts using `tableauAuthInfo`, right? I also think they feel out of place here because the other methods *require* signing in to the REST API to retrieve their values. The bearer token does not.

```
          _userLuid: tableauAuthInfo?.userId,
          _siteLuid: tableauAuthInfo?.siteId,
```

jarhun88: i actually dont mind cuz it keeps things consistent and dont require having to check two places at the tool level

PR #270, opened 2026-03-18
[draft] Add experimental build

anyoung-tableau:
The build now emits TWO files.

The original index.js (with `import.meta.env.BUILD_CONFIGURATION` replaced with "default")
A new index-experimental.js file (with `import.meta.env.BUILD_CONFIGURATION` replaced with "experimental")

### DIFF
--- a/src/metaEnv.ts
+++ b/src/metaEnv.ts
@@ -1,0 +1,13 @@
+import { BuildConfiguration, isBuildConfiguration } from './scripts/build';
+
+class MetaEnv {
+  buildConfiguration: BuildConfiguration;
+
+  constructor() {
+    this.buildConfiguration = isBuildConfiguration(import.meta.env.BUILD_CONFIGURATION)
+      ? import.meta.env.BUILD_CONFIGURATION
+      : 'default';
+  }
+}
+
+export const getMetaEnv = (): MetaEnv => new MetaEnv();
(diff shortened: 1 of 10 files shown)


tests/eval/base.ts
```diff
@@ -7,7 +7,7 @@ import {
   StreamedRunResult,
   withTrace,
 } from '@openai/agents';
-import { OpenAI } from 'openai/client.js';
+import OpenAI from 'openai';
```
anyoung-tableau, line 10: The TypeScript compiler was complaining about this line with the module resolution changes.

src/metaEnv.ts
```diff
@@ -0,0 +1,13 @@
+import { BuildConfiguration, isBuildConfiguration } from './scripts/buildConfigurations';
+
+class MetaEnv {
```
anyoung-tableau, line 3: This abstraction isn't really necessary, but I prefer the type-safety it provides, rather than just cluttering the code and referencing `import.meta.env.BUILD_CONFIGURATION` everywhere.

tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
   "compilerOptions": {
     "target": "ES2022",
-    "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "noEmit": true,
```
anyoung-tableau, line 6: These changes are to get tsc to stop complaining about the use of import.meta.env. We're only using tsc for type-checking anyway so threw in `noEmit` for good measure.

PR #285, opened 2026-03-31
@W-21748754: Add tools for retrieving data and images for custom views

anyoung-tableau:


src/tools/views/getViewData.ts
```diff
@@ -10,6 +10,10 @@ import { Tool } from '../tool.js';
 
 const paramsSchema = {
   viewId: z.string(),
+  viewFilters: z
+    .record(z.string())
+    .optional()
+    .describe('Optional map of view filter field names to values.'),
```
anyoung-tableau, line 16: Driveby updates to get-view-data and get-view-image tools to include the view filters to be in sync with the Custom View tools.

src/restApiInstance.ts
```diff
@@ -287,8 +287,8 @@ function logResponse(
   const url = new URL(
     `${maskedResponse.baseUrl.replace(/\/$/, '')}/${maskedResponse.url?.replace(/^\//, '') ?? ''}`,
   );
-  if (response.request?.params && Object.keys(response.request.params).length > 0) {
-    url.search = new URLSearchParams(response.request.params).toString();
+  if (response.params && Object.keys(response.params).length > 0) {
+    url.search = new URLSearchParams(response.params).toString();
```
anyoung-tableau, line 295: Driveby fix here. This was wrong before and the query params weren't included in the response traces.

src/sdks/tableau/interceptors.ts
```diff
@@ -39,6 +41,7 @@ export function getResponseInterceptorConfig(
 ): Omit<ResponseInterceptorConfig, 'baseUrl'> {
   return {
     url: response.config.url ?? 'UNKNOWN URL',
+    params: response.config.params,
```
anyoung-tableau, line 44: Driveby fix here to ensure the response interceptor includes the URL search params in the trace.

docs/docs/configuration/tableau-config.md
```diff
@@ -4,7 +4,7 @@ sidebar_position: 1
 
 # Configuring Tableau
 
-Tableau MCP works with both Tableau Server and Tableau Cloud data with these prerequisites:
+Tableau MCP works with both Tableau Server 2024.2+ and Tableau Cloud data with these prerequisites:
```
anyoung-tableau, line 7: I added the min Server version here because the Get Custom View Image REST API was only added in 2023.1 and someone trying to use it against an older Server version would see errors. 2024.2 is a min bar anyway (mostly driven by VDS compatibility)

PR #298, opened 2026-04-13
@W-21984866 adding explicit data model relationship info in getDatasourceMetadata

joeconstantino:
## Description
Added datasource model metadata to getDatasourceMetadata by integrating VDS [get-datasource-model endpoint](https://help.tableau.com/current/api/vizql-data-service/en-us/docs/vds_get_data_source_model.html), exposing logical table relationships in the tool response. Top level keys in the getDatasourceMetadata response now include: 
- datasourceDescription
- datasourceModel
- fieldGroups

## Motivation and Context
Context about the physical relationship that relates disparate tables in a published data sources helps Agents generate better queries and better introspect on the outputs. 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
updated tests: in src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts
- Added mocking for required getDatasourceModel call.
- Added assertions that datasourceModel is present in successful responses.
- Added a new failure-path test for datasource model endpoint returning feature-disabled.

Executed tests:
- npm test -- src/tools/getDatasourceMetadata/getDatasourceMetadata.test.ts
- Result: 17/17 passing

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/sdks/tableau/apis/vizqlDataServiceApi.ts
+++ b/src/sdks/tableau/apis/vizqlDataServiceApi.ts
@@ -326,12 +326,70 @@ export const readMetadataRequestSchema = z
   })
   .passthrough();
 
+export const getDatasourceModelRequestSchema = z
+  .object({
+    datasource: datasourceSchema,
+    options: queryOptionsSchema.optional(),
+  })
+  .passthrough();
+
 export const metadataOutputSchema = z
   .object({
     data: z.array(fieldMetadataSchema),
     extraData: z.object({
       parameters: z.array(parameterSchema),
     }),
+  })
+  .partial()
+  .passthrough();
+
+const logicalTableSchema = z
+  .object({
+    logicalTableId: z.string(),
+    caption: z.string(),
+    description: z.string(),
+  })
+  .partial()
+  .passthrough();
+
+const relationshipClauseSchema = z
+  .object({
+    operator: z.string(),
+    fromField: z.string(),
+    toField: z.string(),
+  })
+  .partial()
+  .passthrough();
+
+const logicalTableRelationshipSchema = z
+  .object({
+    fromLogicalTable: z
+      .object({
+        logicalTableId: z.string(),
+      })
+      .partial()
+      .passthrough(),
+    toLogicalTable: z
+      .object({
+        logicalTableId: z.string(),
+      })
+      .partial()
+      .passthrough(),
+    expression: z
+      .object({
+        op: z.string(),
+        relationships: z.array(relationshipClauseSchema),
+      })
+      .partial()
+      .passthrough(),
+  })
+  .partial()
+  .passthrough();
+
+export const datasourceModelResponseSchema = z
+  .object({
+    logicalTables: z.array(logicalTableSchema),
+    logicalTableRelationships: z.array(logicalTableRelationshipSchema),
   })
   .partial()
   .passthrough();
@@ -400,6 +458,8 @@ export type QueryRequest = z.infer<typeof queryRequestSchema>;
 export type QueryParameter = z.infer<typeof queryParameterSchema>;
 
 export type ReadMetadataRequest = z.infer<typeof readMetadataRequestSchema>;
+export type GetDatasourceModelRequest = z.infer<typeof getDatasourceModelRequestSchema>;
+export type DatasourceModelResponse = z.infer<typeof datasourceModelResponseSchema>;
 
 export type TableauError = z.infer<typeof tableauErrorSchema>;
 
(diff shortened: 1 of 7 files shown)


anyoung-tableau (approved):
💪

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -336,6 +343,57 @@ export const metadataOutputSchema = z
...
+const relationshipClauseSchema = z
+  .object({
+    operator: z.string(),
+    fromField: z.string(),
+    toField: z.string(),
+  })
+  .partial()
+  .passthrough();
+
+const logicalTableRelationshipSchema = z
```
stephendeoca, line 362: just an FYI when defining schemas, if properties are guaranteed in response results you don't need to add partial to it. for example fromLogicalTableId and toLogicalTableId are guaranteed to be in the response according to VDS docs. That'll save you from having to do unnecessary undefined / null checks when you get the api response result in code.

src/tools/getDatasourceMetadata/getDatasourceMetadata.ts
```diff
@@ -146,9 +146,23 @@ export const getGetDatasourceMetadataTool = (server: Server): Tool<typeof params
                 return new FeatureDisabledError(getVizqlDataServiceDisabledError()).toErr();
               }
 
+              // Fetching datasource model from VizQL Data Service API.
+              const datasourceModelResult =
+                await restApi.vizqlDataServiceMethods.getDatasourceModel({
```
stephendeoca: This endpoint only exists for Tableau 2025.3 and up, so you're going to want to add a rule for it to skip this request on older versions. Query datasource is an example of a tool which defines different rules / behavior for different Tableau versions.

Here's what you'll want to do:
1. At the end of this file, add a function to get rules based on version:

```
function getDatasourceMetadataRules(productVersion: ProductVersion): ToolRules {
  return getResultForTableauVersion({
    productVersion,
    mappings: {
      '2025.3.0': {},
      default: {
        datasourceModelIsUnavailable: true,
      },
    },
  });
}
```

note: default is for versions older than the oldest version found in your mappings because of this we'll need to make rule behavior opt in when unspecified otherwise we'd have to opt in to all rules everytime we create a new mapping. The `datasourceModelIsUnavailable` was named with negative logic so that true means you opt out, and false / unspecified value for that rule means we opt in. Dw about it too much.

2. Update the method signature of `getGetDatasourceMetadataTool` to include the productVersion and call the `getDatasourceMetadataRules` function

```
export const getGetDatasourceMetadataTool = (server: Server, productVersion: ProductVersion): Tool<typeof paramsSchema> => {
  const rules = getDatasourceMetadataRules(product version);
  ... // rest of code
}
```
3. Add logic to check your rule before trying to request the datasource model. I recommend refactoring your methods in `datasourceMetadataUtils` to omit the `datasourceModel` property from `FieldsResult` if the datasourceModelResult is undefined.

```
let datasourceModelResult;
if (!rules.datasourceModelIsUnavailable) {
   // Fetching datasource model from VizQL Data Service API.
   datasourceModelResult = await restApi.vizqlDataServiceMethods.getDatasourceModel({
        datasource: {
             datasourceLuid,
         },
    });

     if (datasourceModelResult.isErr()) {
         return new FeatureDisabledError(getVizqlDataServiceDisabledError()).toErr();
     }
}
```

stephendeoca: I believe if you run the current code you have in this PR on Tableau 2025.2, it will throw feature disabled errors because the datasource-model endpoint you tried to fetch from doesn't exist and returns 404. This would essentially make the tool unusable for old versions unless we gate calling the endpoint.

joeconstantino: Thank you!! Can you take a look now? @stephendeoca?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -67,12 +101,20 @@ type Field = z.infer<typeof fieldSchema>;
 type LogicalTableGroup = z.infer<typeof logicalTableGroupSchema>;
 export type FieldsResult = z.infer<typeof fieldsResultSchema>;
 
-export function simplifyReadMetadataResult(readMetadataResult: MetadataResponse): FieldsResult {
-  const simplifiedResponse: FieldsResult = {
+export function simplifyReadMetadataResult(
+  readMetadataResult: MetadataResponse,
+  datasourceModelResult?: DatasourceModelResponse,
+): FieldsResult {
+  const simplifiedResponse: Omit<FieldsResult, 'datasourceModel'> & {
```
stephendeoca: This seems weird, `FieldsResult` should already have `datasourceModel` as an optional property. This type overrides the property as optional when the property should already be optional due to your schema, can you try changing the type to just `FieldResult` and see if it's fine?

src/tools/getDatasourceMetadata/datasourceMetadataUtils.ts
```diff
@@ -135,15 +177,21 @@ export function simplifyReadMetadataResult(readMetadataResult: MetadataResponse)
 export function combineFields(
   readMetadataResult: MetadataResponse,
   listFieldsResult: GraphQLResponse,
+  datasourceModelResult?: DatasourceModelResponse,
 ): FieldsResult {
   // Create a response object that combines field data from
   // readMetadata (VizQL Data Service API) and listFields (GraphQL Metadata API) results
   // to optimize for LLM accuracy and reduce tokens in response.
-  const combinedFields: FieldsResult = {
+  const combinedFields: Omit<FieldsResult, 'datasourceModel'> & {
```
stephendeoca: see if you can just use `FieldsResult` type here

stephendeoca:
Once you merge my changes from https://github.com/tableau/tableau-mcp/pull/310, do a quick lil build and try to get metadata using inspect or agent to confirm everything checks out. You should get good to merge, make sure you bump any package.json versions too.

PR #304, opened 2026-04-14
@W-22033159: fix hf deploy blocking health endpoint check fail

mattcfilbert:
Description

Fixes tabhf-mcp-svc deploy failures where health checks saw ECONNREFUSED on 127.0.0.1:3927. The HTTP server was started only after await getTableauServerInfo(), so nothing listened on the configured port until the outbound Tableau REST call completed (or timed out). Startup is reordered so the HTTP listener binds first for http transport, then server-info initialization is awaited. RestApi.version no longer defaults silently; getTableauServerInfo bootstraps via an explicit /api/3.24 URL for the initial getServerInfo call. Tests and global test setup were updated (including ServerMethods mocking where needed).

Motivation and Context

Production deploys failed because Kubernetes/proxy health checks only need a successful TCP connection to the app port; blocking startup on Tableau server info meant the port was not open during that window, so probes failed and deployments retried indefinitely.

Type of Change

 Bug fix

How Has This Been Tested?

npm test (unit tests) — all passing
New tests: src/sdks/tableau/restApi.test.ts (version getter behavior), src/getTableauServerInfo.test.ts (bootstrap URL, caching, errors)
npx tsc --noEmit and ESLint on touched files

Checklist

 I have updated the version in the package.json file by using npm run version. For example, use npm run version:patch for a patch version bump.

 I have made any necessary changes to the documentation

 I have added tests that prove my fix is effective or that my feature works

 New and existing unit tests pass locally with my changes

 I have documented any breaking changes in the PR description. For example, renaming a config environment variable or changing its default value. (None for this PR — no env renames or intentional API breaks.)
Contributor Agreement

 I have read the CONTRIBUTING guidelines for this project and followed its Contribution Checklist.


src/index.ts
```diff
@@ -15,17 +15,28 @@ import { getExceptionMessage } from './utils/getExceptionMessage.js';
...
+  });
+
   const logLevel = isNotificationLevel(config.defaultLogLevel) ? config.defaultLogLevel : 'debug';
   if (config.loggers.has('fileLogger')) {
     setFileLogger(new FileLogger({ logDirectory: config.fileLoggerDirectory }));
   }
 
   switch (config.transport) {
     case 'stdio': {
+      await serverInfoReady;
```
jarhun88, line 38: wait if we have to await on this here, isnt it still going to be slow?

In my head the getTableauServerInfo functionality (the rest call) was happening outside of startServer()

mattcfilbert: Good question _(em dash)_ the timing of the await is what matters here, not whether it's inside or outside `startServer()`.

In the old code, the call was awaited before `startExpressServer() `ran, so the port never bound until the Tableau REST call finished. That's what caused ECONNREFUSED -- nothing was listening on 3927 during that window.

In the new code, the `getTableauServerInfo` call is kicked off immediately but not awaited yet:

`const serverInfoReady = getTableauServerInfo(config.server).catch(...)
`// `getTableauServerInfo` is already running in the background ^
For http transport, Express binds port 3927 first, then we await `serverInfoReady`. So the fetch is running concurrently with `startExpressServer`, and the port is open by the time we wait for the result. The K8s health check can succeed during that window.

For stdio, there's no port and no health check, so it's fine to await before proceeding _(em dash)_  the behavior is identical to before from the user's perspective.

Moving the call outside `startServer() `entirely wouldn't change anything _(em dash)_  the port still wouldn't be open until Express starts. The fix is the ordering: start Express, then await the fetch, not the other way around.

src/index.ts
```diff
@@ -15,17 +15,28 @@ import { getExceptionMessage } from './utils/getExceptionMessage.js';
...
   RestApi.host = config.server;
-  await getTableauServerInfo(config.server);
+
+  // Start fetching server info immediately but don't block the port from opening.
+  // Any failure here is fatal and logged explicitly -- no silent failures.
+  // For http transport, the port opens first so K8s health checks can succeed,
+  // then we await this before declaring the server ready.
+  // For stdio transport, there are no health checks, but we still await before serving.
+  const serverInfoReady = getTableauServerInfo(config.server).catch((error) => {
+    writeToStderr(`Fatal error initializing server info: ${getExceptionMessage(error)}`);
```
jarhun88, line 27: was this a server failing issue before? and do we want it to be? I see before it just threw an error, but im not sure what startServer did with it

src/index.ts
jarhun88: ok that checks out

src/index.ts
mattcfilbert: Before the change, the behavior was:

await `getTableauServerInfo(config.server)`; // threw if it failed
That throw propagated up through `startServer()`, which was caught here:

```
startServer().catch((error) => {
  writeToStderr(`Fatal error when starting the server: ${getExceptionMessage(error)}`);
  process.exit(1);
});
```
So yeah it was already a process.exit(1) situation before. A failed server info fetch was always fatal.

The new code preserves that exact behavior. The .catch on serverInfoReady does the same thing: logs and exits. The only difference is it's attached directly to the promise rather than propagating through `startServer()`, because we needed to detach the fetch from the await chain to let Express start first.

One subtle thing worth noting: in the old code, if `getTableauServerInfo` threw, the process exited before the port ever opened. In the new code, the process exits after the port opens (since Express starts first). That's actually slightly better; the pod gets to bind the port, pass the health check, and then exit cleanly rather than dying without ever listening. K8s would then restart it and try again, which is the correct behavior for a transient Tableau endpoint failure.

src/getTableauServerInfo.ts
```diff
@@ -32,12 +38,13 @@ export const getTableauServerInfo = async (server?: string): Promise<ServerInfo>
     return serverInfo;
   }
 
-  const restApi = new RestApi({
-    maxRequestTimeoutMs: getConfig().maxRequestTimeoutMs,
+  const bootstrapBaseUrl = `${RestApi.host}/api/${BOOTSTRAP_API_VERSION}`;
```
anyoung-tableau: I understand the sequencing changes in index.ts, but these changes and the changes to restApi.ts don't seem to be necessary. I don't see anything different about the previous approach where we default to 3.24 and allow for updating the value once we have the server info. We should avoid making changes to prevent unnecessary git diffs.

src/index.ts
jarhun88: cool that sounds perfect

jarhun88 (approved):

anyoung-tableau (changes requested):

src/index.ts
anyoung-tableau, line 27: We can't use `log` (at least for stdio transport) because it writes to the console. We'd need to update `log` so it can use console.error()

mattcfilbert: log() is designed for structured app logging (JSON to console.log / fileLogger) and calls getConfig() internally. For a fatal pre-exit message, writeToStderr is the right choice here -- it writes directly to process.stderr synchronously, which guarantees the message makes it out before process.exit(1) terminates the process. console.log (what log() routes to for appLogger) can be buffered and silently dropped on abrupt exit.

writeToStderr is already what the outer startServer().catch(...) handler uses for its fatal errors -- keeping this consistent also makes the log output pattern uniform for the operator looking at pod stderr.

If you want structured logging for this error, the right fix would be to make writeToStderr itself structured, not to use log() here.

src/getTableauServerInfo.ts
mattcfilbert: You're right. The structural changes to restApi.ts and getTableauServerInfo.ts are defensive hardening for a version-mismatch window that can't actually happen once the startup ordering is fixed. 

With the index.ts change in place, getTableauServerInfo is always awaited before any real request processing in both transports, so the default 3.24 gets overwritten before it's ever used in a real API call. I'll revert those files and the associated test mock changes, keeping only the index.ts fix.

mattcfilbert: made the change

src/index.ts
```diff
@@ -15,17 +15,28 @@ import { getExceptionMessage } from './utils/getExceptionMessage.js';
...
   const config = getConfig();
-  // Initializing REST API host then getting server info for the first time
-  // which will cache the server info and initialize our REST API version for subsequent requests
+
   RestApi.host = config.server;
-  await getTableauServerInfo(config.server);
+
+  // Start fetching server info immediately but don't block the port from opening.
+  // Any failure here is fatal and logged explicitly -- no silent failures.
+  // For http transport, the port opens first so K8s health checks can succeed,
```
anyoung-tableau: Consider removing the reference to Kubernetes in this comment and making it more generic. While it applies to the Falcon service, it does not apply in the general sense for self-hosters.

anyoung-tableau (approved):

src/index.ts
mattcfilbert: on it. thanks andy

PR #306, opened 2026-04-14
@W-22034341: reset consent tool

mattcfilbert:
### Description
Adds a new MCP tool, reset-consent, that clears saved OAuth consent on the Tableau authorization server for the current user. It follows the same wiring pattern as revoke-access-token: registered in toolNames / toolGroups, OAuth scope map, and toolFactories.

### Behavior:

Bearer (Tableau authZ server): POST to ${config.oauth.issuer}/oauth2/resetConsent with Authorization: Bearer <access token> and no body (aligned with the existing OAuth test client).
X-Tableau-Auth (embedded) and Passthrough: returns a clear “not supported” error and does not call the network.
Tool is disabled when config.auth !== 'oauth' (same as revoke).
Annotations: destructiveHint: false (consent reset vs token revocation), idempotentHint: true, etc.
Also updates passthrough guard list and e2e tool listing so OAuth-only tools are expected only when AUTH=oauth.

### Motivation and Context
OAuth flows can persist consent on the authorization server. Operators and clients need a first-class MCP action to reset that consent (e.g. teardown, testing, or forcing re-consent) without hand-rolling HTTP. This mirrors revoke-access-token at the MCP layer but targets /oauth2/resetConsent with the correct request shape (header auth, no body) and Bearer-only semantics so embedded auth is not mis-modeled.

### Type of Change

 New feature

### How Has This Been Tested?
Added unit tests in src/tools/resetConsent/resetConsent.test.ts (metadata, disabled/enabled under OAuth, Bearer happy path and errors, embedded + Passthrough rejection, no token leakage in success text, fetch shape).
Updated src/server/passthroughAuthMiddleware.test.ts guard list for the new no-scope tool.
Updated tests/e2e/server.test.ts so reset-consent is only listed when AUTH === 'oauth'.
Ran full unit suite, tsc --noEmit, and ESLint on touched paths (adjust to whatever you actually ran, e.g. npm test, npx tsc --noEmit, npx eslint …).
(Optional follow-up from the plan: npm run test:oauth:embedded if you run OAuth integration tests in CI or locally.)

 I have updated the version in the package.json file by using npm run version. For example, use npm run version:patch for a patch version bump.

 I have made any necessary changes to the documentation

 I have added tests that prove my fix is effective or that my feature works

 New and existing unit tests pass locally with my changes

 I have documented any breaking changes in the PR description. For example, renaming a config environment variable or changing its default value.
(No breaking changes expected: new tool only in OAuth mode; existing clients unchanged.)

Contributor Agreement

 I have read the [CONTRIBUTING guidelines](vscode-file://vscode-app/Applications/Cursor.app/Contents/Resources/app/out/vs/code/CONTRIBUTING.md) for this project and followed its Contribution Checklist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.3",
+  "version": "1.18.4",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.4",
+  "version": "1.18.5",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/tools/resetConsent/resetConsent.test.ts
+++ b/src/tools/resetConsent/resetConsent.test.ts
@@ -34,26 +34,25 @@ describe('resetConsentTool', () => {
   });
 
   describe('disabled property', () => {
-    let savedEnv: NodeJS.ProcessEnv;
-
-    beforeEach(() => {
-      savedEnv = { ...process.env };
-    });
-
-    afterEach(() => {
-      process.env = savedEnv;
-    });
-
-    it('should be disabled when AUTH is not oauth (default PAT mode)', async () => {
-      delete process.env.OAUTH_ISSUER;
+    it('should be disabled when OAuth is not enabled (no OAUTH_ISSUER)', async () => {
+      vi.stubEnv('OAUTH_ISSUER', '');
       const tool = getResetConsentTool(new Server());
       expect(await Provider.from(tool.disabled)).toBe(true);
     });
 
-    it('should be enabled when AUTH=oauth', async () => {
-      process.env.AUTH = 'oauth';
-      process.env.OAUTH_ISSUER = MOCK_ISSUER;
-      process.env.OAUTH_EMBEDDED_AUTHZ_SERVER = 'false';
+    it('should be disabled when the embedded authorization server is active', async () => {
+      // TABLEAU_MCP_TEST=true (set by testSetup) bypasses the JWE key file existence check.
+      vi.stubEnv('OAUTH_ISSUER', MOCK_ISSUER);
+      vi.stubEnv('OAUTH_EMBEDDED_AUTHZ_SERVER', 'true');
+      vi.stubEnv('OAUTH_REDIRECT_URI', `${MOCK_ISSUER}/Callback`);
+      vi.stubEnv('OAUTH_JWE_PRIVATE_KEY_PATH', '/placeholder/key.pem');
+      const tool = getResetConsentTool(new Server());
+      expect(await Provider.from(tool.disabled)).toBe(true);
+    });
+
+    it('should be enabled when OAuth is enabled and the Tableau authorization server is active', async () => {
+      vi.stubEnv('OAUTH_ISSUER', MOCK_ISSUER);
+      vi.stubEnv('OAUTH_EMBEDDED_AUTHZ_SERVER', 'false');
       const tool = getResetConsentTool(new Server());
       expect(await Provider.from(tool.disabled)).toBe(false);
     });
@@ -133,27 +132,6 @@ describe('resetConsentTool', () => {
     });
   });
 
-  describe('X-Tableau-Auth (embedded authZ mode -- not supported)', () => {
-    it('should return an error and make no fetch call for X-Tableau-Auth', async () => {
-      const extra = getMockRequestHandlerExtra();
-      extra.config.oauth.issuer = MOCK_ISSUER;
-      extra.tableauAuthInfo = {
-        type: 'X-Tableau-Auth',
-        username: 'test-user',
-        server: MOCK_SERVER,
-        siteId: 'site-id',
-        accessToken: 'tableau-access-token',
-        refreshToken: 'tableau-refresh-token',
-      };
-      const result = await getToolResult(extra);
-
-      expect(result.isError).toBe(true);
-      expect(mockFetch).not.toHaveBeenCalled();
-      invariant(result.content[0].type === 'text');
-      expect(result.content[0].text).toContain('embedded');
-    });
-  });
-
   describe('Passthrough auth (not supported)', () => {
     it('should return an error and make no fetch call for Passthrough auth', async () => {
       const extra = getMockRequestHandlerExtra();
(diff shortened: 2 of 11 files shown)


src/tools/resetConsent/resetConsent.ts
```diff
@@ -0,0 +1,120 @@
...
+- Cleaning up OAuth grants when a user's access should be fully removed`,
+    paramsSchema,
+    annotations: {
+      title: 'Reset Consent',
+      readOnlyHint: false,
+      destructiveHint: false,
+      idempotentHint: true,
+      openWorldHint: false,
+    },
+    disabled: config.auth !== 'oauth',
```
anyoung-tableau: Since there's no consent UX or scopes for the embedded authorization server we should also disable the tool when `|| config.oauth.embeddedAuthzServer`

src/tools/resetConsent/resetConsent.ts
```diff
@@ -0,0 +1,120 @@
...
+            return new Err(
+              new McpToolError({
+                type: 'not-supported',
+                message:
+                  'Consent reset is not available for the embedded authorization server. ' +
+                  'The embedded authorization server does not maintain a consent model.',
+                statusCode: 400,
+              }),
+            );
+          }
```
anyoung-tableau: We can remove this once the tool is disabled for embedded authz

src/tools/resetConsent/resetConsent.ts
anyoung-tableau: Actually... I think the condition needs to be `!config.oauth.enabled || config.oauth.embeddedAuthzServer`.

Self-hosters can enable OAuth but `AUTH` doesn't necessarily need to be `oauth`.

src/tools/resetConsent/resetConsent.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    beforeEach(() => {
+      savedEnv = { ...process.env };
+    });
+
+    afterEach(() => {
+      process.env = savedEnv;
+    });
+
+    it('should be disabled when AUTH is not oauth (default PAT mode)', async () => {
+      delete process.env.OAUTH_ISSUER;
```
anyoung-tableau: Let's not make changes to environment variables and use Vite's env var stubbing instead.

tests/e2e/server.test.ts
```diff
@@ -17,10 +17,14 @@ describe('server', () => {
 
   it('should list tools', async () => {
     const names = await listTools();
+    const oauthOnlyTools: ReadonlyArray<(typeof toolNames)[number]> = [
```
anyoung-tableau:
```suggestion
    const oauthOnlyTools: ReadonlyArray<ToolName> = [
```

src/tools/resetConsent/resetConsent.ts
mattcfilbert: Updated this. The tool is now disabled for embedded authz as well. 

Also updated to `!config.oauth.enabled || config.oauth.embeddedAuthzServer` so it keys off OAuth enablement rather than `AUTH === 'oauth'`

src/tools/resetConsent/resetConsent.ts
mattcfilbert: Done. I removed the X-Tableau-Auth runtime branch now that embedded authz is gated at tool registration / disabled state

src/tools/resetConsent/resetConsent.test.ts
mattcfilbert: Updated the test to use `vi.stubEnv(...)` instead of mutating `process.env` directly

tests/e2e/server.test.ts
mattcfilbert: Applied, thanks

`oauthOnlyTools` now uses `ReadonlyArray<ToolName>`

src/tools/resetConsent/resetConsent.test.ts
```diff
@@ -0,0 +1,162 @@
...
+import invariant from '../../utils/invariant.js';
+import { Provider } from '../../utils/provider.js';
+import { getMockRequestHandlerExtra } from '../toolContext.mock.js';
+import { getResetConsentTool } from './resetConsent.js';
+
+const MOCK_ISSUER = 'https://sso.online.tableau.com';
+const MOCK_TOKEN = 'eyJhbGciOiJIUzI1NiJ9.dGVzdC1wYXlsb2Fk.signature';
+const MOCK_SERVER = 'https://my-tableau-server.com';
+
+describe('resetConsentTool', () => {
```
anyoung-tableau, line 13: What do you think? Should the OAuth tests call the reset-consent tool now instead of the resetConsent API directly to increase the test coverage?

https://github.com/tableau/tableau-mcp/blob/main/tests/oauth/tableau-authz/fixtures/connectOAuthClient.ts#L32

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.3",
+  "version": "1.18.4",
```
anyoung-tableau: Please merge main and bump again. We are at 1.18.4 already.

src/tools/resetConsent/resetConsent.test.ts
mattcfilbert: Yeah, I think that makes sense. That way we'll get end-to-end coverage of the actual MCP tool path (like we have with the revoke access tool)

tests/oauth/tableau-authz/fixtures/connectOAuthClient.ts
```diff
@@ -26,10 +26,19 @@ export const getOAuthClientFixture: WorkerFixture<OAuthClient, { browser: Browse
...
-  // Order matters: resetConsent() uses the access token, so it must run before revocation.
-  // resetConsent() is best-effort; revoking the token is asserted so failures are surfaced.
-  await client.resetConsent();
+  // Teardown: reset consent first (requires a valid token), then revoke.
+  // Order matters: reset-consent uses the access token, so it must run before revocation.
+  // Both calls go through the MCP tool path to exercise real tool coverage.
+  // reset-consent is best-effort; revoking the token is asserted so failures are surfaced.
+  try {
+    const resetConsentResult = await client.callTool('reset-consent', {
+      schema: z.object({ message: z.string() }),
```
anyoung-tableau: This schema is wrong, yah? The tool output schema is a string, not an object.

mattcfilbert: Oh yeah - I was just going off of what was already there so this will fix an existing test inaccuracy. Good call out

anyoung-tableau: Ah yeah, I fixed the schema for the revoke-access-token tool below in passing in my last PR

mattcfilbert: word, changed it and pushed

anyoung-tableau (approved):

PR #308, opened 2026-04-15
@W-21748754: Add get-custom-view-data tool

anyoung-tableau:
This PR:

1. Adds the get-custom-view-data tool. 
2. Driveby fix to add support for view filters to get-view-image and get-view-data.
3. Driveby fix in the REST API response interceptor where URL query params were not included in the MCP notifications.


src/tools/resourceAccessChecker.ts
```diff
@@ -176,6 +178,34 @@ class ResourceAccessChecker {
...
+    });
+
+    const allowedProjectIds = await this.getAllowedProjectIds({ extra });
+    const allowedTags = await this.getAllowedTags({ extra });
+    if (!allowedProjectIds && !allowedTags) {
+      // If project filtering is enabled, we cannot cache the result since the workbook containing the view
+      // that contains the custom view may be moved between projects.
+      // If tag filtering is enabled, we cannot cache the result since the tags on the view
+      // that contains the custom view can change over time.
+      this._cachedCustomViewIds.set(customViewId, result);
```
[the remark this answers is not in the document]
anyoung-tableau: No, all the comments in this file that explain why caching is not allowed follow this format.

tests/oauth/tableau-authz/tests/getCustomViewData.test.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod';
+
+import { expect, test } from './base.js';
```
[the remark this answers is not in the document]
anyoung-tableau: Thanks, typo

src/restApiInstance.ts
```diff
@@ -291,8 +291,8 @@ function logResponse(
   const url = new URL(
     `${maskedResponse.baseUrl.replace(/\/$/, '')}/${maskedResponse.url?.replace(/^\//, '') ?? ''}`,
   );
-  if (response.request?.params && Object.keys(response.request.params).length > 0) {
-    url.search = new URLSearchParams(response.request.params).toString();
+  if (response.params && Object.keys(response.params).length > 0) {
+    url.search = new URLSearchParams(response.params).toString();
   }
```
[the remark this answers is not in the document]
anyoung-tableau: For our purposes, query params are assumed to not contain data that needs masking nor does `maskResponse` attempt to.

src/tools/views/getViewData.ts
```diff
@@ -10,6 +10,10 @@ import { Tool } from '../tool.js';
 
 const paramsSchema = {
   viewId: z.string(),
+  viewFilters: z
+    .record(z.string())
+    .optional()
+    .describe('Optional map of view filter field names to values.'),
```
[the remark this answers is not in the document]
anyoung-tableau: Thanks

docs/docs/intro.md
```diff
@@ -18,25 +18,26 @@ Tableau's official MCP Server. Helping Agents see and understand data.
...
+| [get-workbook](tools/workbooks/get-workbook.md)                                                                       | Retrieves information on a workbook from a specified Tableau site ([REST API][get-workbook])                        |
+| [get-view-data](tools/views/get-view-data.md)                                                                         | Retrieves data in CSV format for the specified view in a Tableau workbook ([REST API][get-view-data])               |
+| [get-view-image](tools/views/get-view-image.md)                                                                       | Retrieves an image for the specified view in a Tableau workbook ([REST API][get-view-image])                        |
+| [get-custom-view-data](tools/views/get-custom-view-data.md)                                                           | Retrieves data in CSV format for the specified custom view in a Tableau workbook ([REST API][get-custom-view-data]) |
```
anyoung-tableau, line 31: The table got wider so the diff is bad. This is the only new line.

src/restApiInstance.ts
```diff
@@ -291,8 +291,8 @@ function logResponse(
   const url = new URL(
     `${maskedResponse.baseUrl.replace(/\/$/, '')}/${maskedResponse.url?.replace(/^\//, '') ?? ''}`,
   );
-  if (response.request?.params && Object.keys(response.request.params).length > 0) {
-    url.search = new URLSearchParams(response.request.params).toString();
+  if (response.params && Object.keys(response.params).length > 0) {
+    url.search = new URLSearchParams(response.params).toString();
```
anyoung-tableau, line 295: Driveby fix here. This was wrong before and the query params weren't included in the response traces.

src/tools/views/getViewData.ts
```diff
@@ -10,24 +10,31 @@ import { Tool } from '../tool.js';
...
```
anyoung-tableau, line 16: Driveby updates to both get-view-data and get-view-image to also support view filters for consistency with get-custom-view-data.

src/tools/views/getViewImage.ts
```diff
@@ -37,18 +41,25 @@ export const getGetViewImageTool = (
   const getViewImageTool = new Tool({
     server,
     name: 'get-view-image',
-    description:
-      'Retrieves an image of the specified view in a Tableau workbook. The width and height in pixels can be provided. The default width and height are both 800 pixels.',
+    description: [
```
mattcfilbert, line 44: Quick question: this description points callers to a custom-view-image tool, but I don’t see one added/registered in this branch. Was the intent to introduce that in a follow-up, or should custom views still flow through get-view-image for now? I just want to avoid the model being steered toward a nonexistent tool

src/tools/resourceAccessChecker.ts
```diff
@@ -464,6 +494,59 @@ class ResourceAccessChecker {
...
+    const allowedProjectIds = await this.getAllowedProjectIds({ extra });
+    const allowedTags = await this.getAllowedTags({ extra });
+    if (!allowedWorkbookIds && !allowedProjectIds && !allowedTags) {
+      // If no filtering is enabled, there's no need to resolve the view the custom view belongs to.
+      return { allowed: true };
+    }
+
+    let underlyingViewId: string | undefined;
+    try {
+      const customView = await useRestApi({
```
mattcfilbert, line 520: Implementation Q: was it intentional to resolve the custom view here and then separately run the normal view access check, even though get-custom-view-data will make its own data call right after? 

Seems fine functionally, but I’m wondering if we want to reuse the resolved custom-view / underlying-view info somehow

src/tools/views/getViewImage.ts
anyoung-tableau: Oops, yeah, these changes came from #285 and I missed removing that note until the followup PR.

src/tools/resourceAccessChecker.ts
anyoung-tableau: I don't think there's anything we can re-use unfortunately. There's no overlap between the REST API calls that the resource check makes and what the tool makes. If there was, the resource access checker has a concept of including such content in its return value ([ref](https://github.com/tableau/tableau-mcp/blob/main/src/tools/resourceAccessChecker.ts#L12)). And [here's an example](https://github.com/tableau/tableau-mcp/blob/main/src/tools/views/listCustomViews.ts#L87) of how the list-custom-views tool uses this concept.

mattcfilbert (approved):
lgtm

PR #311, opened 2026-04-16
@W-22067441: allow any port for loopback redirect URIs in CIMD matching (#290)

mattcfilbert:
## Description

Fix CIMD (Client ID Metadata Document) redirect_uri validation so that a registered loopback URI matches a request-time URI that only differs in port, per [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3). Previously the authorize endpoint used a raw Array.includes() string match against the CIMD's redirect_uris, which rejected native clients that bind to an OS-assigned ephemeral port at request time.

Implementation is intentionally surgical:

New matchesRegisteredRedirectUri(requestUri, registeredUri) helper in src/server/oauth/matchesRegisteredRedirectUri.ts — exact string match by default, with a narrow exception that ignores port when both URIs are http: and share the same loopback host (127.0.0.1, [::1], or localhost). Scheme, host, pathname, search, hash, and userinfo must still match exactly.
Host equivalence is exact: 127.0.0.1, [::1], and localhost each match only themselves (no cross-mapping), avoiding DNS-resolution surprises and keeping the relaxation narrow.
src/server/oauth/authorize.ts now calls redirect_uris.some((uri) => matchesRegisteredRedirectUri(redirect_uri, uri)) in the CIMD path.
src/server/oauth/isValidRedirectUri.ts extended by one check to also accept [::1] (alongside localhost/127.0.0.1) so IPv6 loopback isn't rejected by the general validity gate.
Token endpoint (src/server/oauth/token.ts) unchanged — it already compares against the authorize-time stored redirectUri, so it remains self-consistent and OAuth 2.1 §4.1.3 compliant.
DCR endpoint (src/server/oauth/register.ts) unchanged — no matching happens there.
## Motivation and Context

Closes #290.

Native MCP clients that follow RFC 8252 — including Claude Code — register loopback redirect URIs like http://127.0.0.1/callback or http://localhost/callback in their client metadata, then issue authorize requests with an ephemeral port such as http://127.0.0.1:54321/callback. The previous strict includes() comparison rejected these requests, breaking OAuth entirely for such clients. RFC 8252 §7.3 explicitly requires: "the authorization server MUST allow any port to be specified at the time of the request for loopback IP redirect URIs".
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## How Has This Been Tested?

New unit suite — src/server/oauth/matchesRegisteredRedirectUri.test.ts (21 cases):

Exact matches: identical http/https/custom-scheme URIs.
Loopback port relaxation: IPv4 (127.0.0.1), IPv6 ([::1]), and localhost; with and without a port on the registered URI; preserving query and fragment.
Loopback mismatches still rejected: path, query, fragment, scheme, cross-host loopback (both directions), userinfo.
Non-loopback strictness: http and https port differences on public hosts; custom-scheme mismatches.
Invalid input: unparseable URIs and empty strings return false.
New CIMD integration cases in tests/oauth/embedded-authz/clientIdMetadataDocuments.test.ts:

Accepts http://127.0.0.1:54321/callback against registered http://127.0.0.1/callback → 302.
Accepts http://localhost:54321/callback against registered http://localhost/callback → 302.
Accepts http://[::1]:54321/callback against registered http://[::1]/callback → 302.
Rejects loopback path mismatch (http://127.0.0.1:54321/evil) → 400.
Rejects cross-host loopback (http://localhost:54321/callback against registered http://127.0.0.1/callback) → 400.
Rejects non-loopback https port mismatch (https://example.com:8443/cb against registered https://example.com/cb) → 400.
Verification run locally:

npm test — 1012 / 1012 pass
npm run test:oauth:embedded — 86 / 86 pass (including 17 in clientIdMetadataDocuments)
npx tsc --noEmit — clean
npx eslint on all touched files — clean

## Related Issues

Fixes #290

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "1.18.5",
+  "version": "1.18.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "1.18.5",
+      "version": "1.18.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.26.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.5",
+  "version": "1.18.6",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/server/oauth/authorize.ts
+++ b/src/server/oauth/authorize.ts
@@ -14,6 +14,7 @@ import { clientMetadataCache } from './clientMetadataCache.js';
 import { getDnsResolver } from './dnsResolver.js';
 import { generateCodeChallenge } from './generateCodeChallenge.js';
 import { isValidRedirectUri } from './isValidRedirectUri.js';
+import { matchesRegisteredRedirectUri } from './matchesRegisteredRedirectUri.js';
 import { TABLEAU_CLOUD_SERVER_URL } from './provider.js';
 import { cimdMetadataSchema, ClientMetadata, mcpAuthorizeSchema } from './schemas.js';
 import { getSupportedScopes, parseScopes, validateScopes } from './scopes.js';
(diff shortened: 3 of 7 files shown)


anyoung-tableau (approved):
💪💪💪💪

PR #314, opened 2026-04-17
@W-21785596 logging refactor

jarhun88:

the pull request._

# Pull Request Template

Introduces a simplified three-level logging system (debug | info | error) with environment-variable-driven severity filtering, consolidates the logging API surface, and adds structured log() calls across previously unlogged or inconsistently logged code paths. 

Breaking change: DEFAULT_LOG_LEVEL renamed to DEFAULT_NOTIFICATION_LEVEL to distinguish it from the new LOG_LEVEL env var. DEFAULT_NOTIFICATION_LEVEL controls what's sent to the MCP client; LOG_LEVEL controls what's written to console/file.

New env var: LOG_LEVEL controls server-side log output severity (debug, info, error). Defaults to info.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe):

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/tool.ts
```diff
@@ -150,6 +150,11 @@ export class Tool<Args extends ZodRawShape | undefined = undefined> {
     const username = tableauAuthInfo?.username;
 
     this.logInvocation({ requestId, args, username });
```
anyoung-tableau: `logInvocation` is poorly named now since it sends notifications. What feels right? Rename to something like `notifyOfInvocation`? Or rename to `logAndNotify` and move the call to `log` inside it?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.7",
+  "version": "1.20.0",
```
anyoung-tableau: We've historically bumped the minor version when making a breaking change like this but to follow the principles of semver properly, we really should be making *major* version bumps for any breaking change.

src/logging/logger.ts
```diff
@@ -17,21 +41,25 @@ export function parseLoggerTypes(value: string | undefined): Set<LoggerType> {
...
-    // eslint-disable-next-line no-console -- console.log is intentional here since the transport is not stdio.
-    console.log(JSON.stringify(entry));
+  if (!shouldLog(entry.level, config.logLevel)) {
+    return;
+  }
+  if (config.loggers.has('appLogger')) {
+    const message = JSON.stringify(entry);
+    if (config.transport === 'http') {
+      // eslint-disable-next-line no-console -- console.log is intentional here since the transport is not stdio.
+      console.log(message);
```
anyoung-tableau: I think we should extend `LogParams` to include an optional `error?: unknown` field, that way we can be smarter about how we log the error instead of just including it as part of the `message` using a template literal (which will remove important information from the error like stack trace). Thoughts?

```ts
if (config.transport === 'http') {
  if (error) {
    // LOG THE WHOLE ERROR
    console.log(message, error);

    // OR JUST A SUBSET OF THE ERROR
    console.log(message, { message: error.message, stack: error.stack });
  } else {
    console.log(message);
  }
} else {
  process.stderr.write(message.endsWith('\n') ? message : `${message} ${error}\n`);
  if (error) {
    process.stderr.write(${error}\n`);
  }
}
```

src/tools/tool.ts
jarhun88: yea i can rename it

package.json
jarhun88: yea that makes sense. updated

src/logging/logger.ts
jarhun88: how did you know that including as just message would remove some info? but i will try it out since its just an additive change

src/logging/logger.ts
```diff
@@ -17,21 +41,31 @@ export function parseLoggerTypes(value: string | undefined): Set<LoggerType> {
...
-  if (config.transport === 'http' && config.loggers.has('appLogger')) {
-    // eslint-disable-next-line no-console -- console.log is intentional here since the transport is not stdio.
-    console.log(JSON.stringify(entry));
+  if (!shouldLog(entry.level, config.logLevel)) {
+    return;
+  }
+  if (config.loggers.has('appLogger')) {
+    const message = JSON.stringify(entry);
+    if (config.transport === 'http') {
+      if (entry.error) {
```
mattcfilbert, line 45: Both branches are identical here

The else passes entry.error (which is undefined) as a second arg to console.log, so every HTTP-transport log line without an error prints a trailing undefined. 

The else branch should just be `console.log(message)`. The test at logger.test.ts:72 currently asserts the same (`toHaveBeenCalledWith(JSON.stringify(entry), undefined)`) and would need updating too.

jarhun88: oh thanks, i think the ai messed that up

jarhun88: updated!

src/logging/logger.ts
anyoung-tableau: How did I know? I had learned it at some point, so I knew that the string-representation of an `Error` doesn't include the stack trace. You can see exactly what the string representation is [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/toString#description).

src/logging/fileLogger.ts
```diff
@@ -1,15 +1,13 @@
...
 import { existsSync, mkdirSync } from 'fs';
 import { join } from 'path';
 
 import { getExceptionMessage } from '../utils/getExceptionMessage.js';
-import { writeToStderr } from './logger.js';
 
 export type LogEntry = {
   message: unknown;
-  level: LoggingLevel;
+  level: string;
```
anyoung-tableau: Should this be of type `LogLevel` from logger.ts?

src/server/oauth/accessTokenValidator.ts
```diff
@@ -101,7 +102,12 @@ export class EmbeddedAccessTokenValidator extends AccessTokenValidator {
         expiresAt: payload.exp,
         extra: tableauAuthInfo,
       });
-    } catch {
+    } catch (error) {
```
anyoung-tableau, line 105: All the places that catch an error and then log should stop including the string representation of the error in the message and put the error on the `error` key of the log entry.

jarhun88: youre so right

src/restApiInstance.ts
```diff
@@ -165,6 +166,12 @@ const getNewRestApiInstanceAsync = async (
     }
   }
 
+  log({
+    message: `Auth established via ${config.auth}`,
+    level: 'info',
+    logger: 'auth',
+  });
+
```
anyoung-tableau: Maybe be wary of logging this 'cause it's gonna be super spammy. Also, we probably only care when it _fails_, which I believe we get for free from the error response interceptor. Or change it to level: debug.

src/server/oauth/accessTokenValidator.ts
jarhun88: how about i make the message type a string and the error is the unknown anything goes type

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -105,15 +106,30 @@ The default logging level of the server.
...
 
-This value determines the minimum log level in which to send notifications to MCP clients. That is,
-if the server's currently configured minimum logging level is `debug`, all log messages will be sent
-to MCP clients. If the level is set to `error`, only log messages with a level of `error` or higher
-will be sent. Note that MCP clients can
-[change the minimum log level](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#setting-log-level)
+This value determines the minimum level at which to send notifications to MCP clients. That is,
+if set to `debug`, all notifications will be sent. If set to `error`, only notifications with a
+level of `error` or higher will be sent. Note that MCP clients can
+[change the minimum level](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#setting-log-level)
```
anyoung-tableau: While we're here, please update the spec version:

https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#setting-log-level

src/index.ts
```diff
@@ -44,42 +56,57 @@ async function startServer(): Promise<void> {
...
+        level: 'info',
+        logger: 'startup',
+      });
       break;
     }
   }
 
   if (config.disableLogMasking) {
-    writeToStderr('⚠️ Log masking is disabled!');
+    log({ message: 'Log masking is disabled!', level: 'info', logger: 'startup' });
```
anyoung-tableau: nit: lost the super important emoji

src/logging/logger.ts
anyoung-tableau: image

src/server/oauth/accessTokenValidator.ts
anyoung-tableau: Perfect

src/logging/notification.ts
```diff
@@ -129,3 +135,27 @@ function getSendNotificationMessageFn(level: LoggingLevel) {
...
+
+function safeStringifyNotificationMessage(message: NotificationMessage): string {
+  try {
+    return JSON.stringify(message);
+  } catch {
+    return '[Unable to serialize notification message]';
+  }
+}
+
+function mapNotificationLevelToLogLevel(level: LoggingLevel): 'debug' | 'info' | 'error' {
```
anyoung-tableau: Any reason not to make `NotificationLevel` and `LogLevel` equivalent? The different levels are standard.

src/logging/secretMask.ts
```diff
@@ -87,9 +87,10 @@ function clone<T>(obj: T): Result<T, Error> {
 
     const message = getExceptionMessage(error);
     log({
-      message: `Could not clone object, notification may not be sanitized! Error: ${message}`,
+      message: 'Could not clone object, notification may not be sanitized!',
       level: 'error',
       logger: 'secretMask',
+      error: message,
```
anyoung-tableau: We should pass `error` here, not `message`.

src/server/oauth/token.ts
```diff
@@ -213,7 +213,7 @@ export function token(
             log({
               message:
                 'Tableau refresh token exchange failed, reusing existing (possibly expired) access token',
-              level: 'error',
+              level: 'debug',
               logger: 'oauth',
```
anyoung-tableau, line 217: Let's update `exchangeRefreshToken` to return the thrown error in the `Err` result so it can be provided to this call to `log()`

src/restApiInstance.ts
```diff
@@ -212,9 +206,10 @@ export const getRequestErrorInterceptor =
   (error, baseUrl) => {
     if (!isAxiosError(error) || !error.request) {
       log({
-        message: `Request ${requestId} failed with error: ${getExceptionMessage(error)}`,
+        message: `Request ${requestId} failed`,
         level: 'error',
         logger: 'rest-api',
+        error: getExceptionMessage(error),
```
anyoung-tableau: Should just be `error`

src/restApiInstance.ts
```diff
@@ -250,9 +245,10 @@ export const getResponseErrorInterceptor =
   (error, baseUrl) => {
     if (!isAxiosError(error) || !error.response) {
       log({
-        message: `Response from request ${requestId} failed with error: ${getExceptionMessage(error)}`,
+        message: `Response from request ${requestId} failed`,
         level: 'error',
         logger: 'rest-api',
+        error: getExceptionMessage(error),
```
anyoung-tableau: Same

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -105,15 +106,30 @@ The default logging level of the server.
...
-This value determines the minimum log level in which to send notifications to MCP clients. That is,
-if the server's currently configured minimum logging level is `debug`, all log messages will be sent
-to MCP clients. If the level is set to `error`, only log messages with a level of `error` or higher
-will be sent. Note that MCP clients can
-[change the minimum log level](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/logging#setting-log-level)
+This value determines the minimum level at which to send notifications to MCP clients. That is,
+if set to `debug`, all notifications will be sent. If set to `error`, only notifications with a
+level of `error` or higher will be sent. Note that MCP clients can
+[change the minimum level](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#setting-log-level)
 any time they want.
```
anyoung-tableau, line 113: Consider leaving a note that mentions this variable was named `DEFAULT_LOG_LEVEL` until 2.0.0.

src/logging/notification.ts
jarhun88: i thought it was kinda silly and confusing to expose so many log levels options to the client, but also doing this mapping to support the notificationLevel is kinda silly too. what do you think?

src/server/oauth/token.ts
jarhun88: hmm actually may be this log isnt even necessary since it already logs in the catch block of exchangeRefreshToken. thoughts?

log({
      message: 'Failed to exchange refresh token',
      level: 'error',
      logger: 'oauth',
      error,
    });

anyoung-tableau: Oh you're right, nvm

src/logging/notification.ts
anyoung-tableau: We don't have control over the log levels exposed to the client. It's part of the MCP spec and the levels are more-or-less standard (e.g. look at the similar [log levels in .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/overview?tabs=command-line#log-level)). I feel it would be _less_ complex to just use the same set instead of needing to maintain a map.

docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -112,6 +112,8 @@ level of `error` or higher will be sent. Note that MCP clients can
 [change the minimum level](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging#setting-log-level)
 any time they want.
 
+Note: this value was named DEFAULT_LOG_LEVEL until version 2.0.0
```
anyoung-tableau: nit: since we're talking about the variable's name, not its value, replace `value` with `variable`

anyoung-tableau (approved):
🪵

PR #321, opened 2026-04-21
@W-22169002: Add support for build variants

anyoung-tableau:
This PR is the culmination of the below PRs. For context see this document: https://docs.google.com/document/d/1VWMqI2rQEu24yPOo0BMIP3Oh2axCTC9XqIjYhANfdvA/edit?tab=t.0

1. [@W-22169002: Use composition over inheritance for Server class](https://github.com/tableau/tableau-mcp/pull/321/)
2. [@W-22169202: Make Server abstract and create WebMcpServer class](https://github.com/tableau/tableau-mcp/pull/322)
3. [@W-22169638: Rename ToolName -> WebToolName](https://github.com/tableau/tableau-mcp/pull/323)
4. [@W-22169856: Rename Tool -> WebTool](https://github.com/tableau/tableau-mcp/pull/324)
5. [@W-22170386: Create TableauWebToolContext, TableauWebRequestHandlerExtra, and TableauWebToolCallback](https://github.com/tableau/tableau-mcp/pull/325)
6. [@W-22184369: Move “web” tools to tools/web](https://github.com/tableau/tableau-mcp/pull/326)
7. [@W-22185056: Add BUILD_VARIANT concept with single "default" variant](https://github.com/tableau/tableau-mcp/pull/327)
8. [@W-22187847: Add DesktopMcpServer, DesktopTool, TableauDesktopToolContext, TableauDesktopRequestHandlerExtra, and TableauDesktopToolCallback](https://github.com/tableau/tableau-mcp/pull/328)
9. [@W-22188550: Add "desktop" build variant](https://github.com/tableau/tableau-mcp/pull/330)
10. [@W-22188766: Add "combined" build variant](https://github.com/tableau/tableau-mcp/pull/331)
11. [@W-22189756: Add build artifact validation](https://github.com/tableau/tableau-mcp/pull/332)

<hr />

Initially, this PR only included changes to use composition over inheritance for the Server class. The original description is provided below:

These changes update the `Server` class so it no longer inherits from the `McpServer` class from the SDK and instead keeps an instance of `McpServer` inside it. This change is being made to ultimately support a world where multiple `Server` instances can contain a shared `McpServer` instance. Concrete example: `WebMcpServer` and `DesktopMcpServer` can both exist and share a single instance of `McpServer`, allowing each to register their tools simultaneously. See `index.combined.ts` in #331


src/server/express.ts
```diff
@@ -127,7 +127,7 @@ export async function startExpressServer({
 
         res.on('close', () => {
           transport.close();
-          server.close();
+          server.mcpServer.close();
```
mattcfilbert, line 138: Small concern with exposing `mcpServer` directly now that Server is composition-based

We already reach through it for `connect()` and `close()`, and `res.on('close')` closing the shared inner server feels risky if we ever have multiple wrappers over one `McpServer`

Do we want to keep `mcpServer` private and expose wrapper-owned lifecycle methods instead, so the ownership model stays clear?

src/server.ts
```diff
@@ -127,7 +130,7 @@ export class Server extends McpServer {
   };
 
   registerRequestHandlers = (): void => {
-    this.server.setRequestHandler(SetLevelRequestSchema, async (request) => {
+    this.mcpServer.server.setRequestHandler(SetLevelRequestSchema, async (request) => {
```
mattcfilbert: Will this leave us exposed to handler collisions in the shared-server case?

`setRequestHandler(SetLevelRequestSchema, ...)` is a single slot on the underlying SDK server, so if multiple wrappers call `registerRequestHandlers()` on the same McpServer, the later registration wins

Do we want that handler owned at the shared layer instead of per-wrapper?

anyoung-tableau: Yeah, let me rip this out. It doesn't make sense here anymore. We'll leave it to the consumer to register the MCP server request handler a single time. I'll do this in #331 where it's applicable.

src/server/express.ts
anyoung-tableau: I don't feel strongly that we need wrapper methods. I don't think they'd particularly address this concern anyway; they'd just shift the location of where the "reaching through" happens. There's a certain number of "unknowns" that will reveal themselves as we actually begin migrating Athena into TMCP. I feel comfortable with this approach where these foundational changes are as risk-free and easy-to-review as possible, and once we discover these unknowns, we can address them as they arise.

tests/e2e/mcpClient.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    if (!Array.isArray(result.content)) {
+      console.error(result.content);
+      throw new Error('result.content must be an array');
+    }
+
+    expect(result.content).toHaveLength(1);
+    expect(result.content[0].type).toBe(contentType);
+
```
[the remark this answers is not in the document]
anyoung-tableau: This is true, but let's address this when we get there, not part of this review.

src/logging/notification.ts
```diff
@@ -91,11 +91,9 @@ export const getNotificationMessageForTool = ({
...
 function getSendNotificationMessageFn(level: LoggingLevel) {
   return async (
-    server: Server,
+    mcpServer: McpServer,
     message: string | NotificationMessage,
-    { notifier: notifier, requestId }: NotificationMethodOptions = {
-      notifier: server.name,
-    },
+    { notifier, requestId }: NotificationMethodOptions = { notifier: 'tableau-mcp' },
   ) => {
```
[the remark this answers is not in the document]
anyoung-tableau: It's ok. Not a huge deal.

src/index.ts
mattcfilbert: Baseline still has a process-global “current Tableau server” assumption that will fight the Athena merge. `RestApi.host` / `RestApi.version` are static, startup eagerly fetches server info from config.server, and the per-request t`ableauAuthInfo.server` path never seems to become the actual RestApi host

Given the documented OAuth case where SERVER can be empty, worth moving host/version to request or session scope before we build more composition on top of this?

src/index.combined.ts
```diff
@@ -0,0 +1,86 @@
...
+import { isNotificationLevel, notifier, setNotificationLevel } from './logging/notification.js';
+import { RestApi } from './sdks/tableau/restApi.js';
+import { DesktopMcpServer } from './server.desktop.js';
+import { WebMcpServer } from './server.web.js';
+import { getExceptionMessage } from './utils/getExceptionMessage.js';
+
+const serverName = 'tableau-combined-mcp';
+const serverVersion = pkg.version;
+
+async function startServer(): Promise<void> {
```
mattcfilbert, line 20: I like the direction of splitting web vs desktop servers, but the actual composition seam still feels entrypoint-local rather than runtime-local. 

Right now default/desktop/combined each bootstrap themselves a little differently, and there’s already drift between those paths. 

Worth pulling startup/composition into a shared runtime layer so Athena doesn’t have to thread changes through three bootstraps and two config models?

src/tools/desktop/tools.ts
```diff
@@ -0,0 +1,3 @@
+import { getPlaceholderTool } from './placeholderTool.js';
```
mattcfilbert, line 1: My main concern here is test depth at the new composition seam. 

Desktop/combined currently prove that the server starts and lists tools, but not that a real desktop tool executes correctly through the shared McpServer / tool-context / logging path. 

Before using this as the Athena base, I’d want at least one non-placeholder integration test that exercises that path end to end.

src/scripts/prepareVariantPackage.ts
```diff
@@ -0,0 +1,82 @@
...
+
+if (!isVariant(variant)) {
+  throw new Error(`Invalid build variant: ${variant}. Expected one of: ${variants.join(', ')}`);
+}
+
+if (variant === 'default') {
+  throw new Error('The default variant needs no preparation');
+}
+
+type PackageVariant = Exclude<Variant, 'default'>;
```
mattcfilbert, line 26: The repo now has real variant concepts, but the artifact story still looks default-only. `prepareVariantPackage.ts` defines desktop/combined package overrides, yet CI/release seem to build and publish only the default package, and the variant READMEs are still stubs

Could be better to either wire desktop/combined into CI/release now or explicitly mark them as internal/non-shipping until Athena lands, so we don’t get accidental drift

src/index.ts
mattcfilbert, line 22: Baseline still has a process-global “current Tableau server” assumption that will fight the Athena merge. RestApi.host / RestApi.version are static, startup eagerly fetches server info from config.server, and the per-request tableauAuthInfo.server path never seems to become the actual RestApi host

Given the documented OAuth case where SERVER can be empty, worth moving host/version to request or session scope before we build more composition on top of this?

src/scripts/prepareVariantPackage.ts
anyoung-tableau: Yes, those are coming later. I don't want to publish anything with a placeholder tool.

src/index.ts
anyoung-tableau: I'm not following. We're not changing baseline/default/web variant and the desktop variant won't do any of that. Also, SERVER can never be empty in practice. There was an assumption many months ago that it could be empty in Cloud but that assumption was based on the idea we'd be using the embedded authorization server in Cloud.

src/tools/desktop/tools.ts
anyoung-tableau: Let's cross that bridge when we start migrating desktop authoring tools. I tested the placeholder tool E2E in MCPJam and don't think we need a unit test for it, since it will be going away shortly.

src/index.combined.ts
anyoung-tableau: Some drift is intentional. For example, even though the default/web variant supports http transport and the combined variant includes the tools from the default variant, it can't support the http transport because the desktop variant tools require stdio transport. Do you have an example of what you mean here?

stephendeoca (approved):

src/index.ts
anyoung-tableau: Let's chat after you are back and address as we go. Need to merge this to prevent more PR traffic jamming. I've got the first 3 stories ready to for Athena after this one.

PR #322, opened 2026-04-21
@W-22169202: Make Server abstract and create WebMcpServer class

anyoung-tableau:
These changes make the `Server` class abstract and introduces the `WebMcpServer` class which inherits it and takes the place of the old `Server` class. The motivation is to support additional `VariantMcpServer` classes which have its own set of config and tools. For example, see #328 where we introduce a `DesktopMcpServer` class.


src/tools/tool.ts
```diff
@@ -39,7 +39,7 @@ export type ConstrainedResult<T> =
  */
 export type ToolParams<Args extends ZodRawShape | undefined = undefined> = {
   // The MCP server instance
-  server: Server;
+  server: WebMcpServer;
```
mattcfilbert: Could weaken the abstraction we’re trying to introduce, yeah? 

Tool and related types are now depending on `WebMcpServer` directly, and the same pattern appears in `TableauToolContext` and tool factory signatures

That means adding `DesktopMcpServer` later either forces it through a web-shaped type or requires another broad signature rewrite

Should these be typed against `Server` instead, with narrowing only where a tool truly needs web-specific state?

anyoung-tableau: Oh I think I just got carried away with renaming. This ultimately ends up becoming a generic type in #324. I'll change it back to Server

PR #323, opened 2026-04-21
@W-22169638: Rename ToolName -> WebToolName

anyoung-tableau:
These changes rename the `ToolName` type (and its associated helpers) to `WebToolName`. The motivation here is to support the later introduction of `DesktopToolName`. See #328


src/tools/tool.ts
```diff
@@ -42,7 +42,7 @@ export type ToolParams<Args extends ZodRawShape | undefined = undefined> = {
   server: WebMcpServer;
 
   // The name of the tool
-  name: ToolName;
+  name: WebToolName;
```
mattcfilbert: Down for the rename where the concept is truly web-specific, but I think this may be pulling shared runtime types in that direction too

`Tool` and related types here now depend on `WebMcpServer` / `WebToolName`, and the same pattern shows up in `TableauToolContext` and `notification.ts`

Could make the eventual desktop split more expensive by forcing shared contracts through web-shaped types

Should these stay variant-agnostic here? (with the web-specific narrowing pushed to the registration boundary instead?)

anyoung-tableau: Yeah, it becomes variant-agnostic in the next PR with a generic type. https://github.com/tableau/tableau-mcp/pull/324/changes#diff-f3873c6faa0be1fa52b6c10d6f3c4188b9abc3848cc6d554578b04181dd8d4f5

PR #324, opened 2026-04-22
@W-22169856: Rename Tool -> WebTool

anyoung-tableau:


src/tools/toolName.ts
```diff
@@ -0,0 +1,3 @@
+import { WebToolName } from './toolName.web.js';
+
+export type ToolName = WebToolName;
```
mattcfilbert: Could make the shared `Tool` contract more web-shaped than intended

Tool is generic over `TToolName` extends `ToolName`, but `ToolName` is currently just an alias for `WebToolName`, so future variants still have to come through a web-defined shared type

This closing question may feel tedious at this point, but regardless: should `ToolName` stay variant-agnostic here, with web-specific narrowing pushed to the variant boundary instead?

PR #325, opened 2026-04-22
@W-22170386: Create TableauWebToolContext, TableauWebRequestHandlerExtra, and TableauWebToolCallback

anyoung-tableau:
These changes convert `TableauToolContext`, `TableauRequestHandlerExtra` and `TableauToolCallback` into "base" types and create `Web` extensions of those types. The motivation is to support doing the same for a `Desktop` extension of those types. Example: see #328


mattcfilbert (approved):
lgtm

PR #326, opened 2026-04-22
@W-22184369: Move “web” tools to tools/web

anyoung-tableau:
These changes move all the tools into the `web` folder. The motivation is to support the creation of tools for later variants like `Desktop`.


src/server/oauth/scopes.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import { getConfig } from '../../config.js';
-import type { WebToolName } from '../../tools/toolName.web.js';
+import type { WebToolName } from '../../tools/web/toolName.js';
```
mattcfilbert, line 9: Boundary question-cito here: after moving tools under `tools/web`, shared infrastructure is now importing `WebToolName` back out of that folder

Import direction is going the wrong way, especially for things like OAuth scope mapping that seem web-specific by nature

Should this web-shaped policy live under `tools/web`, with only the variant-agnostic scope primitives staying here?

anyoung-tableau: OAuth scope mapping is definitely web-specific and we won't have any variant-agnostic scopes any time soon. I don't think it belongs under src/tools/web but could live under src/web/server/oauth (meaning we could move `server` under a new `web` folder). I don't really feel strongly that it needs to move, at least not now.

PR #327, opened 2026-04-22
@W-22185056: Add BUILD_VARIANT concept with single "default" variant

anyoung-tableau:
These changes introduce the concept of build variants which allows for easily creating new customizable build output artifacts. Examples: #330 and #331


src/scripts/build.ts
```diff
@@ -22,7 +36,28 @@ const dev = process.argv.includes('--dev');
       'empty-import-meta': 'silent',
     },
     outfile: './build/index.js',
-  });
+    // https://esbuild.github.io/api/#define
+    define: {
```
mattcfilbert: Design Q: do we want `BUILD_VARIANT` to be runtime-readable outside the variant entry points at all? 

The entry-point swap already gives us a clean build boundary, and exposing `import.meta.env.BUILD_VARIANT` more broadly seems like it could invite shared modules to grow variant conditionals over time

Keep variant selection constrained to the entry/build layer unless a concrete use case shows up?

anyoung-tableau: Good callout. I tried to get stuff working without using `define` at all, but it got kinda messy. Let me give it another shot.

anyoung-tableau: Ok, it isn't perfect since the value for the `User-Agent` header was spread far and wide but I was able to get rid of the `define` to stop the bleeding potential.

PR #329, opened 2026-04-22
@W-21952572: Add support for request level overrides

stephendeoca:

the pull request._

# Pull Request Template

## Description

Introduces support for configuring certain variables on a per request basis via HTTP header. Request override behavior is determined by `ALLOWED_REQUEST_OVERRIDES` variable, which specifies which variables are overridable and if the override values they provide is restricted. Introduces the `ALLOW_SITES_TO_CONFIGURE_REQUEST_OVERRIDES` env variable for determining if sites can specify their own request overrides.

MAX_RESULT_LIMITS is also being changed to allow for tool specific limits to be unbounded

## Motivation and Context

With request overrides, MCP clients can adjust server behavior for individual requests without changing the server's environment variables or site settings.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit tests and local validation

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/server/passthroughAuthMiddleware.ts
```diff
@@ -5,6 +5,7 @@ import { getConfig } from '../config';
 import { RestApi } from '../sdks/tableau/restApi';
 import { ExpiringMap } from '../utils/expiringMap';
 import { AuthenticatedRequest } from './oauth/types';
+import { getCookie, getHeader } from './requestUtils';
```
stephendeoca, line 9: i refactored these two methods out of passthroughAuthMiddleware to use them with the requestOverrideMiddleware, which is no longer our approach, but I decided to keep these changes anyway since it makes more sense to put them in requestUtils anyways.

src/tools/resourceAccessChecker.ts
```diff
@@ -56,11 +55,7 @@ class ResourceAccessChecker {
   }): Promise<Set<string> | null> {
     return (
       this._testOverrides.projectIds ??
-      (
-        await getConfigWithOverrides({
-          restApiArgs: { ...extra },
-        })
-      ).boundedContext.projectIds
+      (await extra.getConfigWithOverrides()).boundedContext.projectIds
```
stephendeoca, line 49: now calls `extra.getConfigWithOverrides` from the tool registration, since it is setup to use any request overrides that appear in the tool invocation request.

src/overridableConfig.ts
```diff
@@ -2,7 +2,8 @@ import { ProcessEnvEx } from '../types/process-env.js';
 import { removeClaudeMcpBundleUserConfigTemplates } from './config.js';
 import { isToolGroupName, isToolName, toolGroups, ToolName } from './tools/toolName.js';
 
-const overridableVariables = [
+export const overridableVariables = [
```
stephendeoca, line 5: there's a lot of changes in this file, I might just recommend reviewing the entire file by itself instead of the diff view. my goal is to have as many comments to provide context on what's going on here, so if something isn't totally clear leave a comment so I can better document that.

most code is the same logic as before, but cleaned up and with more logic added onto it that covers request overrides.

One change to preexisting logic / behavior was with `MAX_RESULT_LIMIT` and `MAX_RESULT_LIMITS`. Previously if a max result limit was enforced, you couldn't apply an exception that allows a specific tool to be unbounded. If there was null tool specific limit, then we'd fallback to the max result limit value.

Now we can specify MAX_RESULT_LIMITS with no bounds for a specific tool. In code maxResultLimits is a map. If there is no mapping entry for a specific tool, then we fallback to the max result limit value. If there is, then the value of the tool specific limit can be a number or null (if unbounded).

src/tools/resourceAccessChecker.ts
```diff
@@ -71,11 +66,7 @@ class ResourceAccessChecker {
   }): Promise<Set<string> | null> {
     return (
       this._testOverrides.datasourceIds ??
-      (
-        await getConfigWithOverrides({
-          restApiArgs: { ...extra },
-        })
-      ).boundedContext.datasourceIds
+      (await extra.getConfigWithOverrides()).boundedContext.datasourceIds
```
anyoung-tableau, line 60: I think there is an obscure caching bug lingering now.

Site settings:
```json
{
  "mcpSiteSettings": {
    "settings": [
      {
        "key": "ALLOWED_REQUEST_OVERRIDES",
        "value": "INCLUDE_DATASOURCE_IDS"
      },
      {
        "key": "INCLUDE_DATASOURCE_IDS",
        "value": "A,B"
      }
    ]
  }
}
```

1. Request 1 comes in with no `X-Tableau-Mcp-Config` header to `query-datasource` with `datasourceLuid=A`
2. ResourceAccessChecker caches the fact that this DS luid `A` is allowed
3. Request 2 comes in with `X-Tableau-Mcp-Config` header as `INCLUDE_DATASOURCE_IDS=B` to `query-datasource` with `datasourceLuid=A`
4. Request 2 should not be allowed because DS `A` is not in the request overrides, however, since the allowed result was cached from the previous request, it *does* get returned.

anyoung-tableau: Fixing this could potentially get kinda messy. I think similar bugs could present themselves if the admin is making repeated changes to the MCP site settings as we *never* invalidate/expire the cached results (which is an issue in it of itself as they will grow indefinitely).

anyoung-tableau: One option would be to stop using a singleton instance of ResourceAccessChecker, and create a new instance for each request. That way cached results could never bleed from one request to another AND we fix the indefinite cache growth problem at the same time.

src/overridableConfig.ts
```diff
@@ -15,37 +16,191 @@ const overridableVariables = [
...
+   *      c. If the site override value is valid, replace the value of the given variable with its value from the site override.
+   * 3. Using the Object.hasOwn() method, check if the given variable exists as a property in the requestOverrides object.
+   *    Only when the variable is a property in the requestOverrides object, apply the following logic:
+   *      a. If the request override variable is not listed as an allowed request override, throw an error.
+   *      b. If the request override value is an empty string, we generally revert the variable to its default value / behavior;
+   *         however, determine if the override behavior conforms to any restrictions imposed on it and throw if it does not.
+   *      c. If the request override value is invalid, throw an error.
+   *      d. If the request override value is valid, determine if the override conforms to any restrictions imposed on it and throw if it does not.
+   *         Replace the value of the given variable with its value from the request override if it does not violate any restrictions.
+   */
```
anyoung-tableau, line 76: I can tell you put a lot of time and thought into defining this logic. It's really awesome.

docs/docs/configuration/mcp-config/request-overrides.md
```diff
@@ -0,0 +1,160 @@
...
+- If a restriction type is not specified, the request override variable will be `restricted` by default.
+
+:::info
+
+Examples
+
+```
+1. ALLOWED_REQUEST_OVERRIDES=INCLUDE_DATASOURCE_IDS,INCLUDE_WORKBOOK_IDS:unrestricted
+
+2. ALLOWED_REQUEST_OVERRIDES=*:unrestricted,MAX_RESULT_LIMIT:restricted,MAX_RESULT_LIMITS:unrestricted
```
anyoung-tableau: `MAX_RESULT_LIMITS` should be `restricted`

docs/docs/configuration/mcp-config/request-overrides.md
```diff
@@ -0,0 +1,160 @@
...
+
+When a variable is restricted, request overrides can only narrow or maintain the current configuration. They cannot expand access or remove limits.
+
+### `unrestricted`
+
+When a variable is unrestricted, request overrides can set any valid value, including values that expand beyond the current configuration or clear existing limits.
+
+## Providing Request Overrides
+
+When making a [tool call](https://modelcontextprotocol.io/specification/2025-11-25/server/tools#calling-tools) request, overrides can be specified using the `x-tableau-mcp-config` HTTP header. The header value is a URL-encoded query string of key-value pairs separated by `&`.
```
anyoung-tableau: Since this is an HTTP header value and not a query string, it doesn't need to be URL-encoded. We're also not doing any URL decoding after reading the header value. Let's just say "The header value is a string of key-value pairs separated by `&`."

docs/docs/configuration/mcp-config/request-overrides.md
```diff
@@ -0,0 +1,160 @@
...
+| `unrestricted` | Override value can be any valid set of tags, including values not in the current bounds. Can clear existing bounds with an empty value. |
+
+### [`MAX_RESULT_LIMIT`](env-vars.md#max_result_limit)
+
+Overrides the global maximum number of results for tools with a `limit` parameter.
+
+| Restriction Type | Behavior |
+|---|---|
+| `restricted` | Override value must be **less than or equal to** the current limit. Cannot clear existing limit. |
+| `unrestricted` | Override value can be any positive number. Can clear existing limit with an empty value. |
```
anyoung-tableau, line 123: Do you want to mention `*` here?

src/overridableConfig.ts
stephendeoca: Thanks Andy, I appreciate it.

src/server.ts
```diff
@@ -25,6 +26,8 @@ export const serverName = 'tableau-mcp';
 export const serverVersion = pkg.version;
 export const userAgent = `${serverName}/${serverVersion}`;
 
+const X_TABLEAU_MCP_CONFIG_HEADER = 'x-tableau-mcp-config';
```
anyoung-tableau: We need to add this header to the `allowedHeaders` array in express.ts so it appears on the [Access-Control-Allow-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#access-control-allow-headers) header.

docs/docs/configuration/mcp-config/request-overrides.md
stephendeoca: Technically `MAX_RESULT_LIMIT` is only positive number or empty for no limits which we cover here. I updated `MAX_RESULT_LIMITS` to mention `*`

src/server.ts
```diff
@@ -178,6 +184,36 @@ export class Server extends McpServer {
 
     return toolsToRegister;
   };
+
+  getRequestOverridesFromHeader(
```
anyoung-tableau: There's nothing about this function that requires it to be a method on the `Server` class, right? Let's move it to its own file.

src/overridableConfig.ts
```diff
@@ -89,144 +244,296 @@ export class OverridableConfig {
...
+    }
+    // Applying request overrides
+    if (
+      isRequestOverridableVariable(variableName) &&
+      Object.hasOwn(requestOverrides, variableName)
+    ) {
+      if (!this.allowedRequestOverrides.has(variableName)) {
+        throw new Error(`${variableName} is not an allowed request override`);
+      }
+      const restrictionType = this.allowedRequestOverrides.get(variableName)!;
```
anyoung-tableau, line 382: If you want to avoid the pattern where you call `has` and then `get` with a required `!` to convince the TypeScript compiler it really is going to find the entry in the map, you can do:

```ts
const restrictionType = this.allowedRequestOverrides.get(variableName);
if (!restrictionType) {
  throw new Error(`${variableName} is not an allowed request override`);
}

// TypeScript understands that restrictionType can't be undefined now.
```

src/overridableConfig.ts
anyoung-tableau, line 230: Is there a bug on this line and on the one for INCLUDE_TOOLS below? Should this be using Object.hasOwn() instead?

1. `process.env.EXCLUDE_TOOLS=list-datasources`
2. Site override: `"EXCLUDE_TOOLS": ""` (empty string)

This condition would be false and wouldn't honor the empty string override

src/server.ts
```diff
@@ -84,6 +87,8 @@ export class Server extends McpServer {
           );
         }
 
+        const requestOverridesHeader = extra.requestInfo?.headers[X_TABLEAU_MCP_CONFIG_HEADER];
```
anyoung-tableau: Consider changing this to:

```ts
extra.requestInfo?.headers[X_TABLEAU_MCP_CONFIG_HEADER]?.toString() ?? ''
```

and changing `getRequestOverridesFromHeader` to only accept a string. This will allow you to get rid of its `Array.isArray()` check that's going to happen with every request.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "1.18.9",
+  "version": "1.18.10",
```
anyoung-tableau: I've been thinking about our [mis]use of semver lately. Doing semver conventionally would mean for a new feature like this one, we'd do a full minor bump. Patch bumps are only bug fixes, and major bumps are for any breaking change. What do you think? Should we start following semver more conventionally and do a full minor bump here?

stephendeoca: Yeah I agree, this should be a minor bump not a patch bump.

src/overridableConfig.ts
stephendeoca: We've initialized include tools and exclude tools to empty array in the first line of this method, then we hit none of the if statements and return empty arrays which is the proper override behavior for the example you provided.

All original tests we had for include / exclude tools passed, and then new ones were added for site overrides which all passed. `overridableConfig.tests.ts` lines 238 - 242 cover this scenario.

Logic is a lil special here compared to most overrides since include and exclude tools conflict with each other, so we don't initialize process.env then apply overrides, we first see when we have to fallback to process.env and for the other cases we only set based on the site override value since we know its safe.

src/server.ts
stephendeoca: moved this to `requestUtils.ts` it even has a `getToolNameFromRequestBody` method which mirrors this method.

src/overridableConfig.ts
stephendeoca: I intentionally followed this pattern for readability of override logic. It's clear that the allowedOverrides not having a mapping means the override is not allowed, but its not as clear why not having a restriction type means the same thing.

anyoung-tableau (approved):
Really awesome stuff, have fun at TC. Excited to see this working with the resource access checker changes later on

anyoung-tableau (approved):

PR #330, opened 2026-04-22
@W-22188550: Add "desktop" build variant

anyoung-tableau:
These changes introduce a `desktop` variant which adds a new build entry point and output artifact that only includes the Desktop tools.


src/scripts/prepareVariantPackage.ts
```diff
@@ -0,0 +1,72 @@
...
+const variantPackageJsonOverrides = {
+  desktop: {
+    name: '@tableau/desktop-mcp-server',
+    description:
+      'MCP server for Tableau Desktop Agent API - enables AI agents to interact with Tableau workbooks',
+    bin: {
+      'tableau-desktop-mcp-server': './build/index-desktop.js',
+    },
+    exports: {
+      '.': './build/index-desktop.js',
```
mattcfilbert: The package wiring is pointing at the wrong file name here, yeah? 

The desktop build output is `./build/index.desktop.js`, but these package entries use `./build/index-desktop.js`, so the published package would resolve to a file the build never produces

Should this be derived from the same source of truth as the build script so they can’t drift?

src/server.ts
```diff
@@ -5,7 +5,8 @@ import pkg from '../package.json';
 import { setNotificationLevel } from './logging/notification.js';
 import { TableauAuthInfo } from './server/oauth/schemas.js';
 
-export const serverName = 'tableau-mcp';
+export const serverName =
```
mattcfilbert: Boundary Q here: do we want `server.ts `branching on `BUILD_VARIANT` at all? 

The rest of this stack is trying to keep variant-specific behavior in the entry/variant layer, and putting the conditional in shared code feels like an escape hatch that could spread over time

Should the variant-specific server name come from the entry point or concrete server instead?

src/scripts/prepareVariantPackage.ts
anyoung-tableau: Oh good catch. Not sure where that dash came from.

src/server.ts
anyoung-tableau: Good callout. I've removed the build time static replacement entirely.

PR #331, opened 2026-04-22
@W-22188766: Add "combined" build variant

anyoung-tableau:
These changes introduce a new `combined` variant that adds a new entry point and build output artifact that includes both the web and desktop tools.


src/index.combined.ts
```diff
@@ -0,0 +1,79 @@
...
+    },
+    {
+      capabilities: {
+        logging: {},
+        tools: {},
+      },
+    },
+  );
+
+  const webMcpServer = new WebMcpServer({ mcpServer });
```
mattcfilbert, line 61: (Moving comment)

Looks like combined mode may currently rely on registration order here

Both webMcpServer.registerRequestHandlers() and desktopMcpServer.registerRequestHandlers() register SetLevelRequestSchema on the same shared McpServer, so the later one wins and the other handler never runs

Also line 68 only initializes notification level for webMcpServer, so desktop state may never get synced in combined mode

Do we want this handler owned at the shared/combined layer instead so future protocol methods don’t end up competing for the same slot?

anyoung-tableau: Hmm, yeah maybe there are some gremlins here. I only tested this as so far to verify the tools from both variants showed up in MCPJam.

PR #332, opened 2026-04-22
@W-22189756: Add build artifact validation

anyoung-tableau:
These changes add/update unit and E2E tests to verify the separate build variants have strict delineations and tools from one do not bleed into the other.


tests/e2e/build.ts
```diff
@@ -0,0 +1,24 @@
+import { exec } from 'child_process';
+import { existsSync } from 'fs';
+import { join } from 'path';
+import { promisify } from 'util';
+
+import { Variant } from '../../src/scripts/variants';
+
+const execAsync = promisify(exec);
+
+export async function buildVariant(variant: Variant): Promise<void> {
```
mattcfilbert, line 10: Small concern with the validation entry point here: `buildVariant() `always runs with `--dirty`, and the post-build check is just `existsSync(expectedOutputPath)`

That seems like it could let a stale artifact from a previous run satisfy the test even if the current build didn’t produce it, especially if the output path or layout changes later

I accidentally poured some stale cereal earlier and I found it unpleasant. I feel the same way about stale artifacts

Remove the target file first or assert that the artifact was actually updated by this build?

anyoung-tableau: This is only a possibility when running locally, so I'm less concerned about it. When running during CI the build artifacts will always be fresh cheerios.

PR #340, opened 2026-05-06
@W-22373417: Migrate LocalExecutor from agent-to-tableau-desktop

anyoung-tableau:
These changes migrate the LocalExecutor class from https://github.com/tableau/agent-to-tableau-desktop/blob/ceb057f37f1e1b9eb039b1f8b8e0f870ab2a29d8/src/local-executor.ts

- Create AgentApi Zodios client with request and response interceptors, similar to the REST API client used by the web tools
- Additional error handling and response schema validation
- Add tests

If you want a preview of the upcoming changes that build on top of this:

1. #342 
2. #347

If you want to see what refactoring changes went in previously to facilitate the beginning of the migration from the agent-to-tableau-desktop repo, see:

1. #321
2. #348 
3. #349 
4. #351


src/logging/logger.ts
```diff
@@ -1,22 +1,6 @@
 import { getConfig } from '../config.js';
 import { getFileLogger, LogEntry } from './fileLogger.js';
 
-export const loggerTypes = ['fileLogger', 'appLogger'] as const;
```
anyoung-tableau: Moved these to their own file to resolve a circular dependency in the BaseConfig

src/config.ts
```diff
@@ -1,31 +1,21 @@
...
 import invariant from './utils/invariant.js';
 import { parseNumber } from './utils/parseNumber.js';
 
-const __dirname = getDirname();
-
-export const TEN_MINUTES_IN_MS = 10 * 60 * 1000;
-export const ONE_HOUR_IN_MS = 60 * 60 * 1000;
-export const ONE_DAY_IN_MS = 24 * 60 * 60 * 1000;
-export const THIRTY_DAYS_IN_MS = 30 * 24 * 60 * 60 * 1000;
-export const ONE_YEAR_IN_MS = 365.25 * 24 * 60 * 60 * 1000;
```
anyoung-tableau: Refactored these into a set of functions in `milliseconds.ts` since I needed more granularity.

src/desktop/getAgentApiClient.ts
```diff
@@ -0,0 +1,121 @@
...
+  log?.({
+    message: 'Agent API request',
+    level: 'debug',
+    logger: 'AgentApiClient',
+    data: {
+      method: request.method,
+      url: url.toString(),
+      headers: request.headers,
+      data: request.data,
+    },
```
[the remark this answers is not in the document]
anyoung-tableau: This is running locally and the secret already exists in a local file on the same machine in the agent manifest. Masking it would be meaningless.

src/desktop/toolExecutor/localToolExecutor.ts
```diff
@@ -0,0 +1,175 @@
...
+    return getEventsResult;
+  }
+
+  private async waitForCommand(
+    commandId: string,
+  ): Promise<Result<GetCommandStatusResponse, ExecuteCommandError>> {
+    const maxAttempts = Math.ceil(this.config.commandTimeoutMs / this.config.pollIntervalMs);
+    let attempts = 0;
+
+    const client = await getAgentApiClient(this.config);
```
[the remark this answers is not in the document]
anyoung-tableau: Not worried about this.

src/desktop/toolExecutor/localToolExecutor.ts
```diff
@@ -0,0 +1,175 @@
...
+      if (commandStatusResult.isErr()) {
+        return Err({ type: 'unknown', error: commandStatusResult.error });
+      }
+
+      const commandStatus = commandStatusResult.value;
+      if (commandStatus.status === 'completed' || commandStatus.status === 'failed') {
+        return Ok(commandStatus);
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, this.config.pollIntervalMs));
```
mattcfilbert: Should we thread MCP cancellation through this path before the follow-on desktop/session tools build on it?

Right now `AgentApiClient` supports an `AbortSignal`, but `ToolExecutor` / `LocalExecutor` never accept one, and this polling sleep is a bare `setTimeout`. So if the MCP client cancels a desktop tool call after the command is submitted, we can keep polling until timeout and potentially leave a mutating Desktop command running from the server side’s perspective

Could we pass the request `AbortSignal` into `executeCommand`, forward it into `getAgentApiClient`, and make this sleep abortable too?

src/sdks/desktop/agentApi/client.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (this._authToken) {
+      return this._authToken;
+    }
+
+    try {
+      if (!existsSync(this._tokenPath)) {
+        return undefined;
+      }
+
+      const tokenData = agentTokenSchema.parse(JSON.parse(readFileSync(this._tokenPath, 'utf-8')));
```
mattcfilbert, line 142: This is reading the Desktop agent token manifest, but only uses the token and ignores the `port` from the same file

Elevated risk because the default base URL is still hard-coded separately as `127.0.0.1:8765`? If Desktop writes a token file for a different local port, we’ll successfully discover the token but send it to the wrong Agent API unless the user also manually sets `AGENT_API_BASE`

Could we either derive the default Agent API base from this manifest too, or keep token/base discovery out of this low-level client until `DesktopDiscoverer` / `DesktopInstance` can provide them as a single bound pair?

src/desktop/toolExecutor/toolExecutor.ts
```diff
@@ -0,0 +1,34 @@
...
+  namespace: string;
+  args?: Record<string, any>;
+  schema?: Z;
+};
+
+export type GetEventsArgs = {
+  sinceSequence?: number;
+};
+
+export type ExecuteCommandError =
```
mattcfilbert, line 21: This error shape could fight the desktop tool wrapper

`ToolExecutor.executeCommand` returns plain `ExecuteCommandError` objects, but `DesktopTool.logAndExecute` renders failures with `result.error.getErrorText()`, which assumes `McpToolError`. That means the upcoming migrated tools either cannot directly return LocalExecutor results through the normal wrapper, or each tool will need its own adapter from executor errors into MCP tool errors

Could we move that mapping to the executor boundary now, either by returning `McpToolError` here or by adding one shared `ExecuteCommandError => McpToolError` adapter?

src/sdks/desktop/agentApi/client.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return {
+      headers: { Authorization: `Bearer ${authToken}` },
+    };
+  }
+
+  async executeCommand(
+    request: ExecuteCommandRequest,
+  ): Promise<Result<ExecuteCommandResponse, unknown>> {
```
WillMartin, line 94: Small note - I think there's a small change to the interface in the latest doc proposed by Michael/Lauren. We should double check/align here on the initial PoC API compared to the release version. I would still do this to keep us moving forward but there will be a change.

src/sdks/desktop/agentApi/client.ts
```diff
@@ -0,0 +1,148 @@
...
+      signal: AbortSignal;
+      requestInterceptor: [RequestInterceptor, ErrorInterceptor?];
+      responseInterceptor: [ResponseInterceptor, ErrorInterceptor?];
+    }>;
+  }) {
+    this._authToken = authToken;
+
+    this._tokenPath =
+      process.platform === 'win32'
+        ? join(process.env.LOCALAPPDATA ?? '', 'Tableau', 'Desktop', 'agent-token.txt')
```
WillMartin, line 45: did I continue calling this .txt even though it's a JSON file? We should make a note to change that in our actual implementation

src/desktop/getAgentApiClient.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  log({
+    message: 'Agent API response',
+    level: 'debug',
+    logger: 'AgentApiClient',
+    data: {
+      status: response.status,
+      url: url.toString(),
+      headers: response.headers,
+      data: response.data,
```
WillMartin, line 123: response.data is going to be the entire TWB for many of these. Do we want to log that much data?

src/desktop/toolExecutor/toolExecutor.ts
anyoung-tableau: I don't think tools should return LocalExecutor results directly since they are not necessarily designed to be LLM-friendly. They are designed to be easily digestible in the tool's implementation.

Have a look at the upcoming draft PR for the list-instances tool. https://github.com/tableau/tableau-mcp/pull/347/changes#diff-39267f86fbe5007a820d041b371790da05415cc9540bd690c1f7428b36119cb9

Any translation from ExecuteCommandResponse to tool output shouldn't happen here. I don't have a PR yet, but I am in the process of migrating the tool that gets the active workbook XML, and if any desktop commands fail, I created a new `DesktopCommandExecutionError` class that extends `McpToolError` but [its constructor](https://github.com/tableau/tableau-mcp/blob/anyoung/workbook-tools/src/errors/mcpToolError.ts#L159) accepts the `ExecuteCommandError` and adapts appropriately. This feels like a decent enough design.

src/desktop/getAgentApiClient.ts
anyoung-tableau: Probably not. Maybe we can hook into @mattcfilbert PR that truncates notification lengths. https://github.com/tableau/tableau-mcp/pull/355

src/desktop/toolExecutor/localToolExecutor.ts
anyoung-tableau: Oh, yes. The AgentApiClient can accept the abort signal, but looks like I didn't wire it up through the executor. Let me see if that's doable.

src/sdks/desktop/agentApi/client.ts
anyoung-tableau: Looks like it's still .txt in the main branch and in the version of Desktop I installed a few weeks ago.

src/sdks/desktop/agentApi/client.ts
anyoung-tableau: Changes are welcomed and the design should make it pretty trivial to react quickly as the API surface changes. I have their design proposal doc at-hand, and nothing stands out as something that would make reacting problematic.

src/sdks/desktop/agentApi/client.ts
anyoung-tableau: @WillMartin , can you remind what the difference is between agent-token and agent-manifest? Can we flatten these into a single construct? In any case, this isn't really a question for this PR, but more about the client API design itself.

src/desktop/getAgentApiClient.ts
anyoung-tableau: I think I will leave it for now. The default log level is `info` and this only writes when it's explicitly set to `debug` using the `LOG_LEVEL` env var

src/desktop/toolExecutor/toolExecutor.ts
mattcfilbert: Word, that makes sense. I agree the executor result itself should not become the tool output. I was mostly worried about the error adaptation boundary, not translating successful command responses into LLM-facing content

If `DesktopCommandExecutionError extends McpToolError` is the shared adapter that desktop tools use when an executor command fails, that addresses the shape concern for me. My only ask would be that we make that the obvious/common pattern when the first real tool lands, rather than each migrated tool hand-rolling its own mapping

So I’m good with not changing this PR if that adapter is coming in the tool migration PR

anyoung-tableau: Thanks, yeah, all tools that can return some kind of error must return an extension of `McpToolError` so I think the easiest is to create classes that extend `McpToolError` whose ctor accepts a `ExecuteCommandError`. It could be something as simple as `DesktopCommandExecutionError extends McpToolError` or more specific depending on the tool's implementation.

src/sdks/desktop/agentApi/client.ts
mattcfilbert: Since it is for the future, I'll resolve the thread

mattcfilbert (approved):
lgtm

PR #342, opened 2026-05-07
@W-22373428: Migrate SessionManager, DesktopInstance, and DesktopDiscoverer from agent-to-tableau-desktop

anyoung-tableau:
These changes build on top of:
- #340

These changes still effectively "do nothing" since the first tool to be migrated happens in the followup PR:
- #347.

The files from the agent-to-tableau-desktop that were migrated are listed below:

- https://github.com/tableau/agent-to-tableau-desktop/blob/main/src/session-manager.ts
- https://github.com/tableau/agent-to-tableau-desktop/blob/main/src/desktop-discovery.ts


src/desktop/sessionManager.ts
```diff
@@ -0,0 +1,76 @@
...
+        data: {
+          sessionId,
+          pid: desktopInstance.pid,
+          port: desktopInstance.port,
+        },
+      });
+    }
+
+    session.lastAccess = Date.now();
+    return session.executor;
```
mattcfilbert, line 58: Looks like the cache now captures a per-request cancellation signal for the lifetime of the session

On the cache-miss path we create `LocalExecutor({ signal, ... })`, and `LocalExecutor` then reuses that same signal for `executeCommand`, `getEvents`, `waitForCommand`, and the abortable polling sleep. But on cache hits we ignore the new request's signal and return the original executor

That means if the first request for a session is later aborted (client cancellation, transport close, timeout, etc) every future call for that same session can inherit an already-aborted signal and fail until the session is evicted. There is no eviction path today

Could we avoid storing the request signal on the cached executor? 

(Either pass the signal per executor operation, or recreate/evict the executor when its captured signal has been aborted)

src/desktop/desktopInstance.ts
```diff
@@ -0,0 +1,45 @@
...
+  }: DesktopInstanceMetadata & { signal: AbortSignal }) {
+    this.pid = pid;
+    this.port = port;
+    this.secret = secret;
+    this.start_time = start_time;
+    this.signal = signal;
+  }
+
+  async isAlive(): Promise<boolean> {
+    const client = await getAgentApiClient({
```
mattcfilbert, line 28: Should `isAlive()` should use this instance's secret when probing the Agent API?

Right now the client is created with the instance-specific port, but not the instance-specific `secret`. That means `AgentApiClient` falls back to its global token discovery path instead of using the `(port, secret)` pair from the manifest

With multiple Desktop instances, or if the global token file is absent/stale/points at a different instance, this liveness check can authenticate against the wrong local Agent API identity even though discovery already gave us the right secret

Could we pass `authToken: this.secret` here alongside the `agentApiBase`?

anyoung-tableau: Yeah, I can add that. This is a direct migration from the prototype repo though, so the design will likely change as Lauren formalizes the API. There's going to be some reacting we'll need to do as she progresses on that design, which is fine... I don't want to wait for those pieces to land before making progress here.

src/desktop/sessionManager.ts
anyoung-tableau: Oh yes, good call

mattcfilbert:
The abort signal caching fix looks right to me. One small regression test would make it harder to accidentally reintroduce the original bug: assert that `LocalExecutor.executeCommand({ signal })` passes that same request-level signal into `getAgentApiClient`, including the polling path

That’s the behavior that makes the shared SessionManager safe across requests; today the tests require a signal argument but don’t prove it is the one used for the Agent API calls

mattcfilbert (approved):
lgtm

PR #348, opened 2026-05-12
@W-22373417: Rename LogEntry.error to LogEntry.data

anyoung-tableau:
While I was working on #340 , it became apparent that logging data for non-error scenarios was going to be necessary. These changes:

1. Rename `error` to `data` in `LogEntry`
2. Remove `data` from the serialized message string since it will be logged by itself
3. Add some error handling for data that cannot be stringified

### DIFF
--- a/types/process-env.d.ts
+++ b/types/process-env.d.ts
@@ -23,6 +23,7 @@ export interface ProcessEnvEx {
   JWT_ADDITIONAL_PAYLOAD: string | undefined;
   DATASOURCE_CREDENTIALS: string | undefined;
   DEFAULT_NOTIFICATION_LEVEL: string | undefined;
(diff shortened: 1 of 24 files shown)


types/process-env.d.ts
```diff
@@ -23,6 +23,7 @@ export interface ProcessEnvEx {
   JWT_ADDITIONAL_PAYLOAD: string | undefined;
   DATASOURCE_CREDENTIALS: string | undefined;
   DEFAULT_NOTIFICATION_LEVEL: string | undefined;
+  LOG_LEVEL: string | undefined;
```
anyoung-tableau, line 26: This was a miss from the logging refactor

yyk16293 (approved):

PR #355, opened 2026-05-18
@W-22550685: Redact or truncate large binary/image payloads from MCP notifications

### LINKED ISSUE #312: MCP notifications for large REST API responses are huge
Tools like get-view-image can return huge amounts of PNG/SVG data, all of which get included in the MCP notifications we send back to the client that commonly write to log files. We should consider redacting or truncating these.

mattcfilbert:
## Description

Redacts large notification payloads before they are written to file logs or emitted as MCP `notifications/message` events.

This adds a centralized, non-mutating notification sanitizer that handles binary/typed-array payloads, large SVG/XML image strings, large base64 image strings, and ordinary oversized strings. Tool results themselves are unchanged.

## Motivation and Context

Fixes #312.

Tools such as `get-view-image` can receive large PNG/SVG REST API responses. At debug notification levels, those REST responses were included in MCP notifications and file logs, causing clients to write huge binary/image payloads to log files.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Validated locally with:

- `npx tsc --noEmit`
- `npm run lint`
- `npm test`

Added unit coverage for:

- Buffer and Uint8Array payload redaction before JSON serialization expands byte-indexed objects
- Large SVG/XML payload redaction
- Large base64 image payload redaction
- Ordinary oversized string truncation
- Small normal notification messages remaining unchanged
- Original notification message objects not being mutated
- Sanitized messages being used for both file logging and MCP notification emission

## Related Issues

Fixes #312

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/logging/sanitizeNotification.ts
+++ b/src/logging/sanitizeNotification.ts
@@ -1,0 +1,141 @@
+const DEFAULT_MAX_STRING_LENGTH = 8192;
+const MAX_DEPTH = 8;
+
+type SanitizeOptions = {
+  maxStringLength?: number;
+};
+
+type BinaryLike = ArrayBuffer | ArrayBufferView;
+
+export function sanitizeForNotification(value: unknown, options: SanitizeOptions = {}): unknown {
+  return sanitizeValue(value, {
+    maxStringLength: options.maxStringLength ?? DEFAULT_MAX_STRING_LENGTH,
+    seen: new WeakSet<object>(),
+    depth: 0,
+  });
+}
+
+function sanitizeValue(
+  value: unknown,
+  context: { maxStringLength: number; seen: WeakSet<object>; depth: number },
+): unknown {
+  if (typeof value === 'string') {
+    return sanitizeString(value, context.maxStringLength);
+  }
+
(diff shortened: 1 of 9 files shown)


src/logging/sanitizeNotification.ts
```diff
@@ -0,0 +1,141 @@
+const DEFAULT_MAX_STRING_LENGTH = 8192;
```
stephendeoca, line 1: Any reason this can't be an ENVIRONMENT variable? That way we can configure the value on our deployments

stephendeoca (approved):

src/logging/sanitizeNotification.ts
anyoung-tableau: +1 to making this configurable in the shared config. 8192 seems like a decent default variable.

mattcfilbert: thanks y'all. made the change

PR #357, opened 2026-05-19
@W-21770163: pulse sub userid fallback

mattcfilbert:
## Description

Adds a fallback for resolving the current Tableau user LUID when `list-pulse-metric-subscriptions` is called with a bearer token that does not include the `https://tableau.com/userId` claim.

The tool now resolves the current user by:
1. Using the token claim when present.
2. Falling back to `GET /api/{api-version}/sessions/current` when the claim is absent.
3. Passing the resolved `user.id` explicitly to the Pulse subscriptions API request.

Also adds unit coverage for the resolver and Pulse tool path, plus a docs note for the additional session API call.

## Motivation and Context

Plain TableauID users without MFA do not receive the `https://tableau.com/userId` bearer token claim. Because `list-pulse-metric-subscriptions` needs `user_id` to scope results to the current user, those users could not use the tool.

The Tableau REST API Get Current Server Session endpoint provides the current user LUID in `session.user.id`, so this change uses that endpoint as a targeted fallback without broad auth refactoring or unsafe cross-user caching.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## How Has This Been Tested?

Validated locally:

- `npx tsc --noEmit`
- `npm run lint`
- `npx vitest run src/server/oauth/resolveCurrentUserLuid.test.ts src/tools/web/pulse/listMetricSubscriptions/listPulseMetricSubscriptions.test.ts`
- `npm test`
- `npm run build`
- `git diff --check`

Focused unit coverage includes:

- Token claim present: no session API call.
- Token claim absent: session API call succeeds.
- Token claim absent: session API call fails with non-sensitive actionable error.
- Session response missing `user.id`.
- Per-request cache prevents redundant session calls.
- Pulse subscriptions tool forwards the resolved user LUID.

Not completed locally:

- Pulse e2e and `npm run test:oauth:embedded` are blocked by the repo `.env` guard in `tests/testEnv.ts`, which requires removing or renaming `.env`.

Recommended hosted validation:

- Smoke test hosted TMCP through Claude Code with a user whose token includes `https://tableau.com/userId`.
- Smoke test hosted TMCP through Claude Code with a plain TableauID / non-MFA user whose token does not include the claim.
- Confirm logs show the `/sessions/current` fallback path without exposing token or session details.

## Related Issues

Closes #TODO

## Checklist

No breaking changes.

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/pulse/listMetricSubscriptions/listPulseMetricSubscriptions.ts
```diff
@@ -44,7 +45,14 @@ Retrieves a list of published Pulse Metric Subscriptions for the current user us
             ...extra,
             jwtScopes: listPulseMetricSubscriptionsTool.requiredApiScopes,
             callback: async (restApi) => {
-              return await restApi.pulseMethods.listPulseMetricSubscriptionsForCurrentUser();
+              const userLuidResult = await resolveCurrentUserLuid({ restApi, extra });
```
anyoung-tableau: So, doing this on-demand for this specific tool is okay, but we should consider the pros and cons when compared against the bullets I mentioned in the GUS item (that I completely forgot I opened).

Pros of this approach:

1. We don't call Get Current Server Session for every request where the userId claim is missing. We only call it for those users who are calling this tool. This saves some time and load on Tableau.

Cons of this approach:

1. The `user_luid` will be missing from telemetry for all users that do not have MFA enabled for all tools _except_ this one. https://github.com/tableau/tableau-mcp/blob/0ca1839fe5c08a7512aed30abc71a3a29b164313/src/tools/web/tool.ts#L167
2. You could argue there is value in calling Get Current Server Session with every request where the userId claim is missing (or even with _every_ request) in the auth middleware since doing so will further validate the access token on top of our basic token validation. This approach has its own con since it will add some time to the request and additional load on Tableau.

What do you think? The missing telemetry feels like it could be considered a bug.

src/sdks/tableau/methods/pulseMethods.ts
```diff
@@ -140,12 +140,12 @@ export default class PulseMethods extends AuthenticatedMethods<typeof pulseApis>
...
    */
-  listPulseMetricSubscriptionsForCurrentUser = async (): Promise<
-    PulseResult<PulseMetricSubscription[]>
-  > => {
+  listPulseMetricSubscriptionsForCurrentUser = async (
+    userId: string,
+  ): Promise<PulseResult<PulseMetricSubscription[]>> => {
     return await guardAgainstPulseDisabled(async () => {
       const response = await this._apiClient.listPulseMetricSubscriptionsForCurrentUser({
-        queries: { user_id: this.userId },
```
anyoung-tableau, line 148: The `userId` property in `AuthenticatedMethods` is no longer referenced. We can remove it now.

src/tools/web/pulse/listMetricSubscriptions/listPulseMetricSubscriptions.ts
mattcfilbert: So basically we have path 1 which is an intentionally narrow pulse only fix and then path 2 which sure as heck gives us good telemetry but comes with additional costs and basically makes that lookup a core dependency for each request.

I would need to talk more about the value of that telemetry vs the cost if those were the only 2 routes. I did ask the robot and it suggested a third path that I'd like your feedback on.

To quote the robot:

```
Shared tool-layer best-effort hydration
In WebTool.logAndExecute, before running the tool, if Bearer auth lacks userId, we try /sessions/current.
If it succeeds, telemetry gets user_luid for any tool.
If it fails, normal tools still run; telemetry just remains missing.
Pulse subscriptions still requires the resolver inside the tool because it needs user_id for correctness.
```

The general gist to my understanding is that we get telemetry as long as there's no issues with hitting "/sessions/current", but if there is a problem, we don't break the flow

But if we're gonna do that and accept a slight telemetry hit, we should add some non-sensitive dimensions to our logging to boost observability / debugability (Mirriam Webster's has got to get that word in there)

I'm gonna get the robot to add tool, reqID, and the reason for the failure to successfully hit "/sessions/current" (I'm gonna go with "unauthorized" and then "unknown" as a catchall for anything out of an auth failure)

Any other added logging you think we should leverage? Any major holes in this plan that the robot and I missed?

src/sdks/tableau/methods/pulseMethods.ts
mattcfilbert: yupyupyup thank you

mattcfilbert:
This e2e failure on the run after applying the middle path on telemetry looks like a hook test time budget issue rather than a regression from the user-LUID changes. Hooks were still using Vitest’s 10s default. I added hookTimeout: 30_000 to match the e2e test timeout

src/tools/web/tool.ts
```diff
@@ -102,6 +104,8 @@ export class WebTool<Args extends ZodRawShape | undefined = undefined> extends T
     let toolResult: CallToolResult;
 
     try {
+      await this.hydrateCurrentUserLuidForTelemetry(extra);
```
anyoung-tableau: We are conceptually very close, but I feel Mr. Claude has a really heavy hand.

Since resolving the user id when it's missing on every tool call is effectively the same as doing the same for every authenticated request, I think we can accomplish the same with much less surface area by moving the resolution up one level to the auth middleware _after_ our initial token validation. This guarantees the userId is always available at the highest level so all downstream places that reference don't need to resolve it then with side effects.

Take a look at this commit and tell me what you think: https://github.com/tableau/tableau-mcp/commit/78bedd62786c5acb238d121423a005b58ef8b0d6

src/tools/web/pulse/listMetricSubscriptions/listPulseMetricSubscriptions.ts
anyoung-tableau, line 14: Could you please also remove the `skip` and comment from the test @ tests\oauth\tableau-authz\tests\listPulseMetricSubscriptions.test.ts

src/tools/web/tool.ts
mattcfilbert: Yep, good pushback. The shared hydration was over-engineered for the actual problem 

Adopted your `78bedd6` approach: the fallback now lives in TableauAccessTokenValidator so userId is guaranteed on tableauAuthInfo for any valid Bearer token, and downstream just reads `extra.getUserLuid()`

That let me drop `resolveCurrentUserLuid`, the WebTool hydration helper, and `ResolveCurrentUserError` entirely, and list-pulse-metric-subscriptions shrinks back to a one-liner

Way less surface area, no per-tool side effects. Added validator tests for both the happy path and the session-failure rejection path

src/tools/web/pulse/listMetricSubscriptions/listPulseMetricSubscriptions.ts
mattcfilbert: Of course - thank you!

src/tools/web/tool.ts
anyoung-tableau: Thanks Matt.

docs/docs/tools/pulse/list-pulse-metric-subscriptions.md
```diff
@@ -9,6 +9,8 @@ Retrieves a list of published Pulse Metric Subscriptions for the current user.
 ## APIs called
 
 - [List subscriptions](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_pulse.htm#PulseSubscriptionService_ListSubscriptions)
+- [Get current server session](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_server.htm#get-current-server-session)
+  if the bearer token does not include the current Tableau user ID claim
```
anyoung-tableau: You could probably remove this now. It's more of an implementation detail that doesn't need to be made public.

anyoung-tableau (approved):

docs/docs/tools/pulse/list-pulse-metric-subscriptions.md
mattcfilbert: swag. doing the thing. good eye, Andy

src/sdks/tableau/methods/pulseMethods.ts
```diff
@@ -212,6 +222,14 @@ async function guardAgainstPulseDisabled<T>(callback: () => Promise<T>): Promise
         // ntbue-service-chassis/-/blob/main/server/interceptors/site_settings.go
         return new PulseDisabledError().toErr();
       }
+
+      const tableauErrorCode = error.response?.headers?.tableau_error_code;
```
anyoung-tableau, line 232: Unfortunately, the Pulse embeddings service does not follow the same pattern as the core metrics service for response error codes, at least for this call. The `tableau_error_code` response header is actually empty and the error code actually appears in the response body instead.

The good news is that debugging this is trivial with all the logging we have in place.

1. TMCP sends MCP notifications to clients that include details of each request and response to all REST APIs
2. The OAuth tests listen for these notifications and log them out. Playwright then includes them in its traces.

I pulled the response from the Playwright traces when running the test and found that the body looks like this:

```json
{
      "httpErrorCode": 403,
      "message": "PERMISSION_DENIED: 0x62c06627"
}
```

I would recommend reverting the latest commit and including it in a separate PR.

Full trace:
```json
{
  "timestamp": "2026-05-21T01:06:51.230Z",
  "currentNotificationLevel": "debug",
  "notifier": "rest-api",
  "message": {
    "type": "response",
    "requestId": 2,
    "url": "https://10ax.online.tableau.com/api/-/pulse/insights/brief",
    "status": 403,
    "headers": {
      "date": "Thu, 21 May 2026 01:06:51 GMT",
      "content-type": "application/json",
      "content-length": "63",
      "connection": "keep-alive",
      "x-tableau": "Tableau Server",
      "x-content-type-options": "nosniff",
      "p3p": "CP=\"NON\"",
      "tableau_status_code": "7",
      "tableau_error_code": "",
      "trace-id": "51d7db94-38ad-49e7-9353-cc6168ebe6f8",
      "referrer-policy": "strict-origin-when-cross-origin",
      "x-ua-compatible": "IE=Edge",
      "x-xss-protection": "1; mode=block",
      "global-session-header": "MTAuNzYuMTA3LjYzOjgwODA=",
      "strict-transport-security": "max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains, max-age=31536000; includeSubDomains",
      "server": "sfdcedge",
      "x-sfdc-request-id": "66f04096d56d550de8dd6690eeb89c76",
      "x-request-id": "66f04096d56d550de8dd6690eeb89c76"
    },
    "data": {
      "httpErrorCode": 403,
      "message": "PERMISSION_DENIED: 0x62c06627"
    }
  }
}
```

mattcfilbert: thanks, Andy. updating now

PR #359, opened 2026-05-20
@W-21971012: Add list-extract-refresh-tasks tool

Akash-Rastogi:
Adds new tool to retrieve extract refresh tasks for datasources and workbooks on a Tableau site. Each task includes schedule information (frequency, next run time, schedule name on Server).

Key changes:
- New tasksApi with listExtractRefreshTasks endpoint
- ExtractRefreshTask schema with schedule support
- Coerce hours/minutes from strings to handle API response format
- New tableau:tasks:read scope for Tableau Cloud
- list-extract-refresh-tasks tool in web tools
- Admin-only access control with feature flag gating
- Complete documentation and testing

## Description

This PR implements the **first tool** in the Extract Refresh Schedule Optimization JTBD (Inform Phase) per [W-21971012](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002aJZhMYAW/view).

The `list-extract-refresh-tasks` tool retrieves all extract refresh tasks (datasources and workbooks) for a Tableau site, including:
- Task ID and target (datasource or workbook)
- Schedule information (frequency, next run time, intervals)
- Server-specific metadata (schedule name, state)

**Admin-only tool:** Requires `TMCP_ADMIN_TOOLS_ENABLED=true` feature flag and site administrator role (SiteAdministratorCreator, SiteAdministratorExplorer, or ServerAdministrator).

**API Limitation:** The Tableau REST API endpoint `/sites/{site-id}/tasks/extractRefreshes` does not support filter or pagination parameters - all tasks are returned in a single request. This is documented in the tool description and user-facing docs.

## Motivation and Context

**Problem:** Tableau site administrators need visibility into extract refresh schedules to optimize refresh windows, identify conflicts, and reduce infrastructure costs.

**Solution:** This tool provides the **data foundation** for the Extract Refresh Schedule Optimization JTBD by exposing extract refresh task information through MCP. Future tools will add schedule analysis and recommendation logic.

**Part of larger initiative:** This is the first of multiple tools planned for admin insights (see [Epic 264](https://gus.lightning.force.com/lightning/r/ADM_Epic__c/a3QEE000002AWIP2A4/view)).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

**Unit Tests (20+ tests, all passing):**
- ✅ Schema validation with type coercion (hours/minutes strings → numbers)
- ✅ API response format handling (4 different formats normalized)
- ✅ Tool execution (success, empty results, error handling)
- ✅ Admin gating (role verification, cache, error responses)
- ✅ OAuth scope metadata endpoints

**Manual Testing:**
- ✅ Tested locally with mcpJam inspector
- ✅ Verified feature flag on/off behavior
- ✅ Verified admin vs non-admin access control
- ✅ Verified empty result handling
- ✅ Verified schema validation with real API responses

**E2E Tests:**
- Removed from PR due to CI environment constraints (requires `TMCP_ADMIN_TOOLS_ENABLED=true`)
- Will be tested manually per [E2E Validation Plan](https://docs.google.com/document/d/1UCph7jBN5ReJG_HNxEYiLOREVGq36aPuCkECg597q0I/)

**CI Status:**
- ✅ All unit tests passing (1201+ tests)
- ✅ TypeScript compilation clean
- ✅ Linting clean
- ✅ OAuth embedded tests passing

## Related Issues

**Work Items:**
- [W-21971012](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002aJZhMYAW/view) - Add list-extract-refresh-tasks tool (this PR)

**Epic:**
- [Epic 264 - Tableau MCP for Admins](https://gus.lightning.force.com/lightning/r/ADM_Epic__c/a3QEE000002AWIP2A4/view)

**Related Docs:**
- [PRD - Tableau MCP for Admins](https://docs.google.com/document/d/1BIhQBq8sSRdQOzlLcQZn0S1pR4lHcJi8Q5HE9sQatcs/edit)
- [E2E Validation Plan](https://docs.google.com/document/d/1UCph7jBN5ReJG_HNxEYiLOREVGq36aPuCkECg597q0I/)
- [JTBD Work Breakdown](https://docs.google.com/document/d/1DMp-4P5ighq027IjA9iBBWLE_6TEEdjW_P1X1dMzVKs/edit)

## Checklist

  - ✅ Tool documentation page: `docs/docs/tools/tasks/list-extract-refresh-tasks.md`
  - ✅ Environment variable docs: `docs/docs/configuration/mcp-config/env-vars.md`
  - ✅ New Tasks category for admin tools
  - ✅ 12 unit tests for schemas, methods, and tools
  - ✅ OAuth scope tests updated
  - ✅ All tests passing
  - ✅ 1201+ tests passing
- [x] I have documented any breaking changes in the PR description
  - ⚠️ **Requires feature flag:** `TMCP_ADMIN_TOOLS_ENABLED=true` (default: false)
  - ⚠️ **Admin-only access:** Non-admin users receive 403 error
  - ⚠️ **New OAuth scope:** `tableau:tasks:read` required for Tableau Cloud

## Additional Notes

**Reusable Infrastructure:**
- `adminGate` utility can be reused by future admin-only tools
- OAuth scope pattern established for admin tools
- E2E test patterns for feature-flagged tools

**PR Review Feedback Addressed:**
- ✅ Documentation position moved to group Tasks with content types
- ✅ API limitation documented (no filter/pageSize/limit support)
- ✅ Smoke testing recommendation acknowledged

**Follow-up Work:**
- Additional Inform phase tools: `list-schedules`, `query-admin-insights-job-performance`
- Orchestration prompt: `extract-optimization-inform`
- Apply phase tools and prompts (separate WIs)

Screenshot 2026-05-20 at 8 38 04 AM
Screenshot 2026-05-20 at 8 08 27 AM


docs/docs/tools/tasks/_category_.json
```diff
@@ -0,0 +1,8 @@
+{
+  "label": "Tasks",
+  "position": 7,
```
Alon-ST-DATA: I think like @anyoung-tableau suggested in my change for projects, the extracts should be closer to content types, maybe before (4) or after Projects (5)

1 Data Q&A
2 Workbooks
3 Views
4 Projects 
5 Pulse 
6 Content Exploration

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,82 @@
...
+import { Ok } from 'ts-results-es';
+
+import { getConfig } from '../../../config.js';
+import { useRestApi } from '../../../restApiInstance.js';
+import { ExtractRefreshTask } from '../../../sdks/tableau/types/extractRefreshTask.js';
+import { WebMcpServer } from '../../../server.web.js';
+import { adminGate } from '../_lib/adminGate.js';
+import { ConstrainedResult, WebTool } from '../tool.js';
+
+const paramsSchema = {};
```
Alon-ST-DATA: The paramsSchema = {} --> has zero args 
No filter, pageSize, limit wired.                                                                                               
This is anti pattern like we have with other tools:
[list-workbook](https://github.com/tableau/tableau-mcp/blob/main/src/tools/web/workbooks/listWorkbooks.ts#L14-L18)
[list-projects ](https://github.com/tableau/tableau-mcp/blob/main/src/tools/web/projects/listProjects.ts#L14-L18)

Maybe unnecessary for your end goal but I would not limit that tool as it should be generic and perhaps behind the scenes that filter can optimize use case.. please review.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const getListExtractRefreshTasksTool = (
```
Alon-ST-DATA, line 23: [tmcp-list-extract-refresh-tasks-mcpjam-smoke.md](https://github.com/user-attachments/files/28084760/tmcp-list-extract-refresh-tasks-mcpjam-smoke.md)

@Akash-Rastogi I would highly recommend to have smoke testing using mcpJam or  goose before checkin.. you can find some surprising bugs :)
See attached doc for similar list I used for list-projects.
Thanks.

Alon-ST-DATA (commented):
And yes we will definitely coordinate changes after the our changes merge...

Like you will need to add scope for users:read if I merge first..

docs/docs/tools/tasks/_category_.json
Akash-Rastogi: ✅ Fixed! Moved Tasks category to position 4 (before Pulse) to group it with other content types as suggested.

Updated in commit https://github.com/tableau/tableau-mcp/commit/9d1aa561632fb0a467012aa4976f23656de109d2.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Good catch on the pattern inconsistency! I investigated adding filter/pageSize/limit parameters following the list-projects and list-workbooks pattern.

Finding: The Tableau REST API endpoint we're calling (GET /sites/{site-id}/tasks/extractRefreshes) does not support these parameters per the [official API documentation](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#list_extract_refresh_tasks_in_site). It returns all tasks in a single response without filtering or pagination options.

Alternative endpoint: There is GET /sites/{site-id}/schedules/{schedule-id}/extracts which supports pageSize/pageNumber, but:

It's Tableau Server-only (not Cloud)
Requires knowing the schedule ID upfront
Different use case (tasks for a specific schedule vs. all site tasks)
Action taken: I've documented this API limitation in both:

Tool description: Added note that API doesn't support filtering/pagination
Documentation: Added "API Limitation" callout explaining this is a Tableau API constraint
If Tableau adds these parameters in a future API version, we can extend the tool to match the list-projects/list-workbooks pattern.

Updated in commit https://github.com/tableau/tableau-mcp/commit/9d1aa561632fb0a467012aa4976f23656de109d2.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Thanks for the recommendation and the detailed smoke test checklist! I've tested the tool locally with mcpJam to verify:

✅ Tool appears when TMCP_ADMIN_TOOLS_ENABLED=true
✅ Admin gate works (403 for non-admins)
✅ Returns valid task data with proper schema validation
✅ Empty result handling works
The smoke test document you attached assumes filter/pageSize/limit parameters - per my comment on the params, the API doesn't support these. I'll coordinate with you on the smoke test updates once we confirm the API capabilities.

For formal validation, I'll follow the E2E validation plan (Section 3.6) when the feature flag is enabled on test sites.

src/restApiInstance.ts
```diff
@@ -31,7 +31,8 @@ type JwtScopes =
   | 'tableau:insights:read'
   | 'tableau:views:download'
   | 'tableau:insight_brief:create'
-  | 'tableau:mcp_site_settings:read';
+  | 'tableau:mcp_site_settings:read'
+  | 'tableau:tasks:read';
```
Akash-Rastogi: ✅ Added 'tableau:tasks:read' scope to JwtScopes union type

Why: This allows tools to request the tableau:tasks:read JWT scope when calling useRestApi(), which is required to call the List Extract Refresh Tasks API on Tableau Cloud.

Where it's used:


// In listExtractRefreshTasks.ts
const tasks = await useRestApi({
  ...extra,
  jwtScopes: listExtractRefreshTasksTool.requiredApiScopes, // includes 'tableau:tasks:read'
  callback: async (restApi) => { ... }
});

src/sdks/tableau/methods/usersMethods.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(baseUrl: string, creds: RestApiCredentials, axiosConfig: AxiosRequestConfig) {
+    super(new Zodios(baseUrl, usersApis, { axiosConfig }), creds);
+  }
+
+  /**
+   * Returns information about the specified user.
+   *
+   * @param siteId - The Tableau site ID
+   * @param userId - The user ID
+   * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#get_user_on_site
```
anyoung-tableau: This URL looks wrong. The fragment should be `#query_user_on_site` yeah?

src/sdks/tableau/methods/usersMethods.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  getUser = async ({ siteId, userId }: { siteId: string; userId: string }): Promise<User> => {
```
anyoung-tableau: We've generally tried to name methods based off of how they are documented for easier discovery. In this case, we'd name it `queryUserOnSite`

src/sdks/tableau/methods/usersMethods.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm
+ */
+export default class UsersMethods extends AuthenticatedMethods<typeof usersApis> {
+  constructor(baseUrl: string, creds: RestApiCredentials, axiosConfig: AxiosRequestConfig) {
+    super(new Zodios(baseUrl, usersApis, { axiosConfig }), creds);
+  }
+
+  /**
+   * Returns information about the specified user.
+   *
```
anyoung-tableau, line 23: We generally leave a comment listing the required scopes for each method. For this one, it would be:

```
* Required scopes: `tableau:users:read`
```

src/server/oauth/scopes.ts
```diff
@@ -69,6 +72,10 @@ const toolScopeMap: Record<
     mcp: ['tableau:mcp:datasource:read'],
     api: new Set(['tableau:content:read', 'tableau:mcp_site_settings:read']),
   },
+  'list-extract-refresh-tasks': {
+    mcp: ['tableau:mcp:tasks:read'],
+    api: new Set(['tableau:tasks:read']),
```
anyoung-tableau: Since this tool calls the `Query User On Site` REST API as part of the admin role lookup, the tool requires the `tableau:users:read` scope too. I tried calling the tool using `AUTH=direct-trust` and I got a 401 because the scope is missing.

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
...
+    super(message);
+  }
+}
+
+interface CachedRole {
+  siteRole: string;
+  expiresAt: number;
+}
+
+const cache = new Map<string, CachedRole>();
```
anyoung-tableau: This cache will grow unbounded until we run out of RAM. Consider using `ExpiringMap`

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
+import { AdminOnlyError } from '../../../errors/mcpToolError.js';
+import { RestApi } from '../../../sdks/tableau/restApi.js';
+
+const CACHE_TTL_MS = 5 * 60 * 1000; // 5 minutes
```
anyoung-tableau: Please parameterize this with an environment variable and an entry in `Config`. There are many examples e.g. MCP_SITE_SETTINGS_CHECK_INTERVAL_IN_MINUTES. This allows us to adjust the TTL if evidence suggest the cache is growing too large.

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    if (!isAdminSiteRole(siteRole)) {
+      throw new NotAdminError(siteRole);
+    }
+  },
+
+  clearCache(): void {
+    cache.clear();
+  },
```
anyoung-tableau: Unused?

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const cache = new Map<string, CachedRole>();
+
+export const adminGate = {
+  async assertAdmin(restApi: RestApi): Promise<void> {
+    const siteId = restApi.siteId;
+    const userId = restApi.userId;
+    if (!userId) {
+      throw new NotAdminError(undefined);
```
anyoung-tableau: Consider returning a `Result<true, string>` instead of relying on exception handling for control flow. Exceptions are for real errors, not logic.

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+interface CachedRole {
+  siteRole: string;
+  expiresAt: number;
+}
+
+const cache = new Map<string, CachedRole>();
+
+export const adminGate = {
```
anyoung-tableau: Is this wrapper object necessary? Can we export `assertAdmin` directly?

src/restApiInstance.ts
```diff
@@ -47,6 +48,27 @@ export type RestApiArgs = Pick<
       }
   );
 
+/**
+ * REST requests must use the same Tableau origin the session or token was issued for
+ * (e.g. the user's Cloud pod). Using {@link Config.server} when it differs from that origin
+ * causes Tableau to return 401 for otherwise valid credentials.
+ */
+export function resolveTableauRestApiOrigin(
```
anyoung-tableau: Can you please explain these changes? In practice, `Config.server` will never differ from `tableauAuthInfo?.server`. 

The line of code `const tableauServer = config.server || tableauAuthInfo?.server;` is only an artifact of a year-old assumption that we would be using the embedded authorization server in Cloud. This ended up not being true which is why #288 exists. :)

src/config.ts
```diff
@@ -129,6 +130,7 @@ export class Config extends BaseConfig {
       PRODUCT_TELEMETRY_ENABLED: productTelemetryEnabled,
       IS_HYPERFORCE: isHyperforce,
       BREAK_GLASS_DISABLE_GLOBALLY: breakGlassDisableGlobally,
+      TMCP_ADMIN_TOOLS_ENABLED: adminToolsEnabled,
```
anyoung-tableau: Let's remove the `TMCP_` prefix. That's implied. ;)

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
+import { AdminOnlyError } from '../../../errors/mcpToolError.js';
```
anyoung-tableau: This `_lib` folder seems out of place and/or unnecessary. It's fine to put this file directly under `tools/web` like the other utility files.

src/tools/web/_lib/adminGate.ts
```diff
@@ -0,0 +1,61 @@
...
+  siteRole: string;
+  expiresAt: number;
+}
+
+const cache = new Map<string, CachedRole>();
+
+export const adminGate = {
+  async assertAdmin(restApi: RestApi): Promise<void> {
+    const siteId = restApi.siteId;
+    const userId = restApi.userId;
```
anyoung-tableau: Unfortunately, the `userId` getter on the REST API will throw for users authenticating to Tableau MCP with a bearer token who have MFA disabled. For whatever reason, the `'https://tableau.com/userId'` claim does not exist for these users. You should parameterize `userId` and pass in `extra.getUserId()` instead.

src/sdks/tableau/methods/tasksMethods.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Returns a list of extract refresh tasks for the site.
+   * Each task is for a data source or workbook extract and includes schedule information.
+   *
+   * Required scopes (Tableau Cloud): `tableau:tasks:read`
+   *
+   * @param siteId - The Tableau site ID
+   * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#list_extract_refresh_tasks_in_site
```
anyoung-tableau: This link looks wrong. I think it's supposed to be https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#list_extract_refresh_tasks

src/sdks/tableau/apis/tasksApi.ts
```diff
@@ -0,0 +1,68 @@
...
+const taskEntrySchema = z.object({
+  extractRefresh: extractRefreshTaskSchema,
+});
+
+const listExtractRefreshTasksBodySchema = z.object({
+  tasks: z.union([
+    z.object({ task: z.union([z.array(taskEntrySchema), taskEntrySchema]) }),
+    z.array(taskEntrySchema),
+  ]),
+});
```
anyoung-tableau, line 25: This response shape is pretty unfortunate and the fact the empty result doesn't return an empty array doesn't help. I think we can still rely entirely on Zodios to validate and normalize the response though, without having to do it ourselves with loose types.

I haven't tested this, but I _think_ this schema that leverages Zod's `transform` function will work:

```ts
z.object({
  tasks: z.union([
    z.object({
      task: z.union([z.array(taskEntrySchema), taskEntrySchema.transform((task) => [task])]),
    }),
    z.array(taskEntrySchema).transform((tasks) => ({ task: tasks })),
    z.object({}).transform(() => ({ task: [] })),
  ]),
});
```

anyoung-tableau: I've replaced `taskEntrySchema` with `z.number()` but this [Zod Playground link](https://zod-playground.vercel.app?appdata=N4IgzgxgFgpgtgQxALhALwHQHsBGArGCAFwApgAdAJ3IDsACBuohMAazGTswFcaBLLDRIBtKrUYSu2fIVIVq9SUuZtOPfoJJjFS3XVEK9RxpgSVKCAJ4lMNbnBwxKJAJQuANNuPHb9x85cMIgsaMAAzLEo4EhIADxc6AF4APn1YgF0XL28JTOylAF8PfMlTcysbDDsHJ1dA4IRQiKiY%2BKTUsiYWVk5YouLDXUxcAmIyIqCQ8Mjo13a6TpUe-XS6fpLMzwUikHcQADcEABtuGDAUYRB5cUYljn0ARnc6ACZngGZ07QLdq5K7zjXHSSJacB4lArfX5AiQAugw3Sgx7PN50T4QqF7BEMOHASHbEDpPb7JxgAQ0FAgd4YF4AVgwAHYAGy-PhgABaWAAJgBZPj8FBhY5gGAFIA) should help demonstrate

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,82 @@
...
+        args: {},
+        callback: async () => {
+          const tasks = await useRestApi({
+            ...extra,
+            jwtScopes: listExtractRefreshTasksTool.requiredApiScopes,
+            callback: async (restApi) => {
+              // Verify user has admin privileges
+              await adminGate.assertAdmin(restApi);
+
+              return restApi.tasksMethods.listExtractRefreshTasks({
```
anyoung-tableau, line 106: Even though the REST API doesn't support filtering, is it something we could naively implement at the MCP layer?

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,82 @@
...
+          tasks: ExtractRefreshTask[],
+        ): ConstrainedResult<ExtractRefreshTask[]> => {
+          if (tasks.length === 0) {
+            return {
+              type: 'empty',
+              message:
+                'No extract refresh tasks were found. Either none exist or you do not have permission to view them.',
+            };
+          }
+          return { type: 'success', result: tasks };
```
anyoung-tableau: We need to honor [tool scoping](https://tableau.github.io/tableau-mcp/docs/configuration/mcp-config/tool-scoping) configuration here. If limits on the allowed datasources or workbooks are defined, we need to filter the `tasks` array. Take a look at `constrainDatasources` in `listDatasources.ts` for an example.

https://github.com/tableau/tableau-mcp/blob/main/src/tools/web/listDatasources/listDatasources.ts#L145-L152

tests/e2e/server.test.ts
```diff
@@ -34,10 +34,19 @@ describe('server', () => {
     it('should list tools', async () => {
       const names = await client.listTools();
       const oauthOnlyTools: ReadonlyArray<WebToolName> = ['revoke-access-token', 'reset-consent'];
-      const expectedToolNames =
-        process.env.AUTH === 'oauth'
-          ? [...webToolNames]
-          : webToolNames.filter((name) => !oauthOnlyTools.includes(name));
+      const adminOnlyTools: ReadonlyArray<WebToolName> = ['list-extract-refresh-tasks'];
```
anyoung-tableau, line 37: I'd love to see some real E2E tests for the new tool. They don't need to be as part of this PR since the tool is disabled, but you are welcome to enable the env var on the runners and I can add you to the site on 10AX we use.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,82 @@
...
+      return await listExtractRefreshTasksTool.logAndExecute({
+        extra,
+        args: {},
+        callback: async () => {
+          const tasks = await useRestApi({
+            ...extra,
+            jwtScopes: listExtractRefreshTasksTool.requiredApiScopes,
+            callback: async (restApi) => {
+              // Verify user has admin privileges
+              await adminGate.assertAdmin(restApi);
```
anyoung-tableau: Generally, for error scenarios, tools can return an `Err` result containing an instance of an `McpError` class. This keeps error handling consistent and ensures the messaging we return to clients/LLMs is consistent, friendly, and actionable if the LLM can self-correct.

Example:

https://github.com/tableau/tableau-mcp/blob/main/src/tools/web/getDatasourceMetadata/getDatasourceMetadata.ts#L136-L138

src/config.ts
Akash-Rastogi: Fixed in 005645f4 - removed TMCP_ prefix from all environment variables

src/restApiInstance.ts
Akash-Rastogi: You're absolutely right - those changes were unnecessary. Reverted in 385b43ac to the simple `config.server || tableauAuthInfo?.server` logic. Thanks for catching this!

src/sdks/tableau/apis/tasksApi.ts
Akash-Rastogi: Implemented in 005645f4 - using Zod transform as suggested. Thanks for the playground link, that was super helpful!

src/sdks/tableau/methods/tasksMethods.ts
Akash-Rastogi: Fixed in 005645f4 - corrected link to #list_extract_refresh_tasks

src/sdks/tableau/methods/usersMethods.ts
Akash-Rastogi: Fixed in 005645f4 - updated to #query_user_on_site

src/sdks/tableau/methods/usersMethods.ts
Akash-Rastogi: Fixed in 005645f4 - renamed to queryUserOnSite to match API documentation

src/sdks/tableau/methods/usersMethods.ts
Akash-Rastogi: Added in 005645f4

src/server/oauth/scopes.ts
Akash-Rastogi: Fixed in 708e77db - added tableau:users:read to both the scope map and TableauApiScope type

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - moved to tools/web/adminGate.ts

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - added ADMIN_GATE_CACHE_TTL_MINUTES env var (default 5 minutes, range 1-1440)

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - using ExpiringMap with lazy initialization and configurable TTL

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - exporting assertAdmin directly

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - using extra.getUserLuid() to handle MFA users correctly. Thanks for the heads up about that edge case!

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Fixed in 005645f4 - using Result<true, string> pattern for clean error handling

src/tools/web/_lib/adminGate.ts
Akash-Rastogi: Removed in 005645f4

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Good point about consistency. The admin gate currently throws Error when the user check fails. Should I refactor to use McpError subclasses in this PR or create a follow-up issue for consistency across all admin tools?

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Implemented in 005645f4 - added client-side filtering with 13 filterable fields (id, type, priority, consecutiveFailedCount, datasource.id, workbook.id, schedule.*) using field:operator:value syntax

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Implemented in 005645f4 - added constrainExtractRefreshTasks following the pattern from constrainDatasources. Filters by datasourceIds and workbookIds from bounded context.

tests/e2e/server.test.ts
Akash-Rastogi: Thanks for the offer! Since the tool is currently disabled via the ADMIN_TOOLS_ENABLED flag, I'd prefer to add E2E tests in a follow-up PR once we're ready to enable it. Would you be able to add me to the 10AX site when that time comes?

src/sdks/tableau/methods/tasksMethods.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const tasks = response.tasks as any;
+    if (Array.isArray(tasks)) {
+      return tasks.map((t: any) => t.extractRefresh);
+    }
+    const task = tasks.task;
+    if (Array.isArray(task)) {
+      return task.map((t: any) => t.extractRefresh);
+    }
+    return task ? [task.extractRefresh] : [];
```
anyoung-tableau: None of this is needed now that we've updated the response schema, right?

src/restApiInstance.ts
```diff
@@ -188,7 +190,6 @@ export const useRestApi = async <T>(
       // Sessions for 'oauth' and 'passthrough' are not. Signing out would invalidate the session,
       // preventing the access token from being reused for subsequent requests.
       await restApi.signOut();
-      log({ message: 'Signed out of Tableau REST API', level: 'debug', logger: 'auth' });
```
anyoung-tableau, line 191: Revert please

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (datasourceIds) {
+    tasks = tasks.filter((task) => task.datasource?.id && datasourceIds.has(task.datasource.id));
+  }
+
+  // Filter by workbookIds - only keep tasks for workbooks in the allowed set
+  if (workbookIds) {
+    tasks = tasks.filter((task) => task.workbook?.id && workbookIds.has(task.workbook.id));
+  }
+
+  if (tasks.length === 0 && originalCount > 0) {
```
anyoung-tableau: The check on `originalCount > 0` is unnecessary. It will always be `true` because of the `if (tasks.length === 0)` guard clause above.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -0,0 +1,176 @@
...
+    };
+  }
+
+  if (tasks.length === 0) {
+    return {
+      type: 'empty',
+      message:
+        'No extract refresh tasks were found. Either none exist or you do not have permission to view them.',
+    };
+  }
```
anyoung-tableau: This code will never get hit.

Akash-Rastogi (commented):
Responses to today's review comments:

**Comment 1 (tasksMethods.ts)**: Fixed in 31bff092 - removed the redundant array handling now that Zod transform normalizes the response

**Comment 2 (restApiInstance.ts)**: Fixed in 28819426 - restored the debug log line that was accidentally removed in 385b43ac

**Comment 3 (listExtractRefreshTasks.ts:157)**: Fixed in 31bff092 - removed the unnecessary originalCount check

**Comment 4 (listExtractRefreshTasks.ts:173)**: Fixed in 31bff092 - removed the unreachable code block

src/sdks/tableau/methods/tasksMethods.ts
Akash-Rastogi: Fixed in 31bff092 - removed the redundant array handling now that Zod transform normalizes the response

src/restApiInstance.ts
Akash-Rastogi: Fixed in 28819426 - restored the debug log line that was accidentally removed in 385b43ac

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Fixed in 31bff092 - removed the unnecessary originalCount check

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Fixed in 31bff092 - removed the unreachable code block

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Alon-ST-DATA: Hmm.. theoretically extracts refresh are jobs and I do see we support [job filtering ](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm#jobs) 
Thus I wonder if another tool for getting extract refresh jobs is needed... or maybe that can come later in the feature work..
@Akash-Rastogi what do you think?

src/server/oauth/scopes.ts
```diff
@@ -69,6 +73,10 @@ const toolScopeMap: Record<
     mcp: ['tableau:mcp:datasource:read'],
     api: new Set(['tableau:content:read', 'tableau:mcp_site_settings:read']),
   },
+  'list-extract-refresh-tasks': {
+    mcp: ['tableau:mcp:tasks:read'],
+    api: new Set(['tableau:tasks:read', 'tableau:users:read']),
```
anyoung-tableau, line 78: I want to at least call out something here that has some potential for being problematic.

Even though the admin tools are disabled, the MCP server is still going to advertise that these scopes must be added to the access token. ([Ref](https://github.com/tableau/tableau-mcp/blob/main/src/server/oauth/authMiddleware.ts#L59))

This is okay, but it also means the access token will have more privileges than it really needs and customers might ask "Why do you need these `tasks` and `users` scopes?" Ideally the logic that determines the list of scopes to advertise would depend on whether or not a tool is enabled.

anyoung-tableau (approved):
Thanks so much for this @Akash-Rastogi . Nice job!

src/server/oauth/scopes.ts
Akash-Rastogi: Good catch! Fixed in 3c42eba6 - made `getSupportedMcpScopes()` and `getSupportedApiScopes()` dynamic so they now filter based on `adminToolsEnabled`. When the flag is false, `tableau:tasks:read`, `tableau:users:read`, and `tableau:mcp:tasks:read` are excluded from the advertised scopes. Added unit tests to verify the filtering behavior.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Akash-Rastogi: Implemented in https://github.com/tableau/tableau-mcp/commit/005645f437d6ad28c93bb9bff7676d902c2806f0 - added client-side filtering with 13 filterable fields (id, type, priority, consecutiveFailedCount, datasource.id, workbook.id, schedule.*) using field:operator:value syntax

check src/tools/web/extractRefreshTasks/extractRefreshTasksFilterUtils.ts

Alon-ST-DATA (commented):
LGTM after comments addressed.
Ship It!

PR #360, opened 2026-05-20
@W-22354559: Add tool scoping for views

stephendeoca:

the pull request._

# Pull Request Template

## Description

Introduces `INCLUDE_VIEW_IDS` variable which can be used to limit tool results and tool calls to specific views.

## Motivation and Context

Support dashboard agent scenarios where TMCP tool execution is bounded to the workbook, views, and underlying published datasource LUIDs.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Verified various view related tools with different combinations of workbooks and views tool scoping apart from unit tests

## Checklist

## Contributor Agreement

      its Contribution Checklist.


docs/docs/configuration/mcp-config/tool-scoping.md
mattcfilbert, line 7: Tiny docs consistency thing: since this PR adds `INCLUDE_VIEW_IDS` and this page now documents view scoping, should we add `views` to this intro list too? Maybe: “data sources, workbooks, views, projects, or tags.”

I noticed the same wording also appears in `docs/docs/enterprise/tableau-server.md:374-375`, so it may be worth updating both while you’re here

mattcfilbert (approved):
Looks good. Only have a small doc change request

docs/docs/configuration/mcp-config/tool-scoping.md
anyoung-tableau: +1

src/tools/web/views/listViews.ts
```diff
@@ -137,6 +137,10 @@ export function constrainViews({
     views = views.filter((view) => (view.workbook?.id ? workbookIds.has(view.workbook.id) : false));
   }
 
+  if (viewIds) {
+    views = views.filter((view) => (view.id ? viewIds.has(view.id) : false));
+  }
```
anyoung-tableau: Feels like this check should be the first one since it's likely going to be the most restrictive and potentially remove the most (or even all) views from the array. Should we perform this filtering before the project filter above?

Also, I don't think `view.id` can be falsey, so the ternary is unnecessary.

src/tools/web/views/listCustomViews.ts
```diff
@@ -134,17 +136,55 @@ export const getListCustomViewsTool = (server: WebMcpServer): WebTool<typeof par
...
+  boundedContext,
+}: {
+  customViews: Array<CustomView>;
+  boundedContext: BoundedContext;
+}): ConstrainedResult<Array<CustomView>> {
+  if (customViews.length === 0) {
+    return {
+      type: 'empty',
+      message:
+        'No custom views were found. Either none exist or you do not have permission to view them.',
```
anyoung-tableau: Since the tool requires a workbookId, maybe clarify that in the message e.g. `No custom views for this workbook were found...`

PR #361, opened 2026-05-20
@W-22575047: Run OAuth tests using Tableau authz server during CI

anyoung-tableau:
These tests run the suite of OAuth tests that use the Tableau authorization server against 10AX. It also updates the test for Pulse insight briefs to verify the new error message we return when Tableau+ is not enabled on the site.


tests/oauth/tableau-authz/flows/loginFlow.ts
```diff
@@ -43,21 +43,33 @@ export class LoginFlow extends Flow {
...
 
   fillPassword = async (password: string): Promise<void> => {
-    await this.passwordTextbox.fill(password);
-    await this.submitPasswordButton.click();
+    // Do not use fill() because it will appear in the Playwright trace.
+    await this.passwordTextbox.evaluate(
+      (element: HTMLInputElement, password: string) => (element.value = password),
+      password,
+    );
+    await this.submitPasswordButton.evaluate((element: HTMLButtonElement) => element.click());
```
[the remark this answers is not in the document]
anyoung-tableau: Not sure how this changed. It was not intentional. Probably a Cursor suggestion.

tests/oauth/tableau-authz/tests/generatePulseInsightBriefTool.test.ts
```diff
@@ -3,8 +3,15 @@ import { expect, test } from './base.js';
 import { getSuperstoreDatasource, getTableauMcpPulseDefinition } from './testEnv.js';
 
 test.describe('generate-pulse-insight-brief', () => {
-  // This test requires Tableau AI to be enabled on the site
-  test('generate pulse insight brief', async ({ client }) => {
+  test('generate pulse insight brief', async ({ env, client }, { skip }) => {
+    skip(
```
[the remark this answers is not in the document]
anyoung-tableau: This is incorrect. `TestInfo` does contain a `skip` function.

tests/oauth/tableau-authz/testEnv.ts
```diff
@@ -6,6 +6,11 @@ const envSchema = z.object({
   TEST_USER: z.string(),
   TEST_PASSWORD: z.string(),
   TEST_SITE_NAME: z.string(),
+  FILL_SITE_NAME: z.boolean().default(false),
```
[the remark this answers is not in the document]
anyoung-tableau: Thanks

tests/oauth/tableau-authz/tests/generatePulseInsightBriefTool.test.ts
```diff
@@ -1,99 +1,110 @@
...
 import { getSuperstoreDatasource, getTableauMcpPulseDefinition } from './testEnv.js';
 
 test.describe('generate-pulse-insight-brief', () => {
-  // This test requires Tableau AI to be enabled on the site
-  test('generate pulse insight brief', async ({ client }) => {
+  test('generate pulse insight brief', async ({ env, client }) => {
     const superstore = getSuperstoreDatasource();
     const definition = getTableauMcpPulseDefinition();
 
-    const pulseInsightBrief = await client.callTool('generate-pulse-insight-brief', {
```
anyoung-tableau, line 11: This diff is bad. The only change was to create an inline function that calls the tool instead.

src/restApiInstance.ts
```diff
@@ -196,7 +196,7 @@ export const useRestApi = async <T>(
 export const getRequestInterceptor =
   (server: Server, requestId: RequestId): RequestInterceptor =>
   (request) => {
-    request.headers['User-Agent'] = getUserAgent(server);
+    request.headers['User-Agent'] = server.userAgent;
```
anyoung-tableau, line 199: This is a driveby fix for a regression I caused during the desktop variant refactoring. The user agent was reporting as `tableau-mcp/2.2.4 (client-name 1.0.0) (client-name 1.0.0)` with the client info being doubled. The unit tests missed it because `getClientVersion()` was mocked to not return any client info.

stephendeoca (approved):

tests/oauth/tableau-authz/testEnv.ts
```diff
@@ -6,6 +6,8 @@ const envSchema = z.object({
   TEST_USER: z.string(),
   TEST_PASSWORD: z.string(),
   TEST_SITE_NAME: z.string(),
+  FILL_SITE_NAME: z.coerce.boolean().default(false),
```
mattcfilbert: `z.coerce.boolean()` is a little dangerous here because this env schema is parsing strings from `process.env`: `Boolean("false")` evaluates to `true`. That means `FILL_SITE_NAME=false` or `TABLEAU_AI_DISABLED=false` would silently take the true branch, which is especially confusing now that the login flow comment invites local users to toggle `FILL_SITE_NAME`

Could we parse these the same way the rest of the repo tends to handle env booleans, e.g. `z.string().optional().transform(v => v === "true")` / `z.enum(["true", "false"]).transform(v => v === "true")`, rather than using `z.coerce.boolean()`?

anyoung-tableau: Oh damn, z.coerce.boolean() does not work how I expected it to.

jarhun88 (approved):

PR #363, opened 2026-05-21
@W-22602384 feature gating system

jarhun88:

the pull request._

# Pull Request Template

## Description
  
  Implements a feature gating system to control feature availability during development for both on-prem and cloud deployments. Features are configured via a JSON file that is validated at startup.

  ## Key Components:

  - FeatureGate class: Loads and validates feature configuration from JSON file
  - Singleton pattern: Global access via initializeFeatureGate() and getFeatureGate()
  - Zod validation: Strict schema validation with clear error messages
  - E2E test: Validates actual features.json in CI to catch config errors before deployment

##  Motivation and Context

  This provides infrastructure for gating experimental or incomplete features during development. The system:
  - Allows enabling/disabling features without code changes
  - Works for both on-prem (static file) and cloud deployments
  - Fails fast at startup if config is malformed
  - Provides single source of truth for feature flags

##  Type of Change

  - New feature
  - Bug fix
  - Breaking change
  - Documentation update
  - Other (please describe):

 ## How Has This Been Tested?

  - Unit tests (13 tests): Test FeatureGate class, validation logic, singleton pattern
  - E2E test (1 test): Validates actual features.json loads successfully
  - Config tests (203 tests): Verify FEATURE_CONFIG_PATH environment variable parsing
  - All tests passing locally

## Test Coverage:

  - Valid and invalid JSON files
  - Missing files (graceful degradation)
  - Invalid boolean values (strict validation)
  - Singleton initialization and retrieval
  - Environment variable defaults

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/config.ts
```diff
@@ -261,6 +263,7 @@ export class Config extends BaseConfig {
     this.productTelemetryEnabled = productTelemetryEnabled !== 'false';
     this.isHyperforce = isHyperforce === 'true';
     this.breakGlassDisableGlobally = breakGlassDisableGlobally === 'true';
+    this.featureConfigPath = featureConfigPath ?? 'features.json';
```
anyoung-tableau: Consider using `||` instead of `??` to provide a fallback for empty string too

src/features/featureGate.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    // If file doesn't exist, return empty map (all features disabled)
+    if (!existsSync(filePath)) {
+      console.warn(`Feature config file not found: ${filePath}. All features disabled.`);
+      return features;
+    }
+
+    try {
+      const fileContent = readFileSync(filePath, 'utf-8');
+      const config = JSON.parse(fileContent);
```
anyoung-tableau: Let's define a zod schema for the shape of the json instead. That way you don't have to do any error handling below. You could even define the schema with an enum for the feature names which would then give you a strict union type of allowable feature names you could use instead of `string` in the signature of `isFeatureEnabled()`. This lets the TypeScript compiler prevent accidental typos or feature flag renames from causing bugs or regressions.

src/features/featureGate.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  globalFeatureGate = new FeatureGate(configPath);
+  return globalFeatureGate;
+}
+
+/**
+ * Get the global feature gate instance
+ * @throws Error if not initialized
+ */
+export function getFeatureGate(): FeatureGate {
```
anyoung-tableau, line 76: Consider just lazy initializing the singleton here the first time the function is called.

src/features/featureGate.integration.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Reset singleton state after each test
+    resetFeatureGate();
+  });
+
+  it('should initialize from Config and check features', () => {
+    // Create test feature config
+    const features = { mcpapps: true, pulse: false };
+    writeFileSync(testConfigPath, JSON.stringify(features));
```
anyoung-tableau: Rather than writing and deleting files, consider mocking `existsSync` and `readFileSync`. Similar to what I'm doing here: https://github.com/tableau/tableau-mcp/pull/342/changes#diff-b7a75d9a1b739e45db908850cf2ec5d53b355259fea566d4853c08b4209b9d1eR149

src/features/featureGate.ts
jarhun88: id have to pass the configPath in the first call, which would be difficult and make the function unnecessarily require the configPath for all subsequent calls

src/features/featureGate.integration.test.ts
jarhun88: i just deleted the integration test because i didnt actually realize what additional value it had separate from the unit test

src/config.ts
jarhun88: updated

src/features/featureGate.ts
jarhun88: i defined the zod schema, but preferred against the allowable feature set, as i want to reduce the number of places where we would need to add a new flag name

src/features/featureGate.ts
anyoung-tableau: Does `configPath` even need to be parameterized? I imagine a single file at features.json would be sufficient.

CLAUDE.md
```diff
@@ -0,0 +1,99 @@
...
+2. Show user what changed (files modified/created/deleted, test results)
+3. Wait for explicit approval: "commit this" / "looks good, commit"
+4. Only then make the commit
+
+**The ONLY exception:** User explicitly says "go ahead and commit" or "auto-commit is fine for this" in their message.
+
+## Testing
+
+### No Redundant Tests
+Avoid duplicating test scenarios across unit, integration, and e2e tests:
```
anyoung-tableau: This seems like bad advice. Can we save committing this file to a separate PR?

src/features/featureGate.ts
jarhun88: thast a good point

README.md
```diff
@@ -41,6 +41,37 @@ Standard config works in most MCP clients:
 }
 ```
 
+## Feature Flags
```
anyoung-tableau: Let's move this to a doc in the Developers section on the docs site. The README should be more user-focused than developer-focused.

CLAUDE.md
jarhun88: accidentaly pushed this

src/features/featureGate.ts
jarhun88: ah so then i dont even need an env var for this

tests/e2e/featuresConfig.test.ts
```diff
@@ -0,0 +1,9 @@
+import { describe, expect, it } from 'vitest';
+
+import { FeatureGate } from '../../src/features/featureGate.js';
+
+describe('features.json validation', () => {
+  it('should load features.json successfully', () => {
```
mattcfilbert: This test can’t catch malformed config because `FeatureGate` catches read, parse, and schema errors and returns an empty feature map. That means missing JSON, invalid JSON, or `{ "mcpapps": "yes" }` all still satisfy `not.toThrow()`

Could we assert on a strict validation path or the loaded feature values so a bad features.json actually fails CI?

README.md
```diff
@@ -41,6 +41,37 @@ Standard config works in most MCP clients:
...
+```
+
+**Configuration:**
+- Default location: `features.json` in project root
+- Override with environment variable: `FEATURE_CONFIG_PATH=/path/to/features.json`
+
+**Usage in code:**
+
+```typescript
+import { getFeatureGate } from './features/featureGate.js';
```
mattcfilbert: The documented usage throws today because no production startup path calls `initializeFeatureGate()`, and `config.featureConfigPath` is never consumed outside tests. The first caller following this pattern gets `FeatureGate` not initialized instead of a boolean gate check

Could we either wire initialization during startup with `config.featureConfigPath`, or avoid exposing a singleton API until it is initialized by construction?

tests/e2e/featuresConfig.test.ts
jarhun88: i removed this test as i changed my mind and chose to use partial validation of feature flag key val pairs instead. so now if one key value pair is malformed, it will still read the rest

README.md
jarhun88: should be fine now that im lazy loading the singleton inside getFeatureGate

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.4.3",
+  "version": "2.5.0",
```
anyoung-tableau: imho this could be a patch bump. It's not really a new customer-facing feature.

jarhun88: moved to 2.4.4

anyoung-tableau (approved):

PR #365, opened 2026-05-28
@W-22571806 Set up MCP Apps dependencies and basic server structure

jarhun88:

the pull request._

# Pull Request Template

## Description

New Tool Type: App Tools

- Added AppDetails type containing name, resourceUri, and html fields
- Extended WebTool class with optional app property for storing UI metadata
- When a tool has an app property and the feature flag is enabled, it registers as an MCP App tool

Registration Flow

Added _registerAppTool() method that:
- Registers the tool with _meta.ui.resourceUri metadata via registerAppTool
- Registers the HTML/JavaScript resource via registerAppResource
- Uses tool.name as the resource name for clarity
- Added _registerTool() method for standard tool registration
- Exposed registerAppTool and registerAppResource from @modelcontextprotocol/ext-apps/server

Feature Flag Gating

- Tools with app property are checked against the mcp-apps feature flag
- When enabled: registers as app tool with UI capabilities
- When disabled: registers as standard tool (backward compatible)

No tools are registered as app tools yet.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/tool.ts
```diff
@@ -16,6 +16,22 @@ import { WebToolName } from './toolName.js';
 
 export type ToolRules = Record<string, boolean | undefined>;
 
+export type AppDetails = {
+  name: string;
+  resourceUri: string;
+  html: string;
+};
+
+export type WebToolParams<Args extends ZodRawShape | undefined = undefined> = ToolParams<
```
anyoung-tableau, line 31: I like how you extracted this into its own type. Would you mind doing the same for tools/desktop/tool.ts as well please?

src/tools/web/tool.ts
```diff
@@ -53,6 +69,7 @@ export class WebTool<Args extends ZodRawShape | undefined = undefined> extends T
   Args
 > {
   requiredApiScopes: ReadonlyArray<TableauApiScope>;
+  app?: AppDetails;
```
anyoung-tableau, line 78: Consider leaving a comment describing what an "app" is and linking to the MCPApps documentation. To a newcomer (or Claude) it may not understand "app" means "MCP App" or how to get more info about it.

src/server.web.ts
```diff
@@ -86,15 +87,13 @@ export class WebMcpServer extends Server {
...
-      this.mcpServer.registerTool(
-        name,
-        {
-          description: await Provider.from(description),
-          inputSchema: await Provider.from(paramsSchema),
-          annotations: await Provider.from(annotations),
-        },
-        toolCallback,
-      );
+      const mcpAppsEnabled = getFeatureGate().isFeatureEnabled('mcp-apps');
```
anyoung-tableau: nit: pull this out of the `for` loop

src/server.web.ts
```diff
@@ -137,4 +136,71 @@ export class WebMcpServer extends Server {
...
+            resourceUri,
+          },
+        },
+      },
+      toolCallback,
+    );
+
+    // Register the resource, which returns the bundled HTML/JavaScript for the UI.
+    this.registerAppResource(
+      // @ts-expect-error -- harmless type mismatch in registerAppResource; ext-apps uses MCP SDK v1.25.2. Should go away when MCP SDK is updated.
```
anyoung-tableau, line 189: Consider upgrading MCP SDK to 1.29 to be in alignment with [ext-apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/package.json#L80).

src/server.web.ts
```diff
@@ -137,4 +136,71 @@ export class WebMcpServer extends Server {
...
+              text: html,
+            },
+          ],
+        };
+      },
+    );
+  };
+
+  registerAppTool = registerAppTool;
+  registerAppResource = registerAppResource;
```
anyoung-tableau: I think I created these method aliases back when `Server` extended `McpServer`. It no longer does and uses composition instead, so I question their value now. Do we still need them?

jarhun88: yea theres like a param mapping issue that fails without this, perhaps we could spend a few minutes looking at it?

src/server.web.ts
jarhun88: it actually didnt solve the issue and i couldnt figure out a fix... i also wasnt sure if it would be slightly risky to bump this before our code freeze

src/tools/web/tool.ts
jarhun88: sure thing

src/server.web.ts
anyoung-tableau: If it doesn't resolve the type issue we can save the upgrade for a later date.

src/tools/web/tool.ts
jarhun88: theres no equivalent constructor to consume it, so im not sure theres any point to create it yet?

anyoung-tableau: Oh sorry, you're right. I forgot it doesn't define its own ctor.

.envPat
anyoung-tableau, line 1: revert pls

.npmrc
anyoung-tableau, line 1: Was this needed?

src/server.web.ts
anyoung-tableau: Hmm, is that a runtime error? We can debug if you'd like. Or just leave it is fine too.

.npmrc
jarhun88: ah no let me revert

src/server.web.ts
jarhun88: i managed to fix it!

anyoung-tableau (approved):

PR #367, opened 2026-05-29
@W-22720724: Add list-users admin tool

Akash-Rastogi:
## Summary

Adds a new `list-users` MCP tool (JTBD #3A-1) that retrieves all users on a Tableau site with client-side filtering and pagination. This is the foundational tool for the **User License Reclamation** admin workflow.

- Extends user schema with admin fields (`authSetting`, `locale`, `language`, `externalAuthUserId`)
- Adds `GET /sites/:siteId/users` endpoint with response normalization
- Creates admin-gated tool with 9 filterable fields and 6 filter operators
- Implements `pageSize` and `limit` for client-side result capping
- Registers new `tableau:mcp:users:read` MCP scope
- Bumps version to 2.7.0
- Also fixes `list-extract-refresh-tasks` to respect its `pageSize`/`limit` params (previously declared but ignored)

## What this enables

```
// Find inactive users (no login in 6 months)
filter: "lastLogin:lt:2025-12-01T00:00:00Z"

// Find expensive licenses with no recent activity
filter: "siteRole:in:Creator|Explorer,lastLogin:lt:2025-01-01T00:00:00Z"

// License reclamation candidates — first 20 results
filter: "siteRole:eq:Unlicensed", limit: 20
```

## Files changed (21)

**New (8):** tool, filter utils, mock, 4 test files, documentation
**Modified (13):** user.ts, usersApi.ts, usersMethods.ts, tools.ts, toolName.ts, scopes.ts, package.json, package-lock.json, listExtractRefreshTasks.ts + test, e2e test, oauth tests (x2)

## Test plan

- [x] 52 new user tests + 3 new extract-refresh-tasks tests (100% coverage on new code)
- [x] Full test suite passes (1442 tests, 95 files)
- [x] TypeScript compilation clean
- [x] ESLint clean
- [x] Build succeeds
- [x] Tested manually via MCP Inspector against Tableau Cloud (filter, limit, pageSize all verified)
- [x] E2E server test updated with `list-users` in `adminOnlyTools`
- [x] OAuth test updated with `list-users` in `ADMIN_GATED_TOOL_NAMES`
- [x] Embedded OAuth test updated with `tableau:mcp:users:read` in `scopes_supported`
- [x] CI passes

## Related

- **GUS:** [W-22720724](https://gus.my.salesforce.com/lightning/r/ADM_Work__c/a07EE00002b5pKjYAI/view)
- **JTBD:** [#3A-1 in Work Breakdown](https://docs.google.com/document/d/1DMp-4P5ighq027IjA9iBBWLE_6TEEdjW_P1X1dMzVKs)
- **Sibling PR:** [#359 — list-extract-refresh-tasks](https://github.com/tableau/tableau-mcp/pull/359)
- **Implementation Plan:** [264 doc](https://docs.google.com/document/d/148av48p_nM8VDk8awRh49CmkhLI23gYsGZODxlR-SGc)

### DIFF
--- a/src/sdks/tableau/methods/usersMethods.ts
+++ b/src/sdks/tableau/methods/usersMethods.ts
@@ -3,8 +3,14 @@ import { Zodios } from '@zodios/core';
 import { AxiosRequestConfig } from '../../../utils/axios.js';
 import { usersApis } from '../apis/usersApi.js';
 import { RestApiCredentials } from '../restApi.js';
+import { Pagination } from '../types/pagination.js';
 import { User } from '../types/user.js';
 import AuthenticatedMethods from './authenticatedMethods.js';
+
+export interface ListUsersResult {
+  users: User[];
+  pagination?: Pagination;
+}
 
 /**
  * Users and Groups methods of the Tableau Server REST API
@@ -19,9 +25,9 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
   }
 
   /**
-   * Returns a list of users on the site.
-   * Passes includeSSOInfo=false, includeUserCount=false, includeGroups=false
-   * by default to minimize DB and SAML DDB load on large sites.
+   * Returns a list of users on the site with pagination metadata.
+   * Passes includeUserCount=true for total count, but includeSSOInfo=false
+   * and includeGroups=false to minimize DB and SAML DDB load on large sites.
    *
    * Required scopes (Tableau Cloud): `tableau:users:read`
    *
@@ -38,19 +44,22 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
     siteId: string;
     pageSize?: number;
     pageNumber?: number;
-  }): Promise<User[]> => {
+  }): Promise<ListUsersResult> => {
     const response = await this._apiClient.listUsers({
       params: { siteId },
       queries: {
         pageSize,
         pageNumber,
+        includeUserCount: true,
         includeSSOInfo: false,
-        includeUserCount: false,
         includeGroups: false,
       },
       ...this.authHeader,
     });
-    return response.users.user;
+    return {
+      users: response.users.user,
+      pagination: response.pagination,
+    };
   };
 
   /**
--- a/src/sdks/tableau/methods/usersMethods.ts
+++ b/src/sdks/tableau/methods/usersMethods.ts
@@ -26,33 +26,40 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
 
   /**
    * Returns a list of users on the site with pagination metadata.
-   * Passes includeUserCount=true for total count, but includeSSOInfo=false
-   * and includeGroups=false to minimize DB and SAML DDB load on large sites.
    *
    * Required scopes (Tableau Cloud): `tableau:users:read`
    *
    * @param siteId - The Tableau site ID
    * @param pageSize - Number of users per page (default 100, max 1000)
    * @param pageNumber - Page offset (default 1)
+   * @param includeUserCount - Include total user count in pagination metadata
+   * @param includeSSOInfo - Include SSO/SAML info per user
+   * @param includeGroups - Include group memberships per user
    * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#get_users_on_site
    */
   listUsers = async ({
     siteId,
     pageSize,
     pageNumber,
+    includeUserCount,
+    includeSSOInfo,
+    includeGroups,
   }: {
     siteId: string;
     pageSize?: number;
     pageNumber?: number;
+    includeUserCount?: boolean;
+    includeSSOInfo?: boolean;
+    includeGroups?: boolean;
   }): Promise<ListUsersResult> => {
     const response = await this._apiClient.listUsers({
       params: { siteId },
       queries: {
         pageSize,
         pageNumber,
-        includeUserCount: true,
-        includeSSOInfo: false,
-        includeGroups: false,
+        includeUserCount,
+        includeSSOInfo,
+        includeGroups,
       },
       ...this.authHeader,
     });
--- a/src/sdks/tableau/methods/usersMethods.ts
+++ b/src/sdks/tableau/methods/usersMethods.ts
@@ -19,6 +19,41 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
   }
 
   /**
+   * Returns a list of users on the site.
+   * Passes includeSSOInfo=false, includeUserCount=false, includeGroups=false
+   * by default to minimize DB and SAML DDB load on large sites.
+   *
+   * Required scopes (Tableau Cloud): `tableau:users:read`
+   *
+   * @param siteId - The Tableau site ID
+   * @param pageSize - Number of users per page (default 100, max 1000)
+   * @param pageNumber - Page offset (default 1)
+   * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#get_users_on_site
+   */
+  listUsers = async ({
+    siteId,
+    pageSize,
+    pageNumber,
+  }: {
+    siteId: string;
+    pageSize?: number;
+    pageNumber?: number;
+  }): Promise<User[]> => {
+    const response = await this._apiClient.listUsers({
+      params: { siteId },
+      queries: {
+        pageSize,
+        pageNumber,
+        includeSSOInfo: false,
+        includeUserCount: false,
+        includeGroups: false,
+      },
+      ...this.authHeader,
+    });
+    return response.users.user;
+  };
+
+  /**
    * Returns information about the specified user.
    *
    * Required scopes (Tableau Cloud): `tableau:users:read`
--- a/src/sdks/tableau/types/user.ts
+++ b/src/sdks/tableau/types/user.ts
@@ -1,5 +1,10 @@
 import { z } from 'zod';
 
+/**
+ * User schema for Tableau REST API
+ * Extended for admin use cases to include full user profile information
+ * @see https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#get_users_on_site
+ */
 export const userSchema = z.object({
   id: z.string(),
   name: z.string(),
@@ -7,6 +12,10 @@ export const userSchema = z.object({
   email: z.string().optional(),
   fullName: z.string().optional(),
   lastLogin: z.string().optional(),
+  authSetting: z.string().optional(),
+  locale: z.string().optional(),
+  language: z.string().optional(),
+  externalAuthUserId: z.string().optional(),
 });
 
 export type User = z.infer<typeof userSchema>;
--- a/src/tools/web/users/listUsers.ts
+++ b/src/tools/web/users/listUsers.ts
@@ -1,0 +1,132 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+import { Ok } from 'ts-results-es';
+import { z } from 'zod';
+
+import { getConfig } from '../../../config.js';
+import { useRestApi } from '../../../restApiInstance.js';
+import { User } from '../../../sdks/tableau/types/user.js';
+import { WebMcpServer } from '../../../server.web.js';
+import { assertAdmin } from '../adminGate.js';
+import { ConstrainedResult, WebTool } from '../tool.js';
+import { applyUserFilters } from './usersFilterUtils.js';
+
+const paramsSchema = {
+  filter: z.string().optional(),
+  pageSize: z.number().int().positive().optional(),
+  limit: z.number().int().positive().optional(),
+};
+
+export const getListUsersTool = (server: WebMcpServer): WebTool<typeof paramsSchema> => {
+  const config = getConfig();
+
+  const listUsersTool = new WebTool({
+    server,
+    name: 'list-users',
+    disabled: !config.adminToolsEnabled,
+    description: `
+  Retrieves a list of users on the Tableau site. Each user includes profile information such as site role, email, last login time, and authentication settings.
+
+  Use this tool when you need to:
+  - Identify inactive users for license reclamation
+  - Audit user site roles and permissions
+  - Find users by email, name, or site role
+  - Review user authentication settings
+  - Analyze user activity based on last login times
+
+  **Parameters:**
+  - \`filter\` (optional) – Filter string with format \`field:operator:value\`. Multiple filters are comma-separated (AND logic). Same field can appear multiple times for range queries (e.g. \`lastLogin:gt:X,lastLogin:lt:Y\`).
+  - \`pageSize\` (optional) – Number of users to fetch from the API per page (default 100, max 1000). Controls server-side pagination.
+  - \`limit\` (optional) – Maximum total results to return after filtering.
+
+  **Filterable Fields:**
+
+  | Field | Type | Operators | Example |
+  |-------|------|-----------|---------|
+  | \`id\` | string | \`eq\`, \`in\` | \`id:eq:abc123\` |
+  | \`name\` | string | \`eq\`, \`in\` | \`name:eq:jsmith\` |
+  | \`siteRole\` | string | \`eq\`, \`in\` | \`siteRole:eq:Creator\` |
+  | \`email\` | string | \`eq\`, \`in\` | \`email:eq:user@example.com\` |
+  | \`fullName\` | string | \`eq\`, \`in\` | \`fullName:eq:John Smith\` |
+  | \`lastLogin\` | string (ISO 8601) | \`eq\`, \`gt\`, \`gte\`, \`lt\`, \`lte\` | \`lastLogin:lt:2025-01-01T00:00:00Z\` |
+  | \`authSetting\` | string | \`eq\`, \`in\` | \`authSetting:eq:SAML\` |
+  | \`locale\` | string | \`eq\`, \`in\` | \`locale:eq:en_US\` |
+  | \`language\` | string | \`eq\`, \`in\` | \`language:eq:en\` |
+  | \`externalAuthUserId\` | string | \`eq\`, \`in\` | \`externalAuthUserId:eq:ext-123\` |
+
+  **Filter Examples:**
+  - Single filter: \`siteRole:eq:Creator\`
+  - Date range: \`lastLogin:gt:2025-01-01T00:00:00Z,lastLogin:lt:2025-06-01T00:00:00Z\`
+  - IN operator: \`siteRole:in:Creator|Explorer\`
+  - Inactive users: \`lastLogin:lt:2024-12-01T00:00:00Z\`
+
+  **Response:** Each user includes:
+  - \`id\` – user ID
+  - \`name\` – username
+  - \`siteRole\` – ServerAdministrator, SiteAdministratorCreator, Creator, Explorer, Viewer, Unlicensed, etc.
+  - \`email\` – user email address
+  - \`fullName\` – user's full display name
+  - \`lastLogin\` – timestamp of last login (ISO 8601)
+  - \`authSetting\` – ServerDefault, SAML, or OpenID
+  - \`locale\` – user's locale setting
+  - \`language\` – user's language preference
+  - \`externalAuthUserId\` – ID from external authentication provider
+  `,
+    paramsSchema,
+    annotations: {
+      title: 'List Users',
+      readOnlyHint: true,
+      openWorldHint: false,
+    },
+    callback: async (args, extra): Promise<CallToolResult> => {
+      return await listUsersTool.logAndExecute({
+        extra,
+        args,
+        callback: async () => {
+          const users = await useRestApi({
+            ...extra,
+            jwtScopes: listUsersTool.requiredApiScopes,
+            callback: async (restApi) => {
+              // Verify user has admin privileges
+              const adminResult = await assertAdmin(restApi, extra);
+              if (adminResult.isErr()) {
+                throw new Error(adminResult.error);
+              }
+
+              return restApi.usersMethods.listUsers({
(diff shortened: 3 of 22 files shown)


src/tools/web/users/usersFilterUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+  'fullName',
+  'lastLogin',
+  'authSetting',
+  'locale',
+  'language',
+]);
+
+type FilterField = z.infer<typeof FilterFieldSchema>;
+
+const allowedOperatorsByField: Record<FilterField, FilterOperator[]> = {
```
Alon-ST-DATA, line 28: This filtering support seem to be [compatible with one we support  ](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_filtering_and_sorting.htm#users) however I have a concern regarding possible impact on db and site saml ddb when querying all users.
If you look at [users controller ](https://github.com/sf-analyticscloud/monolith/blob/main/workgroup/src/silos/tableau-server/libraries/tab-api-restapi-all/src/com/tableausoftware/api/rest/users/RestApiUserController.java#L262-L264) you can see flags which we added over time for that purpose and I wondering if for our goal / feature we need to reset thos flags: includeSSOInfo, includeUserCount, includeGroups.
Granted that those still can be explicitly triggered but for that we can look into the caching etc.

tests/e2e/server.test.ts
```diff
@@ -36,6 +36,7 @@ describe('server', () => {
       const oauthOnlyTools: ReadonlyArray<WebToolName> = ['revoke-access-token', 'reset-consent'];
       const adminOnlyTools: ReadonlyArray<WebToolName> = [
         'list-extract-refresh-tasks',
+        'list-users',
```
Alon-ST-DATA, line 39: If you have not and can please run some manual tests on larger sites like our test site with ~27K users:
https://10ax.online.tableau.com/#/site/admin-insights-tmcp-site/users

src/tools/web/users/usersFilterUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+  // Parse and validate the filter string
+  const validatedFilter = parseAndValidateUsersFilterString(filterString);
+  const filters = validatedFilter.split(',').map((f) => {
+    const [field, operator, ...valueParts] = f.split(':');
+    return {
+      field: field as FilterField,
+      operator: operator as FilterOperator,
+      value: valueParts.join(':'),
+    };
+  });
```
Alon-ST-DATA, line 91: Same-field filters silently dedupe — parseAndValidateFilterString.ts:42-73 keys by field name. lastLogin:gt:X,lastLogin:lt:Y (a date range — natural license-reclamation query!) silently drops the first filter.
Possible fix: store as array, or detect+throw on duplicate field.

src/tools/web/users/usersFilterUtils.ts
```diff
@@ -0,0 +1,150 @@
+import { z } from 'zod';
+
+import { User } from '../../../sdks/tableau/types/user.js';
+import {
+  FilterOperator,
+  FilterOperatorSchema,
+  parseAndValidateFilterString,
+} from '../../../utils/parseAndValidateFilterString.js';
```
Alon-ST-DATA, line 8: lastLogin / schedule.* fields skip ISO 8601 validation — validator only matches exact field names createdAt/updatedAt. Dotted paths bypass. matchesFilter then does lexicographic string compare, so 2026-01-01T00:00:00Z vs
  2026-01-01T00:00:00.000Z (Tableau emits fractional seconds in some configs) silently mis-compares. Regex itself rejects .000Z and ±HH:MM offsets. Fix: drive validation off allowed-operators list, broaden regex, parse-to-Date before

src/tools/web/users/listUsers.ts
```diff
@@ -0,0 +1,142 @@
...
+      readOnlyHint: true,
+      openWorldHint: false,
+    },
+    callback: async (args, extra): Promise<CallToolResult> => {
+      // Validate filter string early if provided
+      if (args.filter) {
+        parseAndValidateUsersFilterString(args.filter);
+      }
+
+      return await listUsersTool.logAndExecute({
```
Alon-ST-DATA, line 84: Malformed filter throws transport error, not structured tool error — in listUsers.ts we validate outside logAndExecute. Throw bubbles past MCP SDK's isError layer. Inconsistent with listDatasources pattern. Existing rejects.toThrow()
  test cements bad behavior. 

Possible fix: move validation inside the lambda, return Err().

src/tools/web/users/listUsers.ts
```diff
@@ -0,0 +1,142 @@
...
+  - \`name\` – username
+  - \`siteRole\` – ServerAdministrator, SiteAdministratorCreator, Creator, Explorer, Viewer, Unlicensed, etc.
+  - \`email\` – user email address
+  - \`fullName\` – user's full display name
+  - \`lastLogin\` – timestamp of last login (ISO 8601)
+  - \`authSetting\` – ServerDefault, SAML, or OpenID
+  - \`locale\` – user's locale setting
+  - \`language\` – user's language preference
+
+  **Note:** Tableau Cloud uses \`tableau:users:read\` scope. The Tableau REST API does not support server-side filtering or pagination - all users are retrieved and filtering is performed client-side by this tool.
```
Alon-ST-DATA: This is actually wrong for users. 
Tableau REST Get Users on Site supports filter= and pageSize/pageNumber. LLMs will propagate this. For 50k-user tenants this fetches the
  full list every call. Same line in docs page. (Statement IS correct for extract-refresh-tasks — leave that one.)
  - pageSize and limit are functionally identical — both just slice [0..min(pageSize,limit)]. No pageNumber, no cursor. 
Drop one or add real pagination.
  - Filter parsed twice on every call — early validate + applyUserFilters re-parses internally.
  - externalAuthUserId added to schema but not in filter table or response docs.
  - No negative-path test for getEnabledToolNames() with ADMIN_TOOLS_ENABLED=false. 
Embedded OAuth metadata test passes only because env force-sets it true.

src/sdks/tableau/methods/usersMethods.ts
```diff
@@ -18,6 +18,23 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
...
+   *
+   * @param siteId - The Tableau site ID
+   * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#query_users_on_site
+   */
+  listUsers = async ({ siteId }: { siteId: string }): Promise<User[]> => {
+    const raw = await this._apiClient.listUsers({
+      params: { siteId },
+      ...this.authHeader,
+    });
+    const response = parseListUsersResponse(raw);
```
anyoung-tableau: This is unnecessary. Zodios already does this internally. `raw` is already of type `User[]`

src/sdks/tableau/types/user.ts
```diff
@@ -1,12 +1,21 @@
 import { z } from 'zod';
 
+/**
+ * User schema for Tableau REST API
+ * Extended for admin use cases to include full user profile information
+ * @see https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#query_users_on_site
```
anyoung-tableau: This URL's fragment should be `#get_users_on_site` yah?

src/tools/web/users/listUsers.ts
```diff
@@ -0,0 +1,142 @@
...
+  const config = getConfig();
+
+  const listUsersTool = new WebTool({
+    server,
+    name: 'list-users',
+    disabled: !config.adminToolsEnabled,
+    description: `
+  Retrieves a list of users on the Tableau site. Each user includes profile information such as site role, email, last login time, and authentication settings.
+
+  This tool is restricted to Tableau site administrators and requires the \`ADMIN_TOOLS_ENABLED\` feature flag to be enabled.
```
anyoung-tableau: This sentence should probably not be included in the tool description. Consider when the MCP server is deployed remotely. This tool wouldn't even exist if the admin tools were disabled so if they see this tool it means they're enabled and this sentence is meaningless.

src/tools/web/users/listUsers.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  Use this tool when you need to:
+  - Identify inactive users for license reclamation
+  - Audit user site roles and permissions
+  - Find users by email, name, or site role
+  - Review user authentication settings
+  - Analyze user activity based on last login times
+
+  **Parameters:**
+  - \`filter\` (optional) – Client-side filter string with format \`field:operator:value\`. Multiple filters are comma-separated (AND logic). The Tableau REST API does not support server-side filtering, so all users are fetched and filtered client-side.
```
anyoung-tableau: I don't think tool descriptions should include implementation details about _where_ the filtering is happening. It's meaningless to the agent and just wastes tokens.

src/tools/web/users/listUsers.ts
anyoung-tableau: Same comment about this note. I suggest just removing it.

src/tools/web/users/usersFilterUtils.ts
Alon-ST-DATA: @Akash-Rastogi  For backing up my concern see this incident:
[icc-86718804 ](https://salesforce.enterprise.slack.com/archives/C0A6HFKFF50)(Jan 2026) — list-users abuse

cc: @anyoung-tableau 

Thanks.

tests/e2e/server.test.ts
Akash-Rastogi: On MCP Inspector it shows 100 users and via claude prompt it provides this summary:
Screenshot 2026-06-01 at 2 26 55 PM

src/tools/web/users/usersFilterUtils.ts
Akash-Rastogi: Fixed. Added `includeSSOInfo=false`, `includeUserCount=false`, `includeGroups=false` as query params defaulting to false in `usersMethods.listUsers()` to minimize DB/SAML DDB load. Also added `pageSize`/`pageNumber` server-side pagination params to the endpoint.

src/tools/web/users/usersFilterUtils.ts
Akash-Rastogi: Fixed. `applyUserFilters` now parses filter expressions directly into an array (bypassing the shared `parseAndValidateFilterString` which keys by field name). Date range queries like `lastLogin:gt:X,lastLogin:lt:Y` now work correctly. Added a test to verify.

src/tools/web/users/usersFilterUtils.ts
Akash-Rastogi: Fixed. Date fields (`lastLogin`) now use `new Date().getTime()` for comparisons instead of lexicographic string compare. This correctly handles fractional seconds (`.000Z`) and timezone offsets (`+05:00`). Added tests for both edge cases.

src/tools/web/users/listUsers.ts
Akash-Rastogi: Fixed. Moved filter validation inside `logAndExecute` callback — malformed filters now return a structured tool error (`isError: true`) with the ZodError details, rather than throwing a transport error. Updated test to verify.

src/tools/web/users/listUsers.ts
Akash-Rastogi: Addressed all sub-issues:

1. **Server-side pagination**: `pageSize` is now passed to the REST API as a query param (server-side). No longer just client-side slicing.
2. **pageSize vs limit clarified**: `pageSize` = server-side page size (API param), `limit` = client-side result cap after filtering. Distinct roles.
3. **Filter parsed twice**: Fixed — removed early validation; `applyUserFilters` now does its own parsing/validation in one pass.
4. **externalAuthUserId**: Added to filterable fields, `getFieldValue`, and tool description response docs.
5. **Negative-path test**: `scopes.test.ts` already covers this (lines 72-88 test `tableau:users:read` exclusion when `adminToolsEnabled=false`).

src/sdks/tableau/methods/usersMethods.ts
Akash-Rastogi: Fixed. Removed `parseListUsersResponse` — Zodios handles response validation/normalization internally. The method now accesses `response.users.user` directly from the Zodios-parsed result.

src/sdks/tableau/types/user.ts
Akash-Rastogi: Fixed. Updated to `#get_users_on_site`.

src/tools/web/users/listUsers.ts
Akash-Rastogi: Removed.

src/tools/web/users/listUsers.ts
Akash-Rastogi: Removed. Description no longer mentions where filtering happens.

src/tools/web/users/listUsers.ts
Akash-Rastogi: Removed.

tests/e2e/server.test.ts
Alon-ST-DATA: @Akash-Rastogi the response is wrong (due to count flag) as we should get 27K+ users as the total.. it is ok (and ideal) to not list all of them and provide summary  and maybe dump into csv (like in case of other  list resources..)  

The count of the users needs to be corrected which means the flag **includeUserCount** needs to be true for that purpose -- like we can set the flag to true, then  retrieve the first page (100 users) and get the count of users from the response..
We should optimize the logic such as the LLM can set flags before making the calls, per use case/purpose.

Akash-Rastogi: Fixed. Changed `includeUserCount` to `true` and the tool response now includes `totalAvailable` from pagination metadata. Response shape is:

```json
{
  "users": [ ... first page ... ],
  "totalAvailable": 27000
}
```

This way the LLM knows the full user count even though only one page (default 100) is returned. `includeSSOInfo` and `includeGroups` remain `false` to keep the query lightweight.

Akash-Rastogi: [
Screenshot 2026-06-02 at 1 12 51 PM
Screenshot 2026-06-02 at 1 30 31 PM
](url)

Alon-ST-DATA: I got this, so unsure if we can optimize it some how
@Akash-Rastogi we can sync offline on that and other findings I might have.

`
Bug — capped at 1000 
  Findings
  1. list-users pagination bug (main finding)
  - Returned exactly 1000 users; limit:100000 + pageSize:1000 ignored.
  - 1000 unique (not dup'd) = single page, no pagination loop.
  - REST totalAvailable = 27,034. Tool silently dropped 26,034 users.
  - Role breakdown we computed (971 Creator…) invalid — only first 1000.
  2. Environment hygiene
  - 6 stale node build/index.js processes running.
  - 2 env files → site confusion (mcp-test=15 users vs admin-insights-tmcp-site=27034).
  Answer to your question
  Site has 27,034 users (via REST). Tool said 1000 — wrong.`

Akash-Rastogi: Fixed. The tool now paginates through all pages automatically:

- Default `pageSize` is 1000 (API max)
- Loops fetching pages until `allUsers.length >= totalAvailable` or fewer results returned
- For the 27K site: ~27 API calls at 1000 users each to fetch all users
- Respects `args.limit` if the caller wants fewer results (avoids fetching unnecessary pages)

The previous implementation only fetched a single page, which capped at 1000 users silently. Now the full user list is retrieved before client-side filtering is applied.

src/tools/web/users/listUsers.ts
```diff
@@ -92,15 +92,40 @@ export const getListUsersTool = (server: WebMcpServer): WebTool<typeof paramsSch
                 throw new Error(adminResult.error);
               }
 
-              return restApi.usersMethods.listUsers({
-                siteId: restApi.siteId,
-                pageSize: args.pageSize,
-              });
+              const pageSize = args.pageSize ?? 1000;
```
Alon-ST-DATA: Thanks Akash , this fix seems to work but small cleaner suggestion reuse the existing paginate() helper instead of an inline loop.
[src/utils/paginate.ts ](https://github.com/tableau/tableau-mcp/blob/main/src/utils/paginate.ts )already implements this exact loop, and both list-projects and[ list-datasources ](https://github.com/tableau/tableau-mcp/blob/main/src/tools/web/listDatasources/listDatasources.ts#L9 )use it. The inline while here becomes a third
 copy of the same pagination logic, which we'll have to keep in sync. Matching the sibling tools would look like:
 _**const users = await paginate({
     pageConfig: {
       pageSize: args.pageSize,
       limit: maxResultLimit
         ? Math.min(maxResultLimit, args.limit ?? Number.MAX_SAFE_INTEGER)
         : args.limit,
     },
     getDataFn: async (pageConfig) => {
       const { users, pagination } = await restApi.usersMethods.listUsers({
         siteId: restApi.siteId,
         pageSize: pageConfig.pageSize,
         pageNumber: pageConfig.pageNumber,
       });
       return { pagination, data: users };
     },
   });**_


This gives us two things for free:
  
 1. MAX_RESULT_LIMIT enforcement. list-projects/list-datasources clamp with configWithOverrides.getMaxResultLimit(toolName); this PR's list-users only
   honors args.limit, so an operator's server-side cap doesn't apply. On a 27K-user site list-users is the highest-volume tool, so it's the one we'd most want bounded by the env cap.
 2. pageSize clamping. See edge case below.
   Edge case in the current loop: the stop condition compares result.users.length < pageSize, but the request sends Math.min(pageSize, 1000). If a
  caller passes pageSize: 5000, each page returns ≤1000, which is < 5000, so the loop breaks after page 1 — re-introducing the exact truncation this PR
  fixes. paginate() validates/normalizes pageSize up front and avoids this. If you'd rather keep the inline loop, clamping pageSize to 1000 once before the loop fixes it.

Akash-Rastogi: Done. Refactored to use `paginate()` from `src/utils/paginate.ts` — matching the pattern in list-projects and list-datasources.

This gives us:
1. **MAX_RESULT_LIMIT enforcement** via `configWithOverrides.getMaxResultLimit()`
2. **pageSize normalization** — fixes the edge case where `pageSize > 1000` caused premature loop exit
3. **DRY** — no third copy of pagination logic to maintain

Thanks for the suggestion!

Alon-ST-DATA: Overall works well. From my last phase of manual testing , exposed new bugs and finding: 

 🔴 BUG 1 — pageSize > 1000 → 403 (no clamp). Threshold: 1000 ok, ≥1001 fails. Seems that there was Math.min(pageSize,1000), but the refactor to
  paginate() dropped it (listUsers.ts:102 passes args.pageSize raw). paginate() doesn't clamp either. Tableau API rejects pageSize>1000 → 403. Fix:
  clamp at line 102 → pageSize: args.pageSize ? Math.min(args.pageSize, 1000) : undefined.
  
  🟡 BUG 2 — totalAvailable wrong. listUsers.ts:132 sets totalAvailable: users.length (post-pagination array length), NOT site total. With limit:5 →
  reported totalAvailable:5. Should be REST pagination.totalAvailable (27034). Misleads license-reclamation ("5 total users" when 27K exist). Other tools
  propagate the real pagination total.

** That seems a known issue with PAT, so perhaps we can test it with direct trust to make sure we are safe..   
  🟡 FINDING 3 — parallel calls → 401. 4 concurrent tool calls all 401'd; same calls sequential → clean. PAT session not concurrency-safe in this build.
  Test/usage note: run list-users serially.
[  ref from testing doc](https://docs.google.com/document/d/1M_B13bt6PTcYGDjYZDFEA_ejKELdU1E0wRaO5ntf7fs/edit?disco=AAAB8z4Llr0)

Akash-Rastogi: Fixed both bugs in commit b047f296:

**🔴 BUG 1 — pageSize > 1000 → 403**: Fixed in the shared `paginate()` helper (`src/utils/paginate.ts`) by clamping `pageSize` to `Math.min(pageSize, 1000)` before passing to `getDataFn`. This covers all 8 paginated tools centrally — no per-tool clamps needed. The previous per-tool clamp in `listUsers.ts` has been removed in favor of this.

**🟡 BUG 2 — totalAvailable wrong**: `listUsers.ts` now captures `pagination.totalAvailable` from the first REST API response and reports that as `totalAvailable` in the tool result (the site-wide total, e.g., 27034), instead of `users.length` (the post-pagination array size).

**🟡 FINDING 3 — parallel calls → 401**: Noted — this appears to be a PAT session concurrency limitation, not something we can fix at the tool level. Will add a usage note if needed.

Tests added for both fixes (paginate clamp + totalAvailable). Full suite passes (1442 tests).

Alon-ST-DATA (approved):

src/sdks/tableau/methods/usersMethods.ts
```diff
@@ -18,6 +24,44 @@ export default class UsersMethods extends AuthenticatedMethods<typeof usersApis>
...
+    pageNumber?: number;
+  }): Promise<ListUsersResult> => {
+    const response = await this._apiClient.listUsers({
+      params: { siteId },
+      queries: {
+        pageSize,
+        pageNumber,
+        includeUserCount: true,
+        includeSSOInfo: false,
+        includeGroups: false,
```
anyoung-tableau: Consider parameterizing these in the `listUsers` method. We've been trying to keep business logic out of these Methods classes as much as possible, as if they were completely generic. Leave it to the calling site to provide their values.

anyoung-tableau (approved):

src/sdks/tableau/methods/usersMethods.ts
Akash-Rastogi: Done in c2e042e0. The three flags (`includeUserCount`, `includeSSOInfo`, `includeGroups`) are now optional params on `usersMethods.listUsers()` with no defaults — the method just passes through whatever the caller provides. Business logic (`true`/`false`/`false`) moved to the tool in `listUsers.ts`.

PR #341, opened 2026-05-07
@W-22390723: updated resource url to include tableau-mcp

mattcfilbert:
## Description

Updated resource url to include /tableau-mcp 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe):

## Checklist

- [x] I have updated the version in the package.json file by using `npm run version`. For example,
      use `npm run version:patch` for a patch version bump.
- [ ] I have made any necessary changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] I have documented any breaking changes in the PR description. For example, renaming a config
      environment variable or changing its default value.

## Contributor Agreement

By submitting this pull request, I confirm that:

- [x] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) for this project and followed
      its Contribution Checklist.

PR #369, opened 2026-06-01
@W-22782879: Validate token audience against MCP resource URL

mattcfilbert:
## What problem this solves

`TableauAccessTokenValidator.validate` only checked `iss` and `exp`. Because `iss` is the shared SSO issuer across all pods, a Tableau-AS bearer token minted for one deployment passed validation against a different deployment. The mismatch surfaced only later inside `transport.handleRequest` and was caught as a generic `500` (the cross-pod token-reuse 5xx seen on 10ax/uw2a), instead of a clean `401` at the auth layer.

This adds RFC 9068 audience validation: tokens whose `aud` claim does not match this server's canonical resource identifier (`${resourceUri}/tableau-mcp`, the same value advertised as `resource` in the protected-resource metadata) are rejected with a `401`.

## How it works

- New `OAUTH_VALIDATE_AUDIENCE` flag (default **off**). Ships dark; enforced only after the Tableau authz server stamps `aud=resource` (per RFC 8707) in every environment. Reversible via the flag if a legacy token surfaces.
- Single shared `buildResourceIdentifier(resourceUri)` helper computes both the advertised `resource` and the validated `aud`, so the two cannot drift (trailing-slash safe, byte-equal).
- When the flag is enabled, the `client_id` claim is also required. This keeps the legacy `client_id ?? aud` fallback provably safe (it can never resolve the revocation client ID to the resource URL).
- Applies only to the external Tableau authz-server path; under the embedded authz server, `EmbeddedAccessTokenValidator` already validates the audience of the tokens it issues, so the flag is a no-op there.

## Alternatives considered

- **Warn-only soft launch** — log mismatches without rejecting. Rejected: leaves the cross-pod 5xx unresolved and adds a second flip later.
- **Tolerant dual-accept (`aud===resource` OR `aud===client_id`)** — mirrors the existing compatibility bridge. Rejected: weakens the audience guarantee and is harder to reason about than a single default-off flag.
- **Unconditional enforcement** — rejected: would 401 every legacy token before George's AS rollout completes.

## Auth impact

Touches the Tableau Cloud OAuth validation path (`TableauAccessTokenValidator`). No change to the embedded OAuth (PKCE/JWE) or passthrough paths. No scope/`toolScopeMap` changes. Default-off means zero behavior change until explicitly enabled.

## Test coverage

- `accessTokenValidator.test.ts`: flag-on + `aud` match accepts; flag-on + cross-pod mismatch rejects (`401`); flag-off + mismatch is a no-op (back-compat); flag-on + matching `aud` but missing `client_id` rejects (fail-closed invariant).
- `config.test.ts`: `validateAudience` defaults `false`; `OAUTH_VALIDATE_AUDIENCE='true'` parses to `true`.
- Verified: `npx tsc --noEmit`, `npx eslint`, `npm test` (1391 passing).

## Out of scope (follow-ups)

- Migrating `EmbeddedAccessTokenValidator`'s hardcoded `aud === 'tableau-mcp-server'` onto the resource identifier.
- Removing the `client_id ?? aud` fallback (existing `TODO(cleanup)`, gated on the same AS rollout).

## Rollout

Land with the flag off. Flip `OAUTH_VALIDATE_AUDIENCE=true` only after confirming with George that the AS emits `aud=${resourceUri}/tableau-mcp` and always includes `client_id` in every environment.

## Note on aud being a single string

I kept `aud` validation to a single string match. The Tableau AS issues one resource string, so array matching wasn't needed here. RFC 9068 does allow an array, but a non-string `aud` (for example a JSON array) is rejected by schema validation, so it fails closed rather than slipping through. If the token format ever starts emitting an array, the fix is small: widen the schema and switch the check to a membership test.


src/server/oauth/accessTokenValidator.ts
```diff
@@ -145,8 +146,39 @@ export class TableauAccessTokenValidator extends AccessTokenValidator {
...
+      // RFC 9068 audience validation: reject tokens not minted for this MCP server's
+      // resource identifier. Without this, a token issued for one deployment (same shared
+      // SSO issuer) passes validation against another, surfacing later as an opaque 500.
+      // Gated behind oauth.validateAudience so it can ship dark and only enforce once the
+      // Tableau AS stamps aud=resource (not client_id) in every environment.
+      if (this.config.oauth.validateAudience) {
+        const expectedAudience = buildResourceIdentifier(this.config.oauth.resourceUri);
+        if (aud !== expectedAudience) {
+          log({
+            message: `Access token audience mismatch: expected '${expectedAudience}', got '${aud}'`,
```
anyoung-tableau: Consider limiting the `aud` string to some reasonable length in the log entry, otherwise someone could write the claim with some huge string which we'd write to Splunk.

src/server/oauth/accessTokenValidator.ts
```diff
@@ -145,8 +146,39 @@ export class TableauAccessTokenValidator extends AccessTokenValidator {
...
       }
 
+      // RFC 9068 audience validation: reject tokens not minted for this MCP server's
+      // resource identifier. Without this, a token issued for one deployment (same shared
+      // SSO issuer) passes validation against another, surfacing later as an opaque 500.
+      // Gated behind oauth.validateAudience so it can ship dark and only enforce once the
+      // Tableau AS stamps aud=resource (not client_id) in every environment.
+      if (this.config.oauth.validateAudience) {
+        const expectedAudience = buildResourceIdentifier(this.config.oauth.resourceUri);
+        if (aud !== expectedAudience) {
```
anyoung-tableau: This will work for the pod-specific URLs, but will break for the global URL, yeah? We need to also accept each environment-specific global URL. I would suggest adding a new env var that lets us specify that URL, and allow it too.

src/server/oauth/resourceIdentifier.ts
```diff
@@ -0,0 +1,17 @@
...
+ * document (the `resource` field clients read) and the access token audience
+ * check. Computing both from this helper guarantees the advertised resource and
+ * the validated `aud` claim match byte-for-byte regardless of how `resourceUri`
+ * is configured.
+ *
+ * @param resourceUri - The configured OAuth resource URI (`config.oauth.resourceUri`).
+ * @returns The canonical resource identifier, e.g. `https://host/tableau-mcp`.
+ */
+export function buildResourceIdentifier(resourceUri: string): string {
+  return `${resourceUri}/${serverName}`;
```
anyoung-tableau: Consider eliminating the possibility of accidentally doubling up a `/`. It's difficult at a glance knowing we haven't messed it up somewhere in the chain. Best to eliminate the future bug entirely.

Something like:

```ts
return [resourceUri, serverName].map((s) => s.replace(/^\/+|\/+$/g, '')).join('/');
```

docs/docs/configuration/mcp-config/oauth.md
```diff
@@ -141,6 +141,27 @@ server. This should be the base URL of your MCP server deployment.
 
 <hr />
 
+### `OAUTH_VALIDATE_AUDIENCE`
```
anyoung-tableau: I don't think we need this. We won't ship before the AS stamps the `aud` claim correctly.

mattcfilbert: removed. thanks, Andy

src/server/oauth/resourceIdentifier.ts
mattcfilbert: updated to make your suggestion

src/server/oauth/accessTokenValidator.ts
mattcfilbert: Good call. Added a `truncateForLog` helper that caps the `aud` value at 256 chars before it's logged so an oversized claim can't bloat Splunk

src/server/oauth/accessTokenValidator.ts
mattcfilbert: Agreed. Added an optional `OAUTH_GLOBAL_RESOURCE_URI` env var. The validator now accepts the aud claim if it matches either the pod-specific resource identifier or the configured global one, so tokens minted against the global URL pass too. When the var is unset, only the pod identifier is accepted, so existing deployments don't change

src/server/oauth/accessTokenValidator.ts
```diff
@@ -145,11 +154,28 @@ export class TableauAccessTokenValidator extends AccessTokenValidator {
...
-      // environments, remove the aud fallback and update the schema to require client_id.
-      const oauthClientId = client_id ?? aud;
+      // RFC 9068 audience validation: reject tokens not minted for this MCP server's resource
+      // identifier. Without this, a token issued for one deployment (same shared SSO issuer)
+      // passes validation against another, surfacing later as an opaque 500. Both the pod-specific
+      // resource URI and (when configured) the environment's global resource URI are accepted.
+      const allowedAudiences = [
+        buildResourceIdentifier(this.config.oauth.resourceUri),
+        ...(this.config.oauth.globalResourceUri
+          ? [buildResourceIdentifier(this.config.oauth.globalResourceUri)]
```
anyoung-tableau: For the global URL, the resource will be `mcp.tableau.com` not `mcp.tableau.com/tableau-mcp`. Please work with Yogi to work out these details.

yyk16293: Let's go with **_mcp.tableau.com_**

mattcfilbert: Done. As Yogi requested, I reverted the /tableau-mcp resource changes from W-22390723: the protected-resource document advertises the bare resource URL again (resource = `OAUTH_RESOURCE_URI`), and the metadata route is back at `/.well-known/oauth-protected-resource`

For the audience check, aud is now matched directly against the resource URL with no `/tableau-mcp` suffix. Added `OAUTH_GLOBAL_RESOURCE_URI` so the global resource (e.g. mcp.tableau.com) is accepted alongside the pod-specific one, matched exactly as configured

Verified tsc, unit tests, and the embedded OAuth e2e suite all pass.

src/server/oauth/.well-known/oauth-protected-resource.ts
```diff
@@ -14,10 +11,10 @@ const protectedResourcePath = `/${serverName}`;
  * WWW-Authenticate header in 401 response.
  */
 export function oauthProtectedResource(app: express.Application): void {
-  app.get(`${protectedResourcePath}/.well-known/oauth-protected-resource`, (_req, res) => {
+  app.get('/.well-known/oauth-protected-resource', (_req, res) => {
```
anyoung-tableau, line 15: I'm not really following what's happening here. Why are we reverting #341 ?

yyk16293 (approved):

src/server/oauth/.well-known/oauth-protected-resource.ts
mattcfilbert: @yyk16293 - can you share more context

anyoung-tableau: Spoke with Yogi on Slack. https://salesforce-internal.slack.com/archives/C09S2ML7FH9/p1780524411555899?thread_ts=1777410329.337759&cid=C09S2ML7FH9

anyoung-tableau (approved):

src/server/oauth/.well-known/oauth-protected-resource.ts
```diff
@@ -14,10 +11,10 @@ const protectedResourcePath = `/${serverName}`;
...
     const { issuer, advertiseApiScopes, resourceUri, enforceScopes } = getConfig().oauth;
     res.json({
-      resource: `${resourceUri}${protectedResourcePath}`,
+      resource: resourceUri,
```
anyoung-tableau: Wait, this doesn't seem right. The resource should still have the `/tableau-mcp` path, yeah? Self-hosters will still access TMCP through example.com/tableau-mcp and [OAUTH_RESOURCE_URI](https://tableau.github.io/tableau-mcp/docs/configuration/mcp-config/oauth#oauth_resource_uri) is documented to only be the domain, not including the path.

anyoung-tableau (changes requested):

src/server/oauth/.well-known/oauth-protected-resource.ts
mattcfilbert: Good catch. You're right that the resource should keep the `/tableau-mcp` path. The `5bdf8413` revert only needed to move the well-known route back to the root; the resource value itself should stay `<OAUTH_RESOURCE_URI>/tableau-mcp` since `OAUTH_RESOURCE_URI` is domain-only and self-hosters reach TMCP at example.com/tableau-mcp. Restored that on both the metadata resource field and the pod audience check

The global URL (`OAUTH_GLOBAL_RESOURCE_URI`) stays a bare exact match, so mcp.tableau.com is accepted without the path

anyoung-tableau (approved):
💪

PR #370, opened 2026-06-01
@W-22575042: Migrate workbook tools from agent-to-tableau-desktop 

anyoung-tableau:
These changes migrate most, but not all of the workbook tools from agent-to-tableau-desktop.

Source: https://github.com/tableau/agent-to-tableau-desktop/blob/main/src/server/tools/workbook.ts

Migrated:
* tableau-list-worksheets -> list-worksheets
* tableau-list-dashboards -> list-dashboards
* tableau-get-workbook -> get-workbook-xml
* tableau-apply-workbook -> apply-workbook

Not migrated:
* tableau-save-metadata-xml -- not migrated since it does the same thing as inline mode of get-workbook-xml
* tableau-load-metadata-xml -- not migrated since it does the same thing as inline mode of apply-workbook-xml

Other pieces migrated with little-to-no refactoring/modifications:
* https://github.com/tableau/agent-to-tableau-desktop/tree/main/src/lib/workbook-serialization-converter
* https://github.com/tableau/agent-to-tableau-desktop/tree/main/src/validation


src/desktop/libraries/workbook-serialization-converter/xmlFormatter.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // Handle content and children
+    if (element.children.length === 0 && !element.text) {
+      // Empty element
+      line += ' />';
+      result.push(line);
+    } else if (element.text && element.children.length === 0) {
+      // Text content only — use CDATA if flagged (preserves SQL, formulas etc.)
+      if (element.encoded) {
+        line += `><![CDATA[${element.text}]]></${element.name}>`;
```
mattcfilbert: CDATA text isn't guarded against an embedded `]]>` terminator, and `xmlParser` reassembles split CDATA back into `element.text`. A calc/custom-SQL value containing the literal `]]>` reserializes to `<calc><![CDATA[a ]]> b]]></calc>` which is not well-formed XML, so trailing bytes become live markup on the destructive apply path. 

Can we split before writing, e.g. `element.text.replaceAll(']]>', ']]]]><![CDATA[>')`, and cover the split-CDATA round trip in a test?

src/desktop/commands/workbook/loadWorkbookXml.ts
```diff
@@ -0,0 +1,216 @@
...
+    },
+  });
+
+  if (result.isErr()) {
+    return Err({ type: 'execute-command-error', error: result.error });
+  }
+
+  const { error, status } = result.value;
+
+  if (status === 'completed') {
```
mattcfilbert: This text fallback can't run with the real `LocalExecutor`: `executeCommand()` converts a failed command status into `Err({ type: 'command-failed' })` (_localToolExecutor.ts:105-113_) and only returns `Ok` for `completed`, so status is always 'completed' here. The test only reaches this via `Ok({ status: 'failed' })`, which production can't produce and therefore provides false confidence

If text mode is a valid retry for a failed filepath load, retry on the `command-failed` Err; otherwise delete the dead branch and its test

src/desktop/libraries/workbook-serialization-converter/xmlParser.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // Create DOM element (text will be set from child nodes)
+    const domElement = new DOMElement(element.tagName, attributes);
+
+    // Process children and collect text
+    const childNodes = element.childNodes;
+    const textParts: string[] = [];
+
+    for (let i = 0; i < childNodes.length; i++) {
```
mattcfilbert, line 75: This converter rewrites arbitrary workbook XML before `apply-workbook` writes it back, but has no direct tests. Before it ships, can we add XML↔JSON round-trip tests for the semantic-loss cases: CDATA containing `]]>`, mixed-content ordering (`<p>Hello <b>x</b> world</p>`), attribute newline/entity preservation (`&#10`;→ literal newline reparses as a space), dropped comments/PIs, and malformed-XML handling? 

Relatedly, a non-fatal parse `error` with a `documentElement` still proceeds at _:39_ (and the `getElementsByTagName('parsererror')` check at _:32_ is a browser-DOMParser idiom that's dead under `@xmldom/xmldom`); parser errors should be fatal

Since TWB input shouldn't need DTDs, rejecting `<!DOCTYPE` up front is a cheap XXE/entity-expansion hardening win

src/desktop/commands/workbook/loadWorkbookXml.ts
```diff
@@ -0,0 +1,216 @@
...
+    | { type: 'execute-command-error'; error: ExecuteCommandError }
+    | { type: 'load-underlying-metadata-error'; error: ExecuteCommandResponseError }
+  >
+> {
+  const result = await executor.executeCommand({
+    namespace: 'tabui',
+    command: 'load-underlying-metadata',
+    signal,
+    args: {
+      text: xml,
```
mattcfilbert, line 158: This PR newly routes full workbook XML through the desktop Agent API path (`args.text` here; `get-workbook-xml` returns it via `result.text)`. The desktop Agent API logger logs request/response `headers` and `data` unmasked at debug (_getAgentApiClient.ts:94,115_), unlike the REST path which masks via `secretMask` (_restApiInstance.ts:277,306_); `apply-workbook` also passes `workbookXml` through the invocation notification args (_applyWorkbook.ts:63_) 

Reusing the REST mask helpers would redact `Authorization` but **not** workbook XML, so we'd need desktop-specific redaction/truncation of `args.text`, `workbookXml`, and response `result.text`. 

Medium severity (debug-only), but these tools create the new exposure

src/sdks/desktop/agentApi/types.ts
```diff
@@ -7,7 +7,7 @@ export const getCommandStatusResponseSchema = z.object({
   started_at: z.string().optional(),
   completed_at: z.string().optional(),
   duration_ms: z.number().optional(),
-  result: z.object({ text: z.string().optional() }).optional(),
+  result: z.record(z.string(), z.string()).optional(),
```
mattcfilbert: Quick sanity check because I forgot:

Is the desktop agent contract really `Record<string, string>` for every command result? 

If any command returns numeric/nested fields, this generic schema fails parsing before command-specific schemas run, turning a completed command into an opaque unknown error.

If not string-only, consider `z.record(z.string(), z.unknown())` and let each command's schema validate what it needs

src/tools/desktop/workbook/applyWorkbook.ts
```diff
@@ -0,0 +1,130 @@
...
+          const executor = await extra.getExecutor(session);
+          const result = await loadWorkbookXml({
+            xml: workbookXml,
+            executor,
+            signal: extra.signal,
+          });
+
+          if (result.isErr()) {
+            const { type, error } = result.error;
+            switch (type) {
```
mattcfilbert, line 110: This `switch (type)` has no `default/assertNever`, and control falls through to the success `Ok` after the `if (result.isErr())` block 

The union is exhaustive today, but adding a fourth error variant compiles cleanly and reports a failed destructive apply as success. Could add an `assertNever(type)` default or `return` immediately after the switch

src/desktop/validation/rules/calcFieldNames.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  it('warning issue includes suggestion with [Calculation_*] format', () => {
+    const xml = buildWorkbookXmlWithCalc('[Bad Name]');
+    const issues = calcFieldNamesRule.validate(xml);
+    const warning = issues.find((i) => i.severity === 'warning');
+    expect(warning).toBeDefined();
+    expect(warning!.suggestion?.includes('[Calculation_')).toBe(true);
+  });
+});
+
+describe('calc-field-names rule — false-positive guards', () => {
```
mattcfilbert, line 62: These assert `result.valid === true` for parameters/bins/copies/auto-columns, but `valid` is always true for a warning-only rule; so they pass even though the rule still warns on every one of those names

Could assert `issues.length === 0` for true exemptions, or rename the tests/comments to reflect that the rule intentionally warns on them

mattcfilbert (changes requested):
Solid migration. The bulk of this is clean and well-tested. Requesting changes on a small set, mostly concentrated in the workbook XML converter since it sits on the destructive apply-workbook path

src/desktop/commands/workbook/loadWorkbookXml.ts
anyoung-tableau: Hmm, yeah, it does look like text is a valid fallback for a `tabui:load-underlying-metadata` command with status=failed. Think that just got lost in translation during the migration.

src/desktop/toolExecutor/toolExecutor.ts
```diff
@@ -11,9 +11,15 @@ type WithAbortSignal = {
...
 };
 
-export type ExecuteCommandArgs<Z extends z.ZodTypeAny = z.ZodTypeAny> = {
+type WithExecutor = {
+  executor: ToolExecutor;
+};
+
+export type WithExecutorAndAbortSignal = WithExecutor & WithAbortSignal;
+
+export type ExecuteCommandArgs<Z extends z.ZodTypeAny | undefined = undefined> = {
```
jarhun88, line 20: what is this undefined = undefined?

src/desktop/libraries/workbook-serialization-converter/converter.ts
```diff
@@ -0,0 +1,78 @@
+/**
```
jarhun88, line 1: should we have some tests for this file, functionality doesn't seem trivial

src/desktop/commands/workbook/loadWorkbookXml.ts
anyoung-tableau: I'm not sure we actually need to care about this so much since everything is constrained to local execution. Logs are written locally and the agent, the MCP server, and the client API are running locally as well.

src/sdks/desktop/agentApi/types.ts
anyoung-tableau: I've been mostly doing this black-box since the client API is going to change anyway as Lauren progresses and I don't want to invest too much wasted time into it. Changing it to `z.unknown()` is fine since the local tool executor will parse it based off the provided `schema`.

src/tools/desktop/workbook/applyWorkbook.ts
anyoung-tableau: Good call

src/desktop/toolExecutor/toolExecutor.ts
anyoung-tableau: `Z` is a generic type parameter that extends the type `z.ZodTypeAny | undefined` and defaults to the type `undefined` if not provided.

src/desktop/validation/rules/calcFieldNames.test.ts
anyoung-tableau: Honestly, I copied the validation rules and the tests over from the prototype branch with very little modifications. Let me take a look at these...

src/desktop/libraries/workbook-serialization-converter/converter.ts
anyoung-tableau: Let me see if I can add some basic coverage. This library is copy/pasted and I'm not terribly happy that it's not tested and published elsewhere since now we are on the hook for owning and testing it. :|

src/desktop/libraries/workbook-serialization-converter/xmlFormatter.ts
anyoung-tableau: Good eye

src/desktop/libraries/workbook-serialization-converter/xmlParser.ts
anyoung-tableau: Thanks, I thought I had removed all the DOMParser stuff. Not sure why it was there in the first place since it's not available in NodeJS.

src/desktop/commands/workbook/loadWorkbookXml.ts
mattcfilbert: Fair point, Mr. Young. Local-only kills the exfil angle that `secretMask` covers on the REST path.

Buuuuuut: workbook XML can carry embedded connection creds (`<connection password='…'>`, OAuth tokens). Enable debug to file a bug, zip the logs, and "local" stops being the boundary

Cheap fix could be: truncate `args.text` / `result.text` in `logRequest`/`logResponse` to `<XML, N bytes>`. Also keeps debug logs readable for big workbooks. Whatchu think?

src/desktop/commands/workbook/loadWorkbookXml.ts
mattcfilbert: Yeah, it definitely starred both Bill Murray and ScarJo

Heads up it survived the restructure too: the executor turns a failed command into `command-failed` Err (localToolExecutor.ts:105-113), so in `loadUnderlyingMetadataByText` status is always 'completed' and that branch is still dead. Real failures come through `isErr()`. Fix is the retry, not the branch: lean on the filepath fallback (already wired on `command-failed`) and drop the dead `status === 'failed'` check + the `failed` enum

anyoung-tableau: Thanks, forgot to update the fallback too. I think a lot of this is going to change once Lauren's new client API lands anyway.

src/desktop/commands/workbook/loadWorkbookXml.ts
anyoung-tableau: It someone has access to files on your local machine you're pwned anyway. Lemme add some basic truncation so at least the log files don't balloon like mad.

jarhun88 (approved):

jarhun88:
its been challenging to review this pr as its very big and theres alot of context and functionality that I am not familiar with... im not sure whats a better way to work with this, maybe smaller prs, or a review from the person who wrote this originally?

mattcfilbert (approved):
lgtm

PR #388, opened 2026-06-09
@W-22969168 Refactor getWorkbook and getView tools to return url

jarhun88:
This PR adds URL support to workbook and view responses using the new
`AppToolResult` wrapper pattern, and fetch siteName handling for all auth types (cloud oauth, embedded oauth, passthru auth, pat/connectedapps/uat)

- Wrap `get-workbook` response with `AppToolResult<Workbook>` containing `{ data, url }` where url points to the default view
- Wrap `get-view` response with `AppToolResult<View>` containing `{ data, url }` where url points to the view 
- Add `constructViewWebUrl` helper to build view URLs from contentUrl

## Motivation and Context

**URL support:** The `url` field in `AppToolResult` enables clients to construct proper web URLs for workbooks and views, pointing directly to the interactive UI.


src/server/oauth/accessTokenValidator.test.ts
```diff
@@ -48,7 +48,7 @@ describe('TableauAccessTokenValidator', () => {
 
   describe('client_id claim resolution', () => {
     it('uses the client_id claim as the OAuth client ID', async () => {
-      const token = makeBearer(basePayload({ client_id: MOCK_CLIENT_ID }));
```
jarhun88, line 51: unnecessary as basePayload already uses MOCK_CLIENT_ID by default

src/server/oauth/accessTokenValidator.ts
```diff
@@ -176,31 +176,31 @@ export class TableauAccessTokenValidator extends AccessTokenValidator {
       // (enforced by the schema). aud holds the resource URL and is never used as the client ID.
       const oauthClientId = client_id;
 
-      if (!userId) {
```
jarhun88, line 179: all i did was remove this if. we always need to call this now to get sitename

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -51,59 +71,86 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
+
+          return new Ok({
+            data: workbook,
+            url,
+          });
+        },
+        constrainSuccessResult: (result) => {
+          const { data: workbook, url } = result;
+
+          const filteredWorkbook = filterWorkbookViews({
```
anyoung-tableau, line 139: What if the workbook's default view gets filtered out? I think we need to generate the `url` _after_ we filter the views from the bounded context.

jarhun88: I think at one point you mentioned you didnt like the getDefaultViewWebUrl setting to be in the constrainSuccessResult. So Id have to either pull out filterWorkbookViews to the main logic section, or just do getDefaultViewWebUrl in the constrainSuccessResult after filterWorkbookViews anyways. Do you have any thoughts? im kind of leaning towards the latter right now as its simpler to implement

jarhun88: i also need to check with joe which view gets shown if the default is removed

stephendeoca: I think a reasonable approach would be to check if the default view id is an allowed view id and not include the URL if it isn't allowed. Do this in the tool callback, and not the constrain search result

src/tools/web/utils/viewUrlUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+  // Remove '/sheets/' from contentUrl if present
+  // API returns 'workbook/sheets/Sheet1', URL uses 'workbook/Sheet1'
+  const urlPath = contentUrl.replace(/\/sheets\//g, '/');
+
+  // Default site uses a different URL structure without /site/{siteName}
+  if (!siteName || siteName === 'Default') {
+    return `${server}/#/views/${urlPath}`;
+  }
+
+  return `${server}/#/site/${siteName}/views/${urlPath}`;
```
stephendeoca: This is obviously fine and working, but you might want to consider constructing with a URL object to prevent trailing slashes from causing issues.

because if a server is something like 'https://my-server.com/' you're going to have double slashes

src/server/oauth/accessTokenValidator.ts
```diff
@@ -85,6 +85,24 @@ export class EmbeddedAccessTokenValidator extends AccessTokenValidator {
           return new Err('Invalid or expired access token');
         }
 
+        const restApi = new RestApi({
+          maxRequestTimeoutMs: this.config.maxRequestTimeoutMs,
+        });
+
+        restApi.setCredentials(tableauAccessToken, tableauUserId);
+        const sessionResult = await restApi.authenticatedServerMethods.getCurrentServerSession();
```
stephendeoca, line 93: not really looking for a change here, more so trying to discuss what we do here. I know that passthroughAuthMiddleware will call getCurrentServerSession and cache the result, is there a reason we don't use any cached results or cache here?

@anyoung-tableau @jarhun88

stephendeoca (approved):

src/server/oauth/accessTokenValidator.ts
jarhun88: i wasnt aware of it, seems like the pattern could apply here

src/tools/web/utils/viewUrlUtils.ts
jarhun88: its a good idea!

jarhun88: using URL object

src/tools/web/workbooks/getWorkbook.ts
jarhun88: im going to try to handle it like this https://salesforce-internal.slack.com/archives/C09S2ML7FH9/p1781305909655019?thread_ts=1781294409.483929&cid=C09S2ML7FH9

src/server/oauth/accessTokenValidator.ts
jarhun88: i cant think of any reason why the other auth types would have additional risks of caching if passthru has caching enabled. Same risks of an expired authtoken getting its sitename where the api call would usually fail, but it would just fail further downstream at the api calls anyway which i assume is what happens with passthru cache

jarhun88: actually i notice config.passthroughAuthUserSessionCheckIntervalInMinutes is specific for passthru... i think im actually just going to make a story to the backlog on improving perf instead of handling here as i dont want to make more decisions and add more complexity in this pr.

jarhun88: story created https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002cH2gZYAS/view

src/tools/web/workbooks/getWorkbook.ts
jarhun88: ive implemented

src/server/oauth/schemas.ts
```diff
@@ -130,6 +130,7 @@ export const tableauAuthInfoSchema = z.discriminatedUnion('type', [
     username: z.string(),
     server: z.string(),
     siteId: z.string().optional(),
+    siteName: z.string(),
```
stephendeoca, line 133: Site name can never be empty right? OAuth flow is only available for Tableau Cloud, thus there won't be a default site.

stephendeoca (approved):

PR #392, opened 2026-06-11
@W-22935148: Add delete-extract-refresh-task tool

Akash-Rastogi:
## Summary
- Implements `delete-extract-refresh-task` MCP tool (item 1B-1 from the Extract Refresh Schedule Optimization Apply phase)
- Adds DELETE `/sites/{siteId}/tasks/extractRefreshes/{taskId}` endpoint to the tasks API layer
- Registers `tableau:mcp:tasks:write` and `tableau:tasks:write` OAuth scopes
- Admin-gated with destructive annotations and 6 unit tests

## Notes
- Currently using `tableau:tasks:read` scope as a workaround — Tableau Cloud returns 401 on the DELETE endpoint with PAT auth. Waiting on CP CS-Schedules (SPAM) / Identity team to enable sessionless OAuth for this route.
- Once that dependency is resolved, swap scope back to `tableau:tasks:write` in `src/server/oauth/scopes.ts`.

## Test plan
- [x] `npm run build:dev` passes
- [x] Unit tests pass (`npx vitest run src/tools/web/extractRefreshTasks/`)
- [x] OAuth scopes tests pass
- [x] Lint passes
- [ ] E2E validation on Tableau Cloud (blocked on server-side enablement)
- [ ] Validate on Tableau Server with PAT auth


src/server/oauth/scopes.ts
```diff
@@ -79,6 +82,10 @@ const toolScopeMap: Record<
     mcp: ['tableau:mcp:tasks:read'],
     api: new Set(['tableau:tasks:read', 'tableau:users:read']),
   },
+  'delete-extract-refresh-task': {
+    mcp: ['tableau:mcp:tasks:write'],
+    api: new Set(['tableau:tasks:write', 'tableau:users:read']),
```
anyoung-tableau: Missing `tableau:tasks:delete` scope here

src/server/oauth/scopes.ts
```diff
@@ -79,6 +82,10 @@ const toolScopeMap: Record<
     mcp: ['tableau:mcp:tasks:read'],
     api: new Set(['tableau:tasks:read', 'tableau:users:read']),
   },
+  'delete-extract-refresh-task': {
+    mcp: ['tableau:mcp:tasks:write'],
```
anyoung-tableau: This should probably be tableau:mcp:tasks:delete

src/sdks/tableau/apis/tasksApi.ts
```diff
@@ -54,5 +54,32 @@ const listExtractRefreshTasksEndpoint = makeEndpoint({
...
 
-const tasksApi = makeApi([listExtractRefreshTasksEndpoint]);
+/**
+ * Delete Extract Refresh Task
+ * DELETE /api/api-version/sites/site-id/tasks/extractRefreshes/task-id
+ * Deletes an extract refresh task.
+ * Tableau Cloud scope: tableau:tasks:write
+ * @see https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_extract_and_encryption.htm#delete_extract_refresh_task
+ */
+const deleteExtractRefreshTaskEndpoint = makeEndpoint({
```
Alon-ST-DATA, line 64: We should add [doc for this new tool.](https://github.com/tableau/tableau-mcp/tree/main/docs/docs) Including in[ intro.md](https://github.com/tableau/tableau-mcp/blob/main/docs/docs/intro.md)
unless it is intentional..

src/server/oauth/scopes.ts
Akash-Rastogi: Fixed — updated to `tableau:tasks:delete` in commit 6622eb23.

src/server/oauth/scopes.ts
Akash-Rastogi: Fixed — renamed to `tableau:mcp:tasks:delete` in commit 6622eb23.

src/sdks/tableau/apis/tasksApi.ts
Akash-Rastogi: Good call — added `docs/docs/tools/tasks/delete-extract-refresh-task.md` and updated `intro.md` with the new tool (plus `list-extract-refresh-tasks` and `list-users` which were also missing) in commit c71b6e2a.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
```diff
@@ -0,0 +1,80 @@
...
+        extra,
+        args,
+        callback: async () => {
+          await useRestApi({
+            ...extra,
+            jwtScopes: deleteExtractRefreshTaskTool.requiredApiScopes,
+            callback: async (restApi) => {
+              const adminResult = await assertAdmin(restApi, extra);
+              if (adminResult.isErr()) {
+                throw new Error(adminResult.error);
```
anyoung-tableau: Let's be consistent and avoid throwing unnecessary errors. Other tools return a new `AdminOnlyError`

Akash-Rastogi: Fixed — now using `AdminOnlyError.toErr()` consistent with `adminInsightsToolBase` and `getStaleContentReport`. The `useRestApi` callback returns the Result directly instead of throwing.

src/sdks/tableau/apis/tasksApi.ts
```diff
@@ -54,5 +54,32 @@ const listExtractRefreshTasksEndpoint = makeEndpoint({
   response: listExtractRefreshTasksBodySchema,
 });
 
-const tasksApi = makeApi([listExtractRefreshTasksEndpoint]);
+/**
+ * Delete Extract Refresh Task
+ * DELETE /api/api-version/sites/site-id/tasks/extractRefreshes/task-id
+ * Deletes an extract refresh task.
+ * Tableau Cloud scope: tableau:tasks:write
```
anyoung-tableau: `tableau:tasks:delete`

src/sdks/tableau/methods/tasksMethods.ts
```diff
@@ -39,4 +39,26 @@ export default class TasksMethods extends AuthenticatedMethods<typeof tasksApis>
     const response = parseListExtractRefreshTasksResponse(raw);
     return response.tasks.task.map((t) => t.extractRefresh);
   };
+
+  /**
+   * Deletes an extract refresh task from the site.
+   *
+   * Required scopes (Tableau Cloud): `tableau:tasks:write`
```
anyoung-tableau: `tableau:tasks:delete`

src/sdks/tableau/apis/tasksApi.ts
Akash-Rastogi: Fixed — updated all remaining `tableau:tasks:write` references to `tableau:tasks:delete` in doc comments and tool description (commit 012eec7e).

src/sdks/tableau/methods/tasksMethods.ts
Akash-Rastogi: Fixed in same commit (012eec7e).

src/server/oauth/scopes.ts
```diff
@@ -79,6 +82,10 @@ const toolScopeMap: Record<
     mcp: ['tableau:mcp:tasks:read'],
     api: new Set(['tableau:tasks:read', 'tableau:users:read']),
   },
+  'delete-extract-refresh-task': {
+    mcp: ['tableau:mcp:tasks:delete'],
+    api: new Set(['tableau:tasks:delete', 'tableau:users:read']),
```
Alon-ST-DATA, line 87: The tableau:tasks:delete API scope here is not yet verified against a live Tableau Cloud site — the PR test plan still has "E2E validation on Tableau Cloud (blocked on server-side 
  enablement)" and "Validate on Tableau Server" unchecked, and the PR description notes Cloud returned 401 on this DELETE route with PAT auth (pending CP CS-Schedules / Identity enabling 
  sessionless OAuth).
  
  The unit suite mocks the REST client, so CI passes regardless of whether the connected-app token mint/exchange actually accepts tableau:tasks:delete. If Tableau doesn't recognize that 
  granular scope string, the token mints without delete authority and every real delete 401/403s for legitimate site admins — invisible until production.
  
  Can we confirm the exact scope string against a live Cloud site (mint a token carrying tableau:tasks:delete, run one real delete) before merge? This is the WI's Apply phase, so it'll get 
  exercised. Also worth a one-line code comment recording that the granular :delete scope is Cloud-verified, since the earlier tasks:write/tasks:read churn (commit 012eec7e) shows the right 
  string wasn't obvious.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
```diff
@@ -0,0 +1,81 @@
...
+          return await useRestApi({
+            ...extra,
+            jwtScopes: deleteExtractRefreshTaskTool.requiredApiScopes,
+            callback: async (restApi) => {
+              const adminResult = await assertAdmin(restApi, extra);
+              if (adminResult.isErr()) {
+                return new AdminOnlyError(adminResult.error).toErr();
+              }
+
+              await restApi.tasksMethods.deleteExtractRefreshTask({
```
Alon-ST-DATA, line 64: Maybe out of scope for this WI - but note for thought.
list-extract-refresh-tasks filters results through constrainExtractRefreshTasks(boundedContext.datasourceIds/workbookIds). The delete tool runs assertAdmin then deletes the raw taskId with no allowlist check. An operator who scopes the server to specific content still has every task on the site deletable by id.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  This tool is restricted to Tableau site administrators and requires the \`ADMIN_TOOLS_ENABLED\` feature flag to be enabled.
+
+  Use this tool when you need to:
+  - Remove a scheduled extract refresh that is no longer needed
+  - Disable refresh schedules for under-used or decommissioned content
+  - Optimize site resources by eliminating unnecessary extract refreshes
+
+  **Parameters:**
+  - \`taskId\` (required) – The ID of the extract refresh task to delete. Obtain this from the \`list-extract-refresh-tasks\` tool.
```
Alon-ST-DATA, line 36: idempotentHint: true (line 47) is wrong for a hard delete-by-id

  Second delete of the same taskId 404s → isError: true (your own "handle API errors gracefully" test proves it). A client trusting the idempotent hint and retrying after a transient failure
  gets a spurious error for an operation that already succeeded.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  it('should fail when user is not admin', async () => {
+    mocks.mockAssertAdmin.mockResolvedValue({
+      isErr: () => true,
+      error: 'User is not a site administrator',
+    });
+    const result = await getToolResult({ taskId: 'task-123' });
+    expect(result.isError).toBe(true);
+    invariant(result.content[0].type === 'text');
+    expect(result.content[0].text).toContain('not a site administrator');
```
Alon-ST-DATA: expect(text).toContain('not a site administrator') passes only because the mock feeds that exact string as .error. Real assertAdmin emits "This tool requires site administrator 
  permissions…", which does not contain "not a site administrator." The assertion validates a literal it supplied to itself.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  });
+
+  it('should call assertAdmin before deleting', async () => {
+    await getToolResult({ taskId: 'task-123' });
+    expect(mocks.mockAssertAdmin).toHaveBeenCalled();
+  });
+
+  it('should fail when user is not admin', async () => {
+    mocks.mockAssertAdmin.mockResolvedValue({
+      isErr: () => true,
```
Alon-ST-DATA: Three problems in the one test that's supposed to prove the security gate works:
  - It mocks a hand-rolled { isErr: () => true, error: '…' } instead of a real ts-results-es Err —  It tests the mock's shape, not assertAdmin's real Result contract.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      taskId: 'task-123',
+    });
+  });
+
+  it('should call assertAdmin before deleting', async () => {
+    await getToolResult({ taskId: 'task-123' });
+    expect(mocks.mockAssertAdmin).toHaveBeenCalled();
+  });
+
+  it('should fail when user is not admin', async () => {
```
Alon-ST-DATA: It never asserts mockDeleteExtractRefreshTask was not called. A regression where the delete proceeds despite the admin failure would still pass this test — the exact bug the gate exists to prevent.

docs/docs/tools/tasks/delete-extract-refresh-task.md
```diff
@@ -0,0 +1,67 @@
...
+
+Use this tool when you need to:
+- Remove a scheduled extract refresh that is no longer needed
+- Disable refresh schedules for under-used or decommissioned content
+- Optimize site resources by eliminating unnecessary extract refreshes
+
+## Required permissions
+
+- **Tableau Cloud**: Requires `tableau:tasks:delete` OAuth scope
+- **Tableau Server**: Non-admin users can delete tasks for which they have Read and Delete permissions
```
Alon-ST-DATA: Nit: 
Docs contradict gate (non-admin Server users vs hard assertAdmin)
do we allow non admins to delete in mcp? I do not think we do.

src/sdks/tableau/methods/tasksMethods.ts
```diff
@@ -39,4 +39,26 @@ export default class TasksMethods extends AuthenticatedMethods<typeof tasksApis>
...
   };
+
+  /**
+   * Deletes an extract refresh task from the site.
+   *
+   * Required scopes (Tableau Cloud): `tableau:tasks:delete`
+   *
+   * @param siteId - The Tableau site ID
+   * @param taskId - The extract refresh task ID to delete
+   * @link https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_jobs_tasks_and_schedules.htm#delete_extract_refresh_task
```
Alon-ST-DATA: Nit:
@link wrong anchor (jobs_tasks_and_schedules#delete_extract_refresh_task; rest of PR uses extract_and_encryption)

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
```diff
@@ -0,0 +1,81 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
```
Alon-ST-DATA, line 1: Nit:

Admin-gate boilerplate divergence (this tool returns AdminOnlyError; list sibling throws raw  Error) 
 src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts:103

Alon-ST-DATA:
**Live E2E finding (DE-3): empty `taskId` returns HTTP 405 instead of a clean client-side 400**

  Ran the delete tool against a live Tableau Cloud site (site-admin auth, ADMIN_TOOLS_ENABLED,
  629 real extract refresh tasks). Error-path matrix — bogus/empty IDs only, no real task deleted:

  | Input | Result |
  |-------|--------|
  | "" (empty string) | 405 |
  | not-a-valid-uuid | 404 |
  | 00000000-0000-0000-0000-000000000000 (well-formed, absent) | 404 |
  | '; DROP TABLE tasks;-- | 400 |

  Issue: empty taskId is accepted and builds DELETE /sites/{siteId}/tasks/extractRefreshes/
  (no id segment) → 405. Confusing error for an invalid-argument case, and the tool issues a
  malformed write instead of failing fast.

  Suggested fix — client-side guard on the schema:
      taskId: z.string().uuid()   // or at minimum .trim().min(1)
  .uuid() also tightens the not-a-valid-uuid case (currently relies on server 404).

  Separate from already-noted review concerns (idempotent hint on hard delete; missing
  allowlist on raw taskId) — both also reproduced live (re-deleting absent id → 404 confirms
  delete is not idempotent).

src/server/oauth/scopes.ts
Akash-Rastogi: I verified it again a test site where I did list extract refresh, grabbed a UUID and gave it to delete. Then run list again and noticed the associated task was gone.
Is that what you meant by testing against live Cloud site ^^

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
Akash-Rastogi: Good catch — acknowledged as a valid concern. Out of scope for this WI but worth a follow-up. The admin gate ensures only site admins can reach the delete, but a boundedContext check would add defense-in-depth for scoped deployments.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
Akash-Rastogi: Fixed — changed to `idempotentHint: false` in commit 80c70d52.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
Akash-Rastogi: Fixed — test now uses the real `assertAdmin` message format (`This tool requires site administrator permissions. Your site role is: Viewer`) in commit 80c70d52.

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
Akash-Rastogi: Fixed — now uses a real `new Err(...)` from ts-results-es instead of a hand-rolled object (commit 80c70d52).

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.test.ts
Akash-Rastogi: Fixed — added `expect(mocks.mockDeleteExtractRefreshTask).not.toHaveBeenCalled()` (commit 80c70d52).

docs/docs/tools/tasks/delete-extract-refresh-task.md
Akash-Rastogi: Fixed — removed the non-admin Server claim. MCP enforces admin-only via `assertAdmin` regardless of what the raw REST API allows (commit 80c70d52).

src/sdks/tableau/methods/tasksMethods.ts
Akash-Rastogi: Fixed — updated to `rest_api_ref_extract_and_encryption.htm#delete_extract_refresh_task` (commit 80c70d52).

src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts
Akash-Rastogi: Agreed — this tool already uses the newer `AdminOnlyError` pattern. The divergence is in the `listExtractRefreshTasks` sibling which predates it. Could unify in a follow-up but didn't want to change unrelated files in this PR.

Alon-ST-DATA:
Looking good: Last e2e results:

## Results matrix

| ID | Case | Expected | Run 1 (pre-fix) | Run 2 (post-fix) | Result |
|----|------|----------|-----------------|------------------|--------|
| DE-1 | Enumerate via `list-extract-refresh-tasks` | tasks returned | 629 tasks | 629 tasks | **Pass** |
| DE-2 | Admin gate — current session = site admin | allowed | executed | executed | **Pass** |
| DE-3 | Empty `taskId` (`""`) | clean client-side 400 | **405** (malformed HTTP — no guard) | **`-32602` Zod: "taskId must be a valid UUID"** | **Fixed ✅** |
| DE-4 | Malformed `taskId` (`not-a-valid-uuid`) | rejected before HTTP | 404 (server) | **`-32602` Zod (client-side)** | **Fixed ✅** |
| DE-5 | Well-formed absent UUID (`00000000-…-000`) | 404 graceful | 404 | 404 | **Pass** |
| DE-6 | Injection (`'; DROP TABLE tasks;--`) | rejected, no injection | 400 (server) | **`-32602` Zod (client-side)** | **Pass (hardened ✅)** |
| DE-7 | Idempotency — re-delete absent id | hint matches behavior | 404; `idempotentHint:true` (wrong) | 404; **`idempotentHint:false`** (L47) | **Fixed ✅** |

Alon-ST-DATA:
One follow up item for another Gus story, if possible.

⏺ Follow-up: DE-8 — self-contained E2E test for delete-extract-refresh-task

  What & why: PR 392 added the delete tool. Live E2E proved all the negative paths (empty / malformed / injection ids rejected by
  Zod; absent id → 404). The happy-path was never run — actually deleting a real task and confirming it's gone — because it's
  irreversible and this is a shared live site. The success path is code-reviewed but empirically unproven.
  
  What to build: an automated E2E test that owns its own fixture so it's safe in CI (touches no real content):

  1. Setup — create a throwaway datasource + extract refresh task.
  2. Delete — call delete-extract-refresh-task with that task's id; assert success message.
  3. Verify — re-list; assert task absent + count −1.
  4. Re-delete — same id again → assert 404 (confirms idempotentHint:false).
  5. Teardown — drop the throwaway datasource.

  Location: tests/e2e/deleteExtractRefreshTask.test.ts (mirrors tests/e2e/client.ts harness), green in CI.

Alon-ST-DATA (approved):

PR #400, opened 2026-06-15
@W-22968653 Add Tableau embedding support to MCP app

jarhun88:
## Description
Screenshot 2026-06-16 at 11 42 15 PM

This PR implements full-screen Tableau visualization embedding in the MCP app using the Embedding API v3. The app now retrieves OAuth tokens via the `get-oauth-token` tool and embeds Tableau views received from MCP tool results.

Key changes:
- Add `tableau:views:embed` scope to `get-oauth-token` tool for embedding authentication
- Create `embedTableauViz` utilities for creating and managing viz elements
- Create `getOAuthTokenToolClient` module for retrieving OAuth tokens from MCP server
- Simplify app UI to full-screen layout with no loading hero section
- Dynamically load Tableau Embedding API from configured server URL
- Add comprehensive test coverage (19 new tests) for embedding and token retrieval
- Install jsdom as dev dependency for DOM testing

## Motivation and Context

The MCP app needed the ability to embed Tableau visualizations when users request views through MCP tools. This enables a seamless experience where tool results are automatically rendered as interactive visualizations rather than just returning metadata.

The critical fix was adding the `tableau:views:embed` scope to the OAuth token, which is required by Tableau's Embedding API. Without this scope, the embedding API prompts for login instead of accepting the token.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Added 19 unit tests covering:
  - OAuth token retrieval (7 tests in `getOAuthTokenToolClient.test.ts`)
  - Tableau viz element creation and embedding (12 tests in `embedTableauViz.test.ts`)
- All 1742 tests pass locally
- Manual testing: verified embedding works with OAuth token containing `tableau:views:embed` scope
- Tested full-screen layout and visualization rendering

## Related Issues

Related to the MCP apps feature development.

## Checklist

- [ ] I have updated the version in the package.json file by using `npm run version`. For example, use `npm run version:patch` for a patch version bump.
- [ ] I have documented any breaking changes in the PR description. For example, renaming a config environment variable or changing its default value.

## Contributor Agreement


tests/e2e/server.test.ts
```diff
@@ -129,6 +130,7 @@ describe('server', () => {
         'query-admin-insights-site-content',
         'query-admin-insights-job-performance',
         'get-stale-content-report',
+        'delete-workbook',
```
jarhun88, line 135: someone must have added thistool without updating the e2e test so it was broken

anyoung-tableau: Ah yeah, the author of that PR submitted it from a fork, so the E2E tests didn't run as part of the PR.

src/web/apps/src/mcp-app.ts
```diff
@@ -1,16 +1,73 @@
...
+  url: z.string().url(),
+});
+
+/**
+ * Loads the Tableau Embedding API script from the Tableau server
+ */
+function loadTableauEmbeddingApi(viewUrl: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    // Check if already loaded
+    if (customElements.get('tableau-viz')) {
```
anyoung-tableau, line 40: This might be overly cautious, but clients could sandbox the app iframe in a way such that the custom element registry is intentionally unavailable. Consider adding a guard for this:

```ts
if (!('customElements' in window)) {
  reject(...)
}
```

src/server.web.ts
```diff
@@ -183,14 +183,32 @@ export class WebMcpServer extends Server {
     );
 
     // Register the resource, which returns the bundled HTML/JavaScript for the UI.
+    const config = getConfig();
+    const serverOrigin = new URL(config.server).origin;
+
+    // Allow Tableau Online, Tableau.com domains, and configured server
+    const cspDomains = ['https://*.online.tableau.com', 'https://*.tableau.com', serverOrigin];
```
anyoung-tableau: Consider parameterizing the first two values as an environment variable. In the event we need to react quickly by adding a new domain in the future we can update the config rather than needing to take a code change + release.

src/web/apps/src/mcp-app.ts
```diff
@@ -1,16 +1,73 @@
...
+    }
+
+    // Derive embedding API URL from the view URL
+    const serverOrigin = new URL(viewUrl).origin;
+    const embeddingApiUrl = `${serverOrigin}/javascripts/api/tableau.embedding.3.latest.min.js`;
+
+    const script = document.createElement('script');
+    script.type = 'module';
+    script.src = embeddingApiUrl;
+    script.onload = () => resolve();
```
anyoung-tableau: Not sure if there's any problem with this approach, but note that if the Embedding API loads and executes, the `onload` event will fire _even_ if there is some runtime error. Since agents could be doing "creative" things with the sandboxed iframe which could cause the Embedding API to throw some runtime error, it might be worth instead resolving the promise once the custom element actually gets defined. Worth looking into...

```ts
customElements.whenDefined('tableau-viz').then(() => {
  resolve();
});
```

src/web/apps/mcp-app.html
```diff
@@ -7,12 +7,7 @@
 </head>
 <body>
   <main class="main">
-    <div class="hero">
-      <div class="emoji">📊✨</div>
-      <h1>Tableau MCP App</h1>
-      <p>Your workbook data is being loaded...</p>
-      <div class="sparkle">🌟</div>
-    </div>
+    <div id="tableauVizContainer" class="viz-container"></div>
```
anyoung-tableau, line 10: Mind sharing a screenshot?

src/web/apps/src/lib/getOAuthTokenToolClient.ts
```diff
@@ -0,0 +1,25 @@
...
+  const result = await app.callServerTool({
+    name: 'get-oauth-token',
+    arguments: {},
+  });
+
+  // Parse the result to extract the token
+  const content = result.content[0];
+  if (content.type === 'text') {
+    const data = JSON.parse(content.text);
+    return data.token;
```
anyoung-tableau: Consider adding a zod schema to ensure the format of the result text matches what the app expects.

src/web/apps/src/mcp-app.ts
```diff
@@ -1,16 +1,73 @@
...
+  return validated.url;
+}
+
+// Handle tool results
+app.ontoolresult = async (result: CallToolResult) => {
+  try {
+    const viewUrl = extractUrlObjectFromResult(result);
+    await loadTableauEmbeddingApi(viewUrl);
+    const token = await callGetOAuthTokenTool(app);
+    embedTableauViz(viewUrl, token);
```
anyoung-tableau, line 98: This is fine for V0, but a nice next step will be to listen for auth failures using: https://help.tableau.com/current/api/embedding_api/en-us/docs/embedding_api_auth.html#add-an-event-listener-for-load-and-initialization-errors

src/server/oauth/scopes.ts
```diff
@@ -141,11 +142,9 @@ const toolScopeMap: Record<
       ...RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES,
     ]),
   },
-  // Token retrieval: no Tableau REST API calls, no content scope required.
-  // Any authenticated user may retrieve their own token regardless of granted scopes.
   'get-oauth-token': {
     mcp: [],
-    api: new Set<TableauApiScope>(),
+    api: new Set<TableauApiScope>(['tableau:views:embed']),
```
anyoung-tableau, line 168: We have a small design problem in that this scope is now going to be advertised as necessary when the client requests an access token, even if the tool is disabled or the mcp-apps feature flag is disabled, making the access token more broadly scoped than it needs to be. Compounding that problem, the list of advertised scopes for the global URLs is completely static (which we'll need to update before we enable the feature flag on the hosted service). At a minimum, you could update the `getEnabledToolNames` function below to stop advertising this scope when it's not actually necessary.

src/web/apps/src/mcp-app.ts
anyoung-tableau: You way also want to implement some kind of timer with `setTimeout` to reject the promise if it doesn't resolve within, say, 30 seconds. Otherwise, the user will see nothing with no indication of what's going on.

src/web/apps/mcp-app.html
anyoung-tableau, line 10: Do we have any mockups from design for what we should render when the Embedding API or viz is loading?

jarhun88: just added to the desc

src/web/apps/src/lib/getOAuthTokenToolClient.ts
jarhun88: added and now i remove the ? check on content

src/web/apps/src/mcp-app.ts
jarhun88: thanks ill take note of this for the error handling story potentially

src/web/apps/mcp-app.html
jarhun88: not yet, im syncing with joe and teh designer next week

src/server.web.ts
jarhun88: i did it, but would it be better to leave the three and pass the config list to combine with cspDomains?

src/server/oauth/scopes.ts
jarhun88: thats a good point, ive removed the tool when the flag is off

src/web/apps/src/mcp-app.ts
jarhun88: curious how could this happen?

src/web/apps/src/mcp-app.ts
jarhun88: i vibed it with claude in a way i think makes sense, and used 15s, 30 felt abit long

tests/oauth/tableau-authz/tests/getOAuthToken.test.ts
```diff
@@ -9,6 +10,7 @@ const tokenResponseSchema = z.object({
 
 test.describe('get-oauth-token', () => {
   test('should return Bearer token', async ({ client }) => {
+    test.skip(!getFeatureGate().isFeatureEnabled('mcp-apps'), 'mcp-apps feature is disabled');
```
jarhun88, line 13: this test will fail if its enabled as the right scope doesnt get added unless the FF is on, so i just made it conditional to the ff state

mattcfilbert:
@jarhun88 — 🤖 **MattGPT** (Matt Filbert's agent)

Embeds Tableau views in the MCP app via Embedding API v3. CI is green and the PR adds solid test coverage.

**One gap worth closing:**

- The new CSP config (`connectDomains` / `resourceDomains` / `frameDomains` on the app resource) has no test asserting the directives and domains are actually applied to the resource metadata. _(Inline note on the spot.)_

**Minor:**

- The embedding-API loader resolves via `script.onload` + `customElements.whenDefined` with a 15s timeout. If the script loads but never defines `tableau-viz`, you wait the full timeout and get a generic error. An explicit `customElements.get()` check post-load would fail faster with a clearer message.

src/server.web.ts
```diff
@@ -183,14 +183,32 @@ export class WebMcpServer extends Server {
...
       // @ts-expect-error -- harmless type mismatch in registerAppResource; ext-apps uses MCP SDK v1.25.2. Should go away when MCP SDK is updated.
       this.mcpServer,
       tool.name,
       resourceUri,
-      { mimeType: RESOURCE_MIME_TYPE },
+      {
+        mimeType: RESOURCE_MIME_TYPE,
+        _meta: {
+          ui: {
+            csp: {
```
mattcfilbert, line 200: 🤖 **MattGPT** (Matt Filbert's agent)

This CSP block (`connectDomains` / `resourceDomains` / `frameDomains`) has no test asserting it's actually applied.

Worth a test that mocks the MCP server and checks `_meta.ui.csp` matches `cspAllowedDomains` + `serverOrigin`.

src/web/apps/src/mcp-app.ts
anyoung-tableau: The iframe the client uses to render the MCP App could inject some JS that deletes or overwrites `window.customElements`, similar to how Tableau does it for its sandboxed extensions e.g. https://gitlab.tableausoftware.com/extensibility/extensions-hosting-playground/-/blob/dev/src/sandboxResources/sandbox.js#L54

src/web/apps/src/mcp-app.ts
```diff
@@ -47,14 +79,12 @@ const app = new App({ name: 'Tableau MCP App', version: pkg.version });
...
-    throw new Error('Tool result does not contain text content');
-  }
+  const validated = callToolResultSchema.parse(result);
+  const content = validated.content[0];
 
   const data = JSON.parse(content.text);
-  const validated = urlSchema.parse(data);
-  return validated.url;
+  const url = urlSchema.parse(data);
+  return url.url;
```
anyoung-tableau: Nit: kinda named poorly now. Maybe this?

```ts
const { url } = urlSchema.parse(data);
return url;
```

anyoung-tableau (approved):

src/server.web.ts
anyoung-tableau: Yeah, you could leave the 3 and then concat whatever appears on the env var.

src/web/apps/src/mcp-app.ts
jarhun88: ah gotcha

PR #393, opened 2026-06-11
@W-22947370: Add delete-workbook tool

Alon-ST-DATA:
## Summary

Implements the `delete-workbook` MCP tool — the destructive workbook step of the **Stale Content Cleanup → Apply (2B)** flow (264 "Tableau MCP for Admins", [W-22947370]). Sibling to `delete-extract-refresh-task` (#392).

The tool is **two-phase and safe by construction**:

1. **Preview** (default — `confirm` omitted or `false`): tags the workbook `stale-pending-deletion` (reversible, visible in the Tableau UI), reports name / project / owner, and deletes nothing.
2. **Delete** (`confirm: true`): `DELETE /sites/{siteId}/workbooks/{workbookId}`. On Tableau Cloud the workbook is moved to the [recycle bin](https://help.tableau.com/current/pro/desktop/en-us/recycle_bin.htm) and is restorable for a limited time before permanent removal.

Admin-gated via the existing `assertAdmin()` pattern and `disabled: !config.adminToolsEnabled`; the tool is removed from the manifest in `getEnabledToolNames()` when `ADMIN_TOOLS_ENABLED` is off, and a non-admin caller is rejected before any side effect.

## Scopes

| Action | API scope |
| --- | --- |
| delete | `tableau:content:delete` (MCP: `tableau:mcp:workbook:delete`) |
| tag (preview) | `tableau:content:write` |
| getWorkbook (preview) | `tableau:content:read` |
| owner lookup + admin check | `tableau:users:read` |

The wider `api` set is a deliberate consequence of bundling tag + notify-info + delete in one tool (see "Scope of this tool" below).

## Changes

- **New tool** `src/tools/web/workbooks/deleteWorkbook.ts` (+ registration in `tools.ts` / `toolName.ts`).
- **REST layer**: `deleteWorkbook` (DELETE) and `addTagsToWorkbook` (PUT …/tags) endpoints + methods on `workbooksApi.ts` / `workbooksMethods.ts`; added `owner` to `workbookSchema`.
- **Scopes**: `tableau:content:delete`, `tableau:content:write`, `tableau:mcp:workbook:delete` registered in `scopes.ts` (+ `JwtScopes` in `restApiInstance.ts`); `delete-workbook` added to the admin-gate removal list.
- **Docs**: per-tool page `docs/docs/tools/workbooks/delete-workbook.md` + `intro.md` row.
[...]

PR #402, opened 2026-06-16
@W-22973720: Add delete-datasource tool

Alon-ST-DATA:
## Summary

Adds a `delete-datasource` admin tool to the Tableau MCP server, mirroring the existing `delete-workbook` tool (#393).

Deletes a published data source by ID via the Tableau REST API. Gated behind the admin-tools feature flag and protected by a confirmation-token handshake to prevent accidental deletion.

## Changes

- **New tool** `src/tools/web/listDatasources/deleteDatasource.ts` — `delete-datasource` with confirmation-token gate
- **REST API / methods** `datasourcesApi.ts`, `datasourcesMethods.ts` — delete-datasource endpoint
- **Lineage** `lineageUtils.ts` — surface downstream-dependency warning in the delete confirmation
- **Types** `dataSource.ts`
- **OAuth scopes** `scopes.ts` — new `tableau:mcp:datasource:delete` scope + admin-gate wiring
- **Registration** `tools.ts`, `toolName.ts`
- **Docs** `docs/docs/tools/data-qna/delete-datasource.md`, `intro.md`
- **Version** bump to `2.14.0`

## Testing

- Unit: `src/tools/web/listDatasources/deleteDatasource.test.ts` (300 lines)
- E2E: `tests/e2e/datasources/deleteDatasource.test.ts`
- OAuth scope test updated

**Manual e2e testing**

Tool: `delete-datasource`   Category: admin action (mutating, two-phase)
Server: `tableau-mcp-local` HTTP :3927, PID 11375 (started 22:02), `build/index.js` mtime 15:08 — build unchanged since prior PR402 run; no `.ts` newer than build. Package version **2.12.0**.
Site: `admin-insights-tmcp-site` (10ax), siteId `c04137ab-2fed-4ef4-add8-b740c6918632`
Ground truth: direct REST (signin contentUrl = SITE_NAME) + Metadata API GraphQL reverse-lineage. Binding confirmed: fixture LUIDs published via REST matched MCP `list-datasources`.

## Verdict: PASS (10/10 core + N1/N2/N3 new-surface). No new bugs. Same carry-forward error-polish gap. Reproduces prior PR402 result exactly.

Fixtures: 3 bare `.hyper` (3-row) datasources published via REST `multipart/mixed` into throwaway project `dds-e2e-throwaway`.
- dds_happy `3b366958-34e1-4354-a9ae-32ffdab9862a` (token `43e11758a364`)
- dds_tokenguard `73518a8b-b8d4-46cf-ad69-dde6a8971c3c` (token `1bb07e081065`)
- dds_nodeps `64fe8bab-3199-4f5c-92e5-e1ff6482d044` (token `05f612700fa6`)

All tokens independently recomputed as `sha256("{siteId}:{luid}")[:12]` and matched tool output exactly.

## Core regression matrix

| # | Case | Result |
|---|------|--------|
| 1 | Preview happy → token + tag | **PASS** — token `43e11758a364` matched; REST confirms `pending-deletion` applied to happy only, others untouched |
| 2a | confirm:true, no token | **PASS** — rejected, "Deletion requires the confirmationToken returned by the preview step" |
| 2b | confirm:true, forged token `deadbeef0000` | **PASS** — rejected (same message) |
| 2c | confirm:true, cross-content token (nodeps `05f612700fa6` vs happy) | **PASS** — rejected; tokens content-bound |
| — | happy survived all 3 negatives | **PASS** — REST GET 200 |
| 3 | confirm:true + correct token (happy) | **PASS** — REST GET → 404 (genuinely gone); tool `list` also empty |
| 4a | re-delete deleted LUID (preview) | **PASS** — clean 404, no silent success |
| 4b | nonexistent LUID (all-zeros) | **PASS** — 404 (raw-error gap, see below) |
| 4c | malformed LUID (`not-a-valid-luid`) | **PASS** — 404 (raw-error gap, see below) |
| 5 | custom `tag` (e2e-custom-tag) on tokenguard | **PASS** — token `1bb07e081065` matched; REST shows `e2e-custom-tag`, default NOT applied |

## New-surface cases

| # | Case | Result |
|---|------|--------|
| N1 | Preview TS Users (real, has dependents) → ⚠️ WARNING | **PASS** — tool reports 14 workbooks; name-set EQUAL to Metadata API GT (14/14, exact names); token `4cb49188db76` matched prior run. TS Users restored (untagged → NONE) after test. |
| N2 | Preview with NO dependents (happy/nodeps) | **PASS** — "No workbooks or flows were found that depend on this data source." |
| N3 | Owner-email resolution in preview + delete msgs | **PASS** — `asimantov@tableau.com` throughout |

## Observations this run (not bugs in the tool)
- **Parallel MCP list calls → HTTP 429** (rate limit) on near-simultaneous filtered list-datasources. Confirms the harness "run tools sequentially" rule. Ran everything sequential thereafter.
- **`name:in:[...]` filter rejected** by list-datasources ("Invalid filter expression format"). Used `name:eq:` per-name instead. (list-datasources filter quirk, separate tool.)
- **Fixture LUID reuse:** `overwrite=true` republish reused prior-run LUIDs that still carried `pending-deletion` tags; cleared via REST before case 1 to establish clean baseline.

## Carry-forward issue (not a blocker, unchanged from prior runs)
- **Error-polish gap:** nonexistent & malformed LUIDs surface raw `Request failed with status code 404` (requestId 14/15/16). No up-front LUID-format validation; malformed string reaches the API and 404s rather than a friendly rejection. Inconsistent with the polished token-path messages. Same as delete-workbook PR393 and prior PR402 run.

## Teardown
All fixtures deleted (204), throwaway project deleted (204), `projectName:eq:dds-e2e-throwaway` totalAvailable = 0. TS Users tags restored to NONE. `/tmp/dds_e2e` wiped.


Alon-ST-DATA:
Addressed 2 inputs made by @Akash-Rastogi in staged PR:
[datasource dir rename
](https://github.com/Alon-ST-DATA/tableau-mcp/pull/1#discussion_r3414744028)

[idempotent flag](https://github.com/Alon-ST-DATA/tableau-mcp/pull/1#discussion_r3414752389)

src/tools/web/datasources/deleteDatasource.test.ts
```diff
@@ -0,0 +1,354 @@
...
+  mockAddTagsToDatasource: vi.fn(),
+  mockDeleteDatasource: vi.fn(),
+  mockQueryUserOnSite: vi.fn(),
+  mockGraphql: vi.fn(),
+  mockAssertAdmin: vi.fn(),
+  mockIsDatasourceAllowed: vi.fn(),
+}));
+
+vi.mock('../../../restApiInstance.js', () => ({
+  useRestApi: vi.fn().mockImplementation(async ({ callback }) =>
```
Akash-Rastogi, line 62: The `useRestApi` mock receives `jwtScopes: deleteDatasourceTool.requiredApiScopes` from the tool but no test asserts that the right set was passed. A regression where `toolScopeMap['delete-datasource'].api` drifts from what the tool actually needs (e.g. someone removes `tableau:datasource_tags:update` from scopes.ts) would pass silently here.

Consider adding an assertion like:
```ts
expect(vi.mocked(useRestApi)).toHaveBeenCalledWith(
  expect.objectContaining({
    jwtScopes: expect.arrayContaining([
      'tableau:datasources:delete',
      'tableau:datasource_tags:update',
      'tableau:content:read',
      'tableau:users:read',
    ]),
  }),
);
```
on one of the success-path tests, so the test pins the wired scopes.

src/tools/web/datasources/deleteDatasource.ts
```diff
@@ -0,0 +1,297 @@
...
+    ),
+  confirmationToken: z
+    .string()
+    .optional()
+    .describe(
+      'Required when confirm is true. The confirmationToken returned by the preview step ' +
+        '(confirm omitted/false) for this data source. Deletion is rejected without a matching ' +
+        'token, which guarantees a preview was run first.',
+    ),
+  tag: z
```
Akash-Rastogi, line 69: Nit: `tag: z.string().optional()` has no length bound. A multi-MB `tag` would be sent verbatim to the Tableau REST API (`addTagsToDatasource`). Tableau will likely reject extreme cases at the API layer, but adding `.max(200)` (or a similar reasonable cap) at the schema layer rejects fast and gives a clearer error to the caller.

Not a regression — `delete-workbook` likely has the same shape — so a cross-tool follow-up may be cleaner than fixing only here.

Akash-Rastogi (approved):
LGTM — approving. Solid two-phase pattern, defensive gate ordering (admin → token → scoping → I/O), graceful degradation for Metadata API and owner lookup, comprehensive unit tests, e2e safe-by-default with opt-in destructive path. Resource-specific OAuth scopes look correct.

A few non-blocking suggestions are inline. Two general comments that aren't anchored to a single line:

**Missing scope-tests parity.** `src/server/oauth/scopes.test.ts` exercises `tasks:read` / `jobs:read` admin-tool gating but adds no analogous tests for the new `tableau:mcp:datasource:delete` / `tableau:datasources:delete` / `tableau:datasource_tags:update` scopes. The same gap exists for `delete-workbook` already on main, so this is parity not regression — worth a follow-up to add admin-tool-gated tests for both delete tools.

**Tool description vs Tableau Server (cross-file).** The tool description, the recycle-bin URL in `deleteDatasource.ts`, and the docs at `docs/docs/tools/data-qna/delete-datasource.md` all assume Tableau Cloud (recycle bin → recoverable for a limited time). On Tableau Server (on-prem), deletion is permanent — there is no recycle bin. The tool doesn't gate on Cloud-vs-Server. Same shape as `delete-workbook` so parity not regression, but worth flagging at the program level — especially since `intro.md` says the server "Works with any published data source on either Tableau Cloud or Tableau Server". Consider either gating the tool on Cloud, or softening the recycle-bin language to "on Tableau Cloud, deleted data sources can be restored from the recycle bin; on Tableau Server, deletion is permanent".

src/tools/web/datasources/deleteDatasource.test.ts
Alon-ST-DATA: Fixed: new test "should wire the tool-mapped API scopes into useRestApi" — asserts useRestApi got exactly [datasources:delete, datasource_tags:update, content:read, users:read]. Order-independent (sorted compare), catches both removed AND stray
  scopes. Mutation-checked: flipping a scope fails it. 20/20 pass, tsc clean, eslint clean.  [commit](https://github.com/tableau/tableau-mcp/pull/402/changes/96f5aa58b72f8da23c0b2764ef5319cef9812eec)

src/tools/web/datasources/deleteDatasource.ts
Alon-ST-DATA: Resolved in that [commit](https://github.com/tableau/tableau-mcp/pull/402/changes/b704369162ebb746ca579d753f33d63e0aaab312)

src/tools/web/datasources/deleteDatasource.ts
```diff
@@ -0,0 +1,299 @@
...
+                    'Run delete-datasource with confirm omitted (or false) for this datasourceId ' +
+                    'first, then call again with confirm: true and the confirmationToken from that ' +
+                    'response.',
+                ).toErr();
+              }
+
+              // Honor the same tool-scoping rules the read tools enforce (e.g. get-datasource-metadata):
+              // a data source outside the configured bounded context cannot be tagged or deleted.
+              // Runs before any read/write so a rejected call has zero side effects.
+              const isDatasourceAllowedResult = await resourceAccessChecker.isDatasourceAllowed({
```
anyoung-tableau, line 164: Let's update `resourceAccessChecker.isDatasourceAllowed()` to return an `AllowedResult<Datasource>` (similar to `isWorkbookAllowed`) so you don't necessarily have to call `restApi.datasourcesMethods.queryDatasource` _again_ below. Follow the pattern of the delete-workbook tool.

src/server/oauth/scopes.ts
```diff
@@ -117,6 +121,18 @@ const toolScopeMap: Record<
       'tableau:users:read',
     ]),
   },
+  // Admin-only destructive tool. Preview tags the datasource (datasource_tags:update), resolves the
+  // owner (users:read), and warns about dependent workbooks/flows via the Metadata API (content:read);
+  // confirm deletes it (datasources:delete). adminGate.assertAdmin → GET /users/{id} → users:read.
+  'delete-datasource': {
+    mcp: ['tableau:mcp:datasource:delete'],
+    api: new Set([
```
anyoung-tableau, line 133: I forgot to mention earlier, any tool that passes through the ResourceAccessChecker should include `...RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES` in their Set.

src/tools/web/datasources/deleteDatasource.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  const workbooks = downstream?.workbooks ?? [];
+  const flows = downstream?.flows ?? [];
+  if (workbooks.length === 0 && flows.length === 0) {
+    return 'No workbooks or flows were found that depend on this data source.';
+  }
+
+  const parts: string[] = [];
+  if (workbooks.length > 0) {
+    parts.push(`${workbooks.length} workbook(s): ${workbooks.map((w) => w.name).join(', ')}`);
```
anyoung-tableau: Do we really want to list all the workbooks if there are, say, 10000?

docs/docs/tools/data-qna/delete-datasource.md
```diff
@@ -0,0 +1,99 @@
...
+   argument), reports the data source name, project, and owner, **warns which workbooks and flows
+   depend on it and may break**, returns a `confirmationToken`, and does **not** delete anything.
+2. **Delete** (`confirm: true` + `confirmationToken`): permanently removes the data source. The
+   token from the preview step is **required** — deletion is rejected without a matching token,
+   which guarantees the preview was run first for this data source. On Tableau Cloud the data source
+   is moved to the [recycle bin](https://help.tableau.com/current/pro/desktop/en-us/recycle_bin.htm)
+   and can be restored for a limited time before permanent removal; on Tableau Server there is no
+   recycle bin and deletion is permanent.
+
+:::warning Human confirmation required
```
anyoung-tableau: I think this syntax is not correct. It should be:

```
:::warning

Human confirmation required
...

:::
```

Notice how the warning bubble on this doc page isn't rendering? https://tableau.github.io/tableau-mcp/docs/tools/users/list-users

anyoung-tableau: Also, while you are there, please add a `_category_.json` file under docs/docs/tools/users so "users" in the left nav will get capitalized. Example for Views: https://github.com/tableau/tableau-mcp/blob/main/docs/docs/tools/views/_category_.json

src/tools/web/datasources/deleteDatasource.ts
Alon-ST-DATA: Reuse the access-checked data source instead of re-querying it.
   resourceAccessChecker.isDatasourceAllowed now returns
   AllowedResult<DataSource> (mirrors isWorkbookAllowed); the tool reuses
   that content and only falls back to queryDatasource when no project/tag
   scope forced a fetch.

src/server/oauth/scopes.ts
Alon-ST-DATA: Added the resource-access-checker scopes to both delete tools. Any tool
   that passes through resourceAccessChecker needs
   RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES; delete-datasource and
   delete-workbook were missing tableau:mcp_site_settings:read. Swap the
   bare content:read for ...RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES.

src/tools/web/datasources/deleteDatasource.ts
Alon-ST-DATA: Cap the dependent-content list in the preview warning. List the first
   10 workbook/flow names then "...and N more"; the total count is always
   reported so nothing is silently hidden.
   If all associated workbooks needed in report, user can ask for it. // TODO I will test this case.

docs/docs/tools/data-qna/delete-datasource.md
Alon-ST-DATA: Good point, thanks. 
Fixed the docs admonitions. Docusaurus 3 needs the bracket title form
   :::warning[Title], not :::warning Title (the space form rendered no
   bubble). Fixed delete-datasource.md and list-users.md, and added
   docs/docs/tools/users/_category_.json so "Users" is capitalized in the
   left nav.
 Verified within local docs server. see images below.
image


image

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.13.0",
+  "version": "2.14.0",
```
anyoung-tableau, line 4: Since you are an internal contributor, in the future please create feature branches instead of using a fork. E2E tests don't run in PRs from forked repos since they require access to the project secrets.

anyoung-tableau (approved):

mattcfilbert:
@Alon-ST-DATA — 🤖 **MattGPT** (Matt Filbert's agent)

Adds a `delete-datasource` admin tool that mirrors the existing `delete-workbook` pattern — confirmation-token handshake, admin feature-flag gate, lineage warnings, REST integration. CI is green on both Node versions, and the manual e2e validation in the description is thorough.

**Two things worth a reviewer's eye:**

- The description mentions a "prior PR402 run" — if this is a resubmission, a one-line note on what changed since the last iteration would help reviewers focus.
- The raw 404 on malformed datasource LUIDs is flagged as carry-forward (consistent with `delete-workbook`). Fine to defer — just worth confirming that's intended parity rather than a gap to close here.

_See the inline note on the confirmation-token helper._

package.json
Alon-ST-DATA: Noted.

src/tools/web/datasources/deleteDatasource.ts
```diff
@@ -0,0 +1,313 @@
...
+export const DEFAULT_PENDING_DELETION_TAG = 'pending-deletion';
+
+/**
+ * Deterministic confirmation token derived from the site + datasource. The preview phase returns
+ * it; the delete phase requires it. Because the value is only obtainable by running the preview,
+ * this forces a genuine two-step (preview → confirm) flow and prevents a blind single-call delete.
+ * Stateless by design (no server-side nonce store) so it works across instances and restarts.
+ */
+export function computeConfirmationToken(siteId: string, datasourceId: string): string {
+  return createHash('sha256').update(`${siteId}:${datasourceId}`).digest('hex').slice(0, 12);
```
mattcfilbert, line 47: 🤖 MattGPT (Matt Filbert's agent): the docstring says this token is "only obtainable by running the preview," but it's a pure `sha256(siteId:datasourceId)[:12]` — both inputs are known to any caller (datasourceId is the tool arg; siteId is the connected site). So a client can compute the token without ever running the preview, which means it enforces *token correctness*, not *that a preview actually ran*. If the intent is to guarantee a preview/tag step happened, that needs server-side state (e.g. gating on the pending-deletion tag you set in preview) rather than a deterministic hash. If the real intent is just "make a blind one-shot delete inconvenient / force an explicit second call," that's reasonable — but consider softening the docstring so it doesn't over-claim a safety property. Not blocking; flagging since it's on a destructive path.

Alon-ST-DATA:
> @Alon-ST-DATA — 🤖 Posted by MattGPT (Matt Filbert's agent).
> 
> Adds a `delete-datasource` admin tool mirroring the existing `delete-workbook` pattern (confirmation-token handshake, admin-feature-flag gate, lineage warnings, REST integration). CI is green on both Node versions; the manual e2e validation in the description is thorough.
> 
> Two things worth a reviewer's eye:
> 
>     * The description references a "prior PR402 run" — if this is a resubmission, a one-line note on what changed since the last iteration would help reviewers focus.
> 
>     * The raw 404 on malformed datasource LUIDs is flagged as carry-forward (consistent with delete-workbook) — fine to defer, but worth confirming that's the intended parity rather than a gap to close here.
> 
> 
> (See the inline note on the confirmation-token helper.)

 Thanks Matt.
  1. Yes, resubmission-ish — same branch, force-updated. Changes since the last run: 4 fixes from @anyoung-tableau (reuse access-checked datasource, add resource-access-checker scopes to both delete tools, cap dependent-content list at 10,
 fix Docusaurus admonition syntax + add Users nav category) and 2 from @Akash-Rastogi (scope-pin test, tag .max(200)). Will add a one-line "changed since" note to the description.
  2. Intentional parity. validateDatasourceLuid only enforces non-empty; no UUID-format check — same as delete-workbook and the other datasource tools, so a malformed LUID 404s rather than failing fast. No side effect on the 404 path.
 Deferring as a cross-tool follow-up (add format validation to the shared validator + apply to all workbook/datasource tools) rather than diverging one tool here.
I will file a follow up WI for that change.

src/tools/web/datasources/deleteDatasource.ts
Alon-ST-DATA: Good point. Reworded the JSDoc, the confirmationToken arg description, the tool
  description, the inline gate comment, and the docs pages to state the
  honest property. Applied to both delete-datasource and delete-workbook
  for parity. No behavior change — comments/descriptions only.

PR #409, opened 2026-06-22
@W-22973057 Support multiple resource URIs

stephendeoca:

the pull request._

# Pull Request Template

## Description

BREAKING CHANGE: Updates OAUTH_GLOBAL_RESOURCE_URI to support multiple resource URIs.

## Motivation and Context

Audience in a JWT token may come from different environments deployed for global URL.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [X] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit tests

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.17.0",
+  "version": "2.18.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/config.ts
+++ b/src/config.ts
@@ -49,7 +49,7 @@ export class Config extends BaseConfig {
     issuer: string;
     redirectUri: string;
     resourceUri: string;
-    globalResourceUri: string;
+    globalResourceUris: string[];
     lockSite: boolean;
     jwePrivateKey: string;
     jwePrivateKeyPath: string;
@@ -115,7 +115,7 @@ export class Config extends BaseConfig {
       OAUTH_JWE_PRIVATE_KEY_PATH: oauthJwePrivateKeyPath,
       OAUTH_JWE_PRIVATE_KEY_PASSPHRASE: oauthJwePrivateKeyPassphrase,
       OAUTH_RESOURCE_URI: oauthResourceUri,
-      OAUTH_GLOBAL_RESOURCE_URI: oauthGlobalResourceUri,
+      OAUTH_GLOBAL_RESOURCE_URIS: oauthGlobalResourceUris,
       OAUTH_REDIRECT_URI: redirectUri,
       OAUTH_CLIENT_ID_SECRET_PAIRS: oauthClientIdSecretPairs,
       OAUTH_CIMD_DNS_SERVERS: dnsServers,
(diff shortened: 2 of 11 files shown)


types/process-env.d.ts
```diff
@@ -61,7 +61,7 @@ export interface ProcessEnvWeb extends ProcessEnvBase {
   ADVERTISE_API_SCOPES: string | undefined;
   OAUTH_REDIRECT_URI: string | undefined;
   OAUTH_RESOURCE_URI: string | undefined;
-  OAUTH_GLOBAL_RESOURCE_URI: string | undefined;
+  OAUTH_GLOBAL_RESOURCE_URIS: string | undefined;
```
stephendeoca, line 64: Changing an ENVIRONMENT variable name results in a breaking change as anyone using the old environment variable will no longer work, but this environment variable has only been out for 2 weeks and realistically we would be the only consumers of it.

anyoung-tableau: Agreed

mattcfilbert:
Clean, focused breaking change — the trailing-slash canonicalization is well-justified (and nicely documented in the code), the comma-split parsing is forgiving in the right way, and the rename is propagated through config/types/manifest/docs with tests covering the multi-aud and both slash directions. One question before merge:

**`accessTokenValidator.ts` — the audience-mismatch log was bumped `debug` → `error`.** An `aud` mismatch is a client/token problem (a token minted for a different deployment), not a server fault — and the code comment itself notes these share an SSO issuer, so tokens from sibling deployments will hit the wrong pod routinely. At `error` that becomes log/alert noise an operator can't act on. Was the bump intentional to surface misconfig during this rollout? If so, `warn` might capture "notable but expected + client-correctable" without tripping error-rate alarms; otherwise `debug` matches the "expected rejection" framing. Your call — just flagging the operational cost.

Everything else looks good to me. 🤖 MattGPT (automated review on Matt's behalf)

mattcfilbert (approved):
lgtm

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.17.0",
```
jarhun88, line 4: i think breaking change should be major bump?

src/config.ts
```diff
@@ -196,7 +196,12 @@ export class Config extends BaseConfig {
       embeddedAuthzServer,
       issuer: oauthIssuer ?? '',
       resourceUri: oauthResourceUri ?? `http://127.0.0.1:${this.httpPort}`,
-      globalResourceUri: oauthGlobalResourceUri ?? '',
+      globalResourceUris: oauthGlobalResourceUris
```
jarhun88, line 199: do we care to verify if the uris are actually uris? or leave up to the user's disgression

yyk16293 (approved):

package.json
yyk16293: This is a new field which we just added 2 weeks back. We should be good here.

src/config.ts
anyoung-tableau: We haven't been super strict about config validation since these values are coming from the admin running TMCP and we "trust" them.

yyk16293: We are planning to mainly use this on hyperforce side.

jarhun88 (approved):

PR #411, opened 2026-06-23
@W-23050560: Add extract-optimization-apply MCP prompt

Akash-Rastogi:
## Summary

Adds a new admin-only MCP prompt `extract-optimization-apply` that orchestrates the recommend-then-apply workflow for extract refresh schedule optimization on Tableau Cloud. It pairs with the already-merged inform-side prompt and the four underlying tools, and adds a **two-phase preview/confirm contract** to the two write tools it drives.

- **Inform-side reads:** `list-extract-refresh-tasks`, `query-admin-insights-job-performance`
- **Apply-side writes:** `update-cloud-extract-refresh-task` (downgrade), `delete-extract-refresh-task` (delete)
- **Defaults to `dryRun=true`** so the report is produced with zero writes.
- **Step 4 HITL gate** uses the shared `renderHitlGate` primitive from `src/prompts/_lib/confirm.ts` (introduced in #405) so the wording stays consistent across destructive admin prompts.
- **Two-phase contract on both write tools.** Preview validates + returns a `confirmationToken`; confirmed apply requires a matching token before any Tableau write. For the update tool the token is bound to `(siteId, taskId, schedule)` — a preview of schedule A cannot be applied as schedule B.
- **Step 5 apply leg** is gated on explicit per-turn human approval, refuses to parallelize, and stops on first failure (no partial fan-out).

Reconciles with #414 (server-authoritative tag gate for `delete-workbook` / `delete-datasource`) by widening the shared `renderConfirmInstructions` helper with a `gateKind: 'tag' | 'token'` discriminant. Extract refresh tasks have no Tableau REST tag affordance so this prompt uses the friction-token contract; stale-content continues to use the tag contract, byte-identical to `main`.

Files:
- New: `src/prompts/extractOptimization/apply.ts` (prompt factory) + `apply.test.ts`
- New: `src/prompts/_lib/jobPerformance.ts` (shared with `job-optimization-inform`)
- New: `docs/docs/prompts/extract-optimization-apply.md`
- New: `tests/e2e/prompts/extractOptimization.test.ts`
- Modified: `src/prompts/_lib/confirm.ts` (+ tests) — `gateKind: 'tag' | 'token'` discriminant
- Modified: `src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts` (+ tests) — two-phase contract, schedule-bound token
- Modified: `src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts` (+ tests) — two-phase contract
- Modified: `docs/docs/intro.md` (Prompt List row), `docs/docs/tools/tasks/delete-extract-refresh-task.md`, `docs/docs/tools/tasks/update-cloud-extract-refresh-task.md`
- Modified: `tests/e2e/extractRefreshTasks/*.test.ts` — preview + mismatched-token coverage; disposable-id opt-in test now runs the full two-phase flow
- Modified: `src/prompts/index.ts` (register the factory), `package.json` / `package-lock.json` (2.21.1 → 2.22.0)

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm run lint` clean
- [x] `npm test` — 1996 unit tests pass
- [x] `npm run build` succeeds
- [ ] Eval (`extract-optimization-apply.eval.ts`) — separate work item (1B-4), not in this PR


src/prompts/extractOptimization/apply.ts
```diff
@@ -0,0 +1,190 @@
...
+        'Tableau Cloud caps lookback at 90 (365 with Advanced Management).',
+    ),
+  taskIds: z
+    .string()
+    .optional()
+    .describe(
+      'Optional comma-separated list of extract refresh task IDs (UUIDs) to scope the run to. ' +
+        'When omitted, the workflow analyzes every task returned by ' +
+        `\`${INVENTORY_TOOL}\`.`,
+    ),
```
Akash-Rastogi, line 36: [Akash's agent] `taskIds` is interpolated into the prompt text on line 104 (each id wrapped in backticks) but has no character-class constraint — `staleContent/apply.ts` adds a regex on its `tag` field as defense-in-depth for exactly this reason. Mirror that defense here.
```suggestion
  taskIds: z
    .string()
    .regex(
      /^[A-Za-z0-9, -]+$/,
      'taskIds must contain only letters, numbers, commas, spaces, and dashes',
    )
    .optional()
    .describe(
      'Optional comma-separated list of extract refresh task IDs (UUIDs) to scope the run to. ' +
        'When omitted, the workflow analyzes every task returned by ' +
        `\`${INVENTORY_TOOL}\`.`,
    ),
```

src/prompts/extractOptimization/apply.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    expect(text).toContain('Do **not** call `update-cloud-extract-refresh-task`');
+    expect(text).toContain('Dry run — no changes applied.');
+  });
+
+  it('allows the apply step only after Step 4 confirmation when dryRun = false', async () => {
+    const text = await textOf({ dryRun: 'false' });
+    expect(text).toContain('`dryRun = false`');
+    expect(text).toContain('only after** the human confirms in Step 4');
+    expect(text).toContain('Step 5 — Apply (only after Step 4 approval).');
+    expect(text).not.toContain('Dry run — no changes applied.');
```
Akash-Rastogi: [Akash's agent] This test only checks the Step 5 header — the safety-critical body ("Do **not** parallelize", "stop immediately" on first failure) is unverified, so a future refactor that drops those guards from the prompt wouldn't fail this test.
```suggestion
    expect(text).toContain('Step 5 — Apply (only after Step 4 approval).');
    expect(text).toContain('Do **not** parallelize');
    expect(text).toContain('stop immediately');
    expect(text).not.toContain('Dry run — no changes applied.');
```

src/prompts/extractOptimization/apply.ts
Akash-Rastogi: [Akash's agent] Fixed in commit `5f1ebf87` — added the `[A-Za-z0-9, -]+` regex.

src/prompts/extractOptimization/apply.test.ts
Akash-Rastogi: [Akash's agent] Fixed in commit `5f1ebf87` — added `Do **not** parallelize` and `stop immediately` assertions.

Alon-ST-DATA (commented):
Please add:
 - docs 
 - unit tests 
 - e2e tests

Akash-Rastogi:
[Akash's agent] Addressed in commit `c570cfce`:

- **Docs**: added `docs/docs/prompts/extract-optimization-apply.md`, mirroring `stale-content-cleanup-apply.md`.
- **E2E tests**: added `tests/e2e/prompts/extractOptimization.test.ts` (9 tests), structurally mirroring `staleContent.test.ts` and `jobOptimization.test.ts`.
- **Unit tests**: existing `src/prompts/extractOptimization/apply.test.ts` had 11 tests; expanded to 13 with explicit assertions on the DESTRUCTIVE framing and closing `Fixed notes` safety block.

Also bundled in the same commit while in the file (sibling-alignment pass):
- Extracted shared `JOB_PERFORMANCE_FIELDS` + `EXTRACT_REFRESH_JOB_TYPES` constants into `src/prompts/_lib/jobPerformance.ts` (was duplicated between `jobOptimization/inform.ts` and `extractOptimization/apply.ts`).
- Refactored prompt body from mutable `lines.push` to a single `array.join` literal so it matches `staleContent/apply.ts` shape.
- Added a closing `**Fixed notes**` safety block summarizing the irreversibility guarantee.
- Fixed a pre-existing step-numbering jump (Step 4 → Step 6) in dry-run mode by renumbering the Final report block to Step 5 when `dryRun = true`.

Akash-Rastogi:
[Akash's agent] Fixed in commit `8236c913`:

- **e2e CI fix**: the dry-run `Mode:` line emits one compound sentence (``Do **not** call `<update>` or `<delete>` ``), so the standalone-DELETE_TOOL substring assertion at `tests/e2e/prompts/extractOptimization.test.ts:69` was unsatisfiable and broke CI on `c570cfce`. Collapsed the two `Do **not** call \`<X>\`` asserts into one that matches the actual joined phrase.
- **PR #405 alignment-audit gaps** (the three from the earlier audit comparison):
  - `src/prompts/_lib/jobPerformance.test.ts` (new): 4 tests over `JOB_PERFORMANCE_FIELDS` and `EXTRACT_REFRESH_JOB_TYPES` — exact shape, no dupes, no empties.
  - `apply.test.ts`: HITL-before-apply ordering invariant + `taskIds`-dedup test.
  - `apply.ts`: actually de-duplicate `taskIds` (`new Set<string>` + `: string[]` annotation; tsc otherwise widens to `unknown[]` and breaks `.map`).

Local gates green: tsc, eslint, build, 1862/1862 unit tests. e2e couldn't be validated locally (`tests/.env` missing), but CI on `8236c913` is now fully green — both `build (24.x)` and `build (>=22.7.5 <23)` pass.

Alon-ST-DATA:
**Non-blocking observation** — nice clean refactor; `jobPerformance.ts` extraction is byte-identical, dryRun defaults true, HITL gate sits before any write, and apply is sequential stop-on-error. No correctness concerns.

One thing worth a glance, not a blocker: `extract-optimization-apply` has no explicit large-batch guard analogous to `LARGE_REPORT_THRESHOLD` (=100) in the stale-content apply path (#405). Low impact here — extract-task counts are far smaller than stale-content's scale and the existing defenses (dryRun default, HITL gate, taskId regex, stop-on-error) are strong — so it may not be worth adding. Flagging only for symmetry with the sibling prompt.

Alon-ST-DATA:
## E2E Test Report — `extract-optimization-apply` prompt
_Posted by Alon's Agent_

Ran full e2e on this PR @ `8236c913` against a local MCP server (HTTP) bound to the 10ax admin-insights site, verifying every result against independent REST / VDS ground truth (never the tool's own counts).

**Verdict: all in-scope cases PASS. No defects found.**

| Layer | Result |
|---|---|
| Unit tests | 29 passed (`apply` 15, `jobPerformance` 4, `inform` 10) |
| Prompt render matrix | dryRun gating, `taskIds` dedup + scope, `lookbackDays` DATE/LASTN filter, Missing-tasks section — all correct |
| Schema guards | injection (`"`, backtick, newline) rejected; `lookbackDays` / `dryRun` enums enforced |
| **Step 1 live** | `list-extract-refresh-tasks` = **628** = REST `/tasks/extractRefreshes` 628, all unique ✅ |
| **Step 2 live** | `query-admin-insights-job-performance` (default prompt JSON) = 5585 rows, **0 rows outside the 4 extract job types** (SET filter held exactly); VDS-direct COUNT = 5586 ✅ |
| Apply tool | `update-cloud-extract-refresh-task` 5-min-boundary guard rejects malformed schedule pre-API ✅ |

Step-2 off-by-1 (5585 vs 5586) = live jobs landing between the two calls on an actively-running site — not a tool defect. Per-type counts otherwise match VDS-direct exactly.

**What was verified:**
- dryRun defaults true; Apply step omitted unless `dryRun=false`; HITL break + "previous approval does NOT carry forward" present in every mode; sequential / stop-on-error apply instructions present.
- `taskIds` regex closes the prompt-injection vector at the schema boundary (mirrors the `tag` defense in `stale-content-cleanup-apply`).
- Shared `_lib/jobPerformance.ts` constants drove a live VDS query whose filter held exactly (0 rows leaked) and whose counts match VDS-direct.

**Scope note:** real PUT/DELETE Apply was intentionally **not** exercised — that path is model-driven from the prompt text (fully verified by rendering), and the underlying mutating tools were e2e'd separately (update: #404, delete: #392). #411 adds no code on the mutation path beyond the orchestration text.

### Next actions
Worth looking still - no explicit large-batch guard 
- **None blocking** — safe to merge from an e2e standpoint.

Akash-Rastogi:
[Akash's agent] Rebased onto current `main` (`fd4fbb91`). HEAD is now `a0e337dd`; only conflict was the `version` field in `package.json` / `package-lock.json` — kept the higher branch value (2.20.0). Local gates green: tsc, eslint, build, 1866/1866 unit tests.

On the two comments above:

- **E2E Test Report** (Alon's Agent): thanks — verdict noted, no changes needed.
- **Large-batch guard symmetry** (non-blocking observation): leaving this out for now. Reasoning matches what you noted: extract-task counts are bounded (the live site is at ~628 tasks, well below stale-content's scale), and the existing defenses (`dryRun` default + HITL gate + `taskIds` regex + sequential stop-on-error + the explicit per-batch approval text) already gate every write. A `LARGE_REPORT_THRESHOLD`-style guard would add a numeric ceiling on top of those, but the HITL gate already forces explicit per-batch approval and the user can scope via `taskIds` if they want a smaller batch — so it'd be a second guardrail on a path that's already paved. Happy to add it as a follow-up if you'd rather have the symmetry.

Alon-ST-DATA:
**Re: HITL enforcement parity with the staleContent apply prompt**

The staleContent `apply.ts` imports both `renderHitlGate` and `renderConfirmInstructions`, but this prompt imports only `renderHitlGate`. Flagging the resulting difference in enforcement strength, since this is a destructive admin workflow.

`renderConfirmInstructions` is only meaningful for a **two-phase tool** — one that returns a per-item `confirmationToken` from a preview phase and requires a second call with `confirm: true` echoing that exact token. The staleContent tools support that handshake, so it gets two layers: the prompt-level approval gate *plus* a token the model can't fabricate (it has to come from the preview output).

The tools here are one-phase:

```ts
const UPDATE_TOOL = 'update-cloud-extract-refresh-task';  // { taskId, schedule }
const DELETE_TOOL = 'delete-extract-refresh-task';        // { taskId }
```

Neither accepts `confirm` / `confirmationToken`, so `renderConfirmInstructions` doesn't apply and we're left with **only the prompt-text gate**. If the model ignores the Step 4 gate wording, there's nothing at the tool layer to stop the destructive call — unlike staleContent.

That's an inherent gap (these tools weren't built two-phase), not a bug in this PR. Two questions:

1. Is single-layer (prompt-gate only) acceptable for this workflow, given it can update/delete refresh tasks?
2. If parity is wanted, should we add a `confirm`/`confirmationToken` parameter to `update-cloud-extract-refresh-task` / `delete-extract-refresh-task` as a follow-up, then wire in `renderConfirmInstructions`?

`dryRun` defaulting to `true` is a good backstop in the meantime.

Akash-Rastogi:
[Akash's agent] Both tools are now two-phase — parity achieved on the tip (`76ba8099`, force-pushed after rebase onto current `main`).

**What changed**

- `update-cloud-extract-refresh-task` and `delete-extract-refresh-task` each accept `confirm: boolean` (default `false`) and `confirmationToken: string` (required when `confirm: true`). Preview (`confirm` omitted or `false`) validates inputs and returns a per-task `confirmationToken`; confirmed call re-derives the token from server-side inputs and rejects mismatched/missing tokens before any write. Token derivation is `sha256(`${siteId}:${taskId}`).slice(0, 12)` — the caller can't compute it because `siteId` comes from the authenticated session, not from the tool args.
- The prompt now imports both `renderHitlGate` and `renderConfirmInstructions`, matching the staleContent shape.

**Reconciliation with #414**

#414 landed on `main` in the meantime and rewrote `renderConfirmInstructions` for the server-authoritative pending-deletion-tag contract (`delete-workbook` / `delete-datasource`). Extract refresh tasks have no Tableau REST tag affordance, so the shared helper now takes a `gateKind: 'tag' | 'token'` discriminant:

- `'tag'` (default): unchanged; stale-content output byte-identical to `main`.
- `'token'`: new; used by this prompt. `_lib/confirm.ts` documents both contracts.

Full doc block on `renderConfirmInstructions` explains when to pick which. `dryRun` still defaults to `true` as an extra backstop.

Alon-ST-DATA:
## tmcp review — extract-optimization-apply

Adds an MCP **prompt** + retrofits a two-phase `confirm`/`confirmationToken` contract onto two already-shipped destructive tools. **Verdict: needs-changes.** Prompt logic, two-phase code, and unit tests are well-built — gaps are docs parity + the intro registry, not correctness.

### Findings (most severe first)

**🔴 BLOCKER**
- `docs/docs/intro.md` (Prompt List, ~L117) — table stops at `job-optimization-inform`; the new `extract-optimization-apply` prompt has no row/link-ref. This is the item reviewers always block on. Add a row (admin-only, destructive, applies schedule downgrades/deletions after a required human-confirmation break).

**🟠 MAJOR — doc↔schema parity broken (default behavior changed: delete → preview)**
- `docs/docs/tools/tasks/delete-extract-refresh-task.md` (~L50) — new `confirm`/`confirmationToken` params (schema `deleteExtractRefreshTask.ts:6-19`) and the preview-first default are absent. Arguments table lists only `taskId`; Response still says "successfully deleted." A caller sending `{taskId}` now gets a **preview, not a delete**. Add the params + document the preview→confirm flow.
- `docs/docs/tools/tasks/update-cloud-extract-refresh-task.md` (~L48-51, L108) — same gap: new params + preview-first default missing from Arguments table and Response section. Document the two-phase contract + token requirement.

**🟡 MINOR**
- `src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts:25` — `computeConfirmationToken(siteId, taskId)` does not bind the schedule payload. A caller can preview schedule A (get token), then apply schedule B with the same token — the approved preview need not equal the applied change. Delete has no payload so it's fine there. Consider folding a schedule hash into the token. (Severity limited: admin-gated, and the token is documented as friction-only, not proof-of-approval.)
- `tests/e2e/extractRefreshTasks/` — the two tools' new two-phase behavior has solid unit coverage (success/preview/missing-token/mismatched-token/admin-gate) but no e2e exercising preview→confirm live. Only the prompt got an e2e. Match sibling e2e coverage or note why deferred.

**⚪ NIT**
- PR description says the docs page is a separate WI (not in this PR) and cites a `2.18.0 → 2.19.0` bump, but the PR **includes** the docs page and bumps `2.21.1 → 2.22.0` (`package.json`). Stale pre-rebase description — reconcile so scope isn't misread.

### Clean
- Version bump correct (minor, new prompt).
- Gating tight: build-time `disabled: !config.adminToolsEnabled` + runtime `assertAdmin` before any token/write logic. No scope leak.
- Params all `.describe()`'d + regex-validated (prompt-injection hardening on `taskIds`); no dead params.
- No ISO-date/UTC trap — perf read uses relative `LASTN`/`DAYS`.
- Shared-constant refactor (`_lib/jobPerformance.ts`) dedups the inform prompt's copy; good reuse.
- `idempotentHint` correctly flipped to `false` on the update tool to match the non-idempotent two-phase contract.
- HITL honesty: `confirm.ts` docstring is candid that the token gate is friction, not proof of human approval (real elicitation tracked under W-23125362).

### Merge path
1 blocker: intro.md Prompt List row. Plus the 2 MAJOR doc updates — ship-blockers in practice since the tool defaults changed under callers.

Akash-Rastogi:
[Akash's agent] Addressed all 6 findings in commit `1e4c9539`.

### BLOCKER — `docs/docs/intro.md` Prompt List
- Added a new row for `extract-optimization-apply` after `job-optimization-inform` (~L118): "Admin-only. Destructive. Applies schedule downgrades and deletions to extract refresh tasks after a required human-confirmation break; defaults to a dry-run report."

### MAJOR — doc↔schema parity on both write tools
- **`docs/docs/tools/tasks/delete-extract-refresh-task.md`**: added a `## Two-phase contract` section, a `:::warning Human confirmation required — advisory, not enforced` callout, `confirm`/`confirmationToken` rows in the Arguments table, and split the Response section into Preview vs Delete variants with a mismatched-token error example.
- **`docs/docs/tools/tasks/update-cloud-extract-refresh-task.md`**: same treatment (Two-phase contract section, HITL callout, Arguments rows, split Response section). Also documents that the token binds to the `(taskId, schedule)` pair — see MINOR #1.

### MINOR — schedule not bound into the update token
- `src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts`: `computeConfirmationToken` now takes an optional third `schedule` arg and folds a canonical `stableStringify(schedule)` into the hash: `sha256('${siteId}:${taskId}:${stableStringify(schedule)}').slice(0,12)`. `stableStringify` sorts object keys recursively so `{a,b}` and `{b,a}` hash identically. Preview-A/apply-B is now rejected: token mismatch → `ArgsValidationError` before any Tableau write. Delete continues to call with 2 args (schedule omitted) → token reduces to `sha256('${siteId}:${taskId}')`, byte-identical to prior behavior.
- Unit tests added in `updateCloudExtractRefreshTask.test.ts` under `describe('two-phase contract', ...)`:
  - `'binds the schedule payload into the token so preview A and apply B do not match'` — Weekly-A vs Daily-B, asserts `tokenA !== tokenB`, asserts apply-B-with-tokenA rejected with "taskId + schedule pair", asserts mock update never called.
  - `'hashes semantically equal schedules the same regardless of object key order'` — asserts canonical vs permuted-key stringification produce equal tokens.
- Tool `description` (section 2, Apply) + error message updated to mention "editing the schedule after previewing invalidates the token — re-preview to get a fresh one."

### MINOR — e2e coverage for preview→confirm on the write tools
- `tests/e2e/extractRefreshTasks/deleteExtractRefreshTask.test.ts`: added `'should return a preview with confirmationToken when confirm is omitted'` (no-side-effect preview leg) and `'should reject a confirmed call with a mismatched confirmationToken'` (asserts error contains "confirmationToken returned by the preview step"). Rewrote the disposable-id opt-in test to run the full two-phase flow: preview → regex-extract token (`/confirmationToken:\s*([a-f0-9]{12})/`) → confirm with the extracted token (rather than recomputing — keeps the client honest).
- `tests/e2e/extractRefreshTasks/updateCloudExtractRefreshTask.test.ts`: mirror treatment — preview test + mismatched-token test (asserts "taskId + schedule pair") + rewrite of the disposable-id opt-in test to run the full two-phase flow.

### NIT — PR description stale post-rebase
- Rewrote the PR body to reflect the actual scope: added "Two-phase contract on both write tools" bullet emphasizing schedule binding, updated the file list to include the new docs and tests, updated the version delta to `2.21.1 → 2.22.0`, updated the test count to 1996 unit tests, removed the "docs are separate WI" line. Live on the PR now.

### Validation before commit
- `npx tsc --noEmit` clean
- `npm run lint` clean (after `lint:fix` — the multi-line `computeConfirmationToken` call in one new test was auto-collapsed)
- `npm test` — 1996/1996 unit tests pass
- `npm run build` succeeds
- Code-review agent run on the diff — returned zero findings.

Ready for another look.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.test.ts
```diff
@@ -411,13 +418,165 @@ describe('updateCloudExtractRefreshTaskTool', () => {
...
+        taskId: validTaskId,
+        schedule: validSchedule,
+        confirm: false,
+      });
+      expect(result.isError).toBe(false);
+      invariant(result.content[0].type === 'text');
+      expect(result.content[0].text).toContain('Preview');
+      expect(result.content[0].text).toContain(validTaskId);
+      expect(result.content[0].text).toContain('Weekly');
+      // Token is the deterministic sha256(siteId:taskId)[0..12].
```
Akash-Rastogi: [Akash's agent] Stale comment — `validToken` on line 64 now folds the schedule into the hash (`computeConfirmationToken('test-site-id', validTaskId, validSchedule)`), so the derivation is no longer `sha256(siteId:taskId)[0..12]`.
```suggestion
      // Token is the deterministic sha256(siteId:taskId:stableStringify(schedule))[0..12].
```

Alon-ST-DATA (approved):

PR #416, opened 2026-06-24
@W-23149256 prerelease-aware npm + Docker publish (hotfix without moving latest) + release runbook

mattcfilbert:
## W-23149256 — Formalize the hotfix release path

Today a livesite fix can only ship by pushing to `main` and upgrading Hyperforce to that version — which drags in any unrelated changes that landed on `main` since, and there's no way to ship a surgical fix off the exact deployed version without moving the npm `latest` dist-tag. This formalizes a hotfix path. Andy's spec; continuity-critical (he holds the release process and departs 2026-07-02 — landing this before a livesite issue forces it, and before he's gone).

### What changed
- **`.github/workflows/publish.yml`** — prerelease-aware `npm publish`. A version carrying a SemVer prerelease identifier (a `-`, e.g. `2.18.0-hotfix.1`) publishes under a dedicated **`hotfix-v<version>`** dist-tag and never moves `latest`; the stable path is unchanged (bare `npm publish`). It **fails closed** if the version can't be read (refuses to risk moving `latest`). The dist-tag is deliberately not the bare version — npm rejects a dist-tag that parses as a valid SemVer version.
- **`.github/workflows/docker-publish.yml`** — Docker `latest` now skips prerelease tags (mirrors the npm guard), so a hotfix gets its exact-version Docker tag without stealing `latest`.
- **`docs/RELEASE.md`** (new) — the normal + hotfix runbook (branch off the deployed tag → fix on `main` then cherry-pick → bump to `-hotfix.N` → release from the hotfix branch → pin the exact version in Hyperforce) + the rationale. Linked from CONTRIBUTING.

### Verification
Stable path provably unchanged (bare `npm publish`, `upload-mcpb` untouched); prerelease/stable/empty-version cases table-tested; the chosen dist-tag confirmed non-SemVer (npm-acceptable). This is a CI/process change — no `src/` touched, so `scripts/agent-check` isn't the relevant gate.

A pre-push dual-model self-review (GPT-5.5 + Opus) on the local diff caught and fixed three blockers before the first commit: npm rejecting a valid-SemVer dist-tag (would have failed every hotfix publish), Docker `latest` moving on hotfixes, and a fail-open on an unreadable version.

### One thing worth a maintainer eye
The dist-tag scheme is `hotfix-v<version>` (a hotfix never moves `latest`; consumers pin the exact version). If whoever owns the Hyperforce consume step prefers a single reusable `hotfix` pointer instead, that's a one-line change — flagging since it's a release-ergonomics call.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -19,6 +19,24 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Preflight — release tag must match package.json version
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        # The `latest` guard below keys off the release TAG NAME while the npm
+        # side (publish.yml) keys off package.json — this preflight (same check
+        # in both workflows) makes a mismatch fail loudly instead of letting the
+        # two registries disagree about moving `latest`.
+        run: |
+          VERSION="$(node -p "require('./package.json').version")"
+          if [[ -z "$VERSION" || "$VERSION" == "undefined" ]]; then
+            echo "::error::Could not read version from package.json — aborting"
+            exit 1
+          fi
+          if [[ "$RELEASE_TAG" != "v$VERSION" ]]; then
+            echo "::error::Release tag '$RELEASE_TAG' != 'v$VERSION' (package.json) — aborting. Tag the exact commit whose package.json matches, or fix the release tag."
+            exit 1
+          fi
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -38,7 +38,10 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=raw,value=latest,enable=${{ github.event_name == 'release' }}
+            # Move `latest` only for STABLE releases — a prerelease/hotfix tag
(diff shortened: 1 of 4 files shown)


jarhun88:
can you walk me thru the workflow on how you would release a specific release? im not quite sure im following this

mattcfilbert:
Good followup, @jarhun88 

Say Hyperforce is pinned to 2.18.0 and we have a livesite bug. Today the only path is shipping main, which drags along whatever else landed. With this PR:
  
1. Branch from the deployed tag: `git checkout -b hotfix/2.18.0 v2.18.0`
2. Land the actual fix on main as normal, then cherry-pick it onto the hotfix branch (so main never misses the fix)
3. On the hotfix branch, bump `package.json `to `2.18.0-hotfix.1` and push
4. Tag that commit `v2.18.0-hotfix.1`, push the tag, then create and publish a GitHub Release from it (mark it prerelease). The Release publish is what triggers the workflows and the tag alone does nothing.
5. publish.yml sees the - in the version and runs `npm publish --tag hotfix-v2.18.0-hotfix.1` _entonces_ the package exists on npm at that exact version but latest never moves. The Docker workflow does the same dance: exact-version image, no latest.
6. Hyperforce upgrades to `@tableau/mcp-server@2.18.0-hotfix.1` which is an exact-version pin, dist-tag never matters to the consumer.

A normal release is unchanged: merge to main with the version bump, tag gets cut, publish the Release, bare npm publish moves latest.
The guard rails: if the version can't be read from package.json the job aborts rather than risk a bare publish moving latest, and the dist-tag is `hotfix-v<version>` because npm rejects dist-tags that parse as versions.

jarhun88 (approved):

PR #421, opened 2026-06-24
@W-23146703 Add open-in-Tableau link to embedded MCP app viz

jarhun88:
## Summary

[
Screenshot 2026-06-24 at 4 52 06 PM (2)
Screenshot 2026-06-24 at 4 52 16 PM
](url)
Adds an "Open in Tableau" link to the embedded MCP app viz so users can open the source dashboard in the full Tableau experience.

## Changes

- **New module** `openInTableauLink.ts`: builds the link in code and opens it via the host's `app.openLink()` capability.
- **Graceful degradation**: the link is only rendered when a URL is present and the host advertises the `openLinks` capability.
- **Failure handling**: if the host denies or the open call fails, an inline "The URL was unable to be opened." message is shown (styled in `mcp-app.css`).
- **serverTools guard**: `get-oauth-token` now throws early if the host lacks the `serverTools` capability, so the client's top-level handler can surface an error.
- **Viz sizing fix**: set embedded viz height from `sheetHeight` when `chromeHeight` is 0/absent.
- **Tests**: unit coverage for the link (render, degradation, failure message) and the OAuth serverTools guard.

## Work Items

- W-23146703 (open-in-Tableau link)
- W-23146699 (hide toolbar)


stephendeoca (approved):

src/web/apps/src/mcp-app.css
```diff
@@ -30,3 +30,29 @@ tableau-viz {
...
+
+[hidden] {
+  display: none;
+}
+
+.open-in-tableau {
+  display: inline-block;
+  margin: 16px 0 0 0;
+  padding: 8px 12px;
+  font-family: system-ui, -apple-system, sans-serif;
```
anyoung-tableau, line 42: Please create a ticket to use the Tableau fonts rather than system fonts.

src/web/apps/src/mcp-app.css
```diff
@@ -30,3 +30,29 @@ tableau-viz {
...
+  border-radius: 4px;
+  transition: background-color 0.2s ease;
+}
+
+.open-in-tableau:hover {
+  background-color: #f0f0f0;
+  text-decoration: underline;
+}
+
+.open-in-tableau:focus {
```
anyoung-tableau, line 55: How does the link look after it's been clicked? aka do we need to define a style for `:visited` too?

You can test this by selecting the link in dev tools inspector, then click the `:hov` button on the right side

Image

src/web/apps/src/mcp-app.css
```diff
@@ -30,3 +30,29 @@ tableau-viz {
...
+  font-family: system-ui, -apple-system, sans-serif;
+  font-size: 14px;
+  color: #0066CC;
+  text-decoration: none;
+  border-radius: 4px;
+  transition: background-color 0.2s ease;
+}
+
+.open-in-tableau:hover {
+  background-color: #f0f0f0;
```
anyoung-tableau, line 51: Consider creating a ticket to support when the client requests dark mode.

src/web/apps/src/lib/openInTableauLink.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // Set onclick handler to use host-mediated link opening
+  link.onclick = async (e) => {
+    e.preventDefault();
+
+    try {
+      const result = await app.openLink({ url });
+
+      if (result.isError) {
+        console.warn('Open in Tableau link request denied by host', { url });
```
anyoung-tableau, line 68: For a proper UX, we should probably display a user-friendly error message whenever a link fails to open.

We could use some help from designers, but something like:
"There was a problem opening the link. You can navigate to the viz in your browser by copying and pasting the URL: `<VIZ_URL>`"

src/web/apps/src/mcp-app.css
```diff
@@ -30,3 +30,29 @@ tableau-viz {
...
+[hidden] {
+  display: none;
+}
+
+.open-in-tableau {
+  display: inline-block;
+  margin: 16px 0 0 0;
+  padding: 8px 12px;
+  font-family: system-ui, -apple-system, sans-serif;
+  font-size: 14px;
```
anyoung-tableau: Generally, it's a better idea to use a relative unit of measurement like `rem` instead of `px` for font sizes and margin/padding. That way, the size will respect the user's font size preference in their browser.

src/web/apps/src/lib/openInTableauLink.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export function setupOpenInTableauLink(app: App, url: string, container: HTMLElement): void {
+  // Remove any existing link first (idempotency guard)
+  const existingLink = container.querySelector('#openInTableauLink');
+  if (existingLink) {
+    existingLink.remove();
+  }
+
+  // Don't create link if URL is empty or host lacks openLinks capability
+  const capabilities = app.getHostCapabilities();
```
anyoung-tableau, line 46: I noticed we're not calling `app.getHostCapabilities()` before we call the `get-oauth-token` tool to make sure `capabilities.serverTools` is defined.

What should we do if calling a server tool is not allowed?

1. Still render the viz without a `token`? That would allow folks to click the Sign in to Tableau Cloud button.
2. Show some error?

I'd vote for bullet 1 with potentially a message that says something like "we were unable to authenticate to the viz. You can still sign in to Tableau below"

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: hmm what does this typically look like, some kind of pop up?

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: im not actually sure that the sign in to tableau cloud button will reliably work though, or maybe i need to expose the pop up somehow

src/web/apps/src/mcp-app.css
jarhun88: it does nothing, but im not sure why we would need to define a visted style?

anyoung-tableau: Browsers define their own visited style which can make your links look bad if you don't define your own.

src/web/apps/src/lib/openInTableauLink.ts
anyoung-tableau: In Tableau's design system, yes, we could use the Inline Alert or a Toast: https://tableaudesignsystem.com/71f0857a3/p/970f32-toast

But since we don't access to the widgets at this point, we'd need to figure something else out.

src/web/apps/src/lib/openInTableauLink.ts
anyoung-tableau: Oh that's right, it won't necessarily work because it's a popup. In that case maybe our only fallback is to help the user copy the viz URL to the clipboard.

jarhun88: i could still surface the link to go to the url

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: i might just have to settle with something functional, but not visually the best, and leave astory to improve. I dont want to get too bogged down on epic scope

jarhun88: Screenshot 2026-06-29 at 2 36 05 PM
i implemented this

jarhun88: the url is technically returned by the client as part of the response payload, so they should be able to copy paste it themselves without us specifying it again?

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: heres what im thinking ill do. if the viz render fails for any reason, im going to display an error ui. underneath it ill still have the link to url.

yyk16293 (approved):

src/web/apps/src/lib/openInTableauLink.ts
anyoung-tableau, line 65: In the event the link fails to open once (like maybe the client shows some approval UX they deny, but then they approve it the 2nd time), the error message will still appear from the first time. Consider removing any pre-existing error message if the `openLink()` method succeeds

src/web/apps/src/lib/openInTableauLink.ts
```diff
@@ -1,5 +1,24 @@
...
+  let errorMessage = container.querySelector('.open-in-tableau-error') as HTMLElement;
+
+  if (!errorMessage) {
+    // Create new error message element
+    errorMessage = document.createElement('div');
+    errorMessage.className = 'open-in-tableau-error';
+    container.appendChild(errorMessage);
+  }
+
+  errorMessage.textContent = 'The URL was unable to be opened.';
```
anyoung-tableau, line 17: Since this text is static, it doesn't really make sense to re-set the textContent if the error is already there. Consider moving this into the `if {} block` before appending the child.

anyoung-tableau (approved):
Some non-blocking suggestions

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: good point, i addresesd it

src/web/apps/src/lib/openInTableauLink.ts
jarhun88: good point

src/web/apps/src/mcp-app.css
jarhun88: ive created one

src/web/apps/src/mcp-app.css
jarhun88: done!

PR #423, opened 2026-06-25
Phase 2: embed-token resolver for pat/uat via optional embedding Connected App

jarhun88:
## Phase 2 — embed auth for `pat` / `uat`

Stacked on #422 (Phase 1). **Base branch is `embed-token-resolver-phase1`, not `main`** — review/merge Phase 1 first.

### What this does
Lets the embedded Tableau viz in the MCP Apps UI authenticate under `AUTH=pat` and `AUTH=uat` by signing the embed JWT (`tableau:views:embed` scope) from an **optional, dedicated embedding Connected App** credential. When the credential is not configured, behavior is unchanged (the app falls back to an "Open in Tableau" link).

### Changes
- **config** (`fb23978`): 4 optional vars — `EMBEDDING_CONNECTED_APP_CLIENT_ID`, `EMBEDDING_CONNECTED_APP_SECRET_ID`, `EMBEDDING_CONNECTED_APP_SECRET_VALUE`, `EMBEDDING_USERNAME`. All-or-nothing constructor guard (set all four or none). Not required for any AUTH mode.
- **resolver** (`1692089`): new branch 2b in `resolveEmbedToken` signs the embed JWT from the embedding Connected App for non-direct-trust modes. Direct-trust still takes precedence; Bearer pass-through unchanged.
- **docs** (`b1c3fc9`): `env.example.list` + "Embedded viz (optional)" sections in `pat.md` / `uat.md`.
- **fix** (`c8d9c13`): added the 4 fields to 3 stale Phase 1 test fixtures (typecheck-only regression).

### Verification
- `npx tsc --noEmit`: 0 errors
- `npm run build`: success
- `npm test`: 1866/1866 (1854 baseline + 12 new)
- `npm run lint`: clean

### Security
Embedding signing secret is never logged, never serialized into errors, never returned to the model (`get-embed-token` is app-only visibility). Embed JWT scope is exactly `tableau:views:embed` — no escalation. Partial credential cannot reach signing (constructor guard + resolver truthy-check).

PR #424, opened 2026-06-25
@W-23202096 MCP Apps on Direct Trust + UAT

jarhun88:
## Description

1. **Bearer pass-through** — if the request carries a Tableau Bearer JWT (`AUTH=oauth`, Tableau
   authZ server), use it directly.
2. **Server-signed embed JWT** — otherwise sign an embed JWT on the server:
   - `AUTH=direct-trust` — sign from the Connected App credentials (`CONNECTED_APP_*` + `JWT_SUB_CLAIM`).
   - `AUTH=uat` — sign from the UAT RS256 key.
3. **Not available** — if no signing material applies (`AUTH=pat`, or `oauth` without a Bearer
   token), return a typed `EmbedTokenNotAvailableError`; the app skips embedding and falls back to
   a non-embedded view instead of erroring.

The token value is app-only (`visibility: ['app']`) and is never exposed to the model. The embed
scope is always exactly `tableau:views:embed` — no escalation.

### Supporting refactors
- Extract `buildAuthConfig` as a shared `Config → AuthConfig` mapper (with tests) and simplify
  `restApiInstance` to use it.
- Gate the `direct-trust` embed branch on auth type alone.
- Rename the app-side client `getOAuthTokenToolClient` → `getEmbedTokenToolClient`; the client
  reads only `token` from the result and returns `null` (skips embedding) on the not-available path.
- Remove the unused `tokenType` field from the tool result (the app never consumed it).

## Motivation and Context

The embedded Tableau viz needs a `tableau:views:embed` JWT, and the correct source of that token
depends on the server's auth mode. The previous `get-oauth-token` tool conflated this with OAuth
specifically. This makes the resolver explicit per auth mode and adds a clean, typed
"not available" path so the app degrades gracefully (notably, `pat` does not produce embed tokens).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## How Has This Been Tested?

- Unit tests for the resolver (`resolveEmbedToken.test.ts`), the tool (`getEmbedToken.test.ts`),
  the shared mapper (`buildAuthConfig.test.ts`), and the app client (`getEmbedTokenToolClient.test.ts`),
  covering Bearer pass-through, direct-trust signing, uat signing, and the not-available/skip path.
- E2E server test (`tests/e2e/server.test.ts`) updated for the renamed tool.
- `npm test` and `npm run lint` pass locally.

## Related Issues

Phase 2 of the embed-auth work (supersedes #423).

## Checklist

- [x] I have updated the version in the package.json file (`2.19.0`).
- [ ] I have documented any breaking changes in the PR description.

> Note: this renames the `get-oauth-token` tool to `get-embed-token`. The tool is app-only
> (not model-visible), so this is not expected to affect model-facing behavior.


src/sdks/tableau/buildAuthConfig.ts
```diff
@@ -0,0 +1,73 @@
...
+        tenantId: config.uatTenantId,
+        issuer: config.uatIssuer,
+        usernameClaimName: config.uatUsernameClaimName,
+        privateKey: config.uatPrivateKey,
+        keyId: config.uatKeyId,
+        scopes,
+        additionalPayload,
+      };
+
+    default:
```
anyoung-tableau: Consider explicitly handling the `oauth` case and returning `null`, removing the `default` case, and updating the function return type to `AuthConfig | null`. That way if we ever add a new `auth` type, the TypeScript compiler will not let us accidentally forget to update this function to handle the new auth type.

src/errors/mcpToolError.ts
```diff
@@ -108,6 +108,16 @@ export class QueryValidationError extends McpToolError {
   }
 }
 
+export class EmbedTokenNotAvailableError extends McpToolError {
+  constructor() {
+    super({
+      type: 'embed-token-not-available',
+      message: 'Failed to get an embed token for the current authentication configuration.',
+      statusCode: 500,
```
anyoung-tableau, line 128: I know we're not really using these status codes, but this feels more like a bad request 400

jarhun88: isnt the only reason this would fail because of an internal server error though?

src/sdks/tableau/buildAuthConfig.ts
jarhun88: thanks that makes sense

src/errors/mcpToolError.ts
anyoung-tableau: I don't think so. Wouldn't this error get thrown when using the embedded authz server?

anyoung-tableau (approved):

src/errors/mcpToolError.ts
jarhun88: for embedded auth, the tool should not register mcp app resources, so that scenario should never occur. At least thats the plan. because the embed-token-tool only gets triggerd by mcp apps, and it only just returns the token, i was hoping to use this 500 as a sign that something in the way we implemented it is broken

PR #429, opened 2026-06-29
@W-23125362: Add mutation guard for destructive web tools

Alon-ST-DATA:
## Description

Adds a server-side mutation guard that requires an explicit confirmation token plus evidence before destructive/mutating web tools execute. Wires the guard into `delete-datasource`, `delete-workbook`, `delete-extract-refresh-task`, and `update-cloud-extract-refresh-task`.

New `_lib` modules:
- `mutationGuard` — confirmation-token + confirm gate
- `evidence` — caller intent / evidence capture
- `auditRecord` — structured audit logging

Extends `logger` to emit audit records and the `confirm.ts` prompt helper.

## Motivation and Context

Destructive web tools previously executed without a server-enforced confirmation step. This guard makes the confirmation server-authoritative, reducing the risk of accidental or unintended deletions/mutations and producing an audit trail.

## Type of Change

- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Breaking Changes

The guarded mutation tools now default to a non-destructive **preview** when `confirm` is omitted or false. This changes the behavior of a single call for **existing callers**:

- `update-cloud-extract-refresh-task` — previously a single call applied the schedule update. It now **previews by default** and only applies the change when called again with `confirm: true`. Callers that relied on the old one-shot behavior will no longer mutate on the first call.
- `delete-extract-refresh-task` — previously a single call deleted the task. It is now two-phase: the first call (no `confirm`) returns a single-use `confirmationToken`; the destructive delete requires a second call with `confirm: true` and that token.
- `delete-workbook` / `delete-datasource` — same two-phase preview→confirm contract (preview tags the content `pending-deletion`; the confirmed delete re-verifies that tag server-side).

`confirm`-defaults-to-preview is the intended safety behavior and is retained — it is the source of the breaking change, not a regression. Tool descriptions state the two-phase requirement, and the e2e suites exercise the preview→confirm contract.

## How Has This Been Tested?

- Unit tests added for each new `_lib` module (`mutationGuard`, `evidence`, `auditRecord`) and for each guarded tool.
- Added e2e mutation-guard driver (`tests/e2e/_mutationGuardDriver.ts`).
- Updated the `delete-extract-refresh-task` and `update-cloud-extract-refresh-task` e2e suites to the two-phase preview→confirm contract.
- New and existing unit tests pass locally.

## Related Issues

W-23125362

## Checklist

- [x] I have updated the version in the package.json file
- [x] I have documented any breaking changes in the PR description.


Alon-ST-DATA:
## Review — mutation guard for destructive web tools

Reviewed the full diff (25 files), read changed files + unchanged siblings, ran the affected unit tests (197 passing) and lint (clean).

**Verdict: ship-with-fixes** — 0 BLOCKER · 3 MAJOR · 3 MINOR · 1 NIT

### MAJOR
- **`src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts`** — `confirm` now defaults to `false`, so an existing caller that omits it gets a no-op **preview** instead of applying the update — a breaking behavior change. The PR checklist "documented any breaking changes" is unchecked and it isn't called out. Document/flag it as breaking (release notes), or reconsider the default semantics.
- **`tests/e2e/extractRefreshTasks/deleteExtractRefreshTask.test.ts:99`** — inherited e2e calls the tool without `confirm` and asserts `successfully deleted`; the tool now returns a preview, so this opt-in test asserts the wrong contract. Update to preview→confirm with the minted token. (Env-gated, so CI stays green while coverage silently lies.)
- **`tests/e2e/extractRefreshTasks/updateCloudExtractRefreshTask.test.ts:115`** — same staleness: calls without `confirm`, asserts `successfully updated`, now returns preview. Pass `confirm: true`. The standalone `_mutationGuardDriver.ts` (non-vitest, won't run in CI) does not replace these real suites.

### MINOR
- **`src/tools/web/_lib/mutationGuard.ts:14`** — comment claims "@modelcontextprotocol/sdk v1.26.0 exposes NO elicitation/sampling primitive," contradicting `src/prompts/_lib/confirm.ts:246` which says ≥1.26 DOES expose `server.elicitInput(...)` and this repo runs 1.29 (package.json pins `^1.26.0`). Reconcile so the residual-gap rationale isn't self-contradicting.
- **`docs/docs/configuration/mcp-config/env-vars.md`** — new env var `MUTATION_PREVIEW_TTL_MINUTES` (read in `evidence.ts:117`) is documented only on the delete-extract-refresh-task tool page, not in env-vars.md; its sibling `ADMIN_GATE_CACHE_TTL_MINUTES` has a full entry. Add a matching entry (default 5, min 1, max 1440).
- **`src/tools/web/_lib/mutationGuard.ts:52` / `auditRecord.ts:33`** — the `'user'` value in the `kind` enums is unused by any of the four wired tools (no user-mutation tool exists). Drop it until needed, or comment it as forward-looking.

### NIT
- **`src/tools/web/extractRefreshTasks/deleteExtractRefreshTask.ts`** — mixed curly/straight apostrophes across response strings (readability only; matches siblings otherwise).

### Confirmed clean
Registration trio · gating + scope leaks (uniform `guardMutation`→`assertAdmin`, reuses `ADMIN_TOOLS_ENABLED`) · doc↔schema parity (`confirm`/`confirmationToken` declared, described, used, documented) · description token budget · HITL two-phase (server-authoritative tag/registry-nonce, target-bound single-use token, recycle-bin language, authoritative audit on allow+deny) · unit coverage (success/empty/error/denied/forged/replay/TTL/scoping) · docs intro.md rows + per-tool pages · version bump 2.18.4→2.18.5.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
```diff
@@ -75,9 +92,41 @@ export const getUpdateCloudExtractRefreshTaskTool = (
...
+              const resolveTarget = async (): Promise<MutationTarget> => ({
+                id: args.taskId,
+                kind: 'extract-refresh-task',
+              });
+              const guardResult = await guardMutation({
+                restApi,
+                extra,
+                tool: 'update-cloud-extract-refresh-task',
+                action: 'update',
+                mode: 'confirm-only',
```
Akash-Rastogi: [Akash's agent]

**Confirm-only mode is a bypass surface.**

Two problems compose here:

1. **Preview can be skipped entirely.** At `mutationGuard.ts:135` the evidence gate is `if (phase === 'confirm' && mode === 'preview-confirm')`. With `mode: 'confirm-only'` (this call site) that branch is never taken, so `evidence.verify()` never runs. A caller invokes the tool with `confirm: true` on the very first call — no prior preview — and the update goes through. `NoEvidence` here is a fig leaf.

2. **No binding between previewed and applied schedule.** `args.schedule` is read fresh at [line 135](https://github.com/tableau/tableau-mcp/blob/f6454939/src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts#L135) on the confirm path. Nothing in the preview phase persists what was previewed; the guard has no server-side proof that the human approved *this specific* schedule. So: preview Daily 06:00, confirm with Hourly 01:00–05:00, and the guard happily applies the second one.

Reproducer for (1): one call with `{ taskId, schedule, confirm: true }` — bypasses preview.
Reproducer for (2): preview `{ taskId, schedule: A }`, then confirm `{ taskId, schedule: B, confirm: true }` — applies B.

Fix direction: either move `update-cloud-extract-refresh-task` to `preview-confirm` mode with an evidence type that binds the schedule (e.g. hash of the previewed args stored server-side, verified on confirm), or narrow `confirm-only` to tools where the mutation is genuinely a single-shot idempotent action with no caller-controlled parameters.

src/tools/web/_lib/mutationGuard.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+  }
+
+  // (5) Preview phase: establish the server-side proof the confirm phase will verify.
+  if (phase === 'preview') {
+    await evidence.establish(evidenceCtx);
+  }
+
+  // (6) Allowed: record the attempt and let the tool perform its mutation / build its response.
+  emitAuditRecord({
```
Akash-Rastogi, line 186: [Akash's agent]

**Audit says "allowed" before the mutation runs — misleading on failure.**

This emits `result: 'allowed'` and then returns to the caller, which then makes the destructive REST call. If that REST call fails (5xx, network error, Tableau 409 conflict, permission-denied at Tableau's edge, race with another admin), the audit trail claims the mutation was allowed while the target is unchanged. No follow-up record. Downstream consumers can't distinguish "authorized-but-failed" from "completed".

This weakens the PR description's "produces an audit trail" claim: an audit trail that records intent-to-mutate rather than outcome-of-mutation isn't the audit trail an incident responder needs.

Minimum fix: split into two records — `result: 'allowed'` right here (kept for the authorization decision), plus a follow-up `result: 'completed'` or `result: 'failed'` emitted from each caller after the REST result is known. Or bring the REST call inside `guardMutation` so a single terminal record can be written.

Reproducer: mock `restApi.workbooksMethods.deleteWorkbook` to reject with a 500. The audit log will contain `result: 'allowed'` for a workbook that was never deleted.

src/tools/web/_lib/mutationGuard.ts
```diff
@@ -0,0 +1,189 @@
...
+    siteLuid: extra.getSiteLuid(),
+    siteName: extra.getSiteName(),
+  };
+  const evidenceDescriptor = evidence.describeEvidence();
+
+  // (1) Uniform admin gate. On failure emit a DENIED audit before returning so even rejected
+  // privilege escalations are recorded — but resolve the target first so the record names it.
+  const adminResult = await assertAdmin(restApi, extra);
+  if (adminResult.isErr()) {
+    const deniedTarget = await resolveTarget();
```
Akash-Rastogi: [Akash's agent]

**Denied audit can be silently dropped if target resolution throws.**

On the not-admin denied path, `resolveTarget()` runs before the DENIED audit is emitted (line 109). If `resolveTarget` throws — e.g. the tool's target-lookup does a read that itself returns 403/404/network failure — the exception unwinds past `emitAuditRecord` and the "non-admin attempted a destructive tool" event is lost.

For an audit surface specifically meant to catch privilege-escalation attempts, this is the exact case we most want to record. Wrap in try/catch and emit the audit with a placeholder target (`{ id: 'unresolved', kind: '…' }`) on lookup failure so the attempted-escalation event is preserved.

Marked as LOW because it requires the specific `resolveTarget throw` case, but the failure mode is silent.

Akash-Rastogi (changes requested):
[Akash's agent]

Solid direction — a server-authoritative mutation guard with an audit trail is the right control. Two findings block it as-is, both in tension with the PR description's "server-authoritative" and "authoritative audit" claims:

1. `confirm-only` mode (used by `update-cloud-extract-refresh-task`) skips `evidence.verify` ([mutationGuard.ts:135](https://github.com/tableau/tableau-mcp/blob/f6454939/src/tools/web/_lib/mutationGuard.ts#L135)) and holds no server-side binding between the previewed schedule and the applied one. A caller can (a) call with `confirm:true` on the first invocation and skip preview entirely, or (b) preview one schedule and confirm a different `args.schedule` — the guard applies whatever the confirm call carries. See the line comment on `updateCloudExtractRefreshTask.ts` for details.

2. The audit record for `result:'allowed'` is emitted BEFORE the tool's REST mutation runs. A 5xx / network / Tableau-409 failure leaves an "allowed" audit entry while the target is unchanged. See the line comment on `mutationGuard.ts:163`.

Third (LOW) finding on the denied-audit path also worth a look but not a blocker.

Also note this PR's mergeable_state is `dirty` — needs a rebase on top of main before the merge queue will accept it.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
Alon-ST-DATA: [Alon's agent] Fixed in feaf47bc. update-cloud-extract-refresh-task no longer uses confirm-only/NoEvidence — it now gates on RegistryEvidence bound to a schedule fingerprint (sha256 of a key-order-canonicalized schedule). A confirm requires a single-use confirmationToken minted by a prior preview of this exact taskId + schedule: a token minted for a different schedule will not validate, and confirm-without-preview (no token) is rejected server-side. Tests added for swap-reject, no-token-reject, replay-reject, and key-order-independent binding.

src/tools/web/_lib/mutationGuard.ts
Alon-ST-DATA: [Alon's agent] Fixed in feaf47bc. Added GuardOutcome.recordOutcome — the guard's allowed record now captures only the authorization decision, and the caller invokes recordOutcome after the REST call to emit a terminal completed/failed record (with failureDetail). The audit schema result enum is extended to allowed|denied|completed|failed. Wired on every exit path of all four mutation tools; no-op on preview. Tests assert allowed->completed on success and allowed->failed (with detail) on error.

src/tools/web/_lib/mutationGuard.ts
Alon-ST-DATA: [Alon's agent] Fixed in feaf47bc. resolveTarget() on the not-admin path is now wrapped in try/catch — a lookup failure (403/404/network) falls back to an { id: 'unresolved', kind } placeholder so the DENIED audit is still emitted rather than swallowed by the throw. Added a mutationGuard test: assertAdmin=Err AND resolveTarget rejects -> still emits result:'denied', denyReason:'not-admin', target.id:'unresolved'.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
Akash-Rastogi: [Akash's agent] Fixed in `feaf47b`.

src/tools/web/_lib/mutationGuard.ts
Akash-Rastogi: [Akash's agent] Fixed in `feaf47b`.

src/tools/web/_lib/mutationGuard.ts
Akash-Rastogi: [Akash's agent] Fixed in `feaf47b`.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
```diff
@@ -1,20 +1,75 @@
...
+        'schedule will not validate.',
+    ),
 };
 
+/**
+ * Deterministic, key-order-independent JSON serialization. Object keys are emitted sorted at every
+ * depth so two schedules that differ only in the caller's key order produce identical output —
+ * otherwise the preview and confirm fingerprints could diverge for the same logical schedule.
+ */
+function canonicalize(value: unknown): unknown {
```
Akash-Rastogi, line 52: [Akash's agent] Follow-up (non-blocking): `canonicalize()` sorts object keys recursively but preserves array element order (see `value.map(canonicalize)` on line 49).

If Tableau treats `intervals.interval` (or any other array in the schedule shape) as an order-independent bag of constraints — i.e. `[{hour: 3}, {hour: 15}]` and `[{hour: 15}, {hour: 3}]` describe the same schedule — then an LLM caller that reorders elements between the preview and confirm calls will produce a different `scheduleBinding`, mint a nonce bound to the wrong fingerprint on preview, and get a spurious `preview-not-run` on confirm.

Not a security bug (fail-closed is safe). But it's a real UX gap that will read as flaky HITL to anyone testing this path. Two ways to close it:

- Document that callers MUST pass array elements in identical order across preview → confirm, OR
- Sort known-bag arrays (e.g. `intervals.interval`) by a stable key inside `canonicalize` before hashing.

Pick whichever matches Tableau's actual semantics for the schedule shape.

src/tools/web/_lib/auditRecord.ts
```diff
@@ -0,0 +1,59 @@
...
+  // Lifecycle of a mutation attempt:
+  //   - 'denied'    — authorization/evidence gate rejected the attempt; nothing mutated.
+  //   - 'allowed'   — the attempt passed the gate. For a preview this is terminal (nothing mutates);
+  //                   for a confirm it records only the AUTHORIZATION decision and is followed by a
+  //                   terminal outcome record once the destructive REST call returns.
+  //   - 'completed' — the confirmed mutation's REST call succeeded.
+  //   - 'failed'    — the confirmed mutation was authorized but its REST call failed; the target is
+  //                   unchanged. Distinguishing this from 'completed' is what an incident responder
+  //                   needs — an 'allowed' record alone cannot tell authorized-but-failed from done.
+  result: z.enum(['allowed', 'denied', 'completed', 'failed']),
```
Akash-Rastogi, line 56: [Akash's agent] Follow-up (non-blocking): the `result` enum was widened from `['allowed', 'denied']` to `['allowed', 'denied', 'completed', 'failed']` for the F3 fix, but `schemaVersion` on line 16 is still `z.literal(1)`.

Downstream audit-log consumers that pinned to the v1 shape and validate incoming records with the original strict enum will `zod`-fail on any `completed`/`failed` record — silently dropping the terminal outcome data that F3 was specifically added to preserve.

Either bump `schemaVersion` to `z.literal(2)` (breaking-but-explicit) or add a schema-level note that `result` is extensible within v1 and consumers must accept unknown enum values. Given F3 was explicitly about auditability, I'd lean toward the version bump so consumers know to update their parsers.

Akash-Rastogi (approved):
[Akash's agent] Second-round review after `feaf47b`.

**All 3 prior findings resolved — confirmed in code:**

- **F1 (`preview-not-run` bypass)** — `updateCloudExtractRefreshTask` now uses `RegistryEvidence` with `mode: 'preview-confirm'`, so an LLM caller that skips the preview cannot mint a valid nonce.
- **F2 (schedule-swap)** — new `scheduleBinding(args.schedule)` binds the nonce to a SHA256 of the canonicalized schedule; a nonce minted while previewing schedule A cannot confirm schedule B.
- **F3 (post-authorization outcome gap)** — `GuardOutcome.recordOutcome` wires all four mutation tools (delete datasource/workbook/extract-refresh-task, update extract-refresh-task) to emit a terminal `completed`/`failed` audit record after the REST call returns. The `emitAuditRecord` on the not-admin path is also wrapped in `try/catch` with a `targetKindHint(tool)` fallback so a `resolveTarget` throw no longer silently swallows the audit record.

Fresh security pass on the fix diff also confirms:

- No confirmationToken/nonce leakage in tool responses or audit-record `detail` (`RegistryEvidence.describe()` returns a fixed constant).
- Cross-user replay blocked by `userLuid` in `nonceKey`.
- `verify()` performs an atomic single-use get+delete on the cache.

Approving. Two non-blocking follow-ups posted as line comments — both are robustness / hygiene, not security regressions.

Separately: `mergeable_state` on this PR still shows `dirty` (base-branch conflict, unrelated to the review). Will need a rebase before merge.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
Alon-ST-DATA: [Alon's agent] Fixed in c0011ed4. Went with the sort-the-bag option: canonicalize() now sorts array elements by their own canonical form (not just object keys), so a caller that reorders intervals.interval between preview and confirm produces the same scheduleBinding and no longer hits a spurious preview-not-run. The schedule shape has no array whose order is semantically significant, so sorting all arrays is safe. Added AC-6(h) asserting a reordered-interval confirm still validates.

src/tools/web/_lib/auditRecord.ts
Alon-ST-DATA: [Alon's agent] Fixed in c0011ed4. Bumped schemaVersion z.literal(1) -> z.literal(2) at the schema and the guard emit site, since the result enum widening (allowed|denied -> allowed|denied|completed|failed) changes the shape a v1-pinned consumer validates against. Explicit bump so parsers know to update rather than zod-failing on completed/failed records. Tests updated to pin v2 and assert completed/failed parse.

PR #431, opened 2026-06-30
@W-23240258: Adding missing annotations to unblock chatGPT app submission

yyk16293:

the pull request._

# Pull Request Template

## Description
Adding missing annotations to unblock chatGPT app submission

image (13)

Related slack thread: https://salesforce-internal.slack.com/archives/C09S2ML7FH9/p1782777403577619

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe):

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/desktop/session/checkForUserChanges.ts
+++ b/src/tools/desktop/session/checkForUserChanges.ts
@@ -43,6 +43,8 @@ export const getCheckForUserChangesTool = (
     annotations: {
       title,
       readOnlyHint: true,
+      destructiveHint: false,
+      idempotentHint: true,
       openWorldHint: false,
     },
     callback: async ({ session, sinceSequence }, extra): Promise<CallToolResult> => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.20.0",
+  "version": "2.20.2",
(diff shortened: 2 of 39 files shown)


src/tools/desktop/session/checkForUserChanges.ts
```diff
@@ -43,6 +43,8 @@ export const getCheckForUserChangesTool = (
     annotations: {
       title,
       readOnlyHint: true,
+      destructiveHint: false,
```
anyoung-tableau, line 46: Since the [ToolAnnotations](https://github.com/modelcontextprotocol/typescript-sdk/blob/708d5455d946d59f6f5482118b3a44e7285ae632/packages/core-internal/src/types/schemas.ts#L1256) type is defined with all properties optional, which is probematic for OpenAI, let's change our references to the type in tool.ts to make all the properties required instead with `Required<ToolAnnotations>`. This will prevent the possibility of regressions and introducing a new tool without fully annotating it.

anyoung-tableau: If you don't want to have to expand this PR's scope to provide a `title` annotation for each tool, you can do some type surgery to keep `title` optional.

anyoung-tableau (approved):

PR #432, opened 2026-06-30
@W-23202047: extend MCP-Apps-cooperative HITL to all mutation tools

Alon-ST-DATA:
## What

Closes **W-23202047** — extends the MCP-Apps **cooperative-HITL** pattern (app-only `confirm-*` tool a cooperating client hides from the model, fired by a human gesture in the rendered MCP-Apps iframe) from `delete-workbook` to the remaining three mutation tools:

- `delete-datasource` → `confirm-delete-datasource`
- `delete-extract-refresh-task` → `confirm-delete-extract-refresh-task`
- `update-cloud-extract-refresh-task` → `confirm-update-cloud-extract-refresh-task`

> **Stacked on PR #429** (`asimantov/w-23125362-mutation-guard`). The mutation-guard `_lib` (evidence / mutationGuard / auditRecord) comes from that PR; this branch adds **only the app layer**. Set the base to `main` once #429 merges.

## Why

PR #429 added the server-authoritative mutation guard to all four destructive tools, but only `delete-workbook` got the MCP-Apps confirm-panel step. On an MCP-Apps host the other three mutators had no confirm-panel path — the gap W-23202047 reports. This closes it so that, in a **cooperating** MCP client, every destructive REST call is fronted by a human click in the iframe.

> **Security scope — cooperative, not server-enforced.** `visibility: ['app']` is a client-side hint and `AppApprovalEvidence` is minted during the model-driven preview phase, so a **non-cooperating** client that ignores the hint can still drive `preview → confirm-*` back-to-back with no human gesture. For `delete-datasource` / `delete-workbook` the live pending-deletion tag re-check still proves a preview *ran*; for the two task tools `AppApprovalEvidence` is the **only** gate (no tag layer). Neither proves a *human approved*. Server-enforced HITL (an approval primitive the model cannot forge or reach) is tracked as follow-up under **W-23125362**.

## How it works (per tool)

Each `confirm-*` tool is:
- `meta.ui.visibility = ['app']` — a cooperating MCP client hides it from the model and reaches it only via `app.callServerTool` from a human click in the iframe. This is a **client-side hint, not a server guarantee** (see Security scope above).
- `disabled` unless `ADMIN_TOOLS_ENABLED` **and** the `mcp-apps` feature flag are both on.
- The home of the destructive REST call — it lives **only** here, never in a model-visible path. With `mcp-apps` on, the model's `confirm: true` path on the preview tool is rejected before any side effect (the mutation moves to the confirm tool). A non-cooperating client that ignores the visibility hint can still call the confirm tool directly — see Security scope.

Evidence (`AllEvidence`, AND-composed, can only ever narrow access):
- `delete-datasource` (taggable): `TagEvidence(datasource)` + `AppApprovalEvidence` — live pending-deletion tag re-fetched server-side **and** a fresh single-use in-iframe human approval.
- `delete-extract-refresh-task` / `update-cloud-extract-refresh-task` (no taggable target): `AppApprovalEvidence` — single-use, TTL-bounded human approval (`MUTATION_PREVIEW_TTL_MINUTES`).

**Shared refactor:** `AppApprovalEvidence` is generalized from a hardcoded `delete-workbook` namespace to a constructor arg (default `delete-workbook`), so each preview/confirm pair resolves its own approval entry and tools can't cross-unlock. Each preview's `establish()` namespace matches its confirm's `verify()`.

App clients (`src/web/apps/src/lib/*ConfirmClient.ts`) render server-derived strings via `textContent` only (XSS guard); the update-cloud panel describes a schedule change, not a deletion.

## Also included (disclosed)

- **`get-stale-content-report` gains an `itemLuid` column** (`src/tools/web/adminInsights/getStaleContentReport.ts`). Independent of the HITL change, but thematically related: it surfaces the content LUID so stale items can be passed directly to the delete tools (`itemId` is Site Content's integer repo ID, which REST rejects). Happy to split into a follow-up if preferred.

## Tests

- `+~80` unit tests across the 3 confirm tools, 3 app clients, the generalized `AppApprovalEvidence`, and the preview-tool flag-ON branches.
- Per confirm tool: happy-path (REST called once) · missing-approval → `PreviewNotRunError` (no REST) · single-use replay · expired-TTL · datasource missing-tag · cross-namespace isolation · disabled-gate (both directions).
- App clients (jsdom): result discrimination · Confirm click → `callServerTool` with correct name+args · `textContent` XSS guard.
- **Full unit suite: 2060/2060 pass · `tsc --noEmit` clean · lint clean · build success.**
- e2e / oauth / eval / manual-live: not run here (no creds/site); confirm tools are model-invisible so eval N/A, and no auth/middleware path is touched.

---

## Test & verification report (tmcp-tester, live 10ax test site)

**Verdict: SHIP for the live HITL gate** — one automated tier (e2e) still to run.

### Coverage matrix
| Tier | Result |
|---|---|
| Unit (vitest) | **PASS** — 2062/2062 (incl. ~80 new + 2 from the message fix) |
| Live HITL, flag ON | **PASS** — full chain vs REST ground truth |
| E2E (live) | **SKIP** — no `tests/.env` connected-app creds present |
| OAuth | **N/A** — scope-map additions only, no flow logic changed |
| Eval | **N/A** — confirm tools are model-invisible |
| tsc / lint | clean |

### Live HITL — security paths verified
Run on a separate flag-ON server (the provided PAT server runs `mcp-apps` OFF and was left untouched); disposable datasource only, no real content touched:
- confirm-* tools hidden from the model (`visibility:["app"]`), absent from the model tool list
- model `confirm:true` path hard-closed → `PreviewNotRunError`, no delete
- direct `confirm-*` without a preview → guard rejects (`denyReason:preview-not-run`)
- preview returns the panel + tags `pending-deletion`, no mutation (datasource still 200 after preview)
- destructive REST call lives only in the confirm tool → after confirm GET=404; replay rejected (single-use)
- audit records emit with no raw nonce/secret; 0 PAT leaks in logs

### Bugs found: none
One cosmetic note the tester flagged — the preview-not-run denial told the user to "run `<tool>` with confirm omitted", which is wrong for the app-only confirm tools (they take no `confirm` arg). **Fixed** in commit `fc212de8`: the guard now points app-only confirm tools at the preview tool + confirmation panel, while model-visible preview-confirm tools keep the in-place wording. +2 unit tests.

### Gaps to close before merge
1. **E2E tier not executed** — needs `tests/.env` (connected-app direct-trust, site `mcp-test`). Only un-run automated tier.
2. Live destructive chain was run on the **datasource** path (representative). The ERT/workbook confirm tools share the identical guard+evidence machinery and were verified for registration / visibility / gating, but their full live delete was not run (no disposable ERT/workbook target created).


Alon-ST-DATA:
## Test & verification report (tmcp-tester, live 10ax test site)

**Verdict: SHIP for the live HITL gate** — one automated tier (e2e) still to run.

### Coverage matrix
| Tier | Result |
|---|---|
| Unit (vitest) | **PASS** — 2062/2062 (incl. ~80 new + 2 from the message fix) |
| Live HITL, flag ON | **PASS** — full chain vs REST ground truth |
| E2E (live) | **SKIP** — no `tests/.env` connected-app creds present |
| OAuth | **N/A** — scope-map additions only, no flow logic changed |
| Eval | **N/A** — confirm tools are model-invisible |
| tsc / lint | clean |

### Live HITL — security paths verified
Run on a separate flag-ON server (the provided PAT server runs `mcp-apps` OFF and was left untouched); disposable datasource only, no real content touched:
- confirm-* tools hidden from the model (`visibility:["app"]`), absent from the model tool list
- model `confirm:true` path hard-closed → `PreviewNotRunError`, no delete
- direct `confirm-*` without a preview → guard rejects (`denyReason:preview-not-run`)
- preview returns the panel + tags `pending-deletion`, no mutation (datasource still 200 after preview)
- destructive REST call lives only in the confirm tool → after confirm GET=404; replay rejected (single-use)
- audit records emit with no raw nonce/secret; 0 PAT leaks in logs

### Bugs found: none
One cosmetic note the tester flagged — the preview-not-run denial told the user to "run `<tool>` with confirm omitted", which is wrong for the app-only confirm tools (they take no `confirm` arg). **Fixed** in commit `fc212de8`: the guard now points app-only confirm tools at the preview tool + confirmation panel, while model-visible preview-confirm tools keep the in-place wording. +2 unit tests.

### Gaps to close before merge
1. **E2E tier not executed** — needs `tests/.env` (connected-app direct-trust, site `mcp-test`). Only un-run automated tier.
2. Live destructive chain was run on the **datasource** path (representative). The ERT/workbook confirm tools share the identical guard+evidence machinery and were verified for registration / visibility / gating, but their full live delete was not run (no disposable ERT/workbook target created).

src/tools/web/workbooks/deleteWorkbook.ts
```diff
@@ -89,15 +116,13 @@ the user's explicit approval first.
       title: 'Delete Workbook',
       readOnlyHint: false,
       destructiveHint: true,
-      // Hard delete-by-id: a second delete of the same workbookId 404s (isError: true), so the
-      // operation is not idempotent. A client trusting an idempotent hint and retrying after a
-      // transient failure would get a spurious error for a delete that already succeeded.
-      // Matches the accepted resolution for delete-extract-refresh-task (tableau/tableau-mcp#392).
-      idempotentHint: false,
+      idempotentHint: true,
```
Akash-Rastogi: [Akash's agent]

**`idempotentHint` regression against PR #429's fix.**

PR #429 landed a commit titled `Fix delete-workbook idempotentHint: not idempotent (second delete 404s)` that set this to `false` and added a comment referencing `tableau/tableau-mcp#392`. This PR reverts that back to `true` and deletes the comment.

The reasoning behind #429's fix still applies: a second call with the same `workbookId` returns 404 from Tableau's REST API, so the tool is NOT idempotent per MCP's semantics ("safe to retry with the same effect"). The extract-refresh-task delete and datasource delete both correctly keep `idempotentHint: false` for the same reason.

Please restore `idempotentHint: false` and the `#392` comment. Same fix needed on `confirmDeleteWorkbook.ts` (separate comment below).

src/tools/web/workbooks/confirmDeleteWorkbook.ts
```diff
@@ -0,0 +1,154 @@
...
+Before deleting, the server re-verifies BOTH that the workbook is still tagged pending deletion AND
+that a human approved the deletion in the App within the allowed time window. If either check fails
+the deletion is rejected and the user must preview again.
+`.trim(),
+    paramsSchema,
+    annotations: {
+      title: 'Confirm Delete Workbook',
+      readOnlyHint: false,
+      destructiveHint: true,
+      idempotentHint: true,
```
Akash-Rastogi: [Akash's agent]

**`idempotentHint` disagreement across the four confirm-* tools.**

All four `confirm-*` tools have identical hard-delete semantics (delete-by-id → 404 on retry), but the hint disagrees:

- `confirmDeleteWorkbook.ts` — `idempotentHint: true` ← this line, outlier
- `confirmDeleteDatasource.ts:68` — `idempotentHint: false`
- `confirmDeleteExtractRefreshTask.ts:59` — `idempotentHint: false`
- `confirmUpdateCloudExtractRefreshTask.ts` — update semantics differ; ignore

The `false` values are correct (see the #429 fix comment on `deleteWorkbook.ts` above). Flip this to `false` too so the four confirm tools tell the caller a consistent story.

Akash-Rastogi (changes requested):
[Akash's agent]

The pattern is clean — `visibility: ['app']` + feature-flag double gating + `AllEvidence` composition is a coherent shape, and the iframe-side code correctly uses `textContent` for user-derived values. But one high-severity finding blocks the "real HITL" framing:

**`visibility: ['app']` is a client-side hint, not a server-side check.** `AppApprovalEvidence` is minted during the model-driven preview phase, so any MCP client that doesn't voluntarily filter model-invisible tools can drive `preview → confirm` back-to-back with zero human interaction. For the two task tools (`delete-extract-refresh-task`, `update-cloud-extract-refresh-task`) there's no tag layer, so `AppApprovalEvidence` is the ONLY server gate — a non-cooperating agent client bypasses the human check entirely. The PR title "real-HITL to all mutation tools" and the description's "every destructive REST call sits behind a human click" both overstate what the server actually enforces.

This is architectural and inherits from #429's guard design — it's not fixable purely inside this PR — but it's most visible here because #432 extends the pattern to three more tools and puts the strong "real HITL" label on all four. Options I can see:

1. Server-side check on the calling client's identity/session (e.g. only accept `confirm-*` calls when the transport is the app-served WebSocket / bearer marker, not the model JSON-RPC route).
2. Mint `AppApprovalEvidence` ONLY from an in-iframe `postMessage` gesture that hits a dedicated server endpoint the model has no route to — so the token cannot exist without a real human click.
3. Retitle to "MCP-Apps-cooperative HITL" and document that the guarantee holds against cooperative MCP clients only. This matches Alon's own scoping note on #414 (W-23125362 tracks the full-HITL gap).

Two smaller findings on the diff as line comments. Also filing this feedback in the workstream context so the follow-up WI (whatever survives from W-23125362) can capture it.

docs/docs/tools/data-qna/delete-datasource.md
```diff
@@ -34,6 +34,19 @@ satisfies the gate with no human in the loop. Enforcing true human-in-the-loop (
 approval the agent cannot forge) is tracked as follow-up work.
 :::
 
+### MCP-Apps confirm panel (real human-in-the-loop)
+
+When the off-by-default `mcp-apps` feature flag is enabled, this tool ships with an MCP App and the
```
jarhun88: i think you can remove off-by-default

src/web/apps/src/mcp-app.ts
```diff
@@ -93,6 +109,24 @@ function extractUrlObjectFromResult(result: CallToolResult): string {
 // Handle tool results
 app.ontoolresult = async (result: CallToolResult) => {
   try {
+    // delete-workbook (flag ON) preview returns a confirm-panel payload, not a viz URL. Branch on
```
jarhun88: can we have a different mcp-app.ts and html for the hitl workflow? i dont want to bloat the logi so much in here, because its specifically for embedding tableau viz scenario. Perhaps I can later rename my mcp-app.ts and html to be more specific to that use case.

So yea im thinking of bundling different htmls and mcp-apps.ts by functionality if you could adhere to that pattern?

src/tools/web/workbooks/deleteWorkbook.ts
Alon-ST-DATA: [Alon's agent]

Resolved on the `main` merge (#429). `deleteWorkbook.ts:124` now reads `idempotentHint: false` with the explanatory comment restored, including the `tableau/tableau-mcp#392` reference (lines 120-123) — byte-for-byte consistent with the `deleteDatasource.ts` sibling. The `true` you saw was the pre-merge state; #429's fix is back in.

src/tools/web/workbooks/confirmDeleteWorkbook.ts
Alon-ST-DATA: [Alon's agent]

Fixed in `221e35ad`. Flipped `confirmDeleteWorkbook.ts` `idempotentHint` to `false` with a rationale comment matching the primary delete tool and the #392 resolution. The four confirm tools now tell a consistent story (`confirm-update-cloud-extract-refresh-task` keeps `true` — update semantics, per your note).

docs/docs/tools/data-qna/delete-datasource.md
Alon-ST-DATA: [Alon's agent]

Done in `221e35ad`. Dropped "off-by-default" here and applied the same edit to the two sibling docs that had the identical phrasing (`delete-extract-refresh-task.md`, `update-cloud-extract-refresh-task.md`) so they stay consistent.

src/web/apps/src/mcp-app.ts
Alon-ST-DATA: [Alon's agent]

Done in `221e35ad` — split by functionality as you asked.

- **embed-viz** stays in `mcp-app.html` + `src/mcp-app.ts`, now driven by `handleEmbedResult` (embed-Tableau-viz path only). Yours to rename later to something viz-specific.
- **hitl-confirm** is a new `hitl-confirm.html` + `src/hitl-confirm.ts` entry driven by `handleConfirmResult` (the four delete/update confirm-panel clients). No confirm logic bleeds into the embed entry anymore.

Wiring:
- `getAppConfig(toolName, bundle)` gained an `AppBundle = 'embed-viz' | 'hitl-confirm'` param (defaults to `embed-viz`, so `get-view`/`get-workbook` are byte-unchanged). The four delete/update tools pass `'hitl-confirm'`, so their `resourceUri`/`htmlPath` resolve to the confirm bundle.
- `build.ts` builds and copies both single-file bundles. Note: `viteSingleFile` with a single multi-input build did **not** reliably inline both entries, so it runs one `viteSingleFile` pass per entry (`emptyOutDir: false`). Verified both `build/web/apps/dist/{mcp-app,hitl-confirm}.html` emit fully inlined (no external chunk refs), embed bundle carries none of the confirm code and vice-versa.
- `handleToolResult`/its test split into `handleEmbedResult` + `handleConfirmResult` (+ tests). `npm run build` green, unit suite 2204 passing.

Alon-ST-DATA:
[Alon's agent]

Agree — the critique is accurate and I've grounded it against the code:

- `AppApprovalEvidence.establish()` runs in the model-driven **preview** callback (`evidence.ts`), keyed on `siteId:userLuid:namespace:targetId` — all server-known values, nothing bound to an in-iframe gesture. `verify()` only checks presence + single-use.
- `meta.ui.visibility: ['app']` is a **client-side hint**, not server enforcement; the `confirm-*` tools are still registered and callable.
- For the two task tools there's no tag layer, so `AppApprovalEvidence` is the **sole** gate (`confirmDeleteExtractRefreshTask.ts` / `confirmUpdateCloudExtractRefreshTask.ts` — `AppApprovalEvidence` alone, vs `AllEvidence([TagEvidence, AppApprovalEvidence])` on the workbook/datasource confirm tools).

So a non-cooperating MCP client can drive `preview → confirm-*` back-to-back with no human. Your options 1 (server caller-identity check) and 2 (mint approval only from an in-iframe postMessage to a model-unreachable endpoint) are genuine new server plumbing → follow-up WI, tracked under W-23125362, not this PR.

For #432 I took **option 3** (`221e35ad`):
- Retitled the PR to "MCP-Apps-cooperative HITL for all mutation tools".
- Reframed the three MCP-Apps doc sections to "cooperative human-in-the-loop" and added a **"Cooperative, not server-enforced"** warning to each, explicitly calling out that a non-cooperating client can bypass the human gesture — and that the two task tools have no tag layer so app-approval is the only gate.
- Softened the four `confirm-*` tool JSDoc/inline comments that claimed the tool is "invisible to and uncallable by the model" to state it's a cooperating-client hint and name what the server actually enforces.

This matches the existing scoping note that server-enforced HITL is follow-up work. PTAL.

src/tools/web/workbooks/deleteWorkbook.ts
Akash-Rastogi: [Akash's agent] Fixed in `221e35a`. `deleteWorkbook.ts:124` now reads `idempotentHint: false` with the `#392` rationale comment on lines 120-123 — restored via the main-branch merge as you noted.

src/tools/web/workbooks/confirmDeleteWorkbook.ts
Akash-Rastogi: [Akash's agent] Fixed in `221e35a`. `confirmDeleteWorkbook.ts:73` now `idempotentHint: false` with matching rationale on lines 70-72. Four confirm tools consistent as intended.

Akash-Rastogi (approved):
[Akash's agent] Second-round review after `221e35a` (+ merge with main bringing in the PR #429 mutation guard).

**All 3 prior findings resolved — confirmed in code:**

- **HIGH (framing)** — PR retitled to `MCP-Apps-cooperative HITL to all mutation tools`. All three affected docs (`data-qna/delete-datasource.md`, `tasks/delete-extract-refresh-task.md`, `tasks/update-cloud-extract-refresh-task.md`) added a `:::warning[Cooperative, not server-enforced]` callout, and each explicitly calls out that server-enforced HITL is still tracked as a follow-up (W-23125362). All four confirm tools' JSDocs + the `visibility: ['app']` annotation lines now include the `Cooperative-client hint: cooperating clients hide this from the model (not server-enforced)` disclaimer. The two task tools' warning callouts also explicitly note that `AppApprovalEvidence` is their only gate (no tag layer) — the exact framing gap I flagged.
- **LOW #1** — `deleteWorkbook.ts:124` now `idempotentHint: false` with the `tableau/tableau-mcp#392` rationale comment restored (came in via the main-branch merge).
- **LOW #2** — `confirmDeleteWorkbook.ts:73` now `idempotentHint: false` with matching rationale; the four confirm tools now tell a consistent story (`confirm-update-cloud-extract-refresh-task` correctly stays `true` — update semantics).

**Spot-checks on the new app-bundle split (jarhun88's request, also in `221e35a`):**

- `AppBundle` type in `appConfig.ts` is `'embed-viz' | 'hitl-confirm'`; `BUNDLE_HTML` maps each to a distinct html.
- All four delete/update preview tools pass `'hitl-confirm'` to `getAppConfig` explicitly (`deleteWorkbook.ts:85`, `deleteDatasource.ts:92`, `deleteExtractRefreshTask.ts:65`, `updateCloudExtractRefreshTask.ts:120`). All embed tools (`getWorkbook`, `getView`) use the `'embed-viz'` default. No cross-bundle routing leakage.
- `hitl-confirm.ts` entry imports only `handleConfirmResult` — no embed-viz logic bleeds into the confirm bundle.
- `handleConfirmResult` is pure shape-check + delegate to the four pre-existing renderers (already reviewed on the earlier PRs), with a fall-through error UI for unknown shapes. No new innerHTML surface.
- `build.ts` emits both bundles via per-entry `viteBuild` passes with `emptyOutDir: false`; commented rationale (multi-input viteSingleFile does not reliably inline every entry) is sound.

Approving.

Note: `mergeable_state: blocked` on this PR — presumably a required-check or missing-approval gate; nothing in the diff review requires further changes from my end.

src/tools/web/extractRefreshTasks/confirmUpdateCloudExtractRefreshTask.ts
```diff
@@ -0,0 +1,153 @@
...
+              // Require a fresh, single-use in-iframe human approval recorded by the preview.
+              const guardResult = await guardMutation({
+                restApi,
+                extra,
+                tool: 'confirm-update-cloud-extract-refresh-task',
+                previewTool: 'update-cloud-extract-refresh-task',
+                action: 'update',
+                mode: 'preview-confirm',
+                phase: 'confirm',
+                evidence: new AppApprovalEvidence('update-cloud-extract-refresh-task'),
```
mattcfilbert, line 104: Can we bind the app approval to the previewed *schedule*, not just the task? The flag-off `RegistryEvidence` path already passes `binding: scheduleBinding(args.schedule)`, but this app-only confirm path verifies task-level `AppApprovalEvidence` only — so previewing schedule A for task X authorizes a direct `confirm-update-cloud-extract-refresh-task` call for task X with schedule B, and B gets applied. Suggested fix: include the binding in `AppApprovalEvidence`'s key and pass the same schedule binding from both preview establishment and confirm verification, plus a schedule-swap regression test.

*— MattGPT (automation on Matt's behalf)*

mattcfilbert (commented):
Scope question: should this really say "all mutation tools"? At PR head, `readOnlyHint:false` also includes OAuth/session mutation tools (`revoke-access-token`, `reset-consent`) and Desktop mutation/cache-write tools, while this PR covers the four web admin content/task mutation pairs. If the intended scope is "MCP-Apps cooperative HITL for web admin content/task mutations," I'd tighten the title/body/docs to that — otherwise a rebase onto the newer Desktop authoring surface will quietly add more ungated mutators.

Two inline comments below on the approval binding and the audit contract — those are the ones I'd want addressed before this lands.

*Posted by MattGPT (automation on Matt's behalf; Matt reviewed and approved these comments).*

src/tools/web/datasources/confirmDeleteDatasource.ts
```diff
@@ -0,0 +1,158 @@
...
+                resolveTarget,
+              });
+              if (guardResult.isErr()) {
+                return guardResult.error.toErr();
+              }
+              const { target } = guardResult.value;
+              const projectName = target.project ?? 'unknown project';
+              const ownerText = target.owner ? `owner ${target.owner}` : 'owner unknown';
+
+              await restApi.datasourcesMethods.deleteDatasource({ datasourceId, siteId });
```
mattcfilbert: The app-only confirm tools need to call `recordOutcome` after the REST mutation. `guardMutation`'s contract says the `allowed` record is the authorization decision only and callers MUST emit `completed`/`failed` afterward — these new confirm tools call the destructive REST methods directly after `guardResult.value`, so a Tableau API failure leaves the audit trail at `allowed` with no terminal record (and success lacks `completed`). Applies to all four confirm tools; the existing model-visible confirm tests already pin the `allowed → completed/failed` sequence, so the app-only tests should match.

*— MattGPT (automation on Matt's behalf)*

src/tools/web/extractRefreshTasks/confirmUpdateCloudExtractRefreshTask.ts
Alon-ST-DATA: [Alon's agent] Fixed in c38a4ae6. `AppApprovalEvidence.approvalKey` now folds `ctx.binding` into the key (`site:user:namespace:targetId:binding`, empty segment when absent so the payload-free delete confirms are unchanged). The update-cloud preview establishes the approval with `binding = scheduleBinding(schedule)`, and confirm-update-cloud computes the same fingerprint via the now-exported `scheduleBinding()` and passes it to `guardMutation` — so an approval minted for schedule A no longer satisfies a confirm carrying schedule B.

Regression tests added: approve-A/confirm-B is rejected with no update (evidence + confirm-tool suites), a mismatched-binding verify does not consume the single-use approval, and the absent-binding delete flow is unchanged. Also verified live on an isolated flag-ON server: the swap now fails closed (`preview-not-run`, no REST call) while the matching-schedule happy path still applies once.

src/tools/web/datasources/confirmDeleteDatasource.ts
Alon-ST-DATA: [Alon's agent] Fixed in d88790fd. All four app-only confirm tools now call `recordOutcome` after the REST mutation, mirroring the model-visible sibling tools exactly:

- confirm-delete-datasource / confirm-delete-workbook / confirm-delete-extract-refresh-task: the delete REST call is wrapped in try/catch — `recordOutcome({ ok: false, failureDetail: getExceptionMessage(e) })` then rethrow on failure, `recordOutcome({ ok: true })` on success.
- confirm-update-cloud-extract-refresh-task: inspects the Result — `recordOutcome({ ok: false, failureDetail })` on the tableau-api and generic error branches (before the UnknownError returns), `recordOutcome({ ok: true })` before the success Ok.

Tests now pin the same sequence the model-visible confirm tests assert: happy path emits `[allowed,completed]` (all phase `confirm`), and a REST failure emits `[allowed,failed]` with failureDetail. Denied/not-admin/preview-not-run paths still emit exactly one `denied` record. delete-workbook confirm gained its first REST-error test. Full suite 2210/2210, tsc + lint clean.

src/web/apps/src/mcp-app.ts
```diff
@@ -5,12 +5,12 @@ import type { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
...
 import pkg from '~/package.json';
 
-import { handleToolResult } from './lib/handleToolResult.js';
+import { handleEmbedResult } from './lib/handleEmbedResult.js';
 
 const app = new App({ name: 'Tableau MCP App', version: pkg.version });
 app.ontoolresult = (result: CallToolResult) => {
-  void handleToolResult(app, result).catch((err) => {
-    console.error('[mcp-app] Unhandled error in handleToolResult:', err);
+  void handleEmbedResult(app, result).catch((err) => {
```
jarhun88: rename is not necessary

jarhun88 (approved):

src/web/apps/src/mcp-app.ts
Alon-ST-DATA: [Alon's agent] Reverted the rename back to `handleToolResult` (function, file, and test) in cbb9ae91. Agreed it was unnecessary churn here.

mattcfilbert (approved):
lgtm

mattcfilbert:
Re-review at head d88790fd, focused on the two P1s from the earlier review packet plus the new commits.

**Both P1s verified closed:**

1. **Schedule-binding bypass** — `AppApprovalEvidence` now keys approvals with `ctx.binding` (`src/tools/web/_lib/evidence.ts:211`); the update-cloud preview establishes approval with the binding, and confirm recomputes the same `scheduleBinding(args.schedule)` (`confirmUpdateCloudExtractRefreshTask.ts:92,106`). Tests pin binding A rejecting binding B and non-consumption on mismatch.
2. **Terminal audit outcome** — all four app-only confirm tools now emit `completed`/`failed` via `recordOutcome` after the allowed decision, with tests asserting both `['allowed','completed']` and REST-failure `['allowed','failed']` paths across the confirm suites.

One open question, your call: `update-cloud-extract-refresh-task` is now `idempotentHint: false` while the app-only confirm update remains `idempotentHint: true` (`updateCloudExtractRefreshTask.ts:170` vs `confirmUpdateCloudExtractRefreshTask.ts:64`). Intentional? Both paths consume single-use evidence, so a same-args retry fails closed rather than reapplying — the hints read inconsistently even if the state argument works.

nit (welcome to wont-fix): a couple of confirm tool descriptions still name sibling tools / client-visibility mechanics (`confirmDeleteDatasource.ts:58`, `confirmUpdateCloudExtractRefreshTask.ts:50`); since they're app-only this is low stakes.

Nice work closing these narrowly and test-pinned — the bundle split (`embed-viz` vs `hitl-confirm`) is the right direction too.

---

PR #435, opened 2026-06-30
Dplatt/improve pulse insights error handling

dplatt84:

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/errors/mcpToolError.ts
```diff
@@ -120,6 +120,94 @@ export class QueryValidationError extends McpToolError {
   }
 }
 
+const PULSE_INSIGHTS_ERROR_GUIDANCE: Record<string, string> = {
```
jarhun88: can you move this into a new file to track consts for pulse

dplatt84: Did that in latest commit

src/errors/pulseInsightsApiError.ts
```diff
@@ -0,0 +1,89 @@
...
+  '400000':
+    'General validation error. Check that: version is 1, at least one metric is provided, all metric keys are unique and non-empty, and input counts are within limits.',
+  '404936': 'Missing datasource ID or definition specification.',
+};
+
+export class PulseInsightsApiError extends McpToolError {
+  constructor(statusCode: number, responseData: unknown) {
+    const { errorCode, tabCode, guidance } = PulseInsightsApiError.parseResponse(responseData);
+
+    const parts: string[] = [];
```
jarhun88: i think it would be easier if the logic of formatting the error lived outside the of the PulseInsightsApiError class, as the constructor is really only meant for setting the fields, similar to how other Error classes are implemented. 

This will make it simpler for others using the PulseInsightsApiError to just use it without knowing that their inputs need to be formatted a specific way for the fields to be set as how they expect.

src/errors/pulseInsightsApiError.ts
```diff
@@ -0,0 +1,89 @@
...
+  '400970': 'Unsupported field type for the requested operation.',
+  '400971':
+    'Unknown definition specification type. Use either basic_specification, abstract_query_specification, or viz_state_specification.',
+  '400972': 'Time dimension must be absent when both range and comparison are unspecified.',
+  '400000':
+    'General validation error. Check that: version is 1, at least one metric is provided, all metric keys are unique and non-empty, and input counts are within limits.',
+  '404936': 'Missing datasource ID or definition specification.',
+};
+
+export class PulseInsightsApiError extends McpToolError {
```
jarhun88: the class itself can still live in mcpToolError.ts just without any of the logic so its just a setter

dplatt84: I think latest does this

src/errors/pulseInsightsApiError.ts
dplatt84: I think I rectified this with latest

src/errors/pulseInsightsApiError.test.ts
```diff
@@ -0,0 +1,76 @@
+import { PulseInsightsApiError } from './mcpToolError.js';
+import { formatPulseInsightsApiError } from './pulseInsightsApiError.js';
```
jarhun88, line 1: move these tests to pulseInsightsApiError.test.ts

src/errors/mcpToolError.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    ['400946', 'No granularity specified'],
+    ['400947', 'No range specified'],
+    ['400972', 'Time dimension must be absent when both range and comparison are unspecified'],
+  ])('provides guidance for error code %s', (code, expectedFragment) => {
+    const result = formatPulseInsightsApiError(400, { code, message: '0x00000000' });
+    expect(result.message).toContain(expectedFragment);
+  });
+});
+
+describe('PulseInsightsApiError', () => {
```
jarhun88: a setter test is not necessary

dplatt84: addressed

src/errors/pulseInsightsApiError.test.ts
dplatt84: addressed i believe

src/errors/pulseInsightsApiError.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * Human-readable guidance for Pulse service error codes.
+ * Codes are from ntbue-service-chassis/errors/service_error_codes.go.
```
jarhun88: probably we should not expose this as this repo is public facing and your mentioning an internal codebase

jarhun88 (approved):
approved with one final nit to address!

src/errors/pulseInsightsApiError.ts
dplatt84: Addressed I believe

PR #439, opened 2026-07-02
@W-12345678 Docs/7/1

joeconstantino:
what changed:

better instructions for site wide disable with sample payload


jarhun88 (approved):

package.json
```diff
@@ -107,4 +107,4 @@
   "engines": {
     "node": ">=22.7.5"
   }
-}
+}
```
stephendeoca, line 110: nitpick: can you make sure there's no changes to the end of line for this file and package-lock.json

joeconstantino: Not seeing one.

PR #455, opened 2026-07-07
@W-23336786 Add flexibility to PR title check (docs: prefix and skip-title-check label)

### LINKED ISSUE #453: Add more flexibility to PR title check action
The PR title check action (https://github.com/tableau/tableau-mcp/blob/main/.github/workflows/pr-title-check.yml) exists to ensure we have work item numbers in the title (to match internal Salesforce conventions).

This is still a valuable check, but we want more flexibility and have the check bypass that check and return true if any of these conditions exists:

* the PR title starts with "docs:"
* the PR is labelled with `skip-title-check`

bcantoni:
This change enables the PR title check to pass under either of these conditions:
* the PR title starts with "docs:"
* the PR is labelled with skip-title-check

Otherwise the normal must start with "@W-nnnn" will apply.

Co-authored with Claude Code.

Closes #453


bcantoni:
Claude helped me find the simple fix here - just needed to be updated from main to fix that test failure

---
⏺ Research findings

  I traced this using the systematic-debugging process (Phase 1: root cause investigation). Here's what's actually going on — it's more interesting than a simple flaky test:

  The failing CI run is from your open PR #455 (bcantoni/W-23336786-pr-title-check-flexibility), which only touches .github/workflows/pr-title-check.yml. It doesn't touch getStaleContentReport.test.ts at all — but that file
  still exists on the PR's base commit, because the branch is old (based on commit 12d5386e, well before later main history).

  The actual bug, in that stale copy of the test: the test hardcodes 'Last Accessed At': '2026-04-15T00:00:00Z' for a "recent" fixture row (wb-recent) and computes daysSinceLastUse against the real system clock (new Date()),
  comparing it to a 90-day threshold. The test assumed "now" would always be less than 90 days after 2026-04-15. Once real UTC crossed 2026-07-14 (90 days later), wb-recent aged past the threshold and started showing up in the
  "stale" rows — exactly the assertion failure you saw (expected [...] to not include 'wb-recent').

  This was already found and fixed once, then deliberately reverted:
  - ebd86b44/4cb7d5fb (copilot-agent, 2026-07-15): pinned the clock with vi.useFakeTimers + vi.setSystemTime('2026-05-20') to fix exactly this flake.
  - beda5618 (you, 2026-07-17): reverted it, noting "Still want to fix this, but in a separate PR."

  Key fact that changes the plan: on main, this problem no longer exists at all. The same day (2026-07-14), PR #498 (a5757f10, "remove deprecated admin tool shims Phase 2") deleted getStaleContentReport.ts/.test.ts entirely —
  the tool was consolidated into the polymorphic queryAdminInsights dispatcher. That dispatcher's own stale-content test (queryAdminInsights.test.ts) uses a relative offset (400 days before "today") instead of a fixed date vs.
  a fixed threshold — so it can never go flaky this way. main is clean.

  So this isn't a recurring CI-environment flake to defend against broadly — it's a dead test file that only survives because PR #455's branch predates the consolidation.

yyk16293 (approved):

PR #460, opened 2026-07-08
SEA seam adoption: binder manifests + runtime template reads via embedded assets

mattcfilbert:
Follow-up owed from #459/#433: closes the SEA self-containment hole flagged there.

- The last three runtime `getTemplatePath`+`readFileSync` sites (bind-template inject leg, dashboard-auto-apply per-ask loop, build-and-apply-worksheet) now read through the SEA-aware `readTemplate()` seam, fail-closed per each site's existing error path
- Binder manifest loading (`loadBinderFixture`/`loadManifests`) reads via `readDataAsset`, so a SEA binary with **no disk data dir can run the bind fast path** — `manifest.ts:140,160` disk reads are gone
- Remaining disk reads are by design (user workbooks, tool-arg caches, memo sidecar, CORPUS_PATH override) or flagged follow-up (`intelligence/provider.ts` — propose/factory surface only, not the bind path)

Evidence: 2761 tests green on the merged base, tsc clean; acceptance grep shows zero runtime template/manifest disk reads under `src/desktop/binder` + `src/tools/desktop`.

*Opened by MattGPT (Matt's Claude Code automation) on Matt Filbert's behalf.*


mattcfilbert:
needs a rework

PR #467, opened 2026-07-08
desktop: drop additive-POST workaround now that External Client API upserts by sheet name

lauren-jacksonSFDC:
## Description

The Tableau Desktop External Client API ("Athena V0") `POST /v0/workbook/document` used to be **additive** on sheet-name collision — a sheet whose name already existed was re-added and auto-renamed `Sheet 1 (2)`, `(3)`, … causing the sheet-explosion we worked around client-side. Monolith PR #59484 changes the POST to **upsert**: a colliding sheet is overwritten in place, and live sheets absent from the posted doc are still left untouched (it is *not* a full-workbook replace — same merge-at-sheet-granularity model, minus the duplication).

This removes the now-redundant delete-first workaround:

- **Per-sheet paths** (`loadWorksheetXml.ts`, `loadDashboardXml.ts`) — dropped the pre-apply `deleteLiveSheet()` step; the minimal-doc POST now upserts the edited sheet directly.
- **Whole-workbook path** (`loadWorkbookXml.ts`) — collapsed `resetAndApplyWorkbook` (add `...thinking-` scratch sheet → delete every colliding sheet → POST → delete scratch) into a single `applyWorkbookText` POST under the apply lock.
- Deleted `deleteLiveSheet.ts`; removed `SCRATCH_PREFIX` and the `listWorksheets` scratch filter.
- **Kept `applyMutex`** — per-sheet applies are still fetch-modify-apply (read live workbook → splice edited sheet → POST), so overlapping applies still need serializing. Rewrote its comment to drop the stale delete-race rationale.
- **Kept the `buildMinimal*Doc` builders** — posting only the edited sheet is still the correct scoped-edit contract; the rationale shifts from "avoid duplication" to "upsert in place, leave the rest live."

Only the `externalApiEnabled` (TABLEAU_EXTERNAL_API) transport is affected; the Agent API path is untouched.

## Motivation and Context

The delete-then-POST workaround was a temporary band-aid put in before a demo because the API duplication fix wasn't ready. #59484 ships that fix, so the client-side machinery is now dead weight — extra round-trips, a scratch-sheet dance, and a filter that could hide legitimately-named user sheets.

## Type of Change

- [x] Bug fix (removes a workaround now that the underlying API bug is fixed)

## How Has This Been Tested?

- `npx tsc --noEmit` — clean
- `npx eslint` on touched dirs — clean
- Updated unit tests across the 4 affected files — **52/52 pass**. The per-sheet and whole-workbook tests now assert that **no** `delete-sheet`/`new-worksheet` is issued and that a single upsert POST carries the right minimal doc.
- Ran an adversarial diff review (correctness / concurrency-residual / test-coverage dimensions, each finding independently verified); no confirmed regressions.

> Not yet exercised end-to-end against a live Desktop build carrying #59484 — the behavior change is validated at the unit level against the documented upsert contract.

## Related Issues

- Depends on monolith PR #59484 (additive → upsert for `POST /v0/workbook/document`).

## Checklist

- [ ] Version bump — n/a for this internal feature-branch change
- [x] I have added/updated tests that prove the change works

---

## Update: whole-workbook apply is now authoritative (replace, not merge)

Follow-up commit `6e322a4`. The External Client API POST upserts by name but **never prunes** — so after removing the additive workaround, `apply-workbook` with *fewer* sheets left the omitted ones live (a merge). The Agent API path (the pre-Athena default) replaced the whole model, so omitting a sheet deleted it. This restores that authoritative behavior on the External Client API.

`loadWorkbookXml` (external path only) now calls `replaceWorkbook`: fetch live sheets → upsert POST → `tabdoc:delete-sheet` the live worksheets/dashboards the doc omitted. Two ordering rules, both verified live against a Desktop build:

1. **Delete after the POST** — every posted sheet is present first, so a delete can't hit Tableau's "can't delete the last sheet" refusal.
2. **Delete dashboards before worksheets** — Tableau silently refuses (returns success) to delete a worksheet still referenced by a live dashboard's zone. Deleting worksheets first left them behind; dashboards-first fixes it in a single apply.

**Scope safety:** only `loadWorkbookXml` prunes. The per-sheet paths (`loadWorksheetXml`/`loadDashboardXml`) call `applyWorkbookText` directly and stay pure upserts. All 6 `loadWorkbookXml` callers pass a full live workbook (fetch-mutate-reapply), so `bind-template`/`dashboard-auto-apply`/etc. prune nothing; only an explicit partial workbook via `apply-workbook` triggers deletion — matching the old Agent API.

**Verified live** (Desktop, External Client API): full-replace `[Sheet 1, 2 charts, dashboard] → [Sheet 1]` in one apply; selective drop of one worksheet keeps the rest; `bind`/re-bind still add/overwrite without pruning. Unit tests cover same-set (no deletes), omit-sheets-and-dashboard (deletes in dashboard-first order), POST-failure (no prune), and fetch-failure (no POST). 94/94 workbook-command tests pass.


mattcfilbert (commented):
Reviewed alongside the server-side change (monolith #59484) — I verified at that PR's head that the collision-policy param is honored only while `WorkbookXmlImportCollisionPolicy` is ON, so the flag is a live kill switch, which matters for one of the comments below. The workaround removal itself looks complete: `deleteLiveSheet`, the scratch-sheet machinery, and the old collision tests are fully gone, and the new tests pin the no-delete upsert sequence. Two P1-level items inline.

src/desktop/commands/workbook/loadWorkbookXml.ts
```diff
@@ -229,82 +225,42 @@ async function resetAndApplyWorkbook({
...
-  const result = await executor.executeCommand({
-    namespace: 'tabdoc',
-    command: 'delete-sheet',
-    args: { Sheet: scratchName },
-    signal,
-  });
-  if (result.isErr()) {
-    return result;
-  }
   return Ok.EMPTY;
```
mattcfilbert, line 258: P1: the prune can report success while live ≠ doc. The dashboards-before-worksheets ordering covers *stale* dashboards, but not a *kept* one: if the posted doc keeps a dashboard whose zones still reference a worksheet the doc omits, the upserted dashboard pins that worksheet, `delete-sheet` hits the silent refusal documented in the comment above (refusal returns success, so `deleted.isErr()` never fires), and we return `Ok` with the omitted worksheet still live. No validation rule rejects zone references to worksheets absent from the doc, and there's no post-prune readback.

Suggest a readback set-equality check after the prune before returning `Ok` (and optionally a validation rule rejecting dashboards that reference omitted worksheets). Same apply-must-not-lie bar #466 set for the POST itself.

src/desktop/commands/workbook/loadWorksheetXml.ts
```diff
@@ -71,9 +70,9 @@ export async function loadWorksheetXml({
     });
   }
 
-  // External Client API ("Athena V0") exposes no per-sheet route — tabui:load-worksheet is not
-  // in its command registry, and the whole-workbook POST is additive-only, so applying a single
-  // sheet has to delete the live copy first and re-post a minimal whole-workbook document.
+  // External Client API ("Athena V0") exposes no per-sheet route — tabui:load-worksheet is not in
+  // its command registry, so applying a single sheet re-posts a minimal whole-workbook document.
+  // The POST upserts by name: it overwrites the colliding sheet in place and leaves the rest live.
   return getDesktopConfig().externalApiEnabled
```
mattcfilbert, line 76: P1: with the pre-delete gone, this path hard-assumes the server upserts — but that behavior is feature-flagged (`WorkbookXmlImportCollisionPolicy`, Preliminary; #59484 flips the default, and flag-OFF reverts the server to additive-rename regardless of the caller's policy arg — verified in `ImportWorkbookAction`'s ctor at that PR's head). Against a flag-off or older Desktop, this POST recreates the renamed-duplicate failure mode and we report success; `TABLEAU_EXTERNAL_API=1` is the only gate.

Either fail closed on a capability/version check, or state the MCP↔Desktop build-pairing requirement explicitly (docs and/or error text) so a flag flip server-side can't silently degrade applies.

lauren-jacksonSFDC: Not necessary here - the Desktop code will ship at the same time as this code. There will be no gaps in any released versions as this is a fast-follow

src/desktop/commands/workbook/loadWorkbookXml.ts
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

Fair — but the prune can't create a dangling reference: a kept dashboard's worksheet is either in the doc or left live by the upsert, so it stays intact. A genuinely dangling zone ref can only come from the POST body itself, which `SetWorkbookXmlCommand` applies the same way Tableau opens a `.twb`. So this behaves like twb-open by construction — no reconcile change needed.

PR #472, opened 2026-07-09
desktop: verify every embedded SEA asset against asset-manifest hashes

tableaukyler:
## Description

Fixes a SEA (Single Executable Application) regression from #460 where the desktop binary threw `template asset files are missing from the bundle` (actual error: `SEA data asset 'templates/…' is missing from content-manifest.json`) when an agent used template tools.

- `buildSea.ts`: `asset-manifest.json` now maps each embedded key to `{sha256, bytes}` (was a bare key array), hashing every file as it is embedded. The asset-map builder lives in `src/scripts/seaAssets.ts`, imported by both the production embed path and the e2e guard.
- `assets.ts`: a single `readVerifiedSeaAsset` path verifies **every** embedded asset (`desktop/data/*` and `resources/desktop/*`) against that one manifest. Removed the separate content-manifest-based gate. No verified/unverified fork. Verification hashes raw asset bytes (matching the build-time byte hash), then decodes to text — so a non-UTF-8 asset under either root can't false-reject.
- `assets.test.ts`: rewritten to the new manifest shape; covers fail-closed (missing/corrupt/wrong-shape manifest, hash mismatch, unlisted→null) and raw-byte verification.
- `tests/e2e/seaAssets.test.ts`: runs a real desktop build, calls the shared `buildAssetsMap` over the configured asset roots, and reads every embedded asset back through the verified seam — so a dropped root or keying change trips a test instead of only a runtime throw.

## Motivation and Context

#460 routed the legacy `templates/` reads — plus `corpus.json`, `twb-example-index.json`, `workbook-schema-reference.json`, `tableau-desktop-commands-reference.json`, and `examples/` — through the SEA asset seam. But `readVerifiedSeaDataAsset` enforced integrity against `content-manifest.json`, which by design only covers authoring content (`template-manifests/` + `data-visualization-templates-xml/`). So under a SEA binary all six asset roots failed closed; `templates/` was just the first one an agent exercised.

The pre-existing lockstep test only proved manifest→staged, never staged-runtime-read→manifest, so the gap went uncaught. Hashing at embed time makes coverage self-maintaining: any newly embedded asset is automatically verifiable, and `content-manifest.json` stays a pure served authoring artifact.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Full unit suite (`npm test`); `tsc --noEmit` clean; `eslint` clean on changed files.
- `src/desktop/assets.test.ts` covers the fail-closed gate (tamper→throw, unlisted→null, missing/corrupt/wrong-shape manifest→throw) and raw-byte verification (a non-UTF-8 asset passes when its build-time byte hash matches).
- `tests/e2e/seaAssets.test.ts` builds a real desktop variant, runs the shared `buildAssetsMap`, and reads every embedded asset back through the verified seam. Verified as a real guard: temporarily dropping a root from the shared asset-dir list makes the test fail.
- End-to-end: built a desktop SEA binary (`npm run build:sea -- --variant desktop`) and drove `inject-template` reading `templates/box-plot-chart.xml` (the exact line that threw) — it read the embedded, hash-verified template and injected it successfully.

## Related Issues

Follows up on #460.

## Checklist

- [ ] I have updated the version in the package.json file by using `npm run version`. (Not bumped — feature/authoring integration branch.)
- [x] I have made any necessary changes to the documentation (none needed)
- [x] I have documented any breaking changes in the PR description. (None — internal SEA manifest format only.)


mattcfilbert:
Reviewed with a focus on the integrity pipeline and the read paths. The core design is right: hashing raw bytes at embed time into `asset-manifest.json` `{sha256, bytes}` makes coverage self-maintaining, both read paths route through the single `readVerifiedSeaAsset` gate (unlisted → null, listed-but-tampered → throw), and the `_verifiedAssets` cache keeps it one hash per key. The root-cause writeup (content-manifest only ever covered authoring content, so #460's rerouted roots all failed closed) matches the code.

One gap worth closing: `assets.seaStaged.test.ts` builds its own fake SEA store by walking `DATA_DIR` and constructing keys/manifest inline (`walk(DATA_DIR, [], files)` → `` `desktop/data/${rel}` `` → `store['asset-manifest.json'] = ...`). That proves the *verifier* round-trips, but never exercises the real `buildSea.ts` asset-map builder — if a future change drops a root from `assetDirs` or changes keying, this guard still passes and we're back to discovering it via a runtime throw, which is the failure mode this PR is fixing. Suggest exporting the real asset-map builder (or generating the desktop-variant manifest in the test) and asserting the staged roots against it. Your manual SEA e2e covered this once; the test should keep covering it.

FYI for sequencing: #469 (open) also touches `src/desktop/assets.ts` (adds a repo-root `resources/desktop` source fallback for vitest/source mode). No structural conflict with this PR as shaped — SEA reads return from the verified path before any disk fallback — but whichever lands second should preserve that property (no "verified read returned null → try disk" under SEA).

---

mattcfilbert:
Second-pass review (dual-model, adjudicated). **The gap from my first pass is closed.** `85b12e7c8` extracted the asset-map builder into the shared `src/scripts/seaAssets.ts` imported by *both* `buildSea.ts` (the production embed path) and the new `tests/e2e/seaAssets.test.ts`, and deleted the hand-rolled `assets.seaStaged.test.ts` walk. The keying logic (`relative(buildDir, file).split(sep).join(posix.sep)`) now lives in one place both consume, so a future root-drop or keying change trips the e2e test instead of only surfacing as a runtime throw. Exactly what I was asking for — thanks.

Re-verified after the refactor and all holds: single `readVerifiedSeaAsset` gate for both read paths (unlisted → null, listed-but-tampered → throw), `_verifiedAssets` cache keyed by full key + reset in `_setSeaApiForTest`, manifest not self-listed (built before `assets[MANIFEST_KEY]` is appended), parse hardening rejects array/null/non-object/malformed entries, and `writeFile` still has a live call-site so no unused-import lint fallout.

**One standing item — latent raw-bytes vs. utf8-string hash mismatch (P2, not live today).** The manifest hashes raw file bytes at build:

```ts
const buf = await readFile(assets[key]);
manifest[key] = { sha256: createHash('sha256').update(buf).digest('hex'), bytes: buf.byteLength };
```

but the runtime verifies the utf8-decoded string:

```ts
const asset = sea.getAsset(key, 'utf8');
// ...
const actualHash = sha256(raw);
const actualBytes = Buffer.byteLength(raw);
```

For valid UTF-8 text these are byte-identical, and I confirmed no binary/non-UTF-8 asset lives under `resources/desktop/` or `desktop/data/` today (images are all under `docs/`), so this is latent. But the day a `.png`, a BOM, or any non-UTF-8 byte sequence lands under either root, it fails the integrity check at runtime — re-creating the #460 fail-closed regression this PR fixes. Cheap insurance: either verify raw Buffer bytes at runtime (hash the Buffer, decode to text after the check), or add a build-time guard in `buildAssetsMap` that rejects non-UTF-8 files under these roots and documents the text-only contract. A binary fixture in the e2e builder would pin it (`readResourceAsset` would currently throw).

**Nits:**
- **PR body drift:** "How Tested" still names the deleted `assets.seaStaged.test.ts` and frames the e2e file as asserting fail-closed. The substance is true — fail-closed *is* covered (tamper→throw, unlisted→null, wrong-shape→throw live in `src/desktop/assets.test.ts`; the e2e file does the real-builder round-trip) — but the filename/framing are stale. Worth a quick edit.
- **Residual builder coupling:** the e2e calls `buildAssetsMap(DESKTOP_ASSET_DIRS, 'desktop')` directly rather than via `seaVariants.desktop.assetDirs`. Both bind the same const today (so root-drops are caught), but re-pointing `seaVariants` in `buildSea.ts` would slip past. Much smaller than the original gap — optional.

**Sequencing (still valid):** #469 also touches `assets.ts` adding a repo-root disk fallback. Whichever lands second must preserve "SEA reads return from the verified path before any disk fallback" — no "verified read returned null → try disk" under SEA, or unlisted assets escape the gate.

---

PR #473, opened 2026-07-09
@W-23202054: Stop get-stale-content-report from silently ignoring invalid projectIds

Alon-ST-DATA:
## Bug (W-23202054, P3)

`get-stale-content-report` accepts a `projectIds` list to scope the report. Two resolution paths silently dropped requested IDs, and the worst case silently **widened the report to the entire site**:

1. `resolveProjectScopeIds` dropped IDs outside the server's bounded context (`INCLUDE_PROJECT_IDS`) with no signal.
2. `resolveProjectIdsToNames` intersected requested IDs against real site projects; IDs matching no project vanished.

When **every** requested ID was invalid, `projectNameScope` collapsed to `[]`, and `buildSiteContentQuery` treated an empty scope identically to "no scope requested" → the `else` branch emits the unscoped full-site query. A typo in a project scope silently returned **all site content** — the opposite of what the caller asked for.

## Fix

- **Track dropped IDs** through both resolution paths, distinguishing `unknown-on-site` from `not-permitted-by-config`, and surface them in `mcp.warnings` (reusing the result convention from `query-datasource`).
- **Widening guard (core fix):** when a scope was requested but nothing resolved to a real project name, return an **empty report (0 rows) + warnings** — never fall through to the unscoped full-site query.
- **Back-compat:** no `mcp` key is emitted when all requested IDs are valid; the clean result shape is unchanged.

## Test coverage

Unit (`getStaleContentReport.test.ts`, 37/37 pass):
- **Partial-invalid** — scoped to the valid subset, one `unknown-on-site` warning listing the ignored ID; filter still applied for the valid project.
- **All-invalid (widening-guard regression)** — `rows: []`, totals 0, and **asserts `mockQueryDatasource` was NOT called** (proves no full-site query fires).
- **Bounded-context drop** — `not-permitted-by-config` warning for the out-of-scope ID.
- **All-valid back-compat** — no `mcp` key present.
- Direct unit tests for the new `resolveProjectScopeIds` / `resolveProjectIdsToNames` return shapes and `buildProjectIdWarnings`.

Full suite: 2223/2223 pass · `tsc --noEmit` clean · eslint + prettier clean.

**E2E: N/A** — the existing prompt-level e2e cannot deterministically force an invalid-ID site condition; this is a resolver-level fix owned by the unit tier. The all-invalid unit test is the authoritative regression proof (deterministic, fails without the guard, passes with it).

## Version

Patch bump 2.25.1 → 2.25.2 (CI version-bump-check gate).

### DIFF
--- a/docs/docs/tools/admin-insights/get-stale-content-report.md
+++ b/docs/docs/tools/admin-insights/get-stale-content-report.md
@@ -4,35 +4,38 @@ sidebar_position: 3
 
 # Get Stale Content Report
 
-Builds a deterministic report of stale Tableau Cloud content (workbooks and published
-datasources) by querying the Admin Insights `Site Content` datasource — which exposes a
-`Last Accessed At` field per item — applying the staleness threshold server-side, and returning
-already-filtered rows.
+Builds a deterministic report of stale Tableau Cloud content (workbooks and published datasources)
+by querying the Admin Insights `Site Content` datasource — which exposes a `Last Accessed At` field
+per item — applying the staleness threshold server-side, and returning already-filtered rows.
 
-The server applies the threshold comparison, optional project filter, and sort. Clients receive
-only items where days since last use exceed the threshold. **No client-side math is required.**
+The server applies the threshold comparison, optional project filter, and sort. Clients receive only
+items where days since last use exceed the threshold. **No client-side math is required.**
 
-The tool is admin-only — it is registered only when `ADMIN_TOOLS_ENABLED=true`, and at request
-time it verifies the caller's site role and rejects anything below
-`SiteAdministratorCreator` / `SiteAdministratorExplorer` / `ServerAdministrator`.
+The tool is admin-only — it is registered only when `ADMIN_TOOLS_ENABLED=true`, and at request time
+it verifies the caller's site role and rejects anything below `SiteAdministratorCreator` /
+`SiteAdministratorExplorer` / `ServerAdministrator`.
 
 ## APIs called
 
-- [Query Datasource (VDS)](https://help.tableau.com/current/api/vizql-data-service/en-us/reference/index.html#tag/HeadlessBI/operation/QueryDatasource) — single VDS call against the `Site Content` Admin Insights datasource
-- [Query Data Sources (REST)](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_data_sources.htm#query_data_sources) — used internally to resolve the `Site Content` dataset LUID
-- [Query Projects (REST)](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_projects.htm#query_projects) — used internally to resolve project LUIDs to names when a project scope is set
-- [Get User on Site (REST)](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_users_and_groups.htm#get_user_on_site) — used internally for the admin gate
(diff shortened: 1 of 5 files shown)


Alon-ST-DATA:
[Alon's agent]

## Live E2E verification — v2.25.2, 10ax test site, streamable HTTP, admin=on — 4/4 PASS

The PR body notes prompt-level E2E is N/A. This is manual **resolver-level live verification** against the built server, confirming the fix ships correctly.

| # | Case | Args | Observed | Verdict |
|---|------|------|----------|---------|
| 1 | All-invalid (widening guard) | `projectIds:["not-a-luid"]` | `totalStaleItems:0`, `rows:[]`, `mcp.warnings[0]`: `reason=unknown-on-site`, `ignoredProjectIds=["not-a-luid"]` | ✅ PASS |
| 2 | No-filter baseline | (none) | `totalStaleItems:3449`, no `mcp` key | ✅ PASS |
| 3 | Partial-invalid | valid LUID + bogus | 66 rows all scoped to valid project; `mcp.warnings[0].reason=unknown-on-site`, `ignoredProjectIds=[bogus]` | ✅ PASS |
| 4 | All-valid back-compat | valid LUID | 15 rows; no `mcp` key | ✅ PASS |

**Core proof:** all-invalid → **0 items** vs no-filter → **3449**. The widening guard holds live — a scope of a typo'd LUID returns an empty report + explicit warning, never the full site.

mattcfilbert:
Right fix — invalid `projectIds` are tracked and surfaced, and the all-invalid path no longer widens to a full-site report. The test shape covers the cases that matter (partial-invalid, all-invalid widening guard, all-valid back-compat), and nice use of `exportedForTesting` for the resolver helpers.

One thing to address before merge: the warning message is unconditional, so it can still mislead in exactly the way this PR is fixing. `"The report was scoped to the remaining valid projects."` (and the sibling `"...remaining permitted projects."`) also fires when *no* valid projects remain — e.g. `projectIds: ['bad-1', 'bad-2']` returns a correctly-empty report with a warning implying a valid subset existed. Suggest branching the message: all-invalid → something like "None of the requested projectIds resolved; returning an empty report." That gives an LLM caller a self-correction signal instead of a false premise.

Smaller question, your call: should the `not-permitted-by-config` message name `INCLUDE_PROJECT_IDS`? Useful for an admin reading logs, less so for an LLM caller — "outside this server's configured project scope" carries the same signal without coupling tool output to an env-var name.

---

Alon-ST-DATA:
[Alon's agent]

Thanks Matt — both addressed in b211448b.

**1. Unconditional warning message (real bug).** You're right: `"The report was scoped to the remaining valid projects."` fired even in the all-invalid case, asserting a subset that doesn't exist — the exact false premise this PR targets. The wording now branches on whether any requested project survived both drop paths:

- **some remain** → `"...The report was scoped to the remaining valid projects."`
- **none remain** → `"...None of the requested projectIds resolved to a project you can report on; an empty report (0 rows) was returned instead of the full site."`

So `projectIds: ['bad-1','bad-2']` now returns the correctly-empty report *and* a warning that says it's empty — the self-correction signal you described.

**2. `INCLUDE_PROJECT_IDS` in the message (your call → took it).** Dropped the env-var name from tool output — now `"outside this server's configured project scope"`. Same signal for an LLM caller, no coupling to a config key. (Left the env-var name in the docs page — that prose is admin-facing config reference.)

Tests: added `buildProjectIdWarnings` cases for both message branches + an assertion the env-var name isn't leaked; strengthened the all-invalid callback test to assert the empty-report wording. Full suite 2225/2225, tsc/eslint/prettier clean.

mattcfilbert (approved):

PR #474, opened 2026-07-09
Integrate desktop authoring line into core TMCP

mattcfilbert:
## Description

AI Assisted Authoring via MCP is coming to Tableau Desktop.

- Conflicts resolved keeping main's build infra (tsconfig.providers step, toolName registry, feature-gate layout) and folding in desktop additions (desktop asset staging, desktop agent-check checks, README sections)
- All 45 desktop tools carry the Required<ToolAnnotations> contract; hints chosen per tool semantics (execute-tableau-command destructive, cache mutators non-idempotent, readers read-only/idempotent)
- Serialized desktop tools/list surface fits the 46k deferral-cliff budget (45,572 measured) via schema-description compression only; tool descriptions and expertise:// steering pointers unchanged from authoring
- CI: desktop variant build step added; workflow also triggers on PRs to feature/desktop

### Latest: fix calc definition not traveling with template-applied worksheets (W-23826373)

When a template slot is bound to a user calc field (class='tableau'), the template-inject path renamed the hollow column stub to the bound field name but never carried the `<calculation>` formula body or formula-referenced dep columns into the worksheet's `<datasource-dependencies>`. On a live datasource Tableau resolves the formula from context; on an extract it silently blanks the chart.

Fix: new `spliceBoundCalcDefinitions()` in `groupDefinitionSplice.ts`, wired into both template-inject paths (`injectTemplateCore.ts` and `buildAndApplyWorksheet.ts`). Mirrors the existing `spliceBoundGroupDefinitions()` pattern. Also injects formula-referenced dependency columns that are absent from the sheet.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Validation: tsc clean, lint clean, web + desktop builds green
- 16 tests in groupDefinitionSplice.test.ts (8 new for spliceBoundCalcDefinitions)
- Live Desktop smoke test: trend-line-chart template applied with Profit Ratio (user calc) bound — 192 rows returned, calc definition confirmed in worksheet XML

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/desktop/metadata/fields.ts
```diff
@@ -0,0 +1,1064 @@
...
+
+  // CRITICAL: Update panes array back to worksheet structure
+  // This must be done to ensure the changes are reflected in the parsed structure
+  worksheet.table.panes.pane = panes.length === 1 ? panes[0] : panes;
+
+  // Debug: Verify the encoding was added before serializing
+  const verifyPanes = normalizeArray(worksheet.table.panes.pane);
+  const verifyPane = verifyPanes[0];
+  const verifyEncodings = normalizeArray(verifyPane?.encodings?.[encodingType]);
+  const encodingAdded = verifyEncodings.some((enc: any) => enc['@_column'] === columnRef);
```
tableaukyler, line 199: 🔴 **Verification checks the wrong ref — throws on real success.** The encoding was inserted with `correctedColumnRef` (line 145), but this verifies against the original `columnRef`. For any calc field whose formula already aggregates (the exact case the correction exists for), `correctedColumnRef !== columnRef`, so `encodingAdded` is `false` and this throws `Failed to add encoding...` on a successful add. Verify against `correctedColumnRef`.

src/desktop/metadata/fields.ts
```diff
@@ -0,0 +1,1064 @@
...
+    datasourceDep['column-instance'] =
+      columnInstances.length === 1 ? columnInstances[0] : columnInstances;
+  }
+
+  // CRITICAL: Update dependencies array back to worksheet structure
+  worksheet.table.view['datasource-dependencies'] =
+    dependencies.length === 1 ? dependencies[0] : dependencies;
+
+  // Return the corrected instance name so caller can use it
+  return correctedInstanceName;
```
tableaukyler, line 982: 🔴 **Returns the pre-correction name — dangling field binding.** When the base column isn't yet in `datasource-dependencies` (only in the workbook), the first correction block (~654) is skipped and the second block (~814) creates the instance as `actualColumnInstanceName` (`[usr:...]`) — but this returns `correctedInstanceName` (`[sum:...]`). The caller writes the returned name onto the shelf/encoding, so it references a column-instance that doesn't exist. Return `actualColumnInstanceName` (and ideally collapse the two correction blocks into one path).

src/desktop/externalApi/externalApiClient.ts
```diff
@@ -0,0 +1,223 @@
...
+    options: { signal?: AbortSignal; contentType?: string; body?: string },
+  ): Promise<Result<Response, ExternalApiError>> {
+    const headers: Record<string, string> = {
+      authorization: `Bearer ${this.instance.token}`,
+    };
+    if (options.contentType) {
+      headers['content-type'] = options.contentType;
+    }
+
+    const signal = options.signal ?? AbortSignal.timeout(this.timeoutMs);
```
tableaukyler: 🟠 **Per-request timeout is dead in the production path.** `options.signal ?? AbortSignal.timeout(...)` means the 60s floor only applies when no signal is passed — but `ExternalApiToolExecutor.callEndpoint` always threads a `signal`, so the timeout never engages. If the caller signal carries no timeout, a hung loopback socket leaves `fetch` outstanding indefinitely. Combine them: `AbortSignal.any([options.signal, AbortSignal.timeout(this.timeoutMs)])` so a caller signal augments rather than replaces the floor.

src/desktop/commands/workbook/loadWorkbookXml.ts
```diff
@@ -1,32 +1,148 @@
...
+ * brackets, a dropped worksheet, etc.) is surfaced in the response payload — a
+ * top-level `error`, or a failure signal inside `result` — not in `status`. The
+ * executor only maps `status: 'failed'` to an error, so without this check a
+ * rejected load is relayed as success (the P0 "apply must not lie" bug).
+ *
+ * Conservative by design: it only reports failure on an EXPLICIT failure signal,
+ * so a normal success (empty `result`, no `error`) is never turned into a false
+ * negative.
+ */
+export function interpretLoadOutcome(
```
tableaukyler: 🟠 **Payload-sniffing can convert real successes into errors.** This now gates the agent-API load success paths. Branch 1 (top-level `status.error`) is safe, but branches 2-4 key off generic field *names* (`status`, `error`, `errors`, `success`/`ok`/`loaded`) inside an arbitrary `result` payload the code doesn't control. A successful load that happens to carry one of those fields with unrelated semantics gets relayed as `load-rejected`. Worth confirming against real Desktop payloads before trusting the sniffing branches.

src/desktop/binder/classify.ts
```diff
@@ -0,0 +1,1260 @@
...
+    for (const slot of zeroSlots) {
+      const widened = pickUniqueMaxAffinity(schemaDims, geoAffinityTokens(slot.slot_id));
+      if (widened.kind !== 'ok') return null; // still zero, or now ambiguous → fail closed
+      picks.set(slot.slot_id, widened.field);
+      autoCompleted.set(slot.slot_id, widened.field);
+    }
+  }
+
+  const chosen = [...picks.values()];
+  if (new Set(chosen).size !== chosen.length) return null; // two slots, one field
```
tableaukyler, line 2183: 🟠 **Geo-slot widening can silently double-bind one field (fail-open).** This distinctness check compares geo picks only among themselves, never against fields already consumed by non-geo slots in `used`. The W60 zero-candidate widening (~887) resolves a geo slot over the full `schemaDims`. For a fast-path template with ≥2 geo slots plus another dimension slot, an ask naming both can bind the same field to a geo slot AND a non-geo slot, and validation passes — a silent wrong bind on a path whose contract is "fail closed." Fix: exclude `used` fields from the widened pool, or fold geo picks into a distinctness check that also includes the non-geo bound fields.

src/desktop/metadata/field-resolver.ts
```diff
@@ -0,0 +1,348 @@
...
+    case AggregationType.Avg:
+      return 'avg';
+    case AggregationType.Min:
+      return 'min';
+    case AggregationType.Max:
+      return 'max';
+    case AggregationType.Count:
+      return 'count';
+    case AggregationType.CountDistinct:
+      return 'countdistinct';
```
tableaukyler: 🟠 **Count-distinct emits an invalid derivation.** This returns `'countdistinct'`, producing `[ds].[countdistinct:Region:nk]`. `mapDerivationToProperCase` (in `fields.ts`) has no `countdistinct` key (only `ctd`/`countd`), so it falls through and writes `@_derivation="countdistinct"`, which Tableau won't accept. Every other aggregation round-trips. Use a prefix the map recognizes (`ctd`/`countd`).

src/desktop/metadata/fields.ts
```diff
@@ -0,0 +1,1064 @@
...
+
+  // Check if base column exists to determine if it's a calculated field
+  const columnsArray = normalizeArray(datasourceDep.column);
+  const existingBaseColumn = columnsArray.find((col: any) => col['@_name'] === parsed.column);
+  let correctedInstanceName = columnInstanceName;
+
+  // If it's a calculated field with aggregation, we need to use usr prefix
+  if (existingBaseColumn?.calculation?.['@_formula']) {
+    const formula = existingBaseColumn.calculation['@_formula'];
+    const aggFunctions = [
```
tableaukyler, line 774: 🟡 **Duplicated algorithm (root cause of the two blockers above).** This 13-element `aggFunctions` array + `hasAggregation` scan + `usr:` correction is copy-pasted verbatim again at ~817, and the quantitative-derivation list is duplicated twice more in this function (plus a third `aggFunctions` copy in `field-builder.ts:387`). Changing the recognized aggregation set means editing all copies in lockstep. Extract `formulaHasAggregation(formula)` and a shared `QUANTITATIVE_DERIVATIONS` set.

src/desktop/templates/injectTemplateCore.ts
```diff
@@ -0,0 +1,186 @@
...
+// binder's args as-is (matching what the manual inject-template call would receive).
+
+import { normalizeArray, parseXML, serializeXML } from '../metadata/parser.js';
+import { ParsedWindow, ParsedWorkbook, ParsedWorksheet } from '../metadata/types.js';
+import { wellFormedXmlRule } from '../validation/rules/wellFormedXml.js';
+import { spliceBoundFacet } from './facetSplice.js';
+import { rewriteFieldReferences } from './fieldReferenceRewriter.js';
+import { injectTemplate, InsertPosition, SheetType } from './injectTemplate.js';
+
+/** Escape the five XML metacharacters (identical to the inject-template tool). */
```
tableaukyler, line 70: 🟡 **Comment misdescribes the function below it.** This doc (`Escape the five XML metacharacters...`) sits directly above `ensureUserNamespace`; the function it actually describes (`escapeXml`) is further down with no doc. The two ported comments got interleaved — move it (or drop it per the NO-COMMENTS rule).

src/desktop/binder/classify.ts
```diff
@@ -0,0 +1,1260 @@
+// src/binder/classify.ts
```
tableaukyler, line 1: 🟡 **Stale header path.** Reads `// src/binder/classify.ts` but the tree moved under `desktop/`. Same stale first-line comment is on all 11 `src/desktop/binder/*.ts` files — per the delete-stale-comments rule, drop them.

src/desktop/wrappers/injectViewpoints.ts
```diff
@@ -0,0 +1,58 @@
...
+  workbookXml: string,
+  dashboardName: string,
+  worksheetNames: string[],
+): string {
+  const parser = new DOMParser({
+    errorHandler: (_level, _msg) => {},
+  });
+  const doc = parser.parseFromString(workbookXml.trim(), 'text/xml');
+
+  // Find the <window class="dashboard" name="<dashboardName>"> element
```
tableaukyler, line 23: 🟡 **WHAT-narration comments.** These (`// Find the <window...`, `// Remove any existing <viewpoints>`, `// Build new <viewpoints>`) restate what the next line does — flagged under the near-absolute NO-COMMENTS rule. The function's leading docblock explaining *why* viewpoints matter is the allowed kind and should stay. (Same applies to leftover `[DEBUG]` `console.error`/`console.warn` in the `fields.ts` apply path and the `Lane M6 milestone...` PR-status headers across the intelligence files.)

src/desktop/validation/rules/invalidDerivationString.ts
```diff
@@ -0,0 +1,146 @@
...
+import * as xpath from 'xpath';
+
+import type { ValidationIssue, ValidationRule } from '../types.js';
+
+/**
+ * Canonical `derivation` attribute values, case-sensitive and exact. Any
+ * `column-instance` derivation outside this closed set silently rewrites to None on
+ * load. Exported so other code can share the single allowlist.
+ */
+export const CANONICAL_DERIVATIONS = new Set<string>([
```
tableaukyler: 🟡 **Fail-closed allowlist that hard-ERRORs on unlisted-but-valid content.** Any `derivation` outside `CANONICAL_DERIVATIONS` becomes a blocking error — e.g. spatial `Collect` (MakePoint/MakeLine) isn't listed, so a legit live-readback workbook containing it would be hard-rejected as "invalid derivation." An allowlist that errors is fragile to omissions; consider warn-not-error for the "unknown but not obviously invalid" case. Same pattern in `connectionsNotAuthorable.ts` (terminal reject on any non-`federated` connection). Both are documented as intentional — flagging the fragility, not demanding a change.

tableaukyler (commented):
## Early review — desktop authoring integration

Solid, well-tested port with a clean validation/binder architecture. Two confirmed correctness bugs in the field-metadata apply path should block; a handful of important robustness risks in the network/apply paths are worth confirming before this leaves draft. Scope: I reviewed the ~12.6k lines of desktop-authoring TypeScript logic (skipped knowledge markdown, XML fixtures, generated data, tests).

**🔴 Blocking**
- `fields.ts` encoding-add verification checks the wrong ref — throws on real success (calc-field aggregation case)
- `fields.ts` `ensureColumnInstanceInDependencies` returns the pre-correction name — emits a dangling field binding

**🟠 Important**
- `externalApiClient` per-request timeout is dead in the production path
- `interpretLoadOutcome` payload-sniffing can convert real successes into errors
- `classify` geo-slot widening can silently double-bind one field (fail-open)
- `field-resolver` count-distinct emits an invalid derivation

**🟡 Nits** — duplicated aggregation logic (root cause of the two blockers), a comment that misdescribes the function below it, stale `// src/binder/...` header paths, WHAT-narration comments + leftover `[DEBUG]` logs, and a couple of fail-closed allowlists that hard-error on unlisted-but-valid content.

Great test coverage and the fail-closed security funnel (pack verification, SEA asset integrity, apply mutex) all checked out clean.

tableaukyler (commented):
(duplicate of the review directly above — double-posted by mistake; see the other review for the full findings)

PR #475, opened 2026-07-09
@W-23375797: consolidate admin-gated tools into polymorphic dispatchers

Akash-Rastogi:
[Akash's agent]

## Description

Introduces two new polymorphic MCP tools to reduce the model-visible admin tool
surface, while preserving every existing behavior end-to-end during a shim
window:

| New tool | Discriminator | Merges (still wired as legacy) |
| --- | --- | --- |
| `query-admin-insights` | `kind` | `get-stale-workbooks`, `get-stale-datasources`, `get-content-usage-metrics`, `get-user-activity` |
| `delete-content` | `resourceType` | `delete-workbook`, `delete-datasource`, `delete-extract-refresh-task` |

## Motivation and Context

The admin-gated set has grown to the point where its model-visible tool count
crowds the model's decision space and inflates prompt-time tool schemas.
`W-23375797` targets two consolidations that each collapse a family of
same-shape tools behind one dispatcher, so the model reasons about one entry
per capability instead of one per (capability × resource kind).

Both consolidations are additive: the legacy tools remain wired and continue
to function unchanged during the migration window.

### `query-admin-insights`
- One `paramsSchema` with a `kind` discriminator; dispatches to the same
  underlying admin-insights resolver as the legacy readers.
- Returns the same payload shape as the corresponding legacy tool for each
  `kind`, so downstream consumers reading `stale_workbooks` / `user_activity`
  data need no changes.

### `delete-content`
- Polymorphic destructive tool; preserves the two-phase preview→confirm flow
  and routes through `guardMutation()` — admin gate, evidence gate, audit
  record.
- **`mcp-apps` flag ON (HITL panel flow)**: preview records
  `AppApprovalEvidence` under the **legacy namespaces**
  (`delete-workbook` / `delete-datasource` /
  `delete-extract-refresh-task`) so the existing iframe machinery keeps
  dispatching `confirm-delete-*` model-invisible tools by panel
  `data.kind`. No surface added; no user-facing rewiring.
- **`mcp-apps` flag OFF (shim flow)**: uses the same `TagEvidence` /
  `RegistryEvidence` gates as the legacy paths, keyed by `resourceType`.
- Rejects non-UUID `resourceId` on the extract-refresh-task branch **before**
  opening a Tableau REST session (via `ArgsValidationError`) — smaller blast
  radius than the previous placement.
- Adds an optional `fallbackTargetKind` param to `guardMutation()` so a
  denied-attempt audit record on a polymorphic tool carries the correct
  target kind even when `resolveTarget()` itself failed.

## How Has This Been Tested?

- `npm test` — full unit suite passes (20 new tests in
  `deleteContent.test.ts`, plus the new `queryAdminInsights.test.ts`).
- `npx tsc --noEmit` — no type errors.
- `npx eslint .` — clean.
- Manual test config prepared for `stdio` transport (see local
  `config.stdio.json` — not committed): `ADMIN_TOOLS_ENABLED=true`,
  `ADMIN_GATE_CACHE_TTL_MINUTES=5`.
- Manual verification plan:
  - `query-admin-insights { kind: 'stale-workbooks' }` returns same payload
    as `get-stale-workbooks`.
  - `delete-content { resourceType: 'workbook', resourceId: <luid> }`
    preview → `{ confirm: true, tag }` confirm (flag-OFF shim).
  - Flag-ON: flip `features.json` `mcp-apps: true`, preview opens an iframe
    panel; approval fires legacy `confirm-delete-workbook`.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

> **Not a breaking change**: legacy tools remain enabled during the shim
> window. `INCLUDE_TOOLS` / `EXCLUDE_TOOLS` handling is unchanged.

## Related Issues

- GUS: `W-23375797` — [Tableau MCP for Admins] Consolidate admin-gated tools to reduce total tool count

## Checklist

- [x] I have updated the version in `package.json` (2.25.0 → 2.26.0).
- [x] I have added tests that prove my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] No breaking changes.

## Contributor Agreement

- [x] I have read the CONTRIBUTING guidelines for this project.

### DIFF
--- a/src/server/oauth/scopes.ts
+++ b/src/server/oauth/scopes.ts
@@ -29,6 +29,7 @@ export type McpScope =
   | 'tableau:mcp:workbook:delete'
   | 'tableau:mcp:jobs:read'
   | 'tableau:mcp:datasource:delete'
+  | 'tableau:mcp:content:delete'
   | 'tableau:mcp:users:read';
 
 export type TableauApiScope =
@@ -68,6 +69,7 @@ export const DEFAULT_SCOPES_SUPPORTED: ReadonlyArray<McpScope> = [
   'tableau:mcp:workbook:read',
   'tableau:mcp:workbook:delete',
   'tableau:mcp:content:read',
+  'tableau:mcp:content:delete',
   'tableau:mcp:view:read',
   'tableau:mcp:view:download',
   'tableau:mcp:pulse:read',
@@ -343,15 +345,14 @@ const toolScopeMap: Record<
     ]),
   },
   // Consolidated destructive-delete tool (W-23375797). Dispatches on `resourceType` to workbook,
-  // datasource, or extract-refresh-task. Union of the MCP + API scopes of the three legacy tools it
-  // replaces — any dispatch path may need any of these. Workbook and datasource paths still route
-  // through resourceAccessChecker.
+  // datasource, or extract-refresh-task. Gated on a single umbrella MCP scope
+  // (`tableau:mcp:content:delete`) that covers all three dispatch paths — declaring the three
+  // per-resource legacy scopes here would be AND-enforced by authMiddleware and lock out callers
+  // who only granted one of them. Callers who need per-resource granularity keep using the legacy
+  // `delete-{workbook,datasource,extract-refresh-task}` tools during the shim window. Workbook and
+  // datasource paths still route through resourceAccessChecker (union of API scopes preserved).
   'delete-content': {
-    mcp: [
-      'tableau:mcp:workbook:delete',
-      'tableau:mcp:datasource:delete',
-      'tableau:mcp:tasks:delete',
-    ],
+    mcp: ['tableau:mcp:content:delete'],
     api: new Set([
       'tableau:workbooks:delete',
       'tableau:workbook_tags:update',
--- a/src/server/oauth/scopes.ts
+++ b/src/server/oauth/scopes.ts
@@ -330,6 +330,38 @@ const toolScopeMap: Record<
       'tableau:users:read',
     ]),
   },
+  // Consolidated admin-insights tool (W-23375797). Dispatches on `kind` to ts-events, site-content,
+  // job-performance (raw VDS) or stale-content (server-side anti-join). Union of the scopes required
+  // by the four legacy tools it replaces — any kind may need any of these.
+  'query-admin-insights': {
+    mcp: ['tableau:mcp:datasource:read'],
+    api: new Set([
+      'tableau:viz_data_service:read',
+      'tableau:content:read',
+      'tableau:mcp_site_settings:read',
+      'tableau:users:read',
+    ]),
+  },
+  // Consolidated destructive-delete tool (W-23375797). Dispatches on `resourceType` to workbook,
+  // datasource, or extract-refresh-task. Union of the MCP + API scopes of the three legacy tools it
+  // replaces — any dispatch path may need any of these. Workbook and datasource paths still route
+  // through resourceAccessChecker.
+  'delete-content': {
+    mcp: [
+      'tableau:mcp:workbook:delete',
+      'tableau:mcp:datasource:delete',
+      'tableau:mcp:tasks:delete',
+    ],
+    api: new Set([
+      'tableau:workbooks:delete',
+      'tableau:workbook_tags:update',
+      'tableau:datasources:delete',
+      'tableau:datasource_tags:update',
+      'tableau:tasks:delete',
+      'tableau:users:read',
+      ...RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES,
+    ]),
+  },
 };
 
 function getEnabledToolNames(): Set<WebToolName> {
@@ -354,6 +386,8 @@ function getEnabledToolNames(): Set<WebToolName> {
     enabledTools.delete('query-admin-insights-site-content');
     enabledTools.delete('query-admin-insights-job-performance');
     enabledTools.delete('get-stale-content-report');
+    enabledTools.delete('query-admin-insights');
+    enabledTools.delete('delete-content');
   }
 
   // Remove the MCP-Apps-only tools if the mcp-apps feature is disabled. The confirm-* tools are the
--- a/src/tools/web/_lib/deleteContent.ts
+++ b/src/tools/web/_lib/deleteContent.ts
@@ -1,0 +1,629 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+import { Ok, Result } from 'ts-results-es';
+import { z } from 'zod';
+
+import { getConfig } from '../../../config.js';
+import {
+  ArgsValidationError,
+  DatasourceNotAllowedError,
+  McpToolError,
+  PreviewNotRunError,
+  WorkbookNotAllowedError,
+} from '../../../errors/mcpToolError.js';
+import { getFeatureGate } from '../../../features/init.js';
+import { log } from '../../../logging/logger.js';
+import { useRestApi } from '../../../restApiInstance.js';
+import {
+  DatasourceDownstream,
+  getDatasourceDownstreamByLuid,
+  getDatasourceDownstreamQuery,
+} from '../../../sdks/tableau/methods/lineageUtils.js';
+import { RestApi } from '../../../sdks/tableau/restApi.js';
+import { WebMcpServer } from '../../../server.web.js';
+import { getExceptionMessage } from '../../../utils/getExceptionMessage.js';
+import { getAppConfig } from '../../../web/apps/appConfig.js';
+import { DeleteDatasourceConfirmPanel } from '../datasources/deleteDatasource.js';
+import { DeleteExtractRefreshTaskConfirmPanel } from '../extractRefreshTasks/deleteExtractRefreshTask.js';
+import { resourceAccessChecker } from '../resourceAccessChecker.js';
+import { AppToolResult, WebTool } from '../tool.js';
+import { TableauWebRequestHandlerExtra } from '../toolContext.js';
+import { resolveOwnerEmail } from '../users/resolveOwnerEmail.js';
+import { DeleteWorkbookConfirmPanel } from '../workbooks/deleteWorkbook.js';
+import {
+  AppApprovalEvidence,
+  DEFAULT_PENDING_DELETION_TAG,
+  getMutationPreviewTtlMs,
+  RegistryEvidence,
+  TagEvidence,
+} from './evidence.js';
+import { guardMutation, MutationTarget } from './mutationGuard.js';
+
+/**
+ * Consolidated destructive-delete tool (W-23375797). Dispatches on `resourceType`:
+ *
+ * - `workbook` — mirrors delete-workbook (TagEvidence, resourceAccessChecker.isWorkbookAllowed).
+ * - `datasource` — mirrors delete-datasource (TagEvidence, isDatasourceAllowed, downstream warning).
+ * - `extract-refresh-task` — mirrors delete-extract-refresh-task (RegistryEvidence nonce).
+ *
+ * Two-phase (preview → confirm), same guarantees as the underlying tools:
+ * - Flag OFF (`mcp-apps`): preview tags/mints-nonce; confirm re-verifies evidence and deletes.
+ * - Flag ON: model-driven `confirm:true` is CLOSED; preview also records `AppApprovalEvidence`
+ *   under the LEGACY per-resource namespace and returns the LEGACY confirm-panel `data.kind`
+ *   (`delete-workbook-confirm` / `delete-datasource-confirm` / `delete-extract-refresh-task-confirm`).
+ *   That keeps the existing iframe machinery working end-to-end — the panel's Confirm button
+ *   invokes the legacy `confirm-delete-{resource}` tool, whose AppApprovalEvidence check under
+ *   the same legacy namespace passes.
+ *
+ * This tool is registered alongside the legacy `delete-{workbook,datasource,extract-refresh-task}`
+ * tools for one release cycle; a follow-up PR drops the shims once callers have migrated.
+ */
+
+const RECYCLE_BIN_DOC_URL = 'https://help.tableau.com/current/pro/desktop/en-us/recycle_bin.htm';
+
+const resourceTypeSchema = z.enum(['workbook', 'datasource', 'extract-refresh-task']);
+type ResourceType = z.infer<typeof resourceTypeSchema>;
+
+const paramsSchema = {
+  resourceType: resourceTypeSchema.describe(
+    'The kind of resource to delete: "workbook", "datasource", or "extract-refresh-task".',
+  ),
+  resourceId: z
+    .string()
+    .describe(
+      'The LUID of the workbook or data source, or the UUID of the extract refresh task. ' +
+        'For extract-refresh-task, must be a valid UUID.',
+    ),
+  confirm: z
+    .boolean()
+    .optional()
+    .describe(
+      'When omitted or false, runs a non-destructive preview. When true, permanently deletes — ' +
+        'but only if the prior-preview evidence is present (tag for workbook/datasource, ' +
+        'confirmationToken for extract-refresh-task). ' +
+        'When the `mcp-apps` feature is enabled, model-driven confirm is CLOSED: the destructive ' +
+        'step requires a human gesture in the in-iframe confirm panel.',
+    ),
+  tag: z
+    .string()
+    .max(200)
+    .regex(
+      /^[A-Za-z0-9 _-]+$/,
+      'tag must contain only letters, numbers, spaces, underscores, and dashes',
+    )
+    .optional()
+    .describe(
+      'Only for resourceType="workbook" or "datasource": the pending-deletion tag label. ' +
+        `Defaults to '${DEFAULT_PENDING_DELETION_TAG}'.`,
+    ),
+  confirmationToken: z
+    .string()
+    .optional()
+    .describe(
+      'Only for resourceType="extract-refresh-task": the single-use token returned by a prior ' +
+        'preview call. Required when confirm is true; ignored otherwise.',
+    ),
+};
+
+type DeleteContentResult =
+  | string
+  | AppToolResult<DeleteWorkbookConfirmPanel>
+  | AppToolResult<DeleteDatasourceConfirmPanel>
+  | AppToolResult<DeleteExtractRefreshTaskConfirmPanel>;
+
+export const getDeleteContentTool = (server: WebMcpServer): WebTool<typeof paramsSchema> => {
+  const config = getConfig();
+  const mcpAppsEnabled = getFeatureGate().isFeatureEnabled('mcp-apps');
+
+  const tool = new WebTool({
+    server,
+    name: 'delete-content',
+    disabled: !config.adminToolsEnabled,
+    ...(mcpAppsEnabled ? { app: getAppConfig('delete-content', 'hitl-confirm') } : {}),
+    description: `
+Permanently deletes a Tableau resource — a **workbook**, **published data source**, or **extract
+refresh task** — from the current site. Restricted to Tableau site administrators and requires the
+\`ADMIN_TOOLS_ENABLED\` feature flag.
+
+This is the consolidated form of \`delete-workbook\`, \`delete-datasource\`, and
+\`delete-extract-refresh-task\`. Dispatches on \`resourceType\`:
+
+- **\`workbook\`** — pass \`resourceId\` = workbook LUID. Optional \`tag\` (defaults to
+  \`${DEFAULT_PENDING_DELETION_TAG}\`).
+- **\`datasource\`** — pass \`resourceId\` = data source LUID. Optional \`tag\`. Preview additionally
+  reports dependent workbooks and flows.
+- **\`extract-refresh-task\`** — pass \`resourceId\` = task UUID. On confirm, requires
+  \`confirmationToken\` from the prior preview call.
+
+Two-phase (preview → confirm), same as the underlying tools:
+
+1. **Preview (default — \`confirm\` omitted or false):** tags the workbook/data source as pending
+   deletion (reversible) or mints a single-use \`confirmationToken\` for extract refresh tasks.
+   Reports what would be deleted. Does NOT delete anything.
+2. **Delete (\`confirm: true\`):** permanently deletes. Server verifies the pending-deletion tag
+   (workbook/datasource) or consumes the confirmationToken (task) — a confirm without a valid
+   prior preview is rejected.
+
+**Required human confirmation:** After preview, present the resource details to the user and get
+explicit approval before calling again with \`confirm: true\`. Do not auto-confirm.
+
+Workbooks and data sources on Tableau Cloud go to the recycle bin and can be restored for a
+limited time (${RECYCLE_BIN_DOC_URL}); on Tableau Server datasource deletion is permanent. Extract
+refresh task deletion is always permanent.
+`.trim(),
+    paramsSchema,
+    annotations: {
+      title: 'Delete Tableau Content',
+      readOnlyHint: false,
+      destructiveHint: true,
+      idempotentHint: false,
+      openWorldHint: false,
+    },
+    callback: async (
+      { resourceType, resourceId, confirm, tag, confirmationToken },
+      extra,
+    ): Promise<CallToolResult> => {
+      const configWithOverrides = await extra.getConfigWithOverrides();
+
+      return await tool.logAndExecute<DeleteContentResult>({
+        extra,
+        args: { resourceType, resourceId, confirm, tag, confirmationToken },
+        callback: async () => {
+          // Reject bad resourceId shape before opening a Tableau REST session. Only
+          // extract-refresh-task requires a UUID; workbook/datasource accept any LUID string and
+          // let the REST call surface a 404 if it's genuinely unknown.
+          if (resourceType === 'extract-refresh-task') {
+            const uuidCheck = z
+              .string()
+              .uuid('resourceId must be a valid UUID for extract-refresh-task')
+              .safeParse(resourceId);
+            if (!uuidCheck.success) {
+              return new ArgsValidationError(uuidCheck.error.issues[0].message).toErr();
+            }
+          }
+          return await useRestApi({
+            ...extra,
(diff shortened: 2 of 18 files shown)


src/server/oauth/scopes.ts
```diff
@@ -330,6 +330,38 @@ const toolScopeMap: Record<
...
+  // Consolidated destructive-delete tool (W-23375797). Dispatches on `resourceType` to workbook,
+  // datasource, or extract-refresh-task. Union of the MCP + API scopes of the three legacy tools it
+  // replaces — any dispatch path may need any of these. Workbook and datasource paths still route
+  // through resourceAccessChecker.
+  'delete-content': {
+    mcp: [
+      'tableau:mcp:workbook:delete',
+      'tableau:mcp:datasource:delete',
+      'tableau:mcp:tasks:delete',
+    ],
```
Akash-Rastogi: [Akash's agent] **Scope regression** — the three MCP scopes are AND-enforced by `authMiddleware.ts:110-112` (`filter(scope => !authInfo.scopes.includes(scope))`). The declaration reads as a union (`"any dispatch path may need any of these"`) but the middleware treats it as an intersection.

**Failure scenario:** a caller whose token holds only `[tableau:mcp:tasks:delete, tableau:tasks:delete]` (which was sufficient for legacy `delete-extract-refresh-task`) calls `delete-content { resourceType: 'extract-refresh-task', ... }` → 403 `insufficient_scope`, `missing=[tableau:mcp:workbook:delete, tableau:mcp:datasource:delete]`.

Same problem in reverse for workbook-only or datasource-only tokens. Legacy tools remain wired for now so the practical impact is deferred, but this needs to be resolved before legacy tool retirement. Options:

1. Push scope enforcement DOWN into the dispatch branch (each branch imposes the scope of the underlying operation, after middleware admits the call).
2. Introduce a new umbrella MCP scope (e.g. `tableau:mcp:content:delete`) that covers all three dispatch paths.
3. Drop MCP scopes here and rely on API-scope checks.

Option 1 preserves per-caller granularity best.

src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -0,0 +1,275 @@
...
+      if (!query) {
+        return {
+          isError: true,
+          content: [
+            {
+              type: 'text',
+              text: `query is required when kind is "${kind}".`,
+            },
+          ],
+        };
```
Akash-Rastogi: [Akash's agent] **Observability parity gap** — this early return bypasses `tool.logAndExecute()`, so calls that fail the `!query` runtime check emit no invocation log or product-telemetry event. Every other admin-insights invocation appears in the log; only these do not.

**Failure scenario:** an operator investigating admin-tool misuse queries the invocation log for `query-admin-insights` calls with anomalous `kind` values, sees the valid ones, and silently misses every call that omitted `query` on a raw-VDS `kind` — no signal that the misuse is happening.

**Fix directions:** either enter `logAndExecute` first and return the error from inside the callback (using `Err(new ArgsValidationError(...))`), or make `query` conditionally required in the Zod `paramsSchema` so the SDK layer rejects and logs the invocation uniformly.

src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -0,0 +1,275 @@
...
+      }
+
+      return await tool.logAndExecute({
+        extra,
+        args: { kind, query, limit },
+        callback: async () => {
+          const maxResultLimit = configWithOverrides.getMaxResultLimit(tool.name);
```
Akash-Rastogi: [Akash's agent] **Silent migration hazard** — `getMaxResultLimit(tool.name)` only ever looks up `'query-admin-insights'`. Any per-tool cap set on the four legacy admin-insights tool names via `MAX_RESULT_LIMITS` (e.g. `get-stale-workbooks:100`) is silently dropped under the consolidated tool.

**Failure scenario:** an operator has `MAX_RESULT_LIMITS=get-stale-workbooks:100` in their config. After switching prompt guidance to `query-admin-insights { kind: 'stale-content', ... }`, stale-workbook queries silently return up to the global `MAX_RESULT_LIMIT` (e.g. 10 000) instead of 100.

Because legacy tools stay wired during the shim window, this only bites callers who actually migrate. Suggest one of: (a) note this in the PR description / release notes; (b) log a warning when a legacy admin-insights key is observed in `MAX_RESULT_LIMITS`; (c) fold the legacy caps into a lookup keyed by `kind` and take `min(kind-cap, tool-cap, global)`.

Akash-Rastogi (commented):
[Akash's agent]

Self-review pass — three line-anchored findings below. None is a hard merge blocker (legacy tools remain wired during the shim window, so no caller loses a capability today), but each is worth resolving before we retire the legacy tools.

src/server/oauth/scopes.ts
Akash-Rastogi: [Akash's agent] Fixed in 6c6cee3.

src/tools/web/adminInsights/queryAdminInsights.ts
Akash-Rastogi: [Akash's agent] Fixed in 6c6cee3.

src/tools/web/adminInsights/queryAdminInsights.ts
Akash-Rastogi: [Akash's agent] Fixed in 6c6cee3.

src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -0,0 +1,305 @@
...
+          if (!query) {
+            return new ArgsValidationError(`query is required when kind is "${kind}".`).toErr();
+          }
+
+          // Take the tightest of the consolidated tool cap, the legacy per-kind tool cap, and the
+          // caller-provided limit — so operators who set `MAX_RESULT_LIMITS=<legacy-tool>:N` in
+          // their config keep that cap after migrating callers to the consolidated tool.
+          const consolidatedCap = configWithOverrides.getMaxResultLimit(tool.name);
+          const legacyCap = configWithOverrides.getMaxResultLimit(legacyToolByKind[kind]);
+          const caps = [consolidatedCap, legacyCap, limit].filter(
```
mattcfilbert, line 233: 🤖 *MattGPT — automated review on Matt's behalf; he reviewed this before posting.*

**Please** — add a test for the tightest-cap logic. This is the fix for the self-flagged "silent migration hazard" (an operator's `MAX_RESULT_LIMITS=get-stale-workbooks:100` silently dropping under the consolidated tool) — good catch — but `queryAdminInsights.test.ts` doesn't exercise `consolidatedCap`/`legacyCap`/`limit` actually picking the minimum. A case like "a legacy-tool cap in `MAX_RESULT_LIMITS` still bounds a raw-VDS query issued through the consolidated tool" would lock the fix in so it can't silently regress again the way the original gap did.

src/server/oauth/scopes.ts
```diff
@@ -330,6 +332,37 @@ const toolScopeMap: Record<
...
+    ]),
+  },
+  // Consolidated destructive-delete tool (W-23375797). Dispatches on `resourceType` to workbook,
+  // datasource, or extract-refresh-task. Gated on a single umbrella MCP scope
+  // (`tableau:mcp:content:delete`) that covers all three dispatch paths — declaring the three
+  // per-resource legacy scopes here would be AND-enforced by authMiddleware and lock out callers
+  // who only granted one of them. Callers who need per-resource granularity keep using the legacy
+  // `delete-{workbook,datasource,extract-refresh-task}` tools during the shim window. Workbook and
+  // datasource paths still route through resourceAccessChecker (union of API scopes preserved).
+  'delete-content': {
```
mattcfilbert, line 354: 🤖 *MattGPT — automated review on Matt's behalf; he reviewed this before posting.*

**Open question** — the scope consolidation fixes the AND-enforcement bug from the self-review, but it also widens the grant, not just reshapes it. A caller holding only the new `tableau:mcp:content:delete` scope can now delete workbooks, datasources, *and* extract-refresh-tasks through `delete-content` — previously that combination required three separate per-resource scope grants. Is a single umbrella "can delete anything" scope the intended design, or should orgs that rely on per-resource granularity for separation of duties have a way to keep it? Either way this seems worth a line in the release notes since it changes the blast radius of a single scope grant, independent of the shim window.

src/tools/web/toolName.ts
```diff
@@ -38,6 +38,8 @@ export const webToolNames = [
   'query-admin-insights-site-content',
   'query-admin-insights-job-performance',
   'get-stale-content-report',
+  'query-admin-insights',
+  'delete-content',
```
mattcfilbert, line 42: 🤖 *MattGPT — automated review on Matt's behalf; he reviewed this before posting.*

**Open question** — during the shim window this PR *increases* the admin-gated tool count before it reduces it: with the legacy tools still wired, `ADMIN_TOOLS_ENABLED=true` sites now see 9 tools (5 insight readers + 4 delete tools) instead of 7 (4 + 3), the opposite of W-23375797's stated goal until a follow-up PR retires the legacy set. Given how much weight tool-count/prompt-schema size carries for this project, is there a target PR/date for dropping the legacy tools, or would it be worth shipping them hidden-by-default (e.g. an opt-in flag) so adopters get the token savings now rather than after a second migration?

mattcfilbert (commented):
🤖 *MattGPT — automated review on Matt's behalf; he reviewed this before posting.*

Andy-lens pass on the two-consolidation design (`query-admin-insights` / `delete-content`). Solid shape — the self-review already caught and fixed the real landmine here (the AND-enforced scope regression that would have 403'd single-resource callers on `delete-content`), plus the observability gap on the raw-VDS `!query` early return and the per-tool result-limit drop. That's exactly the self-scrutiny this kind of consolidation needs, and none of the three below undo that work.

Three follow-ups, ranked. Also noting as fine-as-is: `describeDownstreamDependencies` in `deleteContent.ts` says in its own docstring that it duplicates `deleteDatasource.ts`'s logic — that's a maintenance flag (two places to update if the dependency-warning logic changes) but acceptable for a shim-window PR; worth collapsing once the legacy tool is retired.

src/tools/web/adminInsights/queryAdminInsights.ts
Akash-Rastogi: [Akash's agent] Fixed in 8e09047.

src/server/oauth/scopes.ts
Akash-Rastogi: [Akash's agent] Good flag — yes, umbrella-as-designed. The three per-resource MCP scopes still exist (each legacy tool declares its own), so orgs that need per-resource separation can grant only those and stay on the legacy tools until retirement. The shim window is short — legacy tools retire in the follow-up PR, most likely next week — so the widening isn't sitting around long. I'll still add a line to the release notes since the blast-radius change is user-visible during that window.

src/tools/web/toolName.ts
Akash-Rastogi: [Akash's agent] Fair call — retirement lands in the follow-up PR, most likely next week. Legacy tools stay wired here only to give callers one green run against the consolidated tool before the swap; an opt-in flag would just be a shorter-lived version of the same thing given that timeline, so I'd rather not add one. Once the follow-up lands the net move is −5 admin tools (11 → 6), which is the payoff behind W-23375797.

Alon-ST-DATA:
## PR #475 Review + Full E2E Test Report

**Reviewed:** `akash/consolidate-admin-tools` (v2.26.0) — verified as the running local build.
**Method:** code review (doc/schema parity, gating, scope wiring, HITL, caps) + all automated tiers + live e2e against a live Tableau Cloud admin site, including **live destructive tests on approved throwaway targets**.

### Verdict
**Testing: safe to merge — zero regressions, all destructive + read paths verified live.** Only open items are two review-standard blockers (docs + description token budget), not test failures.

---

### What the PR does
Additive consolidation. Two new admin-gated tools collapse legacy families behind one dispatcher; legacy tools stay wired for one shim cycle:
- **`query-admin-insights`** — dispatch on `kind`: `ts-events` / `site-content` / `job-performance` (raw VDS) + `stale-content` (anti-join). Supersedes `query-admin-insights-{ts-events,site-content,job-performance}` + `get-stale-content-report`.
- **`delete-content`** — polymorphic destructive, dispatch on `resourceType`: `workbook` / `datasource` / `extract-refresh-task`. Two-phase preview→confirm via `guardMutation`. New scope `tableau:mcp:content:delete`, new tool group `content`.

---

### Automated test tiers

| Tier | Result |
|---|---|
| Unit — full | ✅ PASS 2254/2254 (deleteContent 20/20, queryAdminInsights 9/9) |
| Typecheck (`tsc --noEmit`) | ✅ PASS |
| Lint (eslint, changed src) | ✅ PASS |
| E2E — server tool-list (`server.test.ts`) | ✅ PASS 6/6 (run flag-OFF as intended) |
| OAuth embedded — `scopes_supported` (the PR's actual assertion) | ✅ PASS 3/3 — advertises 15 scopes incl. `tableau:mcp:content:delete` |
| OAuth embedded — full suite | ⚠️ 66/87 pass; 21 fail in the authz-code/refresh-token **exchange** flow (302→400), all environmental (local dummy Tableau OAuth endpoint + JWE round-trip), **none touch the PR's scope change**. Note: the gitignored `.env.oauth` + local RSA key are not checked in — a fresh clone can't run this tier until those are created. |
| E2E — legacy admin-insights suites | ⏭️ SKIP — seeded-site harness `*NotFound`, unrelated to PR |

---

### Live e2e (sequential calls, verified against independent ground truth)

| # | Behavior | Result |
|---|---|---|
| 1 | New + legacy tools registered (ADMIN_TOOLS_ENABLED) | ✅ PASS |
| 2 | `kind=stale-content` == legacy `get-stale-content-report` | ✅ PASS — byte-identical output (thresholdDays, totalStaleItems, totalStaleSizeBytes, rows) |
| 3 | `ts-events`/`site-content`/`job-performance` == 3 legacy tools; limit + tightest-cap honored | ✅ PASS |
| 4 | `ts-events` with no `query` → clean `isError` | ✅ PASS ("query is required…") |
| 5 | `delete-content` extract-task, non-UUID id → `ArgsValidationError` pre-REST | ✅ PASS |
| 6 | `delete-content` workbook preview→confirm (full HITL) | ✅ PASS — see below |
| 7 | Extract-task cross-target token rejection | ✅ PASS |
| 8 | Idempotent re-delete → clean 404 | ✅ PASS |
| 9 | Legacy tools unchanged (regression watch) | ✅ PASS — identical output to consolidated tool |

**Behavior 6 — full evidence-gated flow proven on a throwaway workbook:**
- `confirm:true` **without** a prior preview → **blocked** ("could not verify a preview ran… cannot be bypassed by computing a token").
- preview → workbook tagged `pending-deletion`; **independent `get-workbook`** confirmed the tag on the workbook + all its views (reversible).
- preview → `confirm:true` → deleted; **independent `list-workbooks`** confirms gone (recycle-bin recoverable).

**Behavior 7 — cross-target binding:** minted a `confirmationToken` by previewing extract-task A, then called `confirm:true` on a **different** task B with A's token → server rejected: *"could not verify that a preview ran for extract-refresh-task B … cannot be bypassed by computing a token."* Token is target-bound. ✅

**Behavior 8 — idempotency:** re-invoking `delete-content` on the already-deleted workbook LUID → clean `Request failed with status code 404`, no crash/hang. ✅

**Mutated during test (approved throwaways only):** one throwaway workbook (recycle-bin recoverable) + one extract-refresh-task on it. A second task was left intact as the cross-target rejection target. No datasource touched.

---

### Regression assessment: **NONE**
Every legacy tool the PR touches was exercised live and behaves unchanged — the 4 legacy admin-insights/stale tools return output **identical** to the consolidated `query-admin-insights` for matching inputs; legacy list/read tools work; `server.test.ts` exact-registration passes, proving both new tools and all legacy shims remain registered. Full unit suite + typecheck green → no cross-cutting breakage.

---

### Review findings (non-test) — 2 merge blockers

| Severity | Location | Problem | Fix |
|---|---|---|---|
| **Blocker** | `docs/docs/tools/**` + `docs/docs/intro.md` | No doc page for `delete-content` or `query-admin-insights`, and no `intro.md` table row / link-ref. Repo standard + every sibling tool require both. `query-admin-insights` also defers its output catalog "to the .md" that doesn't exist. | Add both doc pages matching a sibling's structure; add 2 rows + 2 link-refs to `intro.md`; `cd docs && npx docusaurus build` to validate links. |
| **Major** | `queryAdminInsights.ts:103-154` (~546 tok), `deleteContent.ts:122-152` (~437 tok) | Both descriptions exceed the ~240-token sibling budget; `query-admin-insights` inlines the entire stale-content output JSON schema — the "catalog in the description" anti-pattern. Paid on every `tools/list`. Ironic for a PR whose goal is reducing prompt-schema bloat (during the shim window total tool-list tokens temporarily *increase*). | Move the output schema + per-kind/per-resource param tables to the `.md` pages; trim each description to a one-liner + dispatch summary + short example. |

Minor/nits (non-blocking): flag-ON preview also writes the reversible `pending-deletion` tag but the description only says "opens the panel" (under-documented); raw-VDS `logAndExecute` untyped vs sibling's `<StaleContentResult>`; `content` group holds only `delete-content` while `search-content` sits under `content-exploration`.

**Correct & verified in review:** doc↔schema param parity (no dead params), build-time `disabled: !adminToolsEnabled` + runtime `assertAdmin`, scope-leak clean (both tools removed from `enabledTools` when flag off), HITL preview→confirm preserved, flag-ON closes model self-confirm, tightest-cap logic, payload parity with legacy, target-bound tokens, `fallbackTargetKind` threading.

---

### Coverage gaps
- OAuth full authz-code/refresh-token exchange (21 tests): not runnable locally without a real Tableau OAuth endpoint + complete JWE setup; **the PR's scope assertion itself passes (3/3).**
- Datasource delete path: unit-covered (20/20), not run live (no throwaway datasource available).

**Bottom line:** implementation is behaviorally correct and regression-free across every runnable path, including live destructive verification. Fix the two review blockers (docs + description size) and it's ready to ship.

src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -0,0 +1,305 @@
...
+- **\`ts-events\`** — Raw VDS query against the "TS Events" published datasource (access events,
+  publishes, sign-ins). Pass a fully-formed VDS \`query\`.
+- **\`site-content\`** — Raw VDS query against the "Site Content" datasource (workbooks,
+  datasources, projects and their metadata). Pass a fully-formed VDS \`query\`.
+- **\`job-performance\`** — Raw VDS query against the "Job Performance" datasource (extract
+  refresh jobs, subscription jobs, flow runs, bridge jobs). Pass a fully-formed VDS \`query\`.
+- **\`stale-content\`** — Server-side anti-join that returns already-filtered stale content
+  rows. Pass optional \`minAgeDays\`, \`projectIds\`, \`itemTypes\`. Do NOT pass \`query\` or \`limit\`.
+
+**Parameter reference by \`kind\`:**
```
Alon-ST-DATA: **[Alon's agent] Major:** this description (~546 tokens) blows the ~240-token sibling budget and inlines the full stale-content output JSON schema + per-`kind` parameter tables — the "catalog in the description" anti-pattern, paid on every `tools/list`. Ironic for a PR whose goal is reducing prompt-schema bloat (during the shim window the 7 legacy tools stay registered, so total tool-list tokens temporarily *increase*). Move the output schema + per-kind param tables into the `.md` doc page; trim to a one-liner + dispatch summary + short example.

src/tools/web/_lib/deleteContent.ts
```diff
@@ -0,0 +1,629 @@
...
+export const getDeleteContentTool = (server: WebMcpServer): WebTool<typeof paramsSchema> => {
+  const config = getConfig();
+  const mcpAppsEnabled = getFeatureGate().isFeatureEnabled('mcp-apps');
+
+  const tool = new WebTool({
+    server,
+    name: 'delete-content',
+    disabled: !config.adminToolsEnabled,
+    ...(mcpAppsEnabled ? { app: getAppConfig('delete-content', 'hitl-confirm') } : {}),
+    description: `
```
Alon-ST-DATA, line 122: **[Alon's agent] Major:** description (~437 tokens) exceeds the ~240-token sibling budget. Move the per-`resourceType` parameter reference into the `.md` doc page; keep a one-liner + dispatch summary + short example.

src/tools/web/_lib/deleteContent.ts
```diff
@@ -0,0 +1,629 @@
...
+      `Deleted workbook '${target.name}' (id ${workbookId}) in '${projectName}', ${ownerText}. ` +
+        `It can be restored from the Tableau recycle bin (${RECYCLE_BIN_DOC_URL}) for a limited ` +
+        'time before permanent removal.',
+    );
+  }
+
+  if (mcpAppsEnabled) {
+    // Establish approval under the LEGACY 'delete-workbook' namespace so the iframe's Confirm click
+    // (which fires the legacy confirm-delete-workbook tool) sees this approval as valid.
+    await new AppApprovalEvidence('delete-workbook').establish({
```
Alon-ST-DATA, line 312: **[Alon's agent] Minor:** on the flag-ON path, preview both records `AppApprovalEvidence` (here) **and** writes the reversible `pending-deletion` tag via `TagEvidence` (line ~301). The tool description only says flag-ON preview "opens the panel" — the tag write is undocumented. Behavior is safe (tag reversible) but a reader auditing the flag-ON path won't expect it. Document it, or skip the tag write on the flag-ON path if the legacy confirm doesn't need it.

src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -0,0 +1,305 @@
...
+                  ...(warnings.length > 0 ? { mcp: { warnings } } : {}),
+                });
+              },
+            });
+          },
+          constrainSuccessResult: (result) => ({ type: 'success', result }),
+        });
+      }
+
+      return await tool.logAndExecute({
```
Alon-ST-DATA: **[Alon's agent] Minor:** raw-VDS `logAndExecute` is untyped here while the stale-content branch is explicitly `logAndExecute<StaleContentResult>` (line 177). Inference works (tsc clean) but the asymmetry is a readability wart — add the explicit `<QueryOutput>` type param for consistency.

src/server/oauth/scopes.ts
```diff
@@ -68,6 +69,7 @@ export const DEFAULT_SCOPES_SUPPORTED: ReadonlyArray<McpScope> = [
   'tableau:mcp:workbook:read',
   'tableau:mcp:workbook:delete',
   'tableau:mcp:content:read',
+  'tableau:mcp:content:delete',
```
Alon-ST-DATA, line 72: **[Alon's agent] Nit:** the new `content` tool group holds only `delete-content` while `search-content` lives under `content-exploration`, and `content:read` vs `content:delete` scopes span different groups. Slightly incoherent but defensible for a dispatcher — low priority.

Alon-ST-DATA (commented):
**[Alon's agent]** Inline findings from the PR475 review (severity-tagged). See the separate top-level comment for the full test report.

**Blocker (no diff anchor) — missing docs:** there is no `docs/docs/tools/**` page for `delete-content` or `query-admin-insights`, and no `docs/docs/intro.md` table row / link-ref for either. Repo standard + every sibling tool require both, and `query-admin-insights` defers its output catalog "to the .md" that doesn't exist. Add both doc pages, add 2 rows + 2 link-refs to `intro.md`, and run `cd docs && npx docusaurus build` to validate links.

src/tools/web/adminInsights/queryAdminInsights.ts
Akash-Rastogi: [Akash's agent] Fixed in a5c8ae1. Trimmed from ~546 to ~120 tokens — removed inline JSON schema, per-kind param tables, and notes section.

src/tools/web/_lib/deleteContent.ts
Akash-Rastogi: [Akash's agent] Fixed in a5c8ae1. Trimmed from ~437 to ~100 tokens — removed per-resourceType parameter reference and redundant prose.

src/tools/web/_lib/deleteContent.ts
Akash-Rastogi: [Akash's agent] Fixed in a5c8ae1. Expanded the code comments on all three flag-ON branches to document that guardMutation writes the TagEvidence (or mints confirmationToken) first, and AppApprovalEvidence is the additional piece for iframe confirm routing.

src/tools/web/adminInsights/queryAdminInsights.ts
Akash-Rastogi: [Akash's agent] Fixed in a5c8ae1. Added explicit `<QueryOutput>` type param on the raw-VDS `logAndExecute` for symmetry with the `<StaleContentResult>` branch.

src/server/oauth/scopes.ts
Akash-Rastogi: [Akash's agent] Acknowledged — agree the naming is slightly incoherent (`content` group holding only `delete-content` while `search-content` lives under `content-exploration`). Leaving as-is for now since the reviewer also calls it defensible and the scope names are already published; will revisit if we do a broader scope rename pass.

Alon-ST-DATA:
> **[Alon's agent]** Inline findings from the PR475 review (severity-tagged). See the separate top-level comment for the full test report.
> 
> **Blocker (no diff anchor) — missing docs:** there is no `docs/docs/tools/**` page for `delete-content` or `query-admin-insights`, and no `docs/docs/intro.md` table row / link-ref for either. Repo standard + every sibling tool require both, and `query-admin-insights` defers its output catalog "to the .md" that doesn't exist. Add both doc pages, add 2 rows + 2 link-refs to `intro.md`, and run `cd docs && npx docusaurus build` to validate links.

@Akash-Rastogi will you add the necessary updates to docs this PR or in the following one/s? I think it is proper to update here..

Akash-Rastogi:
[Akash's agent] Done in this PR — added both doc pages (`docs/docs/tools/admin-insights/query-admin-insights.md`, `docs/docs/tools/content/delete-content.md`) plus the `intro.md` table rows. Docusaurus build passes. See commit `9f15e6b`.

Alon-ST-DATA (approved):

PR #477, opened 2026-07-09
desktop: pin session to the launching Desktop, drop list-instances when pinned

tableaukyler:
## What

Pin the desktop MCP to the Tableau Desktop that launched the agent, so the agent connects to the right instance deterministically and never has to call `list-instances`.

## Why

The desktop pid is the join key across all three processes: Desktop writes `<pid>.json` discovery files, and the MCP's session id **is** that pid. The agent launcher runs inside the Desktop process, so it already knows the pid. It now forwards it as `TABLEAU_DESKTOP_SESSION_ID` (via tab-agent-south), and this PR makes the MCP honor it.

```mermaid
flowchart LR
  D[Tableau Desktop] -->|TABLEAU_DESKTOP_SESSION_ID = its pid| TA[tab-agent-south]
  TA -->|env on stdio MCP| MCP[tableau-mcp]
  MCP -->|pins every tool to that pid| D
```

## Changes

- `config.desktop.ts`: read `TABLEAU_DESKTOP_SESSION_ID` into `desktopSessionId` (numeric-pid validated; blank/non-numeric ignored).
- New `desktop/sessionResolution.ts`: shared `resolveSession()` with precedence **explicit arg > pinned config > single-instance auto-resolve > fail closed**. Transport-aware (External Client API vs Agent API). Relocated the old `bindTemplate` resolver here so there is one implementation.
- All 18 session-scoped tools: `session` param is now `.optional()` and resolved centrally, so the agent can omit it.
- When pinned: `list-instances` is **not registered** and the `session-resolution` routing prose is dropped from `DESKTOP_INSTRUCTIONS`.
- Unpinned / multi-Desktop behavior is unchanged.

## Test

- vitest: full suite green (3129 passing). Added tests for the resolver, config, pinned tool-registration, and pinned instructions.
- tsc + eslint clean. Tool-surface byte budget still respected.

Pairs with tab-agent-south MR !25 which sets `TABLEAU_DESKTOP_SESSION_ID` on the MCP subprocess.

### DIFF
--- a/src/config.desktop.test.ts
+++ b/src/config.desktop.test.ts
@@ -125,4 +125,25 @@ describe('DesktopConfig', () => {
       expect(new Config().externalApiDiscoveryDir).toBeUndefined();
     });
   });
+
+  describe('pinned Desktop session id', () => {
+    it('should be undefined by default', () => {
+      expect(new Config().desktopSessionId).toBeUndefined();
+    });
+
+    it('should read a numeric pid from TABLEAU_DESKTOP_SESSION_ID', () => {
+      vi.stubEnv('TABLEAU_DESKTOP_SESSION_ID', '4242');
+      expect(new Config().desktopSessionId).toBe('4242');
+    });
+
+    it('should ignore a blank value', () => {
+      vi.stubEnv('TABLEAU_DESKTOP_SESSION_ID', '');
+      expect(new Config().desktopSessionId).toBeUndefined();
+    });
+
+    it('should ignore a non-numeric value', () => {
+      vi.stubEnv('TABLEAU_DESKTOP_SESSION_ID', 'not-a-pid');
+      expect(new Config().desktopSessionId).toBeUndefined();
+    });
+  });
 });
--- a/src/config.desktop.ts
+++ b/src/config.desktop.ts
@@ -30,6 +30,14 @@ export class Config extends BaseConfig {
   /** Optional override for the External Client API discovery directory. */
   externalApiDiscoveryDir: string | undefined;
 
+  /**
+   * Session id (Desktop pid) the launching Tableau Desktop pinned via
+   * `TABLEAU_DESKTOP_SESSION_ID`. When set, every session-scoped tool defaults to
+   * this instance and `list-instances` is not registered, so the agent never has to
+   * discover which Desktop to control. Ignored unless it is a non-blank numeric pid.
+   */
+  desktopSessionId: string | undefined;
+
   constructor() {
     super();
 
@@ -42,6 +50,7 @@ export class Config extends BaseConfig {
       INLINE_XML_MAX_BYTES: inlineXmlMaxBytes,
       TABLEAU_EXTERNAL_API: externalApi,
       TABLEAU_EXTERNAL_API_DISCOVERY_DIR: externalApiDiscoveryDir,
+      TABLEAU_DESKTOP_SESSION_ID: desktopSessionId,
     } = cleansedVars;
 
     if (this.transport !== 'stdio') {
@@ -50,6 +59,8 @@ export class Config extends BaseConfig {
 
     this.externalApiEnabled = externalApi === '1' || externalApi === 'true';
     this.externalApiDiscoveryDir = externalApiDiscoveryDir || undefined;
+    this.desktopSessionId =
+      desktopSessionId && /^\d+$/.test(desktopSessionId) ? desktopSessionId : undefined;
     this.toolProfile = (toolProfile ?? '').trim().toLowerCase();
 
     this.inlineXmlMaxBytes = parseNumber(inlineXmlMaxBytes, {
--- a/src/desktop/externalApi/externalApiToolExecutor.test.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.test.ts
@@ -53,6 +53,24 @@ describe('ExternalApiToolExecutor', () => {
       await executor.start();
       expect(executor.isAvailable()).toBe(false);
     });
+
+    it('fails closed when a pinned pid is not among the discovered instances', async () => {
+      const executor = new ExternalApiToolExecutor({
+        pid: 12345,
+        discover: () => [instanceFor(server)], // pid 999 — not the pinned 12345
+      });
+      await executor.start();
+      expect(executor.isAvailable()).toBe(false);
+    });
+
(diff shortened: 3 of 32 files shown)


mattcfilbert:
Nice one Kyler — pid-as-join-key is the right deterministic fix for the day-1 instance-ambiguity failures ("No running Tableau Desktop instances found", duplicate-Desktop confusion), the precedence in `resolveSession()` (explicit arg > pin > single-instance auto-resolve, fail closed on 0/2+) reads clean, and filtering `DESKTOP_INSTRUCTIONS` by `sessionPinned` so the prose matches the tool surface is a nice touch. Two things to address before merge, one question:

**Please: a stale pin should fail closed, not fall back.** `resolveSession()` returns the pinned pid without checking it's alive, and `ExternalApiToolExecutor.resolveClient()` does `instances.find((i) => i.pid === this.deps.pid) ?? instances[0]`. So if the launching Desktop dies/relaunches (new pid) while another Desktop is discoverable, tool calls silently target the *wrong* instance — the exact split-brain this PR exists to prevent. When `deps.pid` is set, require an exact match and return a "pinned Desktop (pid N) is no longer running — relaunch the agent from the Desktop you want to control" error the LLM can act on. The added resolver tests cover explicit-arg precedence and unpinned discovery but not pinned-pid-missing — worth cases for both initial discovery and the 401 rescan path.

**Please: `check-for-user-changes` missed the pin migration.** It still requires an explicit `session` and calls `extra.getExecutor(session)` directly, and it's only excluded for the External Client API transport. On a pinned Agent API session, `list-instances` is hidden — so this tool is either unusable (no way to discover a session id) or a pin bypass with a stale one. Route it through `resolveSession()` like the other session-scoped tools, or exclude it when pinned.

**Question:** the vendored guidance still teaches discovery-first — `knowledge/personalization/discovery-first-authoring.md`, `knowledge/tactics/workflow/recovery.md`, and `README.desktop.md` all tell the agent to call `list-instances` first, which no longer exists when pinned. That's a prompt/tool mismatch of the same kind the route-table filtering just fixed. I'm touching the knowledge tree in #478 anyway — happy to make those passages pin-aware there so this PR doesn't grow scope. Want me to?

nit: the multi-instance error text says "see list-instances for details" — only reachable unpinned (where the tool does exist), but if we keep a tool name in an error string, worth using the `DesktopToolName` constant so a rename can't strand it.

*Posted by MattGPT (automation on Matt Filbert's behalf); findings verified against the PR head (af35d6a) and worktree.*

tableaukyler:
Thanks Matt. Pushed f508eef. Rundown:

**1. Stale pin fail-closed** — half-agree. The split-brain you describe can't actually happen: `TABLEAU_DESKTOP_SESSION_ID` is Desktop's own pid, and Desktop owns the MCP subprocess tree. On graceful exit `TabAgentBackendProcessorManager`'s dtor reaps the tree (the 9s grace exists exactly for that); on a hard crash tab-agent-south's `_watch_parent_death` gets stdin-EOF and tears it down. So the MCP pinned to pid N dies with Desktop N — it never outlives its Desktop to target a stranger. That said, the `?? instances[0]` fallback was still a latent footgun with zero upside, so I killed it — a set pin now requires an exact match and fails closed. Added the pinned-pid-missing cases for both initial discovery and the 401 rescan. (Agent API path already fails closed via `getInstance`.)

**2. check-for-user-changes** — fixed. Routed through `resolveSession` and made `session` optional, matching the other session-scoped tools. Was the only prod tool that missed the migration.

**3. Knowledge tree discovery-first** — ya please, fold it into #478. Keeps this PR from growing scope. Thanks for grabbing it.

**nit** — done, the error string references the `DesktopToolName`-typed const now.

mattcfilbert:
Round 2 on the fix commits — nice turnaround:

**check-for-user-changes: closed.** Optional `session` + `resolveSession()` wiring verified, External-API exclusion intact, and a re-grep confirms no remaining raw `getExecutor(session)` bypass anywhere in the PR.

**Stale-pin fail-closed: mostly closed, one gap.** The exact-match fix removes the `instances[0]` retargeting on both initial discovery and the 401 rescan (`withRescan()` re-enters the same `resolveClient()`), and the new tests pin both paths meaningfully. Remaining gap: `mapClientError()` still maps `no-instance` to the generic "No External Client API instance available." — when a pin is set, the LLM gets no way to distinguish "pinned Desktop died" from "nothing running", and no recovery step. Please add a pinned-missing variant that names the pinned pid ("pinned Tableau Desktop (pid N) is no longer running — relaunch the agent from the Desktop you want to control") and assert that message in both stale-pin tests.

**Open question (design, your call):** `resolveSession()` honors an explicit `session` arg before the pin — the PR's documented "caller always wins" precedence. But with `list-instances` hidden, an explicit session in a pinned session can only come from stale model context, and it's now the one remaining path to controlling a different Desktop. If the pin is meant as an invariant rather than a default, resolving/rejecting explicit-but-different sessions against the pin (with a pinned+explicit-other test) would close it; if multi-Desktop escape-hatch is intentional, a one-line comment in `resolveSession()` saying so would stop future reviewers from re-flagging it.

*Posted by MattGPT (automation on Matt Filbert's behalf); verified against PR head f508eef.*

mattcfilbert:
Fold-in complete on #478 as agreed: the vendored knowledge files were made pin-aware earlier (`1b4cebdb`), and `139cabf5` now covers the last static surface (`README.desktop.md`). The routeTable instruction prose was deliberately left to this PR's `sessionPinned` filter — no overlap. *— MattGPT (automation on Matt's behalf).*

tableaukyler:
Round 2 handled in 9cd813f.

**Pinned-missing error variant** — added. The `no-instance` error now carries the pinned pid, and `mapClientError` emits "Pinned Tableau Desktop (pid N) is no longer running — relaunch the agent from the Desktop you want to control." Both stale-pin tests (initial discovery + 401 rescan) now assert the pid shows up in the message.

**Explicit-but-different session** — treating the pin as an invariant, per your read. `resolveSession` now checks the pin first: it always wins, and an explicit `session` naming a different Desktop is rejected with an actionable error rather than overriding the pin. Explicit sessions still work normally when unpinned. Added the pinned+explicit-other (reject) and pinned+explicit-matching (allow) cases, and updated the precedence docstring.

Full unit suite green (3167), typecheck clean.

PR #478, opened 2026-07-10
desktop: day-1 vendor sync — full knowledge corpus + missing templates

mattcfilbert:
## Why

Day-1 Laulima dogfood (2026-07-09) hit two asset-staleness failures live: `Template "ww-ou-diff" not found` mid-build (Super Bowl session), and the agent emitting the invalid `cntd` derivation that `column-instance-prefixes.md` explicitly warns against — because that file (2026-06-25) postdates the last knowledge vendor copy. SEA packaging itself is sound (#433/#460/#472): it ships whatever is on disk. The disk content was a 16-file, pre-refactor knowledge snapshot and a drifted template set.

## What

- **Knowledge:** replace `resources/desktop/knowledge/` with the canonical ~90-file corpus from `agent-to-tableau-desktop/data/knowledge/` (new `tactics/` / `strategy/` / `personalization/` taxonomy) and migrate every hardcoded slug in tool instructions and tests (17 files, audited inventory).
- **Templates:** vendor the missing XMLs into **both** loadable dirs (`ww-ou-diff`, `spatial-filled-map`, `part-to-whole-waterfall-chart`, `ranking-bullet-chart` — plus three more the legacy dir alone was missing) with the `xmlns:user` stamps from a2td `3ee7bb6` (xmldom 0.9.10 NamespaceError fix — without these, inject-template stays broken in the binary regardless of asset coverage).
- **`ww-ou-diff` policy change (flagging for review, Kyler):** it was deliberately golden-only (manifest, no XML). Live evidence says that state strands users — the binder offers it, then inject-template 404s. It now ships, keeping `fast_path_eligible:false`, mirroring ww-ou-arrow's W59 move.
- **Drift prevention:** new `src/desktop/vendoredAssets.test.ts` — manifest↔XML orphan check, template-dir set equality, knowledge-corpus floor (≥80 files).

## Validation

- Full suite: 222 files, 3,134 tests green (3 assertions updated: URI sort order, ww-ou-diff golden-only → shipped, corpus 40→44)
- All 44 templates pass the `invalid-derivation-string` no-self-reject invariant
- `tsc --noEmit` clean

## Follow-up (separate)

Scripted `sync-desktop-assets` (same pattern as `sync-lockstep-core`) + provenance stamp so this copy is never manual again — packaging contract C2/C4.


mattcfilbert:
Two updates after the adversarial self-review + a context-budget check:

**Review finding fixed (`38f1e8d9`):** `content-manifest.json` was stale at 40 XML resources — `getContentManifest()`/`getStatus()` under-reported shipped content. Regenerated (87 resources, `content_version 2.11.0+content.2026-07-10`) and `vendoredAssets.test.ts` now enforces XML→content-manifest parity so a future sync can't skip the regen.

**Code ports (`3d8ab279`):** since `feature/authoring` is the live branch, the three a2td-side day-1 code fixes are ported here too — `ensureUserNamespace` self-heal before both `rewriteFieldReferences` call sites (xmldom 0.9.10 NamespaceError guard for any undeclared fragment), the `undeclared-aggregate-ok-ref` warning rule (21× live "Unknown column [sum:Sales:ok]"), and string-tolerant `search-commands` keywords (live -32602).

**Context-size note (re: the recent trim work):** this PR does not grow per-turn context. Knowledge is pull-only — tool descriptions keep one-line `expertise://` pointers, `DESKTOP_INSTRUCTIONS` is untouched, and file content enters context only on an explicit `read-knowledge-resource` call. The single growth surface is the MCP resources *listing* (16 → 90 one-line entries, ~10KB, served only when a client requests `resources/list`).

Full suite: 224 files, 3,148 tests green.

*Posted by MattGPT (automation on Matt Filbert's behalf).*

PR #502, opened 2026-07-14
@W-23388458 - Add insight-cards MCP tool (deterministic insights from datasource context)

tonythebest:
## Description

Adds a reusable path to `tableau-mcp` for generating deterministic, governed **insights** directly from **datasource context** — no pre-created metric definitions required.

New tools:
- `generate-insight-cards` — resolves a datasource, picks measures + a time field from metadata, calls the **Insight Service** to produce period-over-period insight cards (headline + delta, comparison line, top contributing factors, where it's most/least pronounced, what's unusual, suggested follow-ups, available actions), with drill-down via optional `breakdownDimension` + `filters` validated against datasource metadata.
- `resolve-datasource-luid` — deterministic `contentUrl` → LUID resolver (case-sensitive exact match).

Both tools enforce the bounded-context datasource allow-list (`resourceAccessChecker.isDatasourceAllowed`) before returning identity / metadata / insights. Registration + OAuth scoping wired; tools are opt-in (include-tools/profile), not frontloaded.

## Motivation and Context

Part of the Tableau Studio "insight cards on canvas" effort — governed insights from datasource context (no Insight Service metric/definition required). Base branch is `main` per Matt Filbert's guidance (2026-07-14). W-23388458.

## Type of Change
- [x] New feature

## How Has This Been Tested?
- `tsc --noEmit`, eslint, and `npm run build` all clean
- Unit tests for both tools — 11/11 pass (bounded-context denial with the guard running before any read, all-measures-failed error surfacing, drill-arg validation)
- Verified live against `GUS-Work` and `Superstore` on 10ax

Filter contract confirmed live: categorical filters use `operator: 'OPERATOR_EQUAL'` with `categorical_values: [{ string_value }]` in `metric_specification.filters`.

## Checklist
- [x] Version bumped
- [x] New and existing unit tests pass locally
- [ ] Docs updates — follow-up


src/tools/web/pulse/insights/generateInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+                restApi,
+                datasource,
+              });
+              if (!resolved) {
+                return new ArgsValidationError(
+                  'Could not resolve datasource from provided input',
+                ).toErr();
+              }
+
+              const metadataResult = await restApi.vizqlDataServiceMethods.readMetadata({
```
tonythebest, line 196: **[High] Missing bounded-context datasource guard before reading metadata / generating insights.**

This tool resolves a datasource and immediately calls VDS `readMetadata` + generates Pulse bundles over `resolved.luid`, with no allow-list check. Every sibling data-access tool gates first — e.g. `getDatasourceMetadata.ts:133` and `queryDatasource.ts:90` call `resourceAccessChecker.isDatasourceAllowed({ datasourceLuid, extra })`, and `generatePulseMetricValueInsightBundleTool.ts:154` checks `boundedContext.datasourceIds`.

Because `resourceAccessChecker` reads the bounded context off `extra.getConfigWithOverrides().boundedContext` (already available here), a caller can pass any `contentUrl` or `{ luid }` and receive governed metadata and period-over-period insight values for a datasource outside the server's allowed set. This is a tenant-isolation / access-scoping gap.

**Fix:** after resolving `resolved.luid` and before `readMetadata`, call `resourceAccessChecker.isDatasourceAllowed({ datasourceLuid: resolved.luid, extra })` and return `DatasourceNotAllowedError` on denial. Add a test where `boundedContext.datasourceIds` excludes the resolved LUID and assert neither `readMetadata` nor the bundle call runs.

src/tools/web/datasources/resolveDatasourceLuid.ts
```diff
@@ -0,0 +1,66 @@
...
+            jwtScopes: tool.requiredApiScopes,
+            callback: async (restApi) => {
+              const response = await restApi.datasourcesMethods.listDatasources({
+                siteId: restApi.siteId,
+                filter: `contentUrl:eq:${contentUrl}`,
+                pageSize: 100,
+                pageNumber: 1,
+              });
+
+              const exact = response.datasources.find((d) => d.contentUrl === contentUrl);
```
tonythebest, line 51: **[High] Resolver returns datasource identity outside bounded context.**

`list-datasources` filters returned rows by `boundedContext` (`projectIds` / `datasourceIds` / `tags`) before returning; this resolver returns LUID/name/contentUrl for any exact `contentUrl` match with no such check. Even without row-level data, it's an existence/LUID oracle for datasources outside the allowed set, and it feeds the LUID straight into `generate-insight-cards` (making the guard gap there easier to reach).

**Fix:** after finding `exact`, run `resourceAccessChecker.isDatasourceAllowed({ datasourceLuid: exact.id, extra })` and return `DatasourceNotAllowedError` on denial. Add a bounded-context test.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+                  allowedDimensions: activeDimensions,
+                  filters,
+                });
+                const bundle = await runChironBundle({
+                  extra,
+                  request,
+                  bundleType: CHIRON_BUNDLE_TYPE,
+                  jwtScopes: tool.requiredApiScopes,
+                });
+                if (bundle.isErr()) {
```
tonythebest, line 311: **[Medium] Per-measure bundle failures are silently swallowed.**

`if (bundle.isErr()) { continue; }` means if Pulse rejects the request, auth/scoping fails, the bundle API is down, or every selected measure errors, the tool returns `isError=false` with `cards: []` — indistinguishable from a legitimate no-data result, and it hides exactly the Chiron API-contract failures that are hardest to debug.

**Fix:** collect per-measure errors; return a tool error when zero cards were produced and at least one bundle failed. If partial success is intended, include a `warnings`/`errors` array in the payload with the measure and error text.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+                  'Could not determine numeric measures for this datasource',
+                ).toErr();
+              }
+
+              // The full set of categorical dimensions is always returned to the
+              // client so every (including drilled) insight can propose fresh
+              // "break down by <other dimension>" follow-ups. A drill request
+              // only narrows the *active* set used for this bundle's breakdown.
+              const allDimensions = pickDimensions(metadataResult.value.data ?? []);
+              const activeDimensions = breakdownDimension ? [breakdownDimension] : allDimensions;
```
tonythebest, line 290: **[Medium] Caller-supplied `breakdownDimension` / `filters[].field` aren't validated against metadata.**

These strings are threaded directly into `allowed_dimensions` / `metric_specification.filters`. A typo or stale field name produces an opaque Pulse 400 that is then swallowed by the `continue` above, so it surfaces as an empty `cards: []`. Metadata (`allDimensions`) is already in hand here, so we can fail fast.

**Fix:** verify `breakdownDimension` and each `filters[].field` are present in `allDimensions` (consider validating explicit `measures`/`timeField` too) and return a clear `ArgsValidationError` otherwise. Also make `filters[].value` non-empty in the schema.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+      }))
+      .filter(
+        (item): item is { label: string; value: number; formatted: string | null } =>
+          item.label !== '' && item.value !== null,
+      )
+      .slice(0, 6);
+    if (items.length === 0) {
+      continue;
+    }
+    const total = items.reduce((sum, item) => sum + Math.abs(item.value), 0);
```
tonythebest, line 651: **[Medium] `sharePct` uses `Math.abs`, distorting share-of-lift for mixed-sign contributors.**

The denominator sums `Math.abs(item.value)` (this line) and each `sharePct` divides `Math.abs(item.value)` by it (`:510`). For a mixed-sign contributor set (some up, some down), the shares no longer reflect net contribution and won't sum meaningfully. It's invisible today because the only `topcontributors` fixture is all-positive (900/600/300).

**Fix:** decide the intended semantic — share of *net* change (keep signs) vs share of *gross* movement (keep abs but label/comment it) — and add a mixed-sign fixture. Bounded to Medium since it's a labeled display heuristic on a draft and nothing downstream computes from it.

src/tools/web/pulse/chiron/generateChironInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+        pageSize: number;
+        pageNumber: number;
+      }) => Promise<{ datasources: Array<{ id: string; name: string; contentUrl?: string }> }>;
+    };
+    siteId: string;
+  };
+  datasource: string | { luid: string };
+}): Promise<{ luid: string; name: string; contentUrl: string } | null> {
+  if (typeof datasource !== 'string') {
+    return { luid: datasource.luid, name: datasource.luid, contentUrl: datasource.luid };
```
tonythebest: **[Low] `{ luid }` branch fabricates `name`/`contentUrl` from the LUID.**

This returns `name` and `contentUrl` set to the raw LUID string (contrast `:336-339`, which returns the real `exactMatch.name`/`contentUrl`). Any card header rendered from `datasource.name` would show a GUID. This is also the branch that skips `listDatasources` entirely, so it's the path the bounded-context check (see the High comment above) must still cover.

**Fix:** resolve the LUID to the real datasource (and share the resolver with `resolve-datasource-luid`), or omit `name`/`contentUrl` when unknown rather than aliasing them to the LUID.

src/tools/web/pulse/chiron/generateChironInsightCardsTool.ts
```diff
@@ -0,0 +1,751 @@
...
+
+              return Ok({
+                datasource: {
+                  luid: resolved.luid,
+                  contentUrl: resolved.contentUrl,
+                  name: resolved.name,
+                },
+                dimensions: allDimensions,
+                cards,
+                generatedAt: new Date().toISOString(),
```
tonythebest: **[Low] `generatedAt: new Date().toISOString()` makes output nondeterministic.**

The tool is framed as a *deterministic* insight path, but this wall-clock stamp breaks byte-for-byte reproducibility (snapshot tests, caching, replay). Minor, but worth reconsidering given the deterministic framing — source the timestamp from an injected clock via `extra`, or drop the field.

src/tools/web/pulse/chiron/runChironBundle.ts
```diff
@@ -0,0 +1,30 @@
...
+  pulseBundleRequestSchema,
+  PulseBundleResponse,
+  PulseInsightBundleType,
+} from '../../../../sdks/tableau/types/pulse.js';
+import { TableauWebRequestHandlerExtra } from '../../toolContext.js';
+
+export async function runChironBundle({
+  extra,
+  request,
+  bundleType = 'ban',
```
tonythebest: **[Note] Dead `bundleType = 'ban'` default.**

The only caller always passes `CHIRON_BUNDLE_TYPE` (`'detail'`), so this default never executes and misleadingly implies a BAN path. Make `bundleType` required, or default it to `'detail'` to match actual usage.

src/tools/web/pulse/chiron/requestBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+  datasourceLuid,
+  datasourceName,
+  measure,
+  timeField,
+  allowedDimensions = [],
+  filters = [],
+}: BuildChironBundleRequestArgs): z.infer<typeof pulseBundleRequestSchema> {
+  // Categorical equality filters scope the metric to specific dimension members
+  // (drill-down). Shape matches pulseFilterSchema.
+  // Categorical equality filter scoping the metric to a dimension member.
```
tonythebest: **[Note] Duplicated filter comment + hardcoded aggregation.**

This comment restates the one on lines 24-25 — collapse the two. Separately, `aggregation: 'AGGREGATION_SUM'` (`:55`) is hardcoded; that's probably fine for the draft, but average/count-style measures will produce misleading cards, so it's worth calling out as an explicit known limitation.

tonythebest (commented):
Meta-review of the Chiron insight-cards tools. Verified locally against this exact head (`ba986942`): `npm run build` + eslint clean, 7/7 targeted unit tests pass under Node 22.23.1, and all 7 CI checks green. The design is strong — all four W-23388458 ACs are delivered, the live-discovered Pulse categorical-filter contract (`OPERATOR_EQUAL` + `string_value`-only in `metric_specification.filters`) is correct, OAuth scopes match the call path, and the `contentUrl` type addition is backward-compatible.

Two **blocking** issues before this is merge-safe, both inline below: neither new tool enforces the bounded-context datasource guard that all 16 sibling data-access web tools use (`resourceAccessChecker.isDatasourceAllowed` / `boundedContext.datasourceIds`). A caller can resolve an arbitrary datasource LUID and pull governed metadata + insight values outside the configured allow-list. There are also a few Medium correctness/observability items and some Notes.

Nit (not a code line): the PR title/body still say `generate-chiron-insight-cards` / "Chiron insight-cards", but the registered tool name is now `generate-insight-cards` (the intended user-facing name). Worth updating the title/body; the internal `chiron` codename in file/dir names is fine.

Note: posting these as review comments only — no code changes in this review.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
tonythebest: Fixed in `30d07150`: added `resourceAccessChecker.isDatasourceAllowed({ datasourceLuid: resolved.luid, extra })` right after resolve and before `readMetadata`, returning `DatasourceNotAllowedError` on denial. Added a bounded-context denial test asserting neither `readMetadata` nor the bundle call runs.

src/tools/web/datasources/resolveDatasourceLuid.ts
tonythebest: Fixed in `30d07150`: after the exact contentUrl match the resolver now calls `isDatasourceAllowed({ datasourceLuid: exact.id, extra })` and returns `DatasourceNotAllowedError` on denial. Added a bounded-context denial test.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
tonythebest: Fixed in `30d07150`: `breakdownDimension` and each `filters[].field` are validated against the datasource's categorical fields (`ArgsValidationError` on miss), and `filters[].value` is now `.nonempty()`. Added a validation test.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
tonythebest: Addressed in `30d07150`: documented the intended semantic — `sharePct` is share of *gross* movement (abs contribution / sum of abs), a labeled display heuristic. Left signed/net share as a follow-up since nothing downstream computes from it.

src/tools/web/pulse/insights/generateInsightCardsTool.ts
tonythebest: Fixed in `30d07150`: per-measure errors are collected; when every measure fails the tool returns the real bundle error (not an empty success), and partial failures are reported in a `warnings[]` array. Added a test for the all-fail path.

src/tools/web/pulse/chiron/generateChironInsightCardsTool.ts
tonythebest: Fixed in `30d07150`: the `{ luid }` branch now resolves the real identity via `queryDatasource` (graceful fallback) instead of aliasing name/contentUrl to the raw LUID.

src/tools/web/pulse/chiron/generateChironInsightCardsTool.ts
tonythebest: Fixed in `30d07150`: removed the nondeterministic `generatedAt` field.

src/tools/web/pulse/chiron/runChironBundle.ts
tonythebest: Fixed in `30d07150`: `runChironBundle` default `bundleType` is now 'detail' (matches the only caller).

src/tools/web/pulse/chiron/requestBuilder.ts
tonythebest: Fixed in `30d07150`: collapsed the duplicated filter comment and added an explicit known-limitation note that `aggregation` is hardcoded to `AGGREGATION_SUM`.

mattcfilbert:
Two findings from a review pass against the bounded-context claim (both verified in the head commit, 6aeace4):

**1. `resolve-datasource-luid` still leaks datasource existence outside the bounded context.** An absent contentUrl returns `No datasource matched contentUrl "…"` (`resolveDatasourceLuid.ts:49`), but an existing-out-of-context one returns `DatasourceNotAllowedError` (`:61`) — the error-type distinction lets a caller enumerate which non-allowed datasources exist. The identity-withholding comment at `:55` shows the oracle was considered; the distinguishable errors reopen it. Same shape in `generate-insight-cards`' resolve path, and its `{ luid }` input runs `queryDatasource` before the `isDatasourceAllowed` gate (`generateChironInsightCardsTool.ts:390` vs `:190`) — the deny path doesn't echo the name, so this half is guard-ordering hygiene, but it contradicts the PR's "before identity / metadata / insights" claim. Fix: return one indistinguishable error for absent and not-allowed, and gate before any identity lookup. Test: assert identical visible results for absent vs out-of-context contentUrl, for both tools; assert a denied `{ luid }` never calls `queryDatasource`/`readMetadata`/Pulse.

**2. The tool is frontloaded by default — the "opt-in (include-tools/profile), not frontloaded" claim doesn't hold in code.** The registration loop only skips on a `disabled` provider or include/exclude lists (`server.web.ts:135-136`), and `generateChironInsightCardsTool` sets no `disabled` flag, so an empty `INCLUDE_TOOLS` registers it for everyone. Fix: add a disabled-by-default provider (or profile gate) so registration requires `INCLUDE_TOOLS=generate-insight-cards`/`insights`. Test: default registration excludes it; the include-tools path includes it. When this rebases to feature/desktop, also confirm the tool lands in the lazy web set under `combined-lean` rather than the eager surface.

Minor (no response needed): `measures`/`timeField` overrides bypass the metadata validation that `breakdownDimension`/filters get (`requestBuilder.ts:53,60`), and the determinism fix still depends on VDS field order (dedupe without a stable sort, `generateChironInsightCardsTool.ts:353`).

---
*Posted by MattGPT (automation on behalf of Matt Filbert), review adjudicated by Matt's agent stack.*

tonythebest:
Thanks @mattcfilbert — both findings addressed in `c9eed177`:

**1. Existence oracle + gate ordering.** `resolve-datasource-luid` and `generate-insight-cards` now return an **identical** error for absent vs exists-but-out-of-context (404-style, no distinguishable 403), and the allow-list check runs **before any identity lookup** — the `{ luid }` path gates before `queryDatasource`. Tests assert identical visible results for absent vs out-of-context (both tools) and that a denied `{ luid }` never calls `queryDatasource` / `readMetadata` / the Insight Service.

**2. Frontloading.** `generate-insight-cards` is now **disabled-by-default** and only registers when listed in `INCLUDE_TOOLS` (a `disabled` provider gating on `getOverridableConfig().includeTools`). Tests assert default-off / included-on. I'll confirm it lands in the lazy web set under `combined-lean` when this rebases to `feature/desktop`.

**Minors** also done: `measures`/`timeField` overrides are validated against metadata (same fail-fast as the drill inputs), and the series sort now has deterministic tiebreakers (date → value → label) so ordering no longer depends on VDS row order.

Full suite 2138/2138; typecheck/lint/build clean.

mattcfilbert:
**P2 — deterministic field selection is still metadata-order-dependent (the fix commit covers series points, not field selection).**

The new tie-broken sort makes the *series points* deterministic:

```ts
// generateInsightCardsTool.ts ~585
.sort(
  (a, b) =>
    a.date.localeCompare(b.date) ||
    (a.value ?? Number.POSITIVE_INFINITY) - (b.value ?? Number.POSITIVE_INFINITY) ||
    a.label.localeCompare(b.label),
);
```

But the *selected fields* still depend on VDS metadata response order:

```ts
// pickTimeField — first DATE/DATETIME in response order wins
const field = fields.find(
  (item) =>
    typeof item.fieldCaption === 'string' &&
    (item.dataType === 'DATE' || item.dataType === 'DATETIME'),
);
```

```ts
// pickMeasures — dedup preserves response order, then truncates
return Array.from(new Set(measures)).slice(0, maxCards);
```

If VDS returns equivalent metadata in a different order, the chosen time field, the measure set (after `slice(0, maxCards)`), and therefore the generated cards all change run-to-run.

**Fix:** stable-sort the candidate fields (e.g. by `fieldCaption` with a deterministic tie-breaker) before `find`/dedup/truncate in `pickTimeField`, `pickMeasures`, and the dimension candidate selection.

**Test to add:** shuffled-but-equivalent metadata produces identical selected time field, measures, and card request order.

---
*Posted by MattGPT (Claude automation) on Matt Filbert's behalf.*

mattcfilbert (approved):
lgtm. approved with follow up comments

jarhun88:
can we set a env variable with this disabled by default? we are trying to rollout our mcp server to slackbot and trying to keep the build stable atm. see admin tools env var for example

tonythebest:
Both follow-ups addressed in `f0056623` (rebased onto latest `main`, now 3.1.2):

**@jarhun88 — env-var gate (disabled by default).** Added `INSIGHTS_TOOLS_ENABLED`, mirroring `ADMIN_TOOLS_ENABLED` / `FLOW_TOOLS_ENABLED` (config field + manifest entry + `ProcessEnvWeb`). **Both** insight tools — `generate-insight-cards` **and** `resolve-datasource-luid` — are now `disabled: !config.insightsToolsEnabled`, so they won't register in the Slackbot build unless the env var is `"true"`. `INCLUDE_TOOLS=insights` still works as an additional scope. Registration tests added (`server.web.test.ts`): not registered by default, registered when the env var is set.

**@mattcfilbert — P2 deterministic field selection.** `pickTimeField` / `pickMeasures` / `pickDimensions` now dedupe **and stable-sort** candidate captions (`localeCompare`) before find/slice (shared `sortedCaptionsByType` helper), so the selected time field, measures, and dimensions no longer depend on VDS metadata response order. Added the shuffled-metadata test you suggested: reversed-but-equivalent metadata yields identical time field, measures, and dimensions.

Full suite green (2291 tests); typecheck/lint/build clean.

mattcfilbert (approved):
double approved

PR #513, opened 2026-07-15
@W-23375797: fix unawaited async feature gate in confirmDeleteContent

Akash-Rastogi:
## Summary

- **Fix unawaited `isFeatureEnabled` call** in `confirmDeleteContent.ts` — PR #497 made `getFeatureGate().isFeatureEnabled()` async, but this call site was missed. The result was `!Promise` → always `false`, so the tool was never disabled when the `mcp-apps` feature gate was OFF.
- **Wrap in `Provider(async () => ...)`** to match the pattern already used in `confirmUpdateCloudExtractRefreshTask.ts`.
- **Update test mocks** from `mockReturnValue` to `mockResolvedValue` to align with the async interface.

## Details

The `disabled` field on `confirm-delete-content` was:
```ts
disabled: !config.adminToolsEnabled || !getFeatureGate().isFeatureEnabled('mcp-apps'),
```

Since `isFeatureEnabled` now returns `Promise<boolean>`, `!Promise` is always `false`, meaning the `||` short-circuits and `disabled` is only `true` when `adminToolsEnabled` is `false`. When `adminToolsEnabled` is `true`, the tool always registers regardless of the feature gate.

**Impact:** LOW — the tool leaked into `tools/list` but couldn't complete a delete (the `AllEvidence` guard blocks execution). Correctness/hygiene fix only.

**Grep audit:** Confirmed this was the **only** remaining sync `isFeatureEnabled` call site in the codebase. All other usages properly `await`.

## Test plan

- [x] Existing test "is disabled when the mcp-apps flag is OFF" now exercises the async path via `Provider.from(tool.disabled)`
- [x] All 2125 unit tests pass
- [x] Build, tsc --noEmit, lint all clean
- [x] No other unawaited call sites found via grep audit


Alon-ST-DATA (approved):
APPROVE. This fixes the unawaited async feature-gate regression that live e2e caught on #498. Verified in two independent passes (review + build):

- **Root cause fixed, non-cosmetic** — `confirmDeleteContent.ts:44-47` now `new Provider(async () => !config.adminToolsEnabled || !(await getFeatureGate().isFeatureEnabled('mcp-apps')))`, matching the sibling. Confirmed the resolver is genuinely awaited at registration: `server.web.ts:135` `if (await Provider.from(tool.disabled)) continue;` (+ `provider.ts` awaits `.get()`), so the gate actually takes effect — not just reworded.
- **Completeness** — full-tree grep of `isFeatureEnabled`: all 7 runtime consumers are awaited/Provider-wrapped; this was the only broken site. No sibling left unawaited by #497.
- **Regression test is load-bearing** — `confirmDeleteContent.test.ts:199` switched the mock `mockReturnValue(false)` → `mockResolvedValue(false)` and asserts `await Provider.from(tool.disabled) === true`; reverting the source to the sync form fails it. Plus `init.test.ts:189` async-contract test.
- **Green** — tsc/lint/build clean; affected unit tests 11/11 (39/39 broader) pass.
- **No regressions** — flag-OFF path restored, other gated tools untouched, scope-map parity intact (both admin-off and mcp-apps-off blocks still drop `confirm-delete-content`).

One non-blocking note: version numbering looks inconsistent across the in-flight 3.x branches (commit message references 3.0.2 while the branch carries 3.1.1). Whoever merges should confirm the next-free patch version at merge time. Not a code defect.

0 BLOCKER / 0 MAJOR / 1 NIT. Clean to merge. Running a live e2e pass now to confirm the fix on the built server.

Alon-ST-DATA (approved):

Alon-ST-DATA:
## Live E2E Verification — regression fix confirmed ✅

Ran a live e2e pass on the built PR branch to confirm the fix on a real server.

**Server:** `tableau-mcp v3.1.1`, HTTP transport, `ADMIN_TOOLS_ENABLED=true`, connected to a production Tableau Cloud site. Exactly one feature flag loaded, **mcp-apps OFF**. tools/list fetched by raw JSON-RPC directly from the server (bypasses any harness-side filtering); version confirmed via `initialize` → `serverInfo.version: "3.1.1"`.

| # | Check | Result | Evidence |
|---|-------|--------|----------|
| **1** | **`confirm-delete-content` ABSENT (headline — the v3.0.0 leak)** | **PASS** | Not in tools/list. The app-only sibling `confirm-update-cloud-extract-refresh-task` is also absent — the two now behave identically with mcp-apps OFF, which is the correct post-fix behavior. |
| 2 | Tool count / delta | **PASS** | 27 model-visible tools. The prior v3.0.0 run saw 28 (= 27 + the 1 leaked tool). Delta **28 → 27** — the single leaked tool is gone. |
| 3 | `delete-content` present + model-visible (not app-only) | **PASS** | Present; preview exercised successfully. |
| 4 | Regression-adjacent gating | **PASS** | `get-embed-token` absent (mcp-apps OFF). All admin tools present: `list-extract-refresh-tasks`, `list-jobs`, `list-users`, `update-cloud-extract-refresh-task`, `query-admin-insights`. |
| 5 | Health / auth | **PASS** | `list-workbooks` returns real data; count matches independent REST `totalAvailable`; `limit` honored. |
| 6 | `delete-content` preview non-destructive (the #498 consolidation is unbroken) | **PASS** | Preview tagged `pending-deletion` (reversible), asked for confirmation, deleted nothing. Invalid LUID → clean `isError` 404, no crash. |

**Headline:** the regression is fixed. In v3.0.0 the app-only `confirm-delete-content` leaked into the model-facing tools/list because `!getFeatureGate().isFeatureEnabled('mcp-apps')` evaluated `!Promise` (always false) without `await`. On this v3.1.1 build with mcp-apps OFF it is correctly absent, matching its app-only sibling.

**Safety:** no content was deleted. The preview applied one reversible `pending-deletion` tag to a workbook, which was then removed via REST (verified tags → `[]`). `confirm:true` was never called against any real resource.

_Manual live-verification run; the flag-ON direction was not exercised (only mcp-apps=false is loaded on this build) — that path is covered by the unit test added in this PR._

PR #529, opened 2026-07-16
@W-23377934: fix extract-refresh-task delete preview existence check + confirm double-audit

Alon-ST-DATA:
## Summary

Fixes two defects in the consolidated `delete-content` flow (W-23377934, WAM: Tableau MCP for Admins - Part 1).

### DEFECT 1 — preview mints a confirmation token for a non-existent taskId
`runExtractRefreshTaskBranch`'s `resolveTarget` was a pure echo with no REST read, so a well-formed but non-existent `taskId` passed the UUID-shape check, `guardMutation` minted a `RegistryEvidence` nonce, and the preview returned a live `confirmationToken` for a task that doesn't exist. The workbook/datasource branches don't have this problem — they read the target via `getWorkbook`/`queryDatasource` and 404 before any evidence is established. The task branch never read the target, and there is no single-get task endpoint.

**Fix:** before `guardMutation`, verify the task exists via `listExtractRefreshTasks` (list-and-find, since there is no single-get endpoint). An unknown `taskId` returns a `404` not-found error and **no** `confirmationToken` is minted. Runs unconditionally on both preview and confirm, symmetric with the sibling branches. This read requires `tableau:tasks:read`, added to the `delete-content` scope map (`confirm-delete-content` unchanged — its task branch does no read).

### DEFECT 2 — confirm phase double-logs audit
Audit emission is centralized in `guardMutation`. On a successful confirm, two records fired: the unconditional `emitAuditRecord({result:'allowed'})` **plus** the terminal `recordOutcome` emit (`completed`/`failed`). Because it lives in the shared guard, this double-logged **every** confirmed mutation across all resourceTypes and every preview→confirm tool.

**Fix:** suppress the `'allowed'` record on the confirm phase (`if (phase !== 'confirm')`). A confirm now emits exactly one terminal record (`completed`/`failed`); a preview still emits its single `allowed`; denied paths are unaffected. Audit-lifecycle doc comments updated to the one-record contract.

> Note: this deliberately changes a previously-documented audit contract (a confirm now logs once instead of `allowed`+terminal), per the WI's "logged exactly once". All production confirm call sites invoke `recordOutcome` on both success and failure paths, so no path is left unaudited.

## Tests
6 new pinning unit tests:
- **D1:** preview / confirm on nonexistent task + empty-list → `404`, no token minted, no delete, zero audit records.
- **D2:** successful confirm → exactly one `completed`; failed confirm → exactly one `failed`; preview → single `allowed` (regression guard).

Negative control: reverting each fix makes exactly the pinning tests fail. Full suite passes; targeted delete/guard/oauth suites green. `tsc`, `lint`, `build` all clean. Non-destructive live verify on a Cloud test site: bogus UUID → not-found + no token; real task → token minted, no delete.

## Docs
- `delete-content.md`: documents the extract-refresh-task existence check; corrects the REST-API reference to *List Extract Refresh Tasks* (the previously-cited *Get Extract Refresh Task* endpoint does not exist).

## Version
`3.1.2` → `3.1.3` (next free patch).

### DIFF
--- a/docs/docs/tools/content/delete-content.md
+++ b/docs/docs/tools/content/delete-content.md
@@ -23,8 +23,10 @@ The tool is **two-phase** to keep the destructive action safe:
 1. **Preview** (default — `confirm` omitted or `false`):
    - For `workbook` / `datasource`: tags the resource with `pending-deletion` (reversible,
      visible in the Tableau UI), reports identity, project, and owner.
-   - For `extract-refresh-task`: mints a single-use `confirmationToken` and reports task
-     metadata.
+   - For `extract-refresh-task`: first verifies the task exists on the site (there is no single-get
+     endpoint, so the task list is checked for a matching id) — an unknown `taskId` returns a
+     not-found error and **no** `confirmationToken` is minted. When the task exists, mints a
+     single-use `confirmationToken` and reports task metadata.
    - Does **not** delete anything.
 
 2. **Confirm** (`confirm: true`):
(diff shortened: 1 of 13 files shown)


Akash-Rastogi (approved):
[Akash's agent] LGTM. Both fixes are correct — the list-and-find existence check runs unconditionally (symmetric with the sibling branches), and suppressing the `allowed` audit on confirm eliminates the double-log cleanly. Test coverage is thorough.

PR #530, opened 2026-07-16
@W-23146707 Add telemetry for errors and URL button clicks in MCP app

jarhun88:
## Description

Adds product telemetry for MCP app UI errors via a new **app-only** server tool.

- added New tool `record-mcp-app-error` (`src/tools/web/recordMcpAppError/`): the sandboxed MCP app calls it via `app.callServerTool` to record a `tableau_mcp_event.completed` product-telemetry event through the existing `DirectTelemetryForwarder`. Its also called when url link is clicked

- this is added for any mcp app call that fails (for the embedded viz mcp app scenario only)

- Registration + scopes: added to `toolName.ts` / `tools.ts`, empty mcp/api scopes in `scopes.ts`, and safelisted in the passthrough-auth guard (the tool makes no Tableau REST API calls).

- also added a error_message field to the tool call product telemetry to get better undrestanding of onprem errors.

- added white coloring to the mcp app error background 

## Motivation and Context

MCP app UI errors were invisible to product telemetry. This routes them through the server (never the browser) so the ingest endpoint/credentials stay inside the trust boundary and events are enriched with request/session/site/user/pod context the client can't produce. The tool is hidden from the model and gated on the `mcp-apps` feature flag, mirroring `get-embed-token`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

New/updated unit tests (all passing locally via `npx vitest run`):
- `recordMcpAppError.test.ts` (server tool: properties, app-only visibility, telemetry payload + server enrichment, feature gate)
- `recordMcpAppErrorClient.test.ts` (fire-and-forget client: never throws, no-ops without serverTools, omits message when no cause)
- `showError.test.ts`, `handleToolResult.test.ts`, `handleConfirmResult.test.ts` (wiring across all four scenarios, incl. container-missing)
- `passthroughAuthMiddleware.test.ts` (no-API-scope safelist)

Full suite green apart from one pre-existing, unrelated failure (`adminInsights/getStaleContentReport` date-fixture rot). Lint clean on changed files; MCP app bundles build successfully.

## Related Issues

_None._

## Checklist

- [x] I have documented any breaking changes in the PR description. (None — new app-only tool, no config/API changes.)


src/server/oauth/scopes.ts
```diff
@@ -297,6 +297,11 @@ const toolScopeMap: Record<
     mcp: [],
     api: new Set<TableauApiScope>(),
   },
+  // MCP-app event telemetry relay: no Tableau REST API calls, no content scope required.
+  'record-event': {
+    mcp: [],
```
stephendeoca, line 302: why are mcp scopes empty

src/tools/web/tool.ts
```diff
@@ -162,7 +163,7 @@ export class WebTool<Args extends ZodRawShape | undefined = undefined> extends T
 
     let success = false;
     let errorCode = ''; // HTTP status category: "4xx", "5xx", or empty for successful calls
-    let toolResult: CallToolResult;
+    let toolResult: CallToolResult | undefined;
```
stephendeoca, line 166: Why undefined here? Shouldn't every tool call just return CallToolResult even if its empty

stephendeoca (approved):

src/server/oauth/scopes.ts
jarhun88: `mcp` scopes are the required MCP OAuth scopes the auth middleware checks before allowing the call (`getRequiredMcpScopesForRequest` → `403 insufficient_scope` if any are missing). 


`record-event` is an app-only telemetry relay that makes no Tableau REST API calls — it just forwards a `tableau_mcp_event` through the existing telemetry forwarder — so there is no content permission to gate on, and nothing to mint into the JWT (hence `api` is empty too). 

Same pattern as `revoke-access-token` / `reset-consent` (both empty), and matches `get-embed-token`'s `mcp: []`.

src/tools/web/tool.ts
jarhun88: Good question — the method does still always return a `CallToolResult` (return type is unchanged, and every `return` assigns a real result first). The `| undefined` is only about the local variable, and it's driven by the new `error_message` telemetry line that reads `toolResult` inside the `finally` block:

```ts
error_message: toolResult?.isError ? extractToolErrorMessage(toolResult) : '',
```

A `finally` runs on every exit, including when the `catch` itself throws before its assignment (`toolResult = getErrorResult(...)`) completes — e.g. if `getHttpStatus(error)` or `log(...)` throws first. On that path `toolResult` is genuinely still unset when the `finally` reads it, so TypeScript's definite-assignment analysis flags `let toolResult: CallToolResult;` as TS2454 ("used before being assigned"). This is structural: TS won't trust a try/catch assignment when the read is in `finally`, even though both branches assign it.

I considered initializing it to a default `CallToolResult` to drop the `| undefined`, but a `{ isError: false, content: [] }` sentinel is misleading — it reads as "the call succeeded" on the one path that actually observes it, which is a failure. `| undefined` is the honest type: at that read the result may not exist yet, and `?.` handles it (falling back to `error_message: ''`, same as before).

PR #533, opened 2026-07-16
@W-23125362: complete mutation-tool safety layer — audit target enrichment + centralized HITL text + durability contract

Alon-ST-DATA:
## What & why

Scope-completes GUS **W-23125362** ("Common mutation-tool safety layer: enforced HITL approval + authoritative audit across all TMCP mutation tools"). The shared guard + evidence model + audit lifecycle already landed and cover the four content/task mutators (`delete-content` wb/ds/task, `confirm-delete-content`, `update-cloud-extract-refresh-task`, `confirm-update-cloud-extract-refresh-task`). This PR finishes the in-scope remainder.

Three items were **deliberately forked** to a follow-up story (they require new architecture, not completion of this layer): auditing the OAuth session tools (`revoke-access-token`, `reset-consent`), a durable/tamper-resistant external sink, and the e2e test tiers.

## Changes

**AC-3 (Gap B) — enrich extract-refresh-task audit target**
- Audit records for extract-refresh-task paths carried only `{id, kind}`. Now populate `name`/`project`/`owner`, derived from the underlying datasource (else workbook) — a task's natural identity.
- New shared helper `resolveExtractRefreshTaskTarget.ts`; best-effort, degrades to id-only on Server sites or 403/404 and **never blocks or fails the mutation**.
- Applied to all four task paths. The `delete-content` task branch reuses its existing existence-check task list (no extra round-trip); the other three paths do a best-effort list+content+owner lookup (no single-task GET endpoint exists).

**AC-4 (Gap D) — centralize tool-side HITL text**
- Prompt-side text was already centralized; the tool-side "NEXT STEP — REQUIRED", confirm-closed, and `PreviewNotRunError` recovery strings were hand-rolled per branch.
- Extracted into new `hitlText.ts` (kept in `tools/_lib` to avoid a tools→prompts cross-layer import). Byte-for-byte behavior-preserving; rethreaded `delete-content`, `update-cloud-extract-refresh-task`, and the guard's PreviewNotRun branch.

**AC-3 (Gap C, doc) — durability contract**
- Documented that authoritative audit records emit as structured JSON on the dedicated `audit` logger (bypasses `LOG_LEVEL`), and that durable retention is the deployment's responsibility (ship the stream to an immutable store). Added to `auditRecord.ts` and `delete-content.md`. No new sink built.

**AC-6 (Gap E, unit) — per-tool audit test completeness**
- For the four in-scope mutators: forged-confirm rejected, happy preview→confirm, and audit record fields on **both allow and deny** — including assertions that the newly enriched `name`/`project`/`owner` appear. Unit tier only (e2e forked).

## Testing
- `npx vitest run` — **150 files / 2306 tests, all pass**
- `npx tsc --noEmit`, `npm run lint`, `npm run build` — clean
- Version 3.1.3 → **3.1.4**

## Notes
- No tools added/removed; `tests/e2e/server.test.ts` "should list tools" unaffected.
- Enrichment is best-effort by design — on Server-only sites or lookup failures, task targets degrade to id-only (documented in code + `delete-content.md`).

### DIFF
--- a/src/tools/web/_lib/resolveExtractRefreshTaskTarget.ts
+++ b/src/tools/web/_lib/resolveExtractRefreshTaskTarget.ts
@@ -1,0 +1,96 @@
+import { log } from '../../../logging/logger.js';
+import { RestApi } from '../../../sdks/tableau/restApi.js';
+import { ExtractRefreshTask } from '../../../sdks/tableau/types/extractRefreshTask.js';
+import { getExceptionMessage } from '../../../utils/getExceptionMessage.js';
+import { resolveOwnerEmail } from '../users/resolveOwnerEmail.js';
+import { MutationTarget } from './mutationGuard.js';
+
+/**
+ * Best-effort resolution of an extract-refresh task's audit-target identity (AC-3, W-23125362).
+ *
+ * IDENTITY MAPPING (lead decision): an extract-refresh task has no natural name/project/owner of its
+ * own. The Tableau REST list entry carries ONLY `{ id, schedule?, datasource?: {id}, workbook?: {id} }`
+ * (see src/sdks/tableau/types/extractRefreshTask.ts) and there is NO single-task GET endpoint — so a
+ * task's identity DERIVES from its underlying content:
+ *   - `target.name`    = the underlying datasource/workbook name
+ *   - `target.project` = its project name
+ *   - `target.owner`   = its owner email/name (via resolveOwnerEmail, best-effort)
+ * If the task has no datasource/workbook reference, or the content lookup fails, the field is left
+ * undefined (documented best-effort). This function NEVER throws: every lookup is wrapped so a resolve
+ * failure degrades to an id-only target and can never block or fail the guarded mutation/audit.
+ *
+ * COST: the task list yields only the underlying content ID, so enriching name/project/owner requires
+ * ONE content lookup (queryDatasource | getWorkbook) plus ONE owner lookup (queryUserOnSite) —
+ * mirroring the cost the workbook/datasource delete branches already pay. Callers that ALREADY hold
+ * the task list (the delete-content task branch runs it as its existence check) pass it via `tasks` to
+ * avoid a re-list; the confirm/update paths omit it and this helper lists once itself.
+ *
+ * This is single-sourced across all four extract-refresh-task mutation paths (delete-content task
+ * branch, update-cloud-extract-refresh-task, confirm-delete-content, confirm-update-cloud-extract-refresh-task)
+ * so the identity mapping lives in exactly one place.
+ */
+export async function resolveExtractRefreshTaskTarget({
+  restApi,
+  siteId,
+  taskId,
+  logger = 'resolve-extract-refresh-task-target',
+  tasks,
+}: {
+  restApi: RestApi;
+  siteId: string;
+  taskId: string;
+  logger?: string;
+  /** Pre-fetched task list, if the caller already has it (avoids a re-list). */
+  tasks?: ReadonlyArray<ExtractRefreshTask>;
+}): Promise<MutationTarget> {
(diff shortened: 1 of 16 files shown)


src/tools/web/_lib/resolveExtractRefreshTaskTarget.ts
```diff
@@ -0,0 +1,96 @@
...
+  siteId: string;
+  taskId: string;
+  logger?: string;
+  /** Pre-fetched task list, if the caller already has it (avoids a re-list). */
+  tasks?: ReadonlyArray<ExtractRefreshTask>;
+}): Promise<MutationTarget> {
+  const idOnly: MutationTarget = { id: taskId, kind: 'extract-refresh-task' };
+
+  try {
+    const taskList = tasks ?? (await restApi.tasksMethods.listExtractRefreshTasks({ siteId }));
```
Akash-Rastogi, line 49: [Akash's agent] **NIT (informational, non-blocking):** This `listExtractRefreshTasks` call requires `tableau:tasks:read`, and the downstream `queryDatasource`/`getWorkbook` need `tableau:content:read`. On OAuth deployments, only `delete-content` has both scopes in its scope map — the other three consumers (`confirm-delete-content`, `update-cloud-extract-refresh-task`, `confirm-update-cloud-extract-refresh-task`) lack `tasks:read` (and the update tools also lack `content:read`), so enrichment will silently degrade to id-only on those paths under OAuth.

Not a bug — the catch-and-degrade contract is correct and documented — but if the intent is for enrichment to fire on OAuth too, those scope maps need the additional scopes. Could be a follow-up.

Akash-Rastogi (approved):
[Akash's agent] LGTM. Clean extraction — enrichment helper is well-structured with proper best-effort degradation, HITL text centralization is behavior-preserving, and the durability contract is honestly scoped. One informational NIT on OAuth scope coverage below (non-blocking since degradation is by design).

src/tools/web/_lib/resolveExtractRefreshTaskTarget.ts
Alon-ST-DATA: Good catch on the facts — confirmed the three non-`delete-content` paths do degrade to id-only enrichment under OAuth (and direct-trust). Deferring rather than adding the scopes here, for two reasons:

1. `delete-content` carries `tasks:read` + `content:read` for its **own** core operations, not for enrichment — its extract-refresh-task branch runs `listExtractRefreshTasks` as an existence check (`deleteContent.ts:498`) and its workbook/datasource branches route through `resourceAccessChecker` (`content:read`). So it isn't evidence that full-enrichment-under-OAuth was the intent; delete-content would need both scopes even without enrichment.

2. A one-line scope-map widen isn't a safe, isolated change. `requiredApiScopes` is dual-purpose:
   - It feeds the OAuth hard gate (`authMiddleware.ts:114-119`): under `enforceScopes && advertiseApiScopes`, a token missing any required scope gets 403 **before the tool runs**. Adding `content:read` to the update tools would make it *mandatory* to change a schedule — a least-privilege regression (you shouldn't need `content:read` to update a refresh schedule).
   - It's passed straight into the JWT mint for direct-trust/uat (`buildAuthConfig.ts:43-67`). Per the documented footgun in `getFlow.ts:137-148`, a Connected App rejects a JWT requesting an un-granted scope — so widening the map would make the **entire update tool fail at sign-in** (including the core `tasks:write` op) on any deployment whose Connected App doesn't grant the enrichment scopes. That turns today's "works, enrichment degrades gracefully" into "completely broken."

Today's behavior is safe: the resolver's try/catch (`resolveExtractRefreshTaskTarget.ts:86-95`) catches the over-scope REST calls and degrades to id-only exactly as documented. If full OAuth/direct-trust enrichment is genuinely wanted, it needs the get-flow-style treatment (optional additive scopes, never a mandatory gate) rather than a map edit — tracking that as a separate follow-up rather than riding it in this PR. Thanks!

PR #534, opened 2026-07-16
feat(desktop): route-table gates for authoring skill + plan-before-build

tableaukyler:
## What

Adds two prose entries to `DESKTOP_ROUTE_TABLE` so the generated desktop instructions steer the agent before it starts building:

- **authoring-skill**: load the `tableau-desktop-authoring` skill before building/editing; on repeated unresolved failures, load the debugging skill instead of brute-forcing manual XML.
- **plan-before-build**: for any multi-viz/dashboard ask, plan first — map requirement -> encoding -> rule and classify each as **MAGNITUDE** (continuous quantity) vs **MEMBERSHIP** (which discrete group). Encode membership with a discrete bucketing dimension, never a raw-measure color gradient.

Tests pin the generated instruction text.

## Test
`npx vitest run src/desktop/routeTable.test.ts src/server.desktop.test.ts` — 41 passing.

### DIFF
--- a/src/desktop/routeTable.test.ts
+++ b/src/desktop/routeTable.test.ts
@@ -40,6 +40,11 @@ describe('DESKTOP_ROUTE_TABLE', () => {
     expect(rendered).toContain('tableau-desktop-authoring');
   });
 
+  it('names the debug skill by its exact slug so recovery does not rely on description-matching', () => {
+    const rendered = generateDesktopInstructions(DESKTOP_ROUTE_TABLE);
+    expect(rendered).toContain('tableau-agent-debug');
+  });
+
   it('states a plan-before-build gate with the MAGNITUDE/MEMBERSHIP classification', () => {
     const rendered = generateDesktopInstructions(DESKTOP_ROUTE_TABLE);
     expect(rendered).toContain('MAGNITUDE');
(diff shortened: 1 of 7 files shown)


mattcfilbert:
**[MattGPT, reviewing on Matt's behalf — cc @tabstudiodevs]** Reviewed as a pair with tab-agent-south MR!41 (the MR that actually bundles these skills). Prose gates read well and the scoping is right: "for any multi-viz or dashboard ask" keeps plan-before-build off the plain single-chart fast-path, so no latency regression on the simple binds. The `tableau-desktop-authoring` reference matches the slug MR!41 seeds, exactly. 👍

Two things worth tightening, both about the pair fitting together:

1. **Deploy ordering.** This routing tells the agent to load skills that only exist once MR!41 is deployed. If #534 ships to Desktop before !41 lands, those routes reference skills that aren't installed yet — the agent degrades to unguided building (non-fatal, but noisy). Worth sequencing !41 first, or softening to "if available, load…".
2. **Debug-skill reference is a prose phrase, not the slug.** The authoring line uses the exact slug (`tableau-desktop-authoring`); the recovery line says "load the debugging skill" but !41 registers it as `tableau-agent-debug`. Description-matching will probably resolve it, but the recovery path is exactly where you don't want the agent guessing — recommend the exact slug for determinism.

PR #535, opened 2026-07-17
feat(fields): add verbosity=slim to list-available-fields

nawartamawi:
Adds an optional `verbosity: 'slim' | 'full'` param to `list-available-fields`.

- **`full` (default)** — unchanged: human-readable table + full per-field metadata incl. `column_ref`.
- **`slim`** — no table; fields grouped by datasource:
  ```
  { count, datasources: [{ datasource, contentUrl?, fields: [{ caption, role, datatype }] }] }
  ```
  Each field is just `{ caption, role, datatype }`. The datasource name is carried once per group (not repeated per field), and `contentUrl` is included per **published** datasource (the input `resolve-datasource-luid` needs; omitted for embedded ones). A single-datasource workbook is one group, so callers always parse one shape.

**Why:** the default response returns the field catalog twice (table + full metadata) — ~68KB for a wide datasource (~155 fields), which overflows the inline-output cap and gets truncated. `slim` stays well under the cap for callers that only need to reason over field names/roles/types. Get the exact `column_ref` via `resolve-field`.

Verbosity param rather than a new tool to avoid the `tools/list` schema budget.

**Grouping (addresses @mattcfilbert's review):** slim originally hoisted `fields[0].datasource` and dropped per-field datasource — misattributing every field past the first in multi-datasource workbooks and erasing the disambiguator for same-caption fields. Slim now always groups by datasource, which is correct for 2+ datasources and keeps the payload small (datasource string once per group, not per field). Added multi-datasource + duplicate-caption + embedded (no `contentUrl`) tests; `full` path unchanged.

### DIFF
--- a/src/desktop/metadata/field-builder.ts
+++ b/src/desktop/metadata/field-builder.ts
@@ -280,6 +280,11 @@ export function listAvailableFields(
     const datasourceName = datasource['@_name'];
     if (!datasourceName || datasourceName === 'Parameters') continue;
 
+    // A PUBLISHED datasource carries a repository-location whose id is its
+    // contentUrl (the input resolve-datasource-luid needs). Embedded/local
+    // datasources have no repository-location, so this stays undefined.
+    const contentUrl = datasource['repository-location']?.['@_id'];
+
     // Map to track all columns by name (to deduplicate)
     const columnMap = new Map<
       string,
@@ -444,6 +449,7 @@ export function listAvailableFields(
 
       const fieldRef: FieldReference = {
         datasource: datasourceName,
+        contentUrl: contentUrl,
         columnName: columnName,
         columnInstanceName: constructedInstance,
         derivation: defaultAgg,
--- a/src/desktop/metadata/types.ts
+++ b/src/desktop/metadata/types.ts
@@ -30,6 +30,10 @@ export interface FieldInfo {
 
 export interface FieldReference {
   datasource: string;
+  // Published datasource's contentUrl (from the workbook's repository-location);
+  // the input `resolve-datasource-luid` needs to get the server LUID. Undefined
+  // for embedded/local datasources, which have no server copy.
+  contentUrl?: string;
   columnName: string;
   columnInstanceName: string;
   derivation: AggregationType;
--- a/src/tools/desktop/fields/listAvailableFields.test.ts
+++ b/src/tools/desktop/fields/listAvailableFields.test.ts
@@ -77,14 +77,22 @@ describe('listAvailableFieldsTool', () => {
     vi.clearAllMocks();
   });
 
-  it('should create a tool instance with correct properties', () => {
+  it('should create a tool instance with correct properties', async () => {
     const tool = getListAvailableFieldsTool(new DesktopMcpServer());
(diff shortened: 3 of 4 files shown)


mattcfilbert:
Review — posted by **MattGPT** (automation on Matt Filbert's behalf, approved by Matt).

**P2 — `verbosity=slim` misattributes fields in multi-datasource workbooks** (`src/tools/desktop/fields/listAvailableFields.ts`, slim projection)

`listAvailableFields()` (`src/desktop/metadata/field-builder.ts:265`) iterates **all** datasources (skipping only Parameters) and returns one flat array where each field carries its own `datasource`. The slim path hoists `fields[0].datasource` to the top level and drops per-field `datasource` — so in a workbook with 2+ datasources, every field from the second datasource onward is silently attributed to the first. That's not just lossy; for same-caption fields across datasources it's actively wrong, and it removes the only disambiguator a client has.

Full mode is unaffected: its header does the same hoist for display but still returns the complete per-field array — so the earlier verification of the hoist as "not lossy" was correct for that path; this only bites slim.

Suggested fix: keep `datasource` per-field in slim whenever the workbook has more than one datasource (or group slim output by datasource), and add a multi-datasource / duplicate-caption test — current slim tests only cover the single-datasource case.

mattcfilbert:
Two things before the next round, both verified against current `feature/desktop` (`ca668602`):

**1. Rebase needed — the branch predates the live-session contract.** Current `feature/desktop` makes `workbookFile` optional ("Omit workbookFile to read the live session workbook") so the lean/default profiles can call this tool with only a `session` while the XML/cache tools are hidden; this branch's head still has `workbookFile: z.string()` required and a cache-path callback. Please rebase and keep the live-session path working for both `verbosity=full` and `slim` — and add a slim test for `session` without `workbookFile`.

**2. Run the byte-ratchet tests before merge.** `src/server.desktop.test.ts` gates the 46k full-surface cliff and a 1,200-byte per-tool budget (exceptions are enumerated explicitly, e.g. bind-template). The new `verbosity` enum + description lands on a tool that's in the default profiles; if it trips the ratchet, trim the param text rather than raising a cap.

The multi-datasource slim grouping itself looks right, and dropping `column_ref` from slim is fine given `resolve-field` exists for the follow-up. Nit (no action needed to merge): the `contentUrl` extraction is only exercised through mocks — a real published-datasource XML fixture at the metadata level would make it trustworthy.

*(Posted by MattGPT, Matt's review automation; Matt reviewed the findings.)*

tonythebest:
Non-blocking follow-up: the slim projection tests cover mocked `contentUrl` values, but not extraction from real workbook XML. It would be useful to add a field-builder fixture with one published datasource (`repository-location/@_id`) and one embedded datasource, including an empty-ID case. I don't think this should block the PR.

tonythebest (approved):
The multi-datasource issue is fixed, the grouped slim contract is sound, and CI is green. Approved; please rebase on `feature/desktop` and rerun CI before merge.

mattcfilbert (approved):
approved!

nawartamawi:
good points @tonythebest both concerns addressed.

mattcfilbert (approved):
lgtm

PR #553, opened 2026-07-20
@W-23093467 @W-22551424: enforce stale-content row cap server-side + harden Admin Insights LUID cache

Alon-ST-DATA:
## Summary

Two colocated Admin Insights follow-up work items shipped together (both touch `src/tools/web/adminInsights/`, share test files, both small):

- **@W-23093467 — Enforce the stale-content row cap server-side.** Follow-up from PR #405. Today the `LARGE_REPORT_THRESHOLD = 100` cap that protects the destructive stale-content-cleanup **apply** flow lives ONLY in prompt text (`src/prompts/staleContent/apply.ts`), so the safety bound is only as strong as model compliance — the live e2e report surfaced **3376** stale items. This moves the guarantee server-side so a model that ignores the prompt still cannot receive a huge actionable row set.
- **@W-22551424 — Light hardening of the existing Admin Insights LUID resolver cache.** The resolver + `(siteId, dataset)→LUID` `ExpiringMap` cache already exists (`resolver.ts`); this is the remaining hardening the in-code comment tracks (size bound, eviction, telemetry) — not a new build. (The ticket's named `-ts-events` / `-site-content` tools were long ago consolidated into the polymorphic `query-admin-insights`.)

## Part A — server-side stale-row cap (refuse + count)

- New configurable **`STALE_CONTENT_MAX_ROWS`** (default `100`, range `1`–`10000`), mirroring the existing `STALE_CONTENT_MIN_AGE_DAYS` across all wiring sites: allow-list, constants, class field, constructor, `getStaleContentMaxRowsWithOverrides` resolver (env→site→request precedence), `process-env.d.ts`, and the MCPB manifest.
- Enforced in the `kind: "stale-content"` branch of `query-admin-insights`: when `rows.length > staleContentMaxRows`, the tool returns a **structured success** (not a thrown error, so the read-only inform prompt keeps working) with:
  - `rows: []` — the row payload is withheld,
  - `totalStaleItems` / `totalStaleSizeBytes` — the **true** pre-cap totals, so callers can gauge magnitude,
  - a new `ROW_CAP_EXCEEDED` warning (severity `ERROR`) in `mcp.warnings` guiding the caller to narrow scope (`projectIds` / higher `minAgeDays`).
  - Boundary is strictly `>`, so a report exactly at the cap still returns full rows.
- The prompt-text guard in `staleContent/apply.ts` is **kept as defense-in-depth**; its comment now notes the cap is enforced server-side.

## Part B — LUID resolver cache hardening

- `ExpiringMap` gains an optional `maxSize` ctor param (**default unbounded**, so the `adminGate` and project-name caches are untouched). On overflow it evicts the oldest-inserted key, clearing that key's pending timeout via the existing `delete`. Throws on `maxSize <= 0`.
- The resolver cache is bounded at a plain `256`-entry constant (no new env var / doc surface) and gains hit / miss / resolve-count debug telemetry (no `siteId` / PII).
- Invalidation remains TTL + `siteId`-keying; an explicit refresh hook is intentionally deferred.

## Docs

- `env-vars.md` — new `## STALE_CONTENT_MAX_ROWS` section.
- `request-overrides.md` + `site-settings.md` — the overridable-config trio.
- `tools/admin-insights/query-admin-insights.md` — cap behavior + note that `totalStaleItems` stays true while `rows` is withheld.

## Testing

- `npx tsc --noEmit`, `npm run lint` — clean.
- Full unit suite green: **2361 passed**. New/updated cases: below/at/above-cap + combined-warning (`queryAdminInsights.test.ts`), config precedence/validation (`overridableConfig.test.ts`), `ExpiringMap` eviction (`expiringMap.test.ts`), resolver N+1 eviction (`resolver.test.ts`).
- No new e2e test: the cap / config / eviction paths are deterministic and fully covered at the unit tier; existing e2e prompt tests remain green.

## Version

`3.3.0` → `3.4.0` (minor — both items are features; CI `version-bump-check` requires a bump for `src/**` changes).


Alon-ST-DATA:
## Post-review update — fixed a self-contradicting prompt on the over-cap path

An agent review (tmcp-reviewer) + coverage pass (tmcp-tester) ran against the rebased diff. Coverage came back green (2361 pass); the review found one real MAJOR that undermined this PR's own goal, now fixed.

### MAJOR (fixed) — over-cap path reported "No stale items found"
On the withheld-rows path the tool returns `rows: []` with a large `totalStaleItems` (e.g. 3376) + a `ROW_CAP_EXCEEDED` warning. Both stale-content prompts then hit their `rows` empty → "No stale items found above the threshold." → stop rule, so a prompt-following model printed `total = 3376 items` immediately followed by "No stale items found" — a contradiction that buried thousands of stale items. That's exactly the scenario the cap exists to surface, so the code comment claiming the inform prompt "keeps working" was wrong as written.

**Fix:** both `inform.ts` and `apply.ts` now branch on a `ROW_CAP_EXCEEDED` warning **before** the empty-rows rule — reporting the true totals + narrow-scope guidance (`projectIds` / higher `minAgeDays`) instead of a false "No stale items found". On the destructive `apply` prompt this also stops before any tag/delete. Updated the misleading comment at `queryAdminInsights.ts`.

### Tests
Added a regression case to each prompt test asserting the `ROW_CAP_EXCEEDED` branch appears before the empty-rows rule. Full suite now **2363 pass**; `tsc` + `eslint` clean.

### Everything else verified clean
Structured-success-not-throw contract, strictly-`>` at-cap boundary, `ExpiringMap` default-unbounded (adminGate + project-name caches untouched), timeout-cleared-on-eviction, telemetry logs only `datasetName`/counts (no siteId/LUID/PII), doc↔schema parity across all wiring sites, version bump 3.3.0 → 3.4.0.

Force-pushed the amend (`8f4f6b11` → `377e0b40`).

Akash-Rastogi (approved):
[Akash's agent] LGTM. Row cap logic is clean — strict `>` boundary, structured success (not error) preserving true totals, prompt ordering correctly branches on `ROW_CAP_EXCEEDED` before the empty-rows rule. ExpiringMap eviction is well-bounded — only fires on genuinely new keys, correctly clears evicted timeouts, and the resolver test proves correctness after eviction-induced miss. Config wiring is complete across all layers.

PR #560, opened 2026-07-21
feat(desktop): waterfall anchor_category slot + sort proposal wins over template sort

mattcfilbert:
**Live eval receipts (m1 waterfall song, judge FAIL 34/100, 2026-07-20)** — two independent product gaps, both fixed here:

1. **Sort proposals were inert on the waterfall.** The template hard-codes `computed-sort DESC by SUM(Profit)`, and a requested sort field absent from `datasource-dependencies` had nothing to bind to — the singer's override "hit a template limitation", then burned 3× `tabdoc:sort-nested` 500s + a document round-trip to force P&L order. Now: a sort proposal **replaces** the built-in computed-sort, declaring the sort column/column-instance when missing. Every failure path is fail-open with a `warnings[]` channel on the bind result — the bind never blocks, the template default stands.
2. **Subtotal/total rows double-counted.** New optional **virtual slot `anchor_category`**: bind the row-type column and the splice adds a categorical EXCLUDE filter (`subtotal`/`total`) so anchor rows leave the bridge and the running total counts each increment once. Unbound binds are byte-identical (characterization-pinned).

**Byte discipline:** teaching lives in the manifest slot description (model-facing at bind time), not tools/list. Per-tool ratchets **tightened** (bind-template 1908→1886, validate-proposal 1555→1533); combined-lean at 45,952 of the 46,000 cliff.

Also: served-corpus knowledge twins (waterfall subtotal/total law; zero-denominator ratio-sort guard) — source-corpus siblings landed in a2td #239; the interim error-path teaching for the sort-nested 500s is #559.

**Validated firsthand:** full suite 4608 green (316 files), lint clean, all three bind invariant suites unchanged. v2.27.0. Live re-sing of m1 against this build is the next step and will be reported on the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #561, opened 2026-07-21
feat(desktop): waterfall bind hint names the step-order column

mattcfilbert:
Follow-up to #560 (merged). Live m1 waterfall receipts (N=3: 83 PASS / 38 FAIL / 82 PASS) showed the fragility all traced to one seam: the singer binds the waterfall plain, then tries `refine-worksheet` to sort by `display_order` — which refuses (`unknown sort-by field`) because refine can only sort by a field already on the view. Run 2 gave up (magnitude order → wrong bridge); run 3 forced it with two rounds of `load-underlying-metadata` XML surgery (339s, $2.42).

The bind path ALREADY supports `proposal.sort:{by:"display_order"}` (`ensureSortByColumnDependency` injects the off-view column; tested). The gap was pure discoverability — the generic `WATERFALL_SORT_HINT` ("override with proposal.sort:{by:<field>}") was too vague to connect to the sequence column.

**This change:** when the schema carries an explicit sequence/order column (`display_order`, `sort_order`, `ordinal`, … — matched precisely; `line_item` and `order_count` do NOT false-positive), the waterfall discovery hint now NAMES it and prescribes the exact re-call (`proposal.sort:{by:"display_order",direction:"asc"}`), and says plainly not to use refine. Mirrors the proven `anchor_category` hint. Falls back to the generic hint when no order column exists. This is the deterministic fix for the m1 give-up/XML-surgery seam.

Gates: tsc 0, full vitest 4624 passed / 1 skipped, lint 0.


mattcfilbert:
Superseded by #562 — this branch's base predated #560's squash-merge, causing a duplicate-commit conflict. #562 is #560's merged tip + only the order-hint commit, clean.

PR #565, opened 2026-07-21
feat(desktop): temporal_axis_from_string — bind a string month to a continuous date axis

mattcfilbert:
## What

An opt-in binder capability: a time-series template's temporal slot can accept a **date-like STRING** field (e.g. a `"2025-08"` month) and render a continuous month axis, instead of rejecting it with `kind-mismatch`. On bind, the apply path injects a `DATEPARSE` calc so the template's Month-Trunc axis truncates a real parsed date.

## Why

The judge's `suggested_fix` for the live **e4** eval (MAU-over-time). e4's dataset carries a **string** `month` + two products. The `trend-line-chart` `order_date` slot requires `date`/`datetime`, so the bind escalated and the singer fell back to value-sorted bars (summing both products) or dead-ended asking the user for the month format. `temporal_axis_from_string` closes that: one confident bind renders the line.

## Pieces

- **`stringTemporal.ts`** — `inferStringTemporal`: accepts a STRING dimension whose *name* is date-like (`month`/`date`/`period`/…) and infers a `DATEPARSE` format (`yyyy-MM` monthly, else `yyyy-MM-dd`). Fail-closed: a non-date-like name stays a `kind-mismatch`.
- **`validate.ts` gate 3** — the temporal `kind-mismatch` is waived only for a slot that sets `temporal_from_string` and a string field the inferrer accepts; that slot's `field_mapping` key is then skipped (the splice owns its XML) and a `DateparseAxisSpec` rides out on the result.
- **`dateparseTemporalAxis.ts`** — apply-side splice (modeled on `waterfallAnchorFilter`): converts the template's date base column into a `DATEPARSE` calc + declares the bound source column, leaving the Month-Trunc CI and every shelf/format ref **byte-unchanged** (no repoint → no duplicate-CI risk). Identity no-op when no axis requested; fail-closed otherwise.
- **`trend-line-chart` manifest** — `order_date` opts in; index regenerated.

## Validation

- 9 splice tests + 6 inferrer tests + 2 **end-to-end** (real trend-line template → coherent `DATEPARSE` calc + intact namespaced Month-Trunc axis; real-date path byte-identical).
- Full binder/templates/server suites green (**858**); byte-cliff unchanged (no tool-surface bytes added); `tsc` + `eslint` clean.
- Version 2.27.0 → **2.30.0**.

## ⚠️ Live-render gate (reviewer please note)

The `DATEPARSE` format is **inferred from the field name** — the binder sees schema, not cell values, and a wrong format renders a **blank axis** (DATEPARSE returns NULL silently). XML validity is proven; **value parsing is not yet live-verified**. A live render against e4's real `"2025-08"` data is running; I'll post the result. Do not merge on the strength of unit tests alone — this needs the render receipt.


mattcfilbert:
## Live-render result — feature correct, but NOT yet exercised (upstream discoverability gap)

Ran the e4 sing live against this bundle (sonnet singer, opus judge). **The `temporal_axis_from_string` path never executed** — and the reason is instructive, so recording it honestly rather than claiming a pass:

**Receipt (tas-eval-20260721T133547Z):**
- `trend-line-chart` appears **0 times** in the whole transcript — it was never offered as a Call-1 `propose` candidate for "Show me monthly active users over the last 12 months".
- The singer guessed a **nonexistent** template name `"line"` (slots `columns`/`rows`, not the real `trend-line-chart` / `order_date`). Those binds returned `"Use the general worksheet build tools"` (template-not-resolved prefill) — so the temporal gate was never reached.
- It then fell back to `magnitude-simple-bar` (bars, both products silently summed) → judge FAIL 38.

**Conclusion:** this PR is **necessary but not sufficient** for e4. The binder/apply code is correct (17 unit + 2 end-to-end tests prove the DATEPARSE calc + intact Month-Trunc axis), but the singer never proposes `trend-line-chart` with `order_date`, so the string-month source never reaches the opted-in slot. The blocking gap is **upstream discoverability**: `trend-line-chart` isn't surfacing as a candidate for a "MAU over time / monthly ... over the last 12 months" ask (its `intent_keywords` don't match that phrasing strongly), and there's no `"line"`→`trend-line-chart` alias.

**Merge posture:** the code is safe (opt-in, fail-closed, byte-cliff unchanged, real-date path byte-identical) and correct, so it can land as the apply-side half. But e4 won't pass until the discoverability fix lands too (surface `trend-line-chart` for the MAU phrasing / alias `line`). Filing that as the paired follow-up. Not claiming e4 fixed.

mattcfilbert:
## ✅ LIVE-RENDER PROOF — both halves fired end-to-end

Re-sang e4 with the bundle carrying BOTH the discoverability keywords and the temporal capability (sonnet singer, opus judge). Receipt `tas-eval-20260721T134609Z`:

**The judge's `line_over_continuous_time` gate flipped to YES:**
> *"Yes — real temporal axis, string month was converted to a date"* — evidence: *"list-available-fields grew 6→7 fields adding 'month (parsed date)'; agent states 'plotting it on a time axis required creating a derived date field (month (parsed date))'."*

**Receipt signals (vs the prior run where these were all 0/absent):**
- `trend-line-chart`: **4 mentions** (was 0 — the singer now reaches the template; discoverability fix landed).
- `DATEPARSE`: **1**, `month (parsed date)`: **4** — the temporal_axis_from_string calc injected AND rendered. `DATEPARSE('yyyy-MM', [month])` parsed the real `"2025-08"` values correctly (the format inference was right for this data).
- Route `BIND→BIND→BIND→BIND→REFINE→REFINE`, **112.5s / $0.29** — vs the prior fallbacks' 233s/$0.87 and 415s/$1.92. The failed-bind/search/notional-spec tail is gone.

So this PR (capability + discoverability) is **live-proven**: the string→date wall is down.

**e4 still FAILs (34), but on two DIFFERENT gaps** the judge now names — NOT this one:
1. `multi_product_handling`: the two products are summed into one line → needs a `series_split` (color/small-multiple by product) slot.
2. `time_window`: no "last 12 months" bind-time filter → the singer hunted for a trim slot that isn't offered.

Those are separate follow-up features. This PR does exactly what it claims — bind a string month to a continuous date axis — and the render receipt proves it. Safe to review/merge on that scope; e4-green needs the two follow-ups too.

PR #566, opened 2026-07-21
Fix misleading 'Pulse not available on Tableau Server' error on Cloud

dplatt84:
The 404 handler in guardAgainstPulseDisabled treated all 404 responses as "Pulse is not available on Tableau Server". On Tableau Cloud, a 404 can instead indicate a real Pulse API error (e.g., definition not found, datasource not found, metric not found) which has a Pulse error code in the response body.

Now checks whether the 404 response contains an error code. If it does, it falls through to the PulseInsightsApiError handler which surfaces the actual error with actionable guidance. Only bare 404s (no error code) are treated as "Pulse not available".

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.1.0",
+  "version": "4.1.1",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.1.0",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "3.6.1",
+  "version": "3.6.2",
```
stephendeoca: Our package.json on main has been updated past this version can you update this?

dplatt84: @stephendeoca This should be updated

stephendeoca (approved):

PR #568, opened 2026-07-21
fix(desktop): honor session pin when stamping cache sidecars

tableaukyler:
## What

Four cache-mutation desktop tools stamped the fingerprint sidecar with the raw agent-supplied `session` arg instead of the resolved session pin:

- `write-cached-xml`
- `add-field`
- `remove-field`
- `inject-template`

When `TABLEAU_DESKTOP_SESSION_ID` is set, a mismatched (or absent) `session` arg produced a sidecar fingerprinted to the wrong instance. Downstream `apply-*` tools then either failed closed with `CacheSessionMismatchError` or silently bled cache across instances - which is the "agent connected to a different workbook in a different instance" bug.

## Fix

Each tool now calls `resolveSession(session)` at the top of its callback:
- rejects a conflicting explicit `session` when a pin is set,
- stamps the sidecar with the resolved pin, matching the connect-side invariant everywhere else in the desktop server.

## Flow

```mermaid
flowchart LR
  A[agent session arg] --> B[resolveSession]
  P[TABLEAU_DESKTOP_SESSION_ID] --> B
  B -->|pin wins / conflict rejected| C[writeSidecar resolvedSession]
  C --> D[apply-* verifies sidecar - matches pin]
```

## Tests

Each tool gets two new cases: sidecar stamped with the pin (not the requested session), and reject-without-write on genuine conflict. Full desktop suite green, lint + typecheck clean.

### DIFF
--- a/src/tools/desktop/cache/writeCachedXml.test.ts
+++ b/src/tools/desktop/cache/writeCachedXml.test.ts
@@ -1,6 +1,7 @@
 import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 import { resolve } from 'path';
 
+import * as configModule from '../../../config.desktop.js';
 import { DesktopMcpServer } from '../../../server.desktop.js';
 import invariant from '../../../utils/invariant.js';
 import { Provider } from '../../../utils/provider.js';
@@ -32,10 +33,19 @@ function setupCacheMock(): void {
   );
 }
 
+function mockPinnedSession(desktopSessionId: string | undefined): void {
+  const base = new configModule.Config();
+  vi.spyOn(configModule, 'getDesktopConfig').mockReturnValue({
+    ...base,
+    desktopSessionId,
+  } as configModule.Config);
+}
+
 describe('writeCachedXmlTool', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     setupCacheMock();
+    mockPinnedSession(undefined);
     vi.mocked(writeFileSync).mockImplementation(() => {});
   });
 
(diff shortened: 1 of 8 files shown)


mattcfilbert:
**Review (posted by MattGPT on Matt's behalf)** — dual-model intent; see the honest caveat at the end.

Traced this against the real files at the PR head + `resolveSession`'s precedence. **The implementation is sound and complete** — happy to see it land after the one test-strength fix below.

**What's right (verified):**
- All 4 tools resolve once at the top and return `sessionResult.error.toErr()` on failure (write-cached-xml:56, add-field:75, remove-field:73, inject-template:98) — no `.unwrap()`/throw, no swallow.
- No split-brain: these tools use `session` only for the sidecar fingerprint (they write local cache files, not a session-scoped executor), and every `writeSidecar` now takes `resolvedSession`. The raw arg no longer reaches the stamp.
- **No missed siblings** — grepped every `writeSidecar` call site on the branch (get-worksheet/get-workbook/get-dashboard, batch-create, refresh-workbook-cache); all already pass `resolvedSession`. This PR closes the last 4 raw-session stamps, so the sidecar-fingerprint bug class is fully closed.
- The **conflict tests are strong**: pin `'99999'` + requested default `'12345'` → asserts `writeSidecar` NOT called and no `writeFileSync` (reject-without-write confirmed).

**P2 — the three "stamped with the pin" tests can't actually fail if the bug regressed.** In `writeCachedXml.test.ts`, `addField.test.ts`, `removeField.test.ts`, the pin-wins case does `mockPinnedSession(SESSION)` and asserts `writeSidecar(..., SESSION)` — but the test helper defaults the requested `session` to `SESSION` too (`session = SESSION`, e.g. writeCachedXml.test.ts:284), and `SESSION === '12345' === the pin`. So the assertion holds whether the code stamps the *resolved pin* or the *raw requested session* — it doesn't isolate the fix it's guarding. **Fix:** make the pin and the requested session differ in that one case (pin `'12345'`, request `'99999'` or `undefined`, assert the stamp is the pin) so the test would fail if a future change reverted to stamping the raw arg. The conflict test already proves the reject path; this just hardens the positive path.

---
_Caveat, in the spirit of honest tooling: this was meant as a two-vendor (GPT-5.5 + Opus) review. The GPT leg completed (its quoted lines verified 9/9 against the diff) and the Opus leg stalled mid-run — so the second perspective here is Matt's own line-by-line trace, not an independent model. Not claiming a clean dual-model pass._

mattcfilbert (approved):
lgtm

PR #571, opened 2026-07-21
feat(web): slim param on generate-pulse-metric-value-insight-bundle

nawartamawi:
Adds an optional **`slim`** boolean to `generate-pulse-metric-value-insight-bundle`.

When `slim: true`, return the bundle **without the `viz` (Vega chart-spec) blobs** — a text/card UI renders from `facts`/`markup` alone, so this strips the heaviest part of the payload. Defaults to `false` (response returned verbatim, including `viz`), so existing callers are unaffected.

**Why:** dependency for the tab-agent-south-ui viz-canvas insight bundles (W-23447550) — without `slim` the Vega blobs bloat the headless agent's context.

**Testing:** 17/17 tests green · `tsc --noEmit` + `eslint` clean.

### DIFF
--- a/src/sdks/tableau/types/pulse.ts
+++ b/src/sdks/tableau/types/pulse.ts
@@ -480,29 +480,6 @@ export const pulseInsightBriefResponseSchema = z.object({
 export type PulseBundleResponse = z.infer<typeof pulseBundleResponseSchema>;
 export type PulseInsightBriefResponse = z.infer<typeof pulseInsightBriefResponseSchema>;
 
-/**
- * The per-bundle choices from a `bundleRequest`, surfaced so a card UI can read the metric's
- * identity directly instead of parsing it out of markup: curated flat fields (metric name,
- * measure, time dimension, breakdown dimensions). The request `input` is deliberately NOT echoed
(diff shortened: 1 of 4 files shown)


mattcfilbert (commented):
Reviewed (dual-review lens over the full diff). **Verdict: mergeable with one nit** — the design is right for the extended-bundle-card path: default is backward-compatible (slim absent/false returns the response verbatim, pinned by a test), both shapes are tested, and the single-seam slimming comment is exactly where future field-trims should land.

One wording nit (P3):
- The `slim` param doc says it returns *"only the `facts`/`markup` fields"*, but `slimBundle` strips only the `viz` (Vega spec) blobs and preserves `type`, `version`, `question`, `score`, `id`, `generation_id`, etc. Suggest rewording to "strips the large `viz` blobs" so a consumer doesn't read slim=true as a minimal facts-only contract.

No P0/P1/P2 found. This composes cleanly with the Chiron basic-flow work — slim mode is the right shape for what the chat card carries.

---
*Posted by MattGPT (automation on Matt Filbert's behalf; review adjudicated by him).*

PR #572, opened 2026-07-21
feat: env-gated External-API command-registry guard (wire names, enums, dialog refusals)

mattcfilbert:
Consumes the internal command/param registries at runtime via `EXTERNAL_API_REGISTRY_DIR` (tab-agent-south !66 ships the data + sets the env; nothing internal lands in this repo — synthetic test fixtures only).

- **Wire-name rewrite**: param keys resolve local/CamelToDashed → the registry's registered wire name (`ShowMeType` → `show-me-command-type`). Kills the bare-500 class (unknown key throws past Desktop's 400 handler).
- **Enum validation**: values checked against the serialized enum lists; refusals teach the legal values in-band.
- **Refusals**: registry-known `agent_can_invoke:false` / `opens_blocking_dialog:true` commands; plus the eight live-observed dialog-poppers refused **unconditionally** (they hang the UI thread headlessly and pass every static flag).
- **Required-param check** with context-filled (`UPI_Workspace`/`UPI_IWorkspace`) awareness.
- Env unset / unreadable → identical behavior to today (regression-asserted).

Validated: tsc clean, full suite 4869 passed / 1 skipped (verified firsthand post-edit).

### DIFF
--- a/src/desktop/externalApi/commandRegistry.ts
+++ b/src/desktop/externalApi/commandRegistry.ts
@@ -1,0 +1,247 @@
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+import { log } from '../../logging/logger.js';
+
+const REGISTRY_DIR_ENV = 'EXTERNAL_API_REGISTRY_DIR';
+const COMMAND_PARAM_REGISTRY_FILE = 'command_param_registry.json';
+const CODEGEN_REGISTRY_FILE = 'codegen_registry.json';
+
+export type ExternalApiRegistryParam = {
+  local: string;
+  type: string;
+  required: boolean;
+  wire: string;
+  camelToDashed: string;
+};
+
+export type ExternalApiCommandRegistryEntry = {
+  invocable: boolean;
+  blockingDialog: boolean;
+  requiredParams: ExternalApiRegistryParam[];
+  params: ExternalApiRegistryParam[];
+  paramWireByLocal: ReadonlyMap<string, string>;
+  paramWireByCamelToDashed: ReadonlyMap<string, string>;
+  enumValuesForParamType: ReadonlyMap<string, string[]>;
+};
+
+type ParsedRegistry = {
+  commands: Map<string, ExternalApiCommandRegistryEntry>;
+};
+
+type RegistryCache = {
+  dir: string | null;
+  registry: ParsedRegistry | null;
+};
+
+type RawCommandEntry = {
+  agent_can_invoke?: unknown;
+  opens_blocking_dialog?: unknown;
+  in_params?: unknown;
+};
+
+type RawCommandParam = {
+  local?: unknown;
+  type?: unknown;
+  required?: unknown;
+  wire?: unknown;
+};
+
+type RawCodegenRegistry = {
+  type_of_param?: unknown;
+  enum_vals?: unknown;
+};
+
+let cache: RegistryCache | undefined;
+
+export function lookupExternalApiCommandRegistry(
+  namespace: string,
+  command: string,
+): ExternalApiCommandRegistryEntry | null {
+  const registry = loadRegistry();
+  return registry?.commands.get(`${namespace}:${command}`) ?? null;
+}
+
+export function isExternalApiCommandRegistryEnabled(): boolean {
+  return loadRegistry() !== null;
+}
+
+export function _resetExternalApiCommandRegistryForTest(): void {
+  cache = undefined;
+}
+
+function loadRegistry(): ParsedRegistry | null {
+  const dir = normalizedRegistryDir();
+  if (cache?.dir === dir) {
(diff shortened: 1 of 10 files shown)


src/tools/desktop/commands/executeTableauCommand.ts
```diff
@@ -27,12 +32,30 @@ import { DesktopTool } from '../tool.js';
...
 ]);
+// Live-swept dialog poppers misclassified as callable: never let the External API open these headlessly.
+const LIVE_EXTERNAL_API_DIALOG_BLOCKLIST = new Map<string, string>([
+  [
+    'tabui:copy-zone-to-desktop-clipboard',
+    'use XML/readback or image export paths instead of the Desktop clipboard',
+  ],
+  [
+    'tabui:workgroup-change-site',
+    'connect Desktop to the intended site before starting the MCP session',
```
lauren-jacksonSFDC: The guidance here isn't quite right - the user can change sites, that is fine, but changing the site will open a dialog

src/tools/desktop/commands/executeTableauCommand.ts
```diff
@@ -27,12 +32,30 @@ import { DesktopTool } from '../tool.js';
...
     'tabdoc:sort-nested',
     'FIX: tabdoc:sort-nested is known to fail (HTTP 500) on current Desktop builds regardless of parameters — do not retry it. Sort instead via the bind-template sort proposal (preferred for template-bound sheets) or the document round-trip (tabui:save-underlying-metadata → edit the computed-sort → tabui:load-underlying-metadata).',
   ],
 ]);
+// Live-swept dialog poppers misclassified as callable: never let the External API open these headlessly.
+const LIVE_EXTERNAL_API_DIALOG_BLOCKLIST = new Map<string, string>([
+  [
+    'tabui:copy-zone-to-desktop-clipboard',
+    'use XML/readback or image export paths instead of the Desktop clipboard',
+  ],
```
lauren-jacksonSFDC: I think this might have been a false positive. Best to remove for now

src/tools/desktop/commands/executeTableauCommand.ts
mattcfilbert: Fixed — the refusal text now says changing sites is a fine user action and only the headless dialog is the problem, pointing the user at Desktop's site switcher instead. Thanks for the correction.

src/tools/desktop/commands/executeTableauCommand.ts
mattcfilbert: Removed from the blocklist. If a future sweep reproduces the hang we'll re-add it with the receipt; for now it's callable again.

src/desktop/externalApi/commandRegistry.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  if (paramType.startsWith('DPI_')) {
+    const withoutPrefix = stringArray(enumVals[paramType.slice(4)]);
+    if (withoutPrefix.length > 0) {
+      return withoutPrefix;
+    }
+  }
+
+  const enumName = enumNameFrom(typeOfParam[paramType]);
```
lauren-jacksonSFDC, line 201: ✨ _Lauren's code review bot:_ ✨

`enumNameFrom()` cannot read the array-shaped `type_of_param` entries in the real generated `codegen_registry.json`, so enum-value validation is silently skipped for ~16 of 31 enum parameter types.

**Trigger:** In the real registry, `type_of_param["DPI_ChangePageDirection"]` is `["ChangePageType","enum"]` (enum name in slot 0) and `enum_vals` is keyed by that bare name (`ChangePageType` → `["first","next","previous","last"]`). `enumValuesForType` misses on the direct and `DPI_`-prefix-strip lookups, then calls `enumNameFrom(["ChangePageType","enum"])`; `isRecord([])` is true but the function only probes string keys an array lacks, returns `null`, and the enum ends up absent from `enumValuesForParamType`. So `execute-tableau-command tabdoc:change-page-by-direction {"change-to":"garbage"}` (invocable, not blocklisted) passes the guard and is dispatched to the live host as a bad enum value.

**Why this matters:** The guard fails to catch exactly the bad-enum → HTTP 500 / dialog it exists to prevent, for real commands (`change-to`, `legend-type`, `animation-control`, `dashboard-object-type`, delete-sheet orphans mode, …); the test suite stays green only because its fixture uses the object shape `{enum_name:...}` instead of the real `["Name","enum"]` array. Fix: handle the array form in `enumNameFrom` (return the first string element) or read `typeOfParam[paramType][0]`.

mattcfilbert: Fixed in 4aab2fdc (and carried into the policy-table refactor that followed): clipboard entry removed; change-site refusal now says changing sites is fine

PR #573, opened 2026-07-21
feat: data-first tools — get-summary-data + list-site-datasources

mattcfilbert:
First-class lane of the External Client API integration (backdoor guard = #572). The agent looks at the DATA before deciding — `getworkbookxml` is no longer the first play.

- **get-summary-data** — `GET /v0/workbook/worksheets/{id}/summaryData`: worksheet name→id resolution with honest ambiguity errors, maxRows clamp (200/1000), column filter, `shape` summary line. Description teaches the marks-card mechanic (summary data carries only fields ON the view — Detail first, then read).
- **list-site-datasources** — `GET /v0/site/datasources`: the active-sheet → published-LUID bridge for the Pulse insight-card flow.
- Both in the default dynamic-authoring profile. **Live-probed caveat**: installed Desktops (0.1.0) 404 these routes (they land post-monolith-#60211); get-summary-data returns a named `endpoint-not-in-this-build` error with no-retry guidance, regression-tested.
- Byte budget: full-only tool descriptions tightened so the full surface stays under the 46k cliff while the dynamic profile keeps rich tooling.

Pairs with tab-agent-south !65 (DATA FIRST prompt law). Validated: tsc + eslint clean, full suite 4878 passed.


lauren-jacksonSFDC:
Two things this branch should do that it doesn't yet — both measured against the **live** OpenAPI spec at `http://127.0.0.1:51031/openapi.json` (v0.1.1), not any older capture:

**1. Prefer the first-class routes — stop reconstructing data the API serves directly.**

The read tools currently fetch the whole workbook document and slice it client-side, with comments claiming *"External Client API exposes no per-sheet route."* That's false against the spec — the routes exist. Please rewire:

| Tool | Today | Should call |
|---|---|---|
| `list-worksheets` | `GET /v0/workbook/document` → `listSheets()` | `GET /v0/workbook/worksheets` |
| `list-dashboards` | `GET /v0/workbook/document` → `listWorkbookDashboards()` | `GET /v0/workbook/dashboards` |
| `get-worksheet-xml` | `GET /v0/workbook/document` → `extractSheetXml()` | `GET /v0/workbook/worksheets/{id}/document` |
| `get-dashboard-xml` | `GET /v0/workbook/document` → `extractDashboardXml()` | `GET /v0/workbook/dashboards/{id}/document` |

The list routes return `{id, name}`, so add a name→id resolver and drop `extractSheetXml`/`extractDashboardXml` and the "no per-sheet route" comments once the routes are in. Note: a typed `client.listWorksheets()` already hits the real route, but only `get-summary-data` consumes it — the `list-worksheets` tool doesn't. Also please check `validate-workbook-xml`: it validates locally (well-formed check) instead of calling the server's `POST /v0/workbook/document:validate` — use the endpoint.

**2. Every functional route in the spec should be reachable as an MCP tool.** Right now only `get-summary-data` and `list-site-datasources` ride typed routes; everything else funnels through the generic `POST /v0/app:invokeCommand` escape hatch (which isn't even in the spec). Full functional inventory to cover:

- **Site:** `GET /v0/site`, `GET /v0/site/datasources` ✅ (done), `GET /v0/site/workbooks`
- **Workbook:** `GET /v0/workbook` (inventory), `GET /v0/workbook/datasources`, `GET /v0/workbook/document`, `POST /v0/workbook/document`, `POST /v0/workbook/document:validate`
- **Worksheets:** `GET /v0/workbook/worksheets`, `.../{id}`, `.../{id}/document`, `.../{id}/summaryData` ✅ (done)
- **Dashboards:** `GET /v0/workbook/dashboards`, `.../{id}`, `.../{id}/document`
- **Storyboards:** `GET /v0/workbook/storyboards`, `.../{id}`, `.../{id}/document`
- **Misc:** `GET /v0/app`, `GET /v0/health`

**One hard constraint before adding tools:** the combined-lean `tools/list` is at the 46,000-byte deferral cliff with ~zero headroom. Surfacing new tools needs a byte-budget pass first — trim descriptions, don't raise the cap. If some routes are better as transport-only client methods than as full MCP tools, flag which and why rather than blowing the budget.

Sequence: do #1 (route-first the existing reads) before more client-side XML surgery, then #2 (fill in the missing routes).

Also: the whole-workbook `POST /v0/workbook/document` now replaces natively per the spec ("Replaces the open workbook…") — this is confirmed on the Desktop side. So the `replaceWorkbook` + `tabdoc:delete-sheet` prune loop in `loadWorkbookXml.ts` is now redundant and should be removed — a whole-workbook apply collapses to a single `applyWorkbookText`.

lauren-jacksonSFDC:
Great progress on waves 2–4 — the reads are route-first now and most of the spec is covered. To finish goal #1 (**a tool for every route**), here's the exact remaining gap, checked against the live spec (v0.1.1) and the current branch. **7 functional routes still have no MCP tool.**

**Need a tool only (typed client method already exists):**

| Route | Existing client method |
|---|---|
| `GET /v0/health` | `health()` |
| `GET /v0/workbook/worksheets/{id}` (item) | `getWorksheet()` |
| `GET /v0/workbook/storyboards` | `listStoryboards()` |
| `GET /v0/workbook/storyboards/{id}/document` | `getStoryboardDocument()` |

**Need both a client method AND a tool (no transport yet):**

| Route | Missing client method |
|---|---|
| `GET /v0/site` | `getSite()` |
| `GET /v0/workbook/dashboards/{id}` (item) | `getDashboard()` |
| `GET /v0/workbook/storyboards/{id}` (item) | `getStoryboard()` |

So: add 3 client methods + 7 MCP tools. Every functional route in the spec should be individually reachable as a tool — no route left as transport-only or folded into another tool. (`get-workbook-inventory` lists storyboards, but that doesn't cover the dedicated storyboards list/item/document routes.)

Two notes:
- We can skip /v0/
- Excluded (correctly) as non-functional infra: `GET /openapi.json`, `GET /.well-known/oauth-protected-resource`.
- PR description needs updating

PR #578, opened 2026-07-21
refactor: architectural consolidation — guard extraction, read harness, single-source resolution, fallback gateway

mattcfilbert:
Findings 1-4 of today's principal-lens review over the day's merges, as one PR:

1. **ExternalApiCommandGuard** extracted — command safety leaves tool altitude; `execute-tableau-command` is now resolve → guard → execute → shape (result shaping stays tool-side as response policy). Seam tests on refusal ordering + fail-open.
2. **One read harness** — `runExternalApiReadTool()` owns session/executor/route-miss/error for all 15 first-class read tools; three dialects → one.
3. **Single-source name→id** — the older worksheet/dashboard XML commands ride the shared resolver (entity-decode fallback included); duplicated messages deleted.
4. **WorkbookReadGateway** — one fallback decision (first-class | whole-doc slice | Agent API) with the formerly-implicit config/executor-disagreement mode made explicit and logged.

Net +2 lines across the 2-4 legs. Behavior byte-identical except the disagreement log (intended). Validated firsthand on the union: full suite **5005 passed / 1 skipped**, tsc + eslint clean.


mattcfilbert:
Selector-standardization commit added (finding 5 + Lauren's waves-5+6 review finding): `get-worksheet-info` now resolves name-or-id via the shared resolver exactly like its siblings — a worksheet name gets the helpful candidates message instead of an opaque 404 wrap. The audit table confirmed it was the only outlier. Suite on the branch head: 5007 passed / 1 skipped, verified firsthand.

src/desktop/commands/workbook/getDashboardXml.ts
```diff
@@ -1,219 +1,15 @@
...
-  | { type: 'no-dashboard-found' }
-  | { type: 'multiple-dashboards-found' }
-) & { message: string };
-
-type GetDashboardXmlResult = Result<
-  string,
-  | { type: 'execute-command-error'; error: ExecuteCommandError }
-  | { type: 'get-dashboard-xml-error'; error: GetDashboardXmlError }
->;
+export type { GetDashboardXmlError };
```
lauren-jacksonSFDC, line 9: The exports on L8 and L9 are a bit confusing, what exactly are we doing here?

PR #582, opened 2026-07-21
purge: dead hack-build command dialect — data + knowledge (part A)

mattcfilbert:
The nine hack-build-only commands (`tabui:save/load-underlying-metadata`, the per-sheet Save/Load/List family, `ListAllCommands`) were never in real Desktop builds. Part A purges every **data/knowledge** surface: the two reference entries that existed are deleted (335→333, pinned count updated) and 25+ served knowledge files reworded from command-name dialect to capability language. Round-trip guidance survives via the real document endpoints. Part B (source: the load special case, underlyingMetadataGuard, executor aliasing, FIX texts) stacks on #578 after the transport kill lands — the part-B site list is captured.

Validated firsthand: full suite green, zero dead-command references left in data/knowledge.

### DIFF
--- a/resources/desktop/dashboard-xml-guide.md
+++ b/resources/desktop/dashboard-xml-guide.md
@@ -192,7 +192,7 @@ Note: The worksheet name goes in the `name` attribute on the zone itself, NOT in
 
 1. **Always wrap zones in a container**: The root `<zones>` element should contain one main zone at
    (0,0) with full dimensions and `type-v2="layout-basic"`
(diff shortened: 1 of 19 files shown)


mattcfilbert:
Honesty note: the initial push carried 2 failing tests — my validation pipeline masked the suite tail and I committed red. Fixed in the follow-up commit: the underlying-metadata-guard tests now assert the new truth (the dead command is refused as unknown before guard/dispatch runs). Full suite verified green firsthand on the current head: 4995 passed / 1 skipped, tsc + eslint clean.

PR #592, opened 2026-07-22
fix(binder): classify honors temporal_from_string — string month binds line-chart temporal slot (e4)

mattcfilbert:
## e4: monthly-active-users line chart — string month
The concert exposed e4 ("monthly active users over 12 months", where `month` is a **'YYYY-MM' string**) as route-pass but **judge-FAIL 40**, with a **310s / 11-tool thrash** tail.

## Root cause (GPT-5.6-Sol diagnosed, firsthand-verified)
- `trend-line-chart`'s `order_date` slot ALREADY has `temporal_from_string:true`, and `inferStringTemporal()` + the DATEPARSE apply splice already exist (ETUDE #565).
- BUT classify.ts's role-greedy temporal matcher used `isTemporal(f)` = date/datetime datatype ONLY. A string month never filled the temporal slot → validate Gate-1 `missing-required-slot` → the singer thrashed and fell to a bar-over-string-months (not a line over a continuous axis).

## Fix (classify.ts only)
Temporal slot-fit now accepts `isTemporal(f) || (slot.temporal_from_string && inferStringTemporal(f)!==null)`, and exposes `temporal_from_string` in the propose-slot contract. `validate.ts`/`stringTemporal.ts`/DATEPARSE splice untouched. One confident bind of month→order_date now renders the correct line-over-time, killing both the judge-FAIL and the 310s tail.

## Regression guard (the #1 risk = false-positive format inference)
New test: a non-temporal-named string (`region`) still fails closed (inferStringTemporal is name-gated); real date/datetime fields unchanged.

## Verified
tsc 0 · full binder suite 718 pass (incl. portability + bind-behavior-matrix invariants) · eslint 0 · 2 new tests (string-month binds; non-temporal string fails closed) · v2.35.0→2.36.0. **Live re-sing of e4 pending** (will confirm the 310s→fast + judge flip).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.35.0",
+  "version": "2.36.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/binder/binder.test.ts
+++ b/src/desktop/binder/binder.test.ts
@@ -151,6 +151,50 @@ describe('binder/classifyNoLlm', () => {
   });
 });
 
+// ── e4: string-month temporal slot (temporal_from_string) ─────────────────────
+// trend-line-chart's order_date slot opts in via temporal_from_string:true. A 'YYYY-MM'
+// STRING month must be an acceptable source for it (DATEPARSE'd downstream to a continuous
+// axis) — WITHOUT this the string month never fills the temporal slot, the required-slot
+// gate fails, and the singer thrashes into a bar-over-strings (e4: 310s, judge 40).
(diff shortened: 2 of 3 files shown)


mattcfilbert:
Live re-sing on the fix (branch build): the temporal-axis half is PROVEN — judge confirms *'Delivered a line of SUM(Mau) over a PARSED month date, converting the YYYY-MM string into a real continuous temporal axis'* (the string→date bind now fires, was impossible before). Route improved 310s→198s. BUT still judge-FAIL (40→48) + still 6 tool calls: e4 has OTHER issues (multi-product summing) and the singer still thrashes before landing the confident bind — that thrash is a SYSTEMIC prompt/routing gap (singer doesn't take the now-working confident path first, doesn't consult knowledge on a miss), being investigated separately. This PR correctly fixes the classify temporal_from_string gap; it is necessary but not sufficient for a clean e4.

PR #593, opened 2026-07-22
@W-23540240: Improve performance of list-views when bounded context is applied

yyk16293:
# Pull Request Template

## Description
When bounded context is applied through INCLUDE_VIEW_IDS, Adds a fast path to `list-views`. In this fast path, each view is fetched directly via [GetView](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooks_and_views.htm#get_view) API rather than paging through all the views and filtering in memory. 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe):

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/web/views/listViews.ts
+++ b/src/tools/web/views/listViews.ts
@@ -10,8 +10,10 @@ import {
   getViewLineageQuery,
   mergeViewLineage,
 } from '../../../sdks/tableau/methods/lineageUtils.js';
+import { RestApi } from '../../../sdks/tableau/restApi.js';
 import { View } from '../../../sdks/tableau/types/view.js';
 import { WebMcpServer } from '../../../server.web.js';
+import { isAxiosError } from '../../../utils/axios.js';
 import { getExceptionMessage } from '../../../utils/getExceptionMessage.js';
 import { paginate } from '../../../utils/paginate.js';
 import { genericFilterDescription } from '../genericFilterDescription.js';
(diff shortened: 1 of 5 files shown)


src/tools/web/views/listViews.ts
```diff
@@ -88,26 +90,37 @@ export const getListViewsTool = (server: WebMcpServer): WebTool<typeof paramsSch
...
-                    const { pagination, views: data } =
-                      await restApi.viewsMethods.queryViewsForSite({
-                        siteId: restApi.siteId,
-                        filter: validatedFilter ?? '',
-                        includeUsageStatistics: true,
-                        pageSize: pageConfig.pageSize,
-                        pageNumber: pageConfig.pageNumber,
-                      });
+                const effectiveLimit = maxResultLimit
+                  ? Math.min(maxResultLimit, limit ?? Number.MAX_SAFE_INTEGER)
```
jarhun88, line 94: is this Number.MAX_SAFE_INTEGER necessary? wont one of maxResultLimit or limit guaranteed to be smaller

yyk16293: We need it when limit is undefined.

src/tools/web/views/listViews.ts
```diff
@@ -88,26 +90,37 @@ export const getListViewsTool = (server: WebMcpServer): WebTool<typeof paramsSch
...
+                  : limit;
+
+                // Fast path: when results are scoped to a specific set of view IDs and the user
+                // hasn't supplied a filter, fetch those views directly instead of paging the whole
+                // site. Query Views for Site has no view-ID filter, so the slow path would fetch
+                // every view only to discard all but the allowed ones. A user filter forces the
+                // slow path since Get View can't apply server-side filtering.
+                const viewIds = configWithOverrides.boundedContext.viewIds;
+                const views =
+                  viewIds !== null && !filter
```
jarhun88, line 104: not sure how big a deal this is, but maybe we should be using validatedFilter instead of filter

src/tools/web/views/listViews.ts
```diff
@@ -193,6 +206,60 @@ export function constrainViews({
...
+  limit,
+}: {
+  restApi: RestApi;
+  viewIds: Set<string>;
+  limit: number | undefined;
+}): Promise<Array<View>> {
+  const ids = [...viewIds];
+  const results = await Promise.allSettled(
+    ids.map((viewId) =>
+      restApi.viewsMethods.getView({
```
jarhun88, line 232: perhaps we can request a getViewsByID api from the platform for the future

src/tools/web/views/listViews.ts
```diff
@@ -88,26 +90,37 @@ export const getListViewsTool = (server: WebMcpServer): WebTool<typeof paramsSch
...
+                const effectiveLimit = maxResultLimit
+                  ? Math.min(maxResultLimit, limit ?? Number.MAX_SAFE_INTEGER)
+                  : limit;
+
+                // Fast path: when results are scoped to a specific set of view IDs and the user
+                // hasn't supplied a filter, fetch those views directly instead of paging the whole
+                // site. Query Views for Site has no view-ID filter, so the slow path would fetch
+                // every view only to discard all but the allowed ones. A user filter forces the
+                // slow path since Get View can't apply server-side filtering.
+                const viewIds = configWithOverrides.boundedContext.viewIds;
```
jarhun88, line 102: we dont limit the size of this right? im not sure how likely this is that we have users that set this to thousands+, which might make our parallel api calls hit rate limits or otherwise be flaky with a single failing api failing the tool

yyk16293: Yes, agreed. My expectation is that ideally viewIds won't be more than 30-40.

src/tools/web/views/listViews.ts
yyk16293: Looks like validatedFilter is used when we want to send it as a param to REST API. Here we can use regular filter.

src/tools/web/views/listViews.ts
yyk16293: Batching API in platform needs to be requested.

jarhun88 (approved):

stephendeoca (approved):

PR #597, opened 2026-07-22
feat(desktop): atomic activate-sheet tool (goto-sheet recovery) — replaces get/apply-workbook profile exposure

mattcfilbert:
Reshaped per review (original shape: expose get-workbook-xml + apply-workbook on the dynamic-authoring profile as a goto-sheet recovery fallback — rejected because recovery-only was unenforced and full-workbook mutation re-invites hand-XML authoring).

**New shape:**
- `activate-sheet` — dedicated atomic tool: read → set active-window node → validate → apply, all inside one apply-lock acquisition. No XML in or out; the caller can only name a sheet.
- Unknown sheet → structured error with `availableSheets` (machine-readable, quote/comma-safe).
- Dynamic-authoring profile: lean + 1 (31 tools); get-workbook-xml/apply-workbook stay banished.
- Surface byte budget repointed at the real serving surface: dynamic profile ≤ 30,000 (measured 26,552); full tools/list demoted to a 52,000 sanity cap.

**Verification:** tsc clean; full suite 336 files / 4930 tests green. Two-round independent review (atomicity race and lossy error path found and fixed with regression tests). Live-Desktop round-trip of the active-window mutation NOT yet probed — hold for the live check before relying on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.38.0",
+  "version": "2.39.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/server.desktop.test.ts
+++ b/src/server.desktop.test.ts
@@ -145,20 +145,29 @@ async function serializeDesktopToolSurface(tool: DesktopTool<any>): Promise<stri
 }
 
 describe('desktop tools/list serialized surface', () => {
-  it('stays under the tool-search auto-deferral threshold budget', async () => {
+  it('keeps the served dynamic authoring profile under the tool-search auto-deferral threshold budget', async () => {
     const server = new DesktopMcpServer();
-    let total = DESKTOP_INSTRUCTIONS.length;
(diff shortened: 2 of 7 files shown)


mattcfilbert:
**HOLD — Sol review (HARD gate) found a real design problem: recovery-only is NOT enforced.** Adding apply-workbook (a full-workbook MUTATING tool) to the lean profile means the agent CAN choose get-workbook-xml→edit→apply-workbook for AUTHORING, bypassing bind-template — re-opening the exact hand-XML thrash the profile was kept lean to prevent. The 'recovery-only' comment + xml-free descriptions are NOT enforcement. Also UNKNOWN whether apply-workbook actually carries the active-window node, and it's insufficient for >16KiB workbooks (needs the cache pair). 

**Recommendation: do NOT merge as-is.** The PRIMARY goto-sheet fix already works on the singable surface (execute-tableau-command tabdoc:goto-sheet {Sheet} + commandPolicy window-id override + TAS !69), so this XML fallback may not be needed for the demo at all. If we DO want it, the right shape is a dedicated atomic active-sheet tool (or a runtime gate/token issued only after a goto-sheet failure), NOT exposing full apply-workbook. Reframing before any merge.

PR #598, opened 2026-07-22
fix(binder): latlon resolver binds WIDE real map schemas (best detail dim), not fail-closed on 3+ dims

mattcfilbert:
## Live-caught on Blake's REAL data
Blake's actual World Cup workbook (teams.csv: team_id/team_api_id/group_name/country_code/team_name + lat/long + flag/color/source — 11 cols, 5 categoricals) **THRASHED** (route `BIND→BIND→BIND`, 29.6s, FAIL) where the clean synthetic s7 proxy binds single-BIND <20s. **The proxy was too clean; real map data is always wide.**

## Root cause (Sol-diagnosed, firsthand-verified)
resolveLatLonSymbolMap failed closed on 3+ categorical dims (the detail-selection guard) — but the mark identity is ONE label dimension (team_name), not every descriptive attribute. So the confident bind never fired → propose → the model thrashed across the 3 map templates.

## Fix
For 3+ categoricals, `pickBestDetailDim` scores: +2 ask-token overlap (team↔'World Cup team'), +1 'name' label, −2 per technical token (id/code/hex/url/emoji/source). Binds the unique top scorer to detail, coords to axes. **1–2 dims unchanged** (bind all — no collapse). A scoring **tie still fails closed** (a wrong grain is worse than a propose). Blake's schema → team_name on detail, noise ignored → confident single map.

## Tests
Replaced the old '3+ dims → fail closed' test with two: (a) wide World Cup schema → confident bind with team_name detail; (b) genuine tie (alpha/beta/gamma) → still fail closed. Full binder suite 718 green (invariants intact), tsc 0, eslint 0. v2.39.0→2.40.0.

This is the fix for why Blake sees inconsistency on his REAL data even post-#591 — s7 was the unrealistic case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.38.0",
+  "version": "2.40.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/binder/binder.test.ts
+++ b/src/desktop/binder/binder.test.ts
@@ -307,26 +307,57 @@ describe('binder/classifyNoLlm — measure-free lat/long symbol map (Blake wall 
     expect(classifyNoLlm('Build me a Tableau map of the office locations', forced, s)).toBeNull();
   });
 
-  it('BACK-DOOR CLOSED: with 3+ dimensions the resolver declines AND the generic keyword loop must NOT bind latlon (no axis-swap/dropped-dim escape hatch)', () => {
-    // latlon carries generic map keywords ('map'/'symbol-map'/'spatial'); if it were left
-    // in the generic scoring loop, a resolver-declined ask could still win latlon by keyword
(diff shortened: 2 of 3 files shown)


mattcfilbert:
Sol review found a real HIGH (addressed in 73779c4b): my ask-overlap scorer could pick a COARSE dim (tournament_stage) over the fine label (venue_name) when the ask names the coarse one → per-stage centroid collapse (the exact original bug). Fix: COARSE_GRAIN_TOKENS (group/stage/category/region/type/class/status/segment/tier/…) penalized −3, and ask-overlap does NOT rescue a coarse field; the winner must be POSITIVE + unique or it fails closed. Added a test proving 'map tournament stage venue locations' still grains by venue_name. Remaining Sol notes (all addressed or accepted): finest-grain can't be metadata-proven → the positive-fine-label-or-fail-closed rule is the conservative stance; resolver-precedence-over-choropleth is enforced by the early-return before generic scoring (classify.ts) — a both-templates-eligible test is a good follow-up but the forced-LATLON test + the resolver-only exclusion cover the demo path.

PR #599, opened 2026-07-22
feat(desktop): pinned session is a default, not an invariant

tableaukyler:
## What changed

This PR now goes the **opposite** direction from where it started. Instead of hiding the `session` param when a Desktop is pinned, we make the pin a **default the agent can override**, so the MCP can still talk to other running Desktop instances while knowing which one is current.

## Why

The pinned session was an invariant: `session` was stripped from every tool schema when pinned, and an explicit `session` that differed from the pin was rejected. But the PM wants the MCP to be able to target other open Desktops too. The pin should be a default, not a lock.

The original bug that #599 worked around was smaller than the fix: only **3** tools (`add-field`, `remove-field`, `write-cached-xml`) made `session` **required**, which forced the model to jam worksheet names like `"Sheet 1"` into it. Every other tool already had it optional.

## How

1. **Revert the original #599 commit** (the 49-tool schema-stripping mechanism).
2. **Fix the real bug**: make `session` optional on the 3 tools that were required.
3. **Unwind the invariant** so the pin is a default:
   - `resolveSession` uses the pin when `session` is omitted; an explicit `session` is validated against live instances and only rejected if it is not a running pid (fails open when discovery is empty).
   - `list-instances` is no longer hidden when pinned.
   - Session-resolution guidance is now pin-aware instead of dropped.

```mermaid
flowchart TD
    A[session param] --> B{omitted or == pin?}
    B -- yes --> C{pin set?}
    C -- yes --> D[use pin]
    C -- no --> E{how many Desktops running?}
    E -- exactly 1 --> F[use it]
    E -- 0 or 2+ --> G[fail closed]
    B -- "no (explicit other pid)" --> H{is it a running instance?}
    H -- yes --> I[target it]
    H -- "no (and discovery non-empty)" --> J[reject, hint list-instances]
    H -- discovery empty --> I
```

## Testing

- `npm test`: 335 files, 4929 pass, 0 errors
- `npm run lint`: clean
- `npm run build:desktop`: clean

### DIFF
--- a/src/desktop/desktopDiscoverer.ts
+++ b/src/desktop/desktopDiscoverer.ts
@@ -12,19 +12,6 @@ const manifestSchema = z.object({
 
 export type DesktopInstanceManifest = z.infer<typeof manifestSchema>;
 
-/**
- * Agent-actionable recovery text for a stale session (Desktop quit/restarted, or a
- * cache file produced by a different Desktop instance). Shared by SessionManager's
- * freshness gate and the cache-fingerprint bleed guard so both speak the same recovery.
- */
-export function staleSessionRecoveryMessage(sessionId: string | number): string {
-  return (
-    `Session ${sessionId} is stale: that Tableau Desktop process is no longer reachable or was restarted. ` +
-    'Call list-instances and retry with the current session id. ' +
-    'Cached XML files from the old session may not match the current workbook — re-read before applying.'
-  );
-}
-
 export class DesktopDiscoverer {
   private readonly isPidAlive: (pid: number) => boolean;
 
--- a/src/desktop/routeTable.test.ts
+++ b/src/desktop/routeTable.test.ts
@@ -6,6 +6,8 @@ import {
   generateDesktopInstructions,
   renderInstructionEntry,
   SESSION_RESOLUTION_ID,
+  SESSION_RESOLUTION_TEXT_PINNED,
+  SESSION_RESOLUTION_TEXT_UNPINNED,
 } from './routeTable.js';
 
 const routes = DESKTOP_ROUTE_TABLE.filter(
@@ -121,24 +123,25 @@ describe('generateDesktopInstructions', () => {
 });
 
 describe('buildDesktopInstructions', () => {
-  const sessionResolutionText = DESKTOP_ROUTE_TABLE.find(
+  const sessionResolutionEntry = DESKTOP_ROUTE_TABLE.find(
     (entry) => entry.id === SESSION_RESOLUTION_ID && entry.kind === 'prose',
   );
 
-  it('keeps the session-resolution guidance when no session is pinned', () => {
+  it('keeps the unpinned session-resolution guidance when no session is pinned', () => {
     const unpinned = buildDesktopInstructions({ sessionPinned: false });
     expect(unpinned).toBe(generateDesktopInstructions(DESKTOP_ROUTE_TABLE));
+    expect(unpinned).toContain(SESSION_RESOLUTION_TEXT_UNPINNED);
     expect(unpinned).toContain('list-instances');
   });
 
-  it('drops the session-resolution guidance when a session is pinned', () => {
+  it('swaps in pin-aware session guidance when a session is pinned', () => {
     const pinned = buildDesktopInstructions({ sessionPinned: true });
-    expect(sessionResolutionText?.kind).toBe('prose');
-    if (sessionResolutionText?.kind === 'prose') {
-      expect(pinned).not.toContain(sessionResolutionText.text);
-    }
-    expect(pinned).not.toContain('list-instances');
-    // The other routes must survive the filter untouched.
+    expect(sessionResolutionEntry?.kind).toBe('prose');
+    expect(pinned).toContain(SESSION_RESOLUTION_TEXT_PINNED);
+    // Still names list-instances — the pin is a default, and the agent may target another Desktop.
+    expect(pinned).toContain('list-instances');
+    expect(pinned).not.toContain(SESSION_RESOLUTION_TEXT_UNPINNED);
+    // The other routes must survive untouched.
     expect(pinned).toContain('You control Tableau Desktop.');
   });
 });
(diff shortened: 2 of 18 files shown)


mattcfilbert:
On the pinned-restart recovery path: when the pinned Desktop restarts, `SessionManager.getExecutor` → `ExternalApiToolExecutor.isAvailable() === false` throws `ExternalClientApiUnavailableError` — "This Tableau Desktop build does not serve the External Client API — update Desktop." For a restart, "update Desktop" points the caller at the wrong action; in pinned mode there's also no `list-instances`/session param to recover with, so relaunching the agent from Desktop seems like the only real recovery — could the error say that instead?

Related: `staleSessionRecoveryMessage()` in `src/desktop/desktopDiscoverer.ts` (which does carry retry guidance) appears to have zero callers in this tree, though its docstring names two — worth either wiring it into this path or removing it.

tableaukyler:
Both good catches — fixed in 8f8e8bcd.

1. **Pinned-restart message** — `getExecutor` now checks whether the failing session is the pinned pid. When it is, it throws `PINNED_DESKTOP_UNREACHABLE_MESSAGE`: "The pinned Tableau Desktop is no longer reachable — it was closed or restarted. Relaunch the agent from Tableau Desktop to reconnect." The old update-Desktop message stays for the unpinned case where a stale/old build genuinely can't serve the API. Added a test for the pinned path.

2. **`staleSessionRecoveryMessage`** — removed. Confirmed zero callers; the cache-fingerprint bleed guard carries its own recovery text and there's no freshness gate. Its "call list-instances" guidance is exactly wrong for the pinned path (list-instances is hidden when pinned), so it couldn't be wired in as-is.

mattcfilbert:
**P1 (verified on this branch's head) — `src/desktop/sessionResolution.ts`: empty discovery accepts any explicit session.**

When `discoverInstances(...)` returns empty, the guard `runningPids.length > 0 && !runningPids.includes(requestedSession)` short-circuits and any `requestedSession` is accepted. Concrete scenario: Desktop closes (or the discovery dir is missing) → discovery is empty → a stale session id from earlier in the conversation is accepted → local-FS tools (`write-cached-xml`, `add-field`, `inject-template`) write sidecars/XML under a dead session with no liveness check, bypassing the stale-session recovery machinery and risking wrong-workbook application when Desktop relaunches.

Open question rather than a verdict: if fail-open-on-empty-discovery is deliberate (e.g. pinned-session launches where the discovery dir legitimately isn't populated), a one-line comment saying so would prevent this report from recurring — and a narrower guard (fail closed for explicit sessions unless a pin is present) might give both properties. Either way, a regression test for the stale-session/no-instance case would pin whichever behavior is intended (current tests mock discovery to `[]` and don't cover rejection).

For transparency: an earlier automated finding claimed a deleted export (`staleSessionRecoveryMessage`) breaks consumers — checked against the current base head, it has no external consumers; the deletion is clean and no action is needed.

tableaukyler:
Deliberate fail-open — documented it and pinned it with tests (cc7f67b6).

Empty discovery can't confirm *or* deny the pid, so `resolveSession` proceeds rather than blocking blind. The intent is already in the JSDoc (precedence point 3): "if discovery is empty we cannot confirm, so we proceed and let the executor's unreachable handling report a truly dead pid — same posture as the cache-fingerprint guard." Narrowing the guard would break the legit case where Desktop is reachable but the discovery dir isn't populated yet (e.g. pinned-session launches), and it'd be redundant with the loud downstream failures.

On the safety concern — a stale pid can't silently corrupt the wrong workbook:
- `write-cached-xml` only writes a *working copy* inside the sandboxed cache dir (`isWithinCacheDir` guard). It never touches Desktop.
- Apply-* → `SessionManager.getExecutor` throws `ExternalClientApiUnavailableError` with a relaunch-recovery message when the pid is truly dead (`sessionManager.ts:50-54`).
- `checkSidecar` (cache-fingerprint guard) rejects a stale sidecar when Desktop relaunches under a new pid, telling the agent to re-read in the current session — so no wrong-workbook application.

You're right the case wasn't tested (existing tests all mock discovery to non-empty). Added two regression tests for empty-discovery — pinned and unpinned — locking the fail-open behavior.

Thanks also for the `staleSessionRecoveryMessage` transparency note — agreed, clean deletion, no action.

mattcfilbert:
**P2 (verified on this branch's head) — `src/desktop/sessionManager.ts`: stale *non-pinned* sessions get the wrong error class.**

`ExternalClientApiUnavailableError(sessionId, { pinned: config.desktopSessionId === sessionId })` — when a pinned config targets one Desktop but the caller resolved a *different* explicit session that then closes between discovery and `executor.start()`, `pinned` is `false` and the agent is told: *"This Tableau Desktop build does not serve the External Client API — update Desktop."* That's misleading in this scenario — Desktop is fine; the session is just gone — and can send a user (or an agent) down an update path instead of `list-instances` + re-resolve.

Suggested shape: a third message class for stale-but-not-pinned ("session N is no longer running — call list-instances and retry with a current session"), which is essentially the `staleSessionRecoveryMessage` this PR removes as unused — it may have just found its consumer. A regression for stale-explicit-≠-pin would pin it (existing tests cover only the stale *pinned* case).

tableaukyler:
Good catch — fixed in 36ef0ca9.

You're right that `pinned: false` conflated two distinct failures. `getExecutor` now picks the reason from discovery state, so a stale non-pinned session no longer gets sent down the update path:
- **pinned-unreachable** — the pinned pid is gone → relaunch from Desktop
- **stale-session** — this pid is gone but *other* instances are running → `list-instances` + retry with a current session
- **no-api** — discovery is empty → build likely too old, update Desktop

I went with a 3-way reason enum on `ExternalClientApiUnavailableError` rather than resurrecting `staleSessionRecoveryMessage` verbatim — the old string's guidance was written for the pinned path, so its wording didn't fit. The new `STALE_SESSION_MESSAGE` carries the same `list-instances`-and-retry intent you're pointing at.

One nuance worth flagging: I distinguish stale-session from no-api by whether discovery lists *other* running instances. If the requested Desktop closed *and* it was the only instance, that still reports no-api (update Desktop) — can't tell "just closed" from "never served the API" when discovery is empty. The common multi-instance case is now correct.

Added regression tests for stale-explicit-≠-pin (both unpinned, and pinned-at-a-different-live-pid).

PR #601, opened 2026-07-22
fix: read single sheets/dashboards as fragments and apply them without pruning siblings

lauren-jacksonSFDC:
Two related fixes for how the desktop server reads and writes single sheets/dashboards over the External Client API ("Athena V0"), whose only write route is the authoritative whole-workbook `POST /v0/workbook/document` (there is no per-sheet/per-dashboard write route).

## 1. Slice per-sheet/dashboard document reads to a fragment

The per-sheet and per-dashboard `GET .../document` routes return a whole `<workbook>` scoped to the requested item, not a bare fragment. Callers that feed that XML to apply-worksheet/apply-dashboard or a readback verifier need a single `<worksheet>`/`<dashboard>` fragment.

- **`getWorksheetFragment` / `getDashboardFragment`** — wrappers over the read gateway that slice the result to a single fragment (no-op for transports that already return one).
- **`worksheetDocumentToFragment` / `dashboardDocumentToFragment`** (`metadata/sheets.ts`, `metadata/dashboards.ts`) — slice a `<worksheet>`/`<dashboard>` out of a whole-workbook document; a bare fragment is returned unchanged, `null` if absent.
- Callers switched to the fragment getters: `loadWorksheetXml` readback, `batchCreateAndCacheSheets`, `refineWorksheet`, and the get-xml tools.

## 2. Apply by upserting into the whole live workbook (sibling-loss fix)

**Bug:** Applying one sheet/dashboard built a *minimal* whole-workbook doc — one element, siblings and the other type deleted — on the assumption the POST would "upsert by name and leave the rest live." That server-side upsert is gone: the POST now **replaces the open workbook wholesale**, so everything omitted from the minimal doc was pruned. Editing a worksheet wiped sibling sheets and all dashboards; editing a dashboard would have wiped every worksheet.

**Fix:** We already GET the whole live workbook before applying. Splice the edited element into that full document (replace-by-name, or append if new) and POST it, preserving everything else.

- `buildMinimalSheetDoc` → **`upsertSheetIntoWorkbook`**; `buildMinimalDashboardDoc` → **`upsertDashboardIntoWorkbook`**.
- Updated both callers (`loadWorksheetXml`, `loadDashboardXml`) and the tests that encoded the old "omits siblings" contract; added direct unit tests for both new functions.

## Testing

- Unit tests for both fragment getters and both upsert functions; affected suites green. `tsc --noEmit` and `eslint` clean.
- Verified live against Tableau Desktop (External Client API): refining a sheet with a sibling present, adding a new sheet, and refining a sheet with a dashboard + 2 other sheets present all preserve every sheet and the dashboard.

### DIFF
--- a/src/desktop/commands/workbook/getDashboardXml.test.ts
+++ b/src/desktop/commands/workbook/getDashboardXml.test.ts
@@ -2,7 +2,7 @@ import { Err, Ok } from 'ts-results-es';
 
 import { ExternalApiToolExecutor } from '../../externalApi/externalApiToolExecutor.js';
 import { ExecuteCommandError } from '../../toolExecutor/toolExecutor.js';
-import { getDashboardXml } from './getDashboardXml.js';
+import { getDashboardFragment, getDashboardXml } from './getDashboardXml.js';
 
 describe('getDashboardXml', () => {
   it('resolves dashboard name to id and fetches the per-item document', async () => {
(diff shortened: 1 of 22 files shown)


mattcfilbert:
Verified one structural point on the fragment-apply path.

The merge itself is structurally sibling-preserving — `upsertSheetIntoWorkbook`/`upsertDashboardIntoWorkbook` splice the edited sheet/dashboard into the *entire* live workbook by name and leave every other sheet, window, and dashboard untouched.

But the round-trip through it isn't lossless: `parseXML` (in `src/desktop/metadata/parser.ts`) is configured with `trimValues: true`, and every fragment apply (`upsertSheetIntoWorkbook` at sheets.ts:~155, `upsertDashboardIntoWorkbook` at dashboards.ts:~205) parses and re-serializes the *whole* workbook through that same shared parser/builder instance — including sibling sheets it never intended to touch. Any whitespace-significant `<run>` text (leading/trailing spaces in formatted text, titles, tooltips) in an unrelated sheet can silently get trimmed away the next time any single sheet is applied.

Suggest either splicing the raw target `<worksheet>`/`<dashboard>` element as a string (skip the full re-parse) or switching to a whitespace-preserving parse (`trimValues: false` + explicit trim only where whitespace isn't significant), plus a fixture that pins formatted text/windows/thumbnails on a sibling sheet across an unrelated single-sheet apply.

(Checked the apply-lock question too — `loadWorksheetXml.ts`'s fragment apply and `activateSheet.ts` (#597) both go through the same shared `withApplyLock`/`applyMutex.ts` singleton, so no lock-mismatch concern there.)

lauren-jacksonSFDC:
Good catch — confirmed and fixed in 8a825f1c.

Reproduced the trim: with formatted `<run>` text on a sibling sheet, editing an unrelated sheet dropped `<run>Sales: </run>` → `<run>Sales:</run>` and leading spaces on the next run, exactly as you described.

Went with the whitespace-preserving parse (`trimValues: false` in `parser.ts`) over the raw-string splice: it fixes the round-trip losslessly for *all* consumers (windows, thumbnails, datasource formatting), not just run text, and avoids hand-matching XML spans. `parseTagValue`/`parseNodeValue` are already `false`, so this only affects whitespace — no numeric/boolean coercion change. Ran the full metadata/binder/validation/commands/tools suites both ways: identical results, zero new failures.

Added the fixtures you suggested — sibling formatted `<run>` text with significant leading/trailing spaces asserted intact across an unrelated single-sheet apply, on both the worksheet and dashboard paths. Thanks also for checking the apply-lock path.

mattcfilbert:
Confirmed — `trimValues: false` in the shared parser plus the sibling `<run>`-preservation regressions on both upsert paths closes the round-trip concern. Nothing further from this side.

PR #603, opened 2026-07-22
fix(binder): live map ask confident-binds — cross-base tie + decline-reason observability

mattcfilbert:
Closes the remaining propose round-trip on the demo map ask (the <10s lever). Root cause from live receipts: singer rephrasings carry the bare token 'name', so `Player Name` ties the `Team Name` federated trio cross-base — #600's same-base collapse correctly refuses that tie. Fix: explicit-directive (+6) and full-name-phrase (+3, multi-token only, coarse-exempt) evidence bonuses; genuine ambiguity still fails closed (two full-named fields tie → null). Resolver ordering revert-proven load-bearing; generic-scorer latlon exclusion restored and regression-locked (#598 back-door). Adds machine-readable `decline_reason` to escalation payloads.

Review trail: Sol root-cause (the exact tie line), two-model convergent build, confirmation pass — 3 PASS / 2 FAIL round closed with regressions. Tests: three live-observed phrasings all confident-bind on the real 42-field fixture; cross-base two-full-names tie fails closed; full suite 335 files / 4938 green, lint clean.

Live verification: timed pristine re-sing queued — the sub-10s receipt (or its miss) lands in this PR's thread.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.43.0",
+  "version": "2.45.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/binder/binder.test.ts
+++ b/src/desktop/binder/binder.test.ts
@@ -376,6 +376,23 @@ ${dimensionColumns}
     expect(classifyNoLlm('Build me a Tableau map of the office locations', forced, s)).toBeNull();
   });
 
+  it('does not reopen latlon through generic scoring after the coordinate resolver declines', () => {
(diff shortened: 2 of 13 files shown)


mattcfilbert:
Live receipt (pristine instance, PR-branch build, literal demo ask): **Call 1 confident-bound — no propose, no decline_reason, single bind executing in 376ms.** Route fixed as designed. Timing: 11.35s TAS-duration / 12.2s wall — over the <10s bar, but the decomposition is now clean: three LLM turns (4.10s discovery-decision + 3.72s bind-decision + 3.50s closing) account for essentially all latency; tool execution totals ~0.4s. Notable: the closing turn is a full model call — TAB_AGENT_ONE_TURN_BIND did not synthesize `done` locally on this path in this config. Context 80,804 tokens (down from ~117k on propose routes). Structural render PASS via /v0 readback.

PR #608, opened 2026-07-22
fix(commands): refuse raw goto-sheet at the execute boundary — steer to activate-sheet

mattcfilbert:
Demo punch-list rank 4 (the 47BF7751 modal class). Audit receipts: the execute boundary validates command identity, key names, and required presence — never whether the sheet value exists — so a well-formed `tabdoc:goto-sheet {"Sheet":"nonexistent"}` reaches Desktop and pops the blocking modal (live-receipted repeatedly today via an agent retry loop). The registry's `opens_blocking_dialog: false` describes the happy path only, and post-dispatch conversion can't un-pop a modal.

Fix: refuse raw `tabdoc:goto-sheet` on BOTH guard paths (registry-enabled and registry-free — the live-dialog refusal now runs before registry dispatch) with a structured steer to `activate-sheet` (#597), which validates against live workbook windows and returns `sheet-not-found` structurally, pre-mutation. All agent-facing guidance swept (route census, recovery guidance, command examples) plus the knowledge corpus (execute-recipes → activate-sheet; XML button literals kept, annotated XML-only).

Review trail: audit → build → sweep → confirmation pass (3 PASS; one documented residual: internal post-apply focus uses lower-level goto-sheet with names the apply itself just created — self-owned ms-window, out of scope). Full suite 336/4944 green.


lauren-jacksonSFDC:

mattcfilbert:
Pushed to this branch: removed the stale post-apply `tabdoc:goto-sheet` focus mover (`focusAppliedSheet.ts` deleted; worksheet + dashboard apply paths no longer navigate; active-window preservation pinned by tests). This was the W66-era focus-follows — the same raw goto-sheet class this PR refuses at the boundary, and the cause of Desktop snapping back off the user's active sheet after applies. Applies are now focus-neutral, which makes the planned bind-integrated activation (via the activate-sheet machinery) the single sanctioned mover — clean ground for that work to land on. Full suite green (337 files / 4993+ tests) on the branch head.

mattcfilbert:
Fresh-eyes review of the focus-ripout commit (`4a1bef22`): no blockers. Three follow-ups that fit naturally with the bind-integrated activation work planned for this branch:

1. **Compose-end focus (Medium):** with the mover gone, a dashboard-compose flow now ends on whatever sheet was active at start, not the new dashboard. The activation machinery landing here should own that seam too (activate the composed dashboard at flow end) — same single-sanctioned-mover principle.
2. **Test strength (Medium):** the preservation tests pin the absence of `goto-sheet` and stable windows XML, but not the post-apply active window on the *creation* path (brand-new sheet/dashboard introducing a new `<window>`). Worth adding alongside the activation tests.
3. **`suppressFocus` is now a no-op (Low):** kept only so callers compile — remove the parameter and its remaining callers when activation lands, so no caller keeps expressing the obsolete "final apply owns focus" contract.

mattcfilbert:
@kyler — pushed: binding a template now navigates to the created sheet, **live-verified** on real Desktop (bound sheet becomes the active window with zero extra calls). The mechanism journey is worth knowing: windows-flag mutations via workbook apply turn out to be a live no-op (Desktop accepts the document but never moves the UI — two full implementations proved this before we caught it), so activation is a validated internal `tabdoc:goto-sheet` — fresh read, exact-name check against live sheets (the modal class only fires on bad values), boundary guard still refuses agent-supplied goto-sheet. Navigation policy per Matt: standalone binds activate; dashboard/batch intermediates stay neutral; a composed dashboard activates once at the end. Re-test when convenient — your "didn't go to my sheet after binding" case is the exact scenario now pinned live.

mattcfilbert:
Snap-back root-caused and fixed, live-verified: **all 47 template XMLs carried a stray `maximized='true'` window** that injection appended verbatim — Desktop's async apply-settle then honored that stale focus instruction and stomped the goto a moment later (matching Matt's live report: navigates correctly, then snaps to the blank first sheet). Fix: injection strips focus flags from appended windows + the corpus is swept clean (test-pinned so it can't regress), and activation now does settle (500ms) → validated goto → verify (700ms) → one bounded reissue. Live probe: bound sheet still holds focus 5 seconds post-apply. The old W66 mover was confirmed already dead — this was a second, independent mechanism.

PR #610, opened 2026-07-22
fix(templates): Miller spiral train — preflight rule scoping, honest field resolution, loud substitution guards, donor-name hygiene (2.49.0)

mattcfilbert:
## What / why

Matthew Miller's live session today (48-tool-call spiral for one bar chart) exposed four stacked defects in the template path. Each is fixed with the regression that proves it:

1. **`connections-not-authorable` scoped to top-level datasources** — the rule ran document-wide XPath and flagged the worksheet-view datasource references every template legitimately carries. This false error blocked all 9 of his auto-applies (reproduced RED structurally, GREEN after scoping). Companion: preflight failure messages now carry only error-severity findings — warnings can no longer masquerade as the blocker.
2. **build-and-apply-worksheet field resolution** — the legacy role prepass compared captions to canonical column_refs (impossible match) and reported false "field dropped" warnings on populated charts. Resolution now goes caption/local-name/column_ref with ambiguity refusal, invalid qualified refs are rejected with the nearest canonical ref, matching is trim/case-aware, all-dropped hard-fails before apply, and partial drops lead the message.
3. **Substitution guards (engine-level)** — surviving required placeholders fail loudly naming the user's bound fields (never donor names), unused optional slots are pruned (no phantom `[Facet]` columns), missing datasource fails before injection. Wired through binder, explicit inject, worksheet, and dashboard paths.
4. **Donor-name hygiene** — ranking templates' Superstore-era `Region`/`Sales` placeholders renamed to neutral `Category`/`Measure` (slot ids/derivations unchanged; routing unaffected), and `listXmlTemplates` now lists only manifest-backed templates, hiding the three raw orphans that fail silently.

## Verification
- Full suite green at each leg firsthand; final combined run pending one fixture fix on leg 4 (pushing shortly).
- Adversarially reviewed: an independent refute pass on the field-resolution change found two real breaks (wrong-derivation qualified refs, case sensitivity) — both fixed with regressions in this train.
- Live baseline captured pre-fix: packaged twbx applies cleanly on the unfixed build; the blocker reproduces on ingest-shaped workbooks — the live acceptance run replays Miller's exact asks against an ingested workbook.

Root-cause credit: three of the four defects trace to one commit (`af82a71f`); the windowless-sheet crash that ate his fallbacks was fixed separately today (#607).


mattcfilbert:
**Status (in flight):** one leg-4 follow-up just landed — `injectTemplateCore.test.ts`'s partial-mapping guard fixture reads the *shipped* ranking XML but still declared its slots/mapping with the old `Region`/`Sales` donor names, so after the rename the guard had nothing to catch and the expected throw never fired. Fixture now speaks `Category`/`Measure` (the guard-behavior assertions are unchanged, including the donor-name-never-leaks check, updated to the current donor vocabulary).

Remaining before this leaves draft: full-suite rerun on the final head (running), adjudication of an independent whole-train review focused on cross-leg interactions (survivor guard × donor rename identity-collisions, `listXmlTemplates` blast radius, the `canAutoApply` manifest narrowing), and the live acceptance replay of the original failing session's asks against an ingest-shaped workbook.

mattcfilbert:
Delta review adjudication (commits `2d91e046`…`8e127b55`): both actionable findings fixed and pushed — duplicate requested fields now count against claimed consumption (one applied + one dropped, never double-applied or silently vanished), and the constructed-workbook warning log is capped at 5 sanitized issue summaries.

Noted, deliberately not blocking this train:
- The correlation-scatter regression mocks `bindExplicitTemplate`; a binder-level consumption assertion would strengthen it (follow-up test).
- The Miller test's negative donor-name assertions are exact-string; if the donor vocabulary ever renames again, shelf/encoding-node assertions would be sturdier.
- Dropped-field slot-count wording can cite an imprecise reason when categorical/temporal slots differ (consumption itself is correct).
- Custom-vs-bundled template name collision precedence is unpinned (custom dirs were just made discoverable again; no current failure).

docs/proposed/template-placeholder-generalization.md
```diff
@@ -0,0 +1,141 @@
...
+- Formula/caption rewriting and computed sorts: unchanged structurally; their field token is now a placeholder.
+- Final residue guard: catches undeclared, missed, or malformed migration paths even when no literal donor name is known.
+
+## Pilot
+
+### `ranking-ordered-bar` — clean rename
+
+- `region` → `{{field_base_1}}`: ranked category.
+- `sales` → `{{field_base_2}}`: bar length and sort measure.
+- `facet_row` → `{{field_base_3}}`: optional row facet.
```
WillMartin, line 70: "Facet" seems like an artificial bit that we don't want in any templates. That seems like we've "partially fixed" the templates to include facets. But really the pre-existing entries were good metadata. So for example:

region, sales, state

is more meaningful than 

region, sales, "facet"

if it were trying to match something against the template

src/desktop/binder/manifest-types.ts
```diff
@@ -128,12 +128,22 @@ export type Derivation =
 
 export interface SlotSpec {
```
WillMartin, line 129: Mine included "example" which I think might be valuable. for the template manifests.

Is that maintained somewhere?

src/desktop/binder/manifest-types.ts
```diff
@@ -128,12 +128,22 @@ export type Derivation =
...
   slot_id: string; // stable id used by the LLM contract, e.g. "region", "order_date_month"
-  template_field: string; // bare field name AS IT APPEARS in the template <column name='[...]'>
+  /**
+   * Exact base token AS IT APPEARS in the template <column name='[...]'>.
+   * Migrated bindable slots use `{{field_base_N}}`; concrete names remain
+   * temporarily valid only for templates on the migration grandfather list.
+   */
+  template_field: string;
```
WillMartin, line 136: I'm confused what this is. Is slot_id not the id in the template?

mattcfilbert:
@WillMartin — the bot has now properly read your branch (`dev/will.martin/template-uplift-part-2`). Full comparison against this PR's schema v2, receipted both sides. The honest gap table:

| Dimension | Your uplift | This PR (schema v2) | Ahead | Adoption cost |
|---|---|---|---|---|
| **Datasource context** | Explicit DS groups; scoped IDs, mappings, placeholders, resolver | Slots carry no datasource/group field; singular `{{DATASOURCE}}` | **You** | L (real multi-DS support) |
| **Parameters** | Reserved `PARAMETERS` group, `PARAM_INSTANCE_N`, binds existing workbook params | Nothing | **You** | L |
| **Examples** | Required non-empty `examples[]` | None | **You** | **S — adopting now** |
| **Semantic matching** | purpose/examples/description/shelf as agent-facing metadata; resolver is exact-name; fast path zips positionally | Stable semantic `slot_id` as public identity + kind/role/purpose | Split: your annotations, our identity | S–M |
| **Placeholder grammar** | Group-qualified full-instance + base/derivation/type/aggregation forms; processor reconstructs column-instance structure | Base token only; DOM rewriter owns Tableau structure | Divergent — base-only is safer for this engine | not recommended wholesale |
| **Required/optional** | All instances mandatory; no optional pruning | Explicit `required` + optional-slot pruning | This PR | — |
| **Lint/validation** | Broad shape/enum/declaration checks; scoped resolver | Representative binding, donor-residue scan, monotonic grandfather ratchet | This PR overall | M (selected checks worth taking) |
| **Migration** | One-shot converter + manual refinement | 3 pilots, migration classes, hashes/render evidence | This PR | M (your converter as scaffolder) |

**On your two critiques, with receipts:**
1. *"slots lose DS context"* — **correct at the schema level.** Single-DS execution loses nothing today (the XML keeps qualification), but the manifest cannot express which datasource owns a slot, and singular `{{DATASOURCE}}` can't distinguish same-named fields across sources. Your model guarantees this structurally (`types.ts:4-19`, `resolver.ts:48-82`). One nuance: your shipped corpus doesn't yet exercise it either — no manifest declares `DATASOURCE_2`, the converter emits only `DATASOURCE_1`, and the coordination fast path assigns the first workbook DS to every group (`coordination.ts:777-826`).
2. *"needs multiple to reason about parameters"* — **agreed, and your PARAMETERS-group shape is the right foundation.** Proposed tmcp form: a `parameter_ref` slot kind + `datasource_id`, base-token XML (`[Parameters].[{{parameter_base_1}}]`), with one addition your resolver currently skips: existence/datatype validation against the parameter inventory (yours passes names through on faith, `resolver.ts:61-76`).

**Adoption order proposed:** `examples[]` immediately → `datasource_id` groups before any multi-DS eligibility → `parameter_ref` slots on that foundation (pilot: pareto) → controlled `derivation_flexible` → your converter as a scaffolder gated by this PR's lint. Kept from this PR: semantic `slot_id` as the stable binding API (positional `FIELD_INSTANCE_N` pushes meaning into metadata; we want it in the identity — which is also, we think, the real answer to your "facet" objection), and base-token substitution.

mattcfilbert:
@WillMartin on the multi-datasource question — here's the plan, and how this PR sets it up.

**Where we are today (traced firsthand):** each manifest-backed template application is strictly single-data-datasource. The binder flattens fields from all non-Parameters datasources, binds slots without DS grouping, and *rejects* a result whose resolved fields span more than one datasource (`cross-datasource-binding` escalation) — so multi-DS is a **capability gap, not a correctness hole**. It binds all-slots-from-any-one-DS fine (including a non-primary DS); it just can't span. Parameters aren't bindable (Pareto carries a hardcoded param ref).

**The middle path in this PR:** rather than graft the full DS-group model in now (and hold up the Miller P0 fix), #610 adds the schema-v2 foundation — stable `slot_id` + `kind`/`role`/`purpose`/`examples` — and I'd like to land a small forward-compatible **`datasource_id` hook on each slot** here so the follow-up doesn't need a second schema migration. It's inert until the group machinery lands.

**The follow-up PR (with you):** your DS-group model is the target — template-level datasource groups, per-slot group ownership, group-qualified placeholders, multiple `{{DATASOURCE_N}}` substitutions, and a distinct PARAMETERS group. The research confirmed a bare `datasource_id` alone isn't enough — it needs the group registry + multi-substitution, which is the L-effort unit that deserves its own review with your eyes. Your `template-uplift-part-2` branch is the reference.

So: **#610 lands the Miller fix + the schema hook; multi-DS group support lands next, built with you.** Full design comparison (yours vs schema-v2, gap table, adoption order) is written up and I'll walk you through it whenever — no rush given your afternoon.

_Posted by MattGPT (automation) on Matt Filbert's behalf._

PR #612, opened 2026-07-22
fix(refine-worksheet): sort_by_field no longer crashes when targetField is omitted

lauren-jacksonSFDC:
## Summary

`refine-worksheet` with `operation: "sort_by_field"` threw an unhandled `TypeError: Cannot read properties of undefined (reading 'trim')` when the caller omitted `targetField`. The schema marks `targetField` optional, so an agent can legally omit it — and then the tool crashed instead of returning a clean result.

Root cause: `resolveCiByCaption` called `normalizeCaption(caption)` (which does `value.trim()`) **before** its own `if (!wanted)` empty-caption guard, so an `undefined`/absent `targetField` threw before the guard could catch it.

Observed call (from a Tableau Studio session): `operation:"sort_by_field"`, `sortByField:"Goals For"`, `direction:"desc"`, **no** `targetField` (the dimension `Country`).

## What changed

**1. Stop crashing — and make the omitted case succeed**
- `planSortByField` now accepts an optional `targetField`. When it's absent/blank, it routes to the sibling planner `planSortByFieldOnCategoricalAxis`, which auto-detects the single categorical shelf axis (exactly the observed one-dimension case) — the same planner `bind-template` already uses. It refuses cleanly (never throws) when the axis is missing or ambiguous.
- Hardened `normalizeCaption`/`resolveCiByCaption` to accept `string | undefined`, so no future caller can reintroduce the crash.
- Dropped the unsafe `targetField as string` cast in the tool wrapper.

**2. Fixed the param contract**
- Every param now carries real `.describe(...)` text stating which `operation` requires it (all were empty strings, violating AGENTS.md).
- Added early per-operation `ArgsValidationError` guards in the callback (mirroring `add-field`'s `encodingType` pattern): `top_n` requires `topN`, `sort_direction` requires `sortDirection`, `sort_by_field` requires `sortByField` — rejected before touching Tableau.

## Tests

Added 7 colocated vitest cases (planner + tool level), including the **exact observed call** now resolving to a non-error `CallToolResult`, empty/whitespace `targetField` treated as omitted, and an ambiguous-axis refusal that never throws.

## Verification

- All 72 tests pass in the two affected files (65 existing + 7 new).
- `tsc --noEmit` clean; `eslint` clean.

_Note: the broader `agent-check` suite reports pre-existing failures on `feature/desktop` (stamp-hash gate, lockstep-core gate, discovery/stageReopen/resources) in files this PR does not touch — verified identical with these changes stashed._


mattcfilbert:
Carried this home while Lauren's out 💍🎉. Pushed the byte-budget balance so the fix fits the surface budget:

- Trimmed refine-worksheet's own schema descriptions **-543 bytes** (LLM-lean; the omitted-`targetField` axis auto-detect behavior stays discoverable) — 2126 → 1583.
- Raised the grandfather cap to exactly **1583** with inline justification (authorized).
- Lauren's `sort_by_field`-without-`targetField` tests are **unchanged**.
- Dynamic profile total **28425 / 30000**, full **47938 / 52000**. Full suite 339 files / 5054 green.

Rebased onto current feature/desktop; **MERGEABLE**.

PR #614, opened 2026-07-22
fix(summary-data): make empty and no-data results terminal so the agent stops retrying

mattcfilbert:
## Decision — rule we keep

A summary-data call that cannot produce data — empty sheet, no datasource, zero rows, or an invalid worksheet — now returns a stop marker instead of plain success. The rule: only the tool result can stop the agent's loop; the description cannot. This fires because live testing showed the agent calling get-summary-data ~17 times in a row against an empty sheet — each empty success read as an invitation to try again.

## Scope

summary-data only. Two parts:
1. Empty sheet / no datasource / zero rows / invalid worksheet now return `status` / `reason` / `guidance` with an explicit stop and a done next-action marker.
2. A repeat guard: identical calls (same args) within 30s hit a per-session 8-entry cache and get the same terminal result back, so the agent never sees fresh-looking emptiness.

Non-productive results stay terminal. Do not reintroduce a bare-success empty path.

## Risk

An intentional identical refresh within 30s gets the cached terminal result. The window is short by design and the state is process-local.

## Reviewer decision

Approve = empty and no-data summary-data results become terminal, plus a short repeat guard. Full suite 339 files / 5024 tests green firsthand.

Companion (separate, TAS-side): one backstop sentence for the system prompt — "Call get-summary-data only once per data-value ask and only for a populated worksheet; after any tool failure or terminal/no-data result, do not call it again — report the outcome."

Posted by MattGPT (automation) on Matt Filbert's behalf.


mattcfilbert:
**Hold — round 2 in flight.** Adversarial review found the round-1 repeat guard creates false terminals (signature lacks ask/workbook identity so a changed-view re-ask within 30s gets refused, with the timer refreshing on each rejection; a parallel first call is told to use a prior result that doesn't exist yet; transient Desktop failures were classified terminal). The redesign in progress flips the guard to **replay-not-refuse**: only completed outcomes are recorded, identical repeats get the recorded result replayed verbatim (uniform envelope, no fabricated priors, failures never recorded so retries stay open), no timer refresh, and transient errors are explicitly retryable. Do not merge this head.

mattcfilbert:
**Hold lifted.** Round 2 landed: the guard now replays completed results instead of refusing (in-flight/failed calls never recorded — no fabricated priors, retries stay open), the window anchors to original completion with no refresh (15s, shortened because no cheap workbook-revision fingerprint exists on this path), transient failures are explicitly retryable, and envelopes are discriminated with the standard done-label dialect. Full suite green firsthand. Delta review queued.

src/tools/desktop/data-source/getSummaryData.ts
```diff
@@ -42,64 +79,262 @@ export const getSummaryDataTool = (server: DesktopMcpServer): DesktopTool<typeof
...
+                      {
+                        maxRows: resolvedMaxRows,
+                        ...(columns && columns.length > 0
+                          ? { columnsToIncludeByFieldName: columns.join(',') }
+                          : {}),
+                      },
+                      signal,
+                    ),
+                );
+                if (summaryResult.isErr()) {
```
tableaukyler, line 154: Can you check to see if any other tool_calls are doing something similar? i.e. we're not returning an error in the response when we should be

mattcfilbert:
Round 3 landed — simpler than where this started: the replay cache is **deleted** (a second adversarial review found all four remaining P1s lived inside it — stale replays, parallel-completion races, envelope drift, payload retention). What ships: terminal structured results for non-productive outcomes (the mechanism that provably kills spirals), a bytes-only transient-retry counter (first transient failure → retry once; second consecutive → terminal), reason-specific next actions, honest labels. Full suite green firsthand.

@tableaukyler re your inline question — a full audit of every other tool callback for the same ok-when-should-error class is running now; the inventory (pattern: swallowed failures, non-answer successes, partial-failure-as-success) will land as a reply on your thread, and whether empty-sheet here should flip to `isError:true` outright will be decided with that audit's cross-tool convention rather than one-off.

src/tools/desktop/data-source/getSummaryData.ts
mattcfilbert: Audit complete — you were pulling a real thread. Static sweep of all 66 desktop tools: **19 further P1s** (beyond this PR's summary-data case), 5 P2s, 41 correctly classified. The worst classes:

1. **apply-worksheet / build-and-apply-worksheet** — tests literally pin `isError:false` around bodies saying `HOST VERIFICATION — failed`; the lead sentence still says "Successfully applied."
2. **bind-template / dashboard-auto-apply** — inject/validation/apply failures serialized into successful fallback bodies (`applied:false`, `apply_error`, `applied:"partial"` — all inside `Ok`). An empty dashboard can be described to the user as complete.
3. **batch-create-and-cache-sheets** — missing caches and failed fetches downgraded to warnings, still "Ready for Phase 2."
4. **viewpoint injection** — silently no-ops when the dashboard window is absent, then reports the requested viewpoint count applied.
5. **delete-worksheet / activate-sheet** — claim effects without verifying them (we reproduced both live today: `deleted:true` with the sheet persisting, `activated:true` with no navigation).

Cross-cutting fix (so the class dies once, not 20 times): a typed incomplete-operation error handled at the shared `DesktopTool.logAndExecute` funnel — forces `isError:true` while preserving the JSON recovery payload (guidance/nextAction/fallback args), plus required-answer validators at the read boundary (blank HTTP 200 bodies currently become successful empty XML in four tools), and a contract test banning failure-shaped `Ok` values (`applied:false`, `deleted:false`, `refined:false`, nested `isError:true`…).

TAS-compat checked before anyone flips bits: the only externally pinned envelope is bind-template's successful fast-path result (`applied:true, status:"bound"`, exact key set) — untouched by the proposal. The failure outcomes never satisfied that predicate, so flipping them to `isError:true` is safe against the host validator. Several local tests pin the buggy classification and will need intentional updates.

Full 66-row inventory with file:line receipts is in the session artifacts; fix wave is scoped and ready to run as its own PR train when Matt calls it.

PR #622, opened 2026-07-23
@W-23401091: Add always-on site_luid/user_luid log fields

jarhun88:
## Description

- Add a `Logger` class that always emits `site_luid`/`user_luid` fields via lazy getters
- Bind a per-request child logger onto the web tool context (`extra.logger`)
- Route request-scoped tool logs (Bucket A) and REST leaf-site logs (Bucket B: sign-out + request/response error interceptors) through the bound logger
- Retain the module-level `log()` for the context-less call sites

## Motivation and Context

Log lines emitted during web tool requests could not be reliably correlated to a site or user. This threads a per-request logger that always carries `site_luid`/`user_luid`. The getters resolve lazily so the fields populate correctly even when the LUIDs are set after sign-in.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Unit + integration tests: Logger unit tests (incl. lazy-backfill and anti-spoofing), `restApiInstance` integration tests (sign-out success/failure + request/response error interceptors), and `tool.test.ts` asserting the serialized LUID fields on stderr. Full suite green: 2510 tests pass, lint and build clean.

## Related Issues

@W-23401091

## Checklist

- [x] I have updated the version in the package.json file by using `npm run version` (3.5.5 → 3.5.6, patch).


docs/superpowers/plans/2026-07-22-always-on-luid-log-fields.md
```diff
@@ -0,0 +1,620 @@
+# Always-on `site_luid` / `user_luid` Log Fields Implementation Plan
```
stephendeoca: possibly remove these md files

src/logging/logger.ts
```diff
@@ -22,7 +22,7 @@ export function parseLogLevel(value: string | undefined): LogLevel {
  * Custom JSON.stringify replacer that serializes Error objects properly.
  * Removes the config field from AxiosError to avoid logging sensitive headers.
  */
-function errorReplacer(data: unknown): unknown {
+export function errorReplacer(data: unknown): unknown {
```
stephendeoca: Double check if this export is needed. It's being exported, but I don't see it referenced anywhere

src/restApiInstance.ts
```diff
@@ -164,30 +167,31 @@ export const useRestApi = async <T>(
...
-      // Tableau REST sessions for 'pat' and 'direct-trust' are intentionally ephemeral.
-      // Sessions for 'oauth' and 'passthrough' are not. Signing out would invalidate the session,
-      // preventing the access token from being reused for subsequent requests.
-      //
-      // Isolate the sign-out so a teardown failure can NEVER mask the callback's real result or
-      // error. A throw inside `finally` replaces whatever the `try` was returning or throwing, so an
-      // un-caught sign-out error would clobber the real outcome — e.g. a callback that 404s on a
-      // missing resource surfaces to the caller as the sign-out's 401 once the session is torn down
-      // (W-23202034). Swallow-and-log instead: sign-out is best-effort cleanup, and the ephemeral
-      // session expires on its own regardless.
```
stephendeoca, line 180: Why remove this comment, is this no longer true?

stephendeoca (approved):

src/logging/logger.ts
jarhun88: good catch

src/restApiInstance.ts
jarhun88: im not sure why the ai removed it, but good catch

PR #632, opened 2026-07-23
feat(desktop): surface server luid on list-workbook-datasources

nawartamawi:
## What

Surfaces the new server `luid` on the `list-workbook-datasources` desktop tool.

The External Client API now returns a `luid` on each item of `GET /v0/workbook/datasources` — the server-assigned LUID for a published, non-federated datasource, and `null` otherwise. (Monolith side: sf-analyticscloud/monolith#60876, which corrects that field to resolve the datasource's contentUrl to the real server LUID so it mirrors `/v0/site/datasources`.) This PR wires that field through the MCP so `list-workbook-datasources` reports the LUID inline, instead of a consumer needing a second `list-site-datasources` call (or the `resolve-datasource-luid` web tool) to map a workbook connection to its published LUID.

I'm picking up the MCP follow-up while Lauren is out; this stacks on the desktop external-API read routes she wired in earlier.

## How

Mirrors how `siteDatasourceItemSchema` / `list-site-datasources` already handle `luid`:

- **types** — add a nullable `luid` (`z.string().nullish()`) to `datasourceItemSchema`.
- **list-workbook-datasources** — project `luid` only when it is a non-null string (the API emits `luid: null` for embedded/federated datasources); refresh the tool description.
- **tests/fixtures** — the mock workbook-datasources fixture now has one published datasource (`luid` set) and one embedded (`luid: null`); client + tool assertions updated; added `luid` (as the nullable `["string","null"]` shape the API emits) to the captured `externalClientApi-openapi.json` so the contract drift-guard test stays green.

## Testing

- Affected vitest suites green: `listWorkbookDatasources`, `externalApiClient`, `externalApiContract`, `externalApiToolExecutor` (121 tests).
- `eslint` and `tsc --project tsconfig.providers.json` clean.

## Notes

- Targets `feature/desktop` (where the desktop external-API surface lives), matching the base of the related desktop work.
- Depends on the monolith build that ships the `luid` field (#60876); until that lands the field is simply absent/`null` on the wire, which this handles gracefully (projected out).

### DIFF
--- a/src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
+++ b/src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
@@ -1423,6 +1423,9 @@
           "id": {
             "type": "string"
           },
+          "luid": {
+            "type": ["string", "null"]
+          },
           "name": {
             "type": "string"
           },
--- a/src/desktop/externalApi/mockExternalApiServer.ts
+++ b/src/desktop/externalApi/mockExternalApiServer.ts
@@ -95,11 +95,15 @@ const DEFAULT_STORYBOARDS = [
 const DEFAULT_WORKBOOK_DATASOURCES = [
   {
     id: 'wb-ds-superstore',
+    // Published, non-federated: the server resolves a real LUID.
+    luid: 'luid-superstore',
     name: 'Sample - Superstore',
     caption: 'Sample - Superstore',
   },
   {
     id: 'wb-ds-quota',
+    // Embedded/federated: the API emits luid: null.
+    luid: null,
     name: 'Quota Targets',
     caption: 'Quota Targets',
   },
--- a/src/desktop/externalApi/mockExternalApiServer.ts
+++ b/src/desktop/externalApi/mockExternalApiServer.ts
@@ -107,6 +107,14 @@ const DEFAULT_WORKBOOK_DATASOURCES = [
     name: 'Quota Targets',
     caption: 'Quota Targets',
   },
+  {
+    // luid absent entirely (older API build that predates the field): exercises
+    // the `undefined` leg of `luid: z.string().nullish()` — the reason it's
+    // nullish() rather than nullable(). Projected out of the tool output like null.
+    id: 'wb-ds-legacy',
+    name: 'Legacy Extract',
+    caption: 'Legacy Extract',
+  },
 ];
 const DEFAULT_SUMMARY_DATA = {
   columns: [
--- a/src/desktop/externalApi/types.ts
+++ b/src/desktop/externalApi/types.ts
@@ -266,6 +266,9 @@ export type WorkbookInventory = z.infer<typeof workbookInventorySchema>;
 export const datasourceItemSchema = z
   .object({
     id: z.string().optional(),
+    // Server LUID of the datasource; present only for a published, non-federated datasource and null
+    // otherwise. Mirrors the luid on `GET /v0/site/datasources` (nullable string in OpenAPI 3.1).
+    luid: z.string().nullish(),
     name: z.string().optional(),
     caption: z.string().optional(),
   })
--- a/src/desktop/externalApi/types.ts
+++ b/src/desktop/externalApi/types.ts
@@ -267,7 +267,9 @@ export const datasourceItemSchema = z
   .object({
     id: z.string().optional(),
     // Server LUID of the datasource; present only for a published, non-federated datasource and null
-    // otherwise. Mirrors the luid on `GET /v0/site/datasources` (nullable string in OpenAPI 3.1).
+    // otherwise. Same field as the luid on `GET /v0/site/datasources`, but nullable here (["string",
+    // "null"]) because the workbook endpoint emits null for embedded/federated datasources, whereas
+    // the site endpoint's luid is a plain string.
     luid: z.string().nullish(),
     name: z.string().optional(),
     caption: z.string().optional(),
--- a/src/desktop/externalApi/externalApiClient.test.ts
+++ b/src/desktop/externalApi/externalApiClient.test.ts
@@ -212,8 +212,13 @@ describe('ExternalApiClient', () => {
 
     expect(result.isOk()).toBe(true);
     expect(result.unwrap().datasources).toEqual([
-      { id: 'wb-ds-superstore', name: 'Sample - Superstore', caption: 'Sample - Superstore' },
-      { id: 'wb-ds-quota', name: 'Quota Targets', caption: 'Quota Targets' },
+      {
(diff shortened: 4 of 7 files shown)


src/desktop/externalApi/types.ts
```diff
@@ -266,6 +266,9 @@ export type WorkbookInventory = z.infer<typeof workbookInventorySchema>;
 export const datasourceItemSchema = z
   .object({
     id: z.string().optional(),
+    // Server LUID of the datasource; present only for a published, non-federated datasource and null
+    // otherwise. Mirrors the luid on `GET /v0/site/datasources` (nullable string in OpenAPI 3.1).
```
tonythebest: Small doc nit: the captured fixture types `SiteDatasourceItem.luid` as plain `"string"`, not nullable — only the workbook `DatasourceItem` is `["string","null"]`. So "mirrors the luid on `GET /v0/site/datasources` (nullable string in OpenAPI 3.1)" reads as if the site schema were nullable too. Maybe something like "mirrors the luid on `GET /v0/site/datasources`; nullable here (`[\"string\",\"null\"]`) because the workbook endpoint emits `null` for embedded/federated"?

src/desktop/externalApi/mockExternalApiServer.ts
```diff
@@ -95,11 +95,15 @@ const DEFAULT_STORYBOARDS = [
...
     id: 'wb-ds-superstore',
+    // Published, non-federated: the server resolves a real LUID.
+    luid: 'luid-superstore',
     name: 'Sample - Superstore',
     caption: 'Sample - Superstore',
   },
   {
     id: 'wb-ds-quota',
+    // Embedded/federated: the API emits luid: null.
+    luid: null,
```
tonythebest, line 106: Optional: with `luid` now on every fixture item (string or `null`), the `undefined` leg of the `typeof === 'string'` guard isn't exercised anywhere — which is the case that motivates `nullish()` over `nullable()` in the first place. A third item with no `luid` at all (or a `datasourceItemSchema.safeParse({ id: 'x' })` assertion) would pin that choice down against a future refactor to `nullable()`.

src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
```diff
@@ -1423,6 +1423,9 @@
           "id": {
             "type": "string"
           },
+          "luid": {
+            "type": ["string", "null"]
```
tonythebest, line 1427: Optional: the harness's provenance note still describes this as a verbatim `info.version 0.1.0` capture, but `DatasourceItem.luid` is hand-applied here. That's the documented maintenance path and the shape looks right, so the drift guard still does its job — the only edge is that if the real 0.1.x spec ever marks workbook `luid` **required**, this hand-edit would mask it until the next capture. A short annotation on the header ("+ luid on DatasourceItem, hand-applied ahead of next capture") would keep that honest.

tonythebest (approved):
Approving — the wiring looks right and I verified it end to end.

I checked out `b7b24a14` and ran the affected suites (`listWorkbookDatasources`, `externalApiClient`, `externalApiContract`, `externalApiToolExecutor`): 121/121 green, matching your Testing notes. I also ran the `projectDatasource` luid spread against a value matrix (real UUID / `null` / absent / empty string) side by side with `listSiteDatasources`' `!== undefined` guard. Worth calling out that `typeof datasource.luid === 'string'` is deliberately *not* the site tool's idiom: it drops both `null` and `undefined`, so `luid: null` can't leak into the tool output for embedded/federated datasources. Copying the site guard would have leaked it. Nice catch.

One thing I checked because your Notes flag it as a dependency: the pre-#60876 slug-as-`luid` risk is moot. sf-analyticscloud/monolith#60876 merged to `main` as `5957dd41` carrying Lauren's `a8bac22` plus the resolution fix in one commit, and #60837 was closed as superseded — so no reachable build has the field without the resolution. Good call bundling those rather than pushing onto her branch. I also see `ResolveServerLuid` picked up the memo cache and the swallow-on-`TableauException`, so a single failed lookup won't sink the inventory.

Three non-blocking notes inline — all fine as follow-ups if you'd rather land this as is.

src/desktop/externalApi/types.ts
nawartamawi: Fixed in 3bd06a5b — reworded so it no longer reads as if the site schema were nullable. Now: the site endpoint's luid is a plain string; this workbook field is nullable ([\"string\",\"null\"]) because the workbook endpoint emits null for embedded/federated datasources.

src/desktop/externalApi/mockExternalApiServer.ts
nawartamawi: Good call — done in 3bd06a5b. Added a third workbook-datasource fixture item (wb-ds-legacy) with luid absent entirely, which exercises the undefined leg of z.string().nullish() (the reason it's nullish() over nullable()). Client + tool assertions updated to confirm it round-trips / projects out just like the null case, so a future refactor to nullable() would now fail a test.

src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
nawartamawi: Annotated in 3bd06a5b. The JSON fixture itself can't carry a comment, so I put the note on the provenance block in externalApiContract.test.ts: it now records that DatasourceItem.luid is hand-applied ahead of the next capture, and explicitly flags that if a future 0.1.x marks workbook luid required, this hand-edit would mask it until re-capture.

tonythebest (approved):
Re-approving `3bd06a5b`. All three nits are addressed, and the delta is comments plus test fixtures only — `projectDatasource` and the schema logic are untouched from the version I already approved.

The fixture addition is the one that carries real weight: `wb-ds-legacy` with `luid` absent entirely now exercises the `undefined` leg of `z.string().nullish()`, asserted in both directions — the client round-trips it as-is, and the tool projects it out like the `null` case. That's what pins `nullish()` against a future refactor to `nullable()`, which would now fail the client test rather than silently narrowing the accepted wire shape. Good call routing the fixture provenance note into `externalApiContract.test.ts` since JSON can't carry a comment; the caveat about a future 0.1.x marking workbook `luid` required is stated where the next person to re-capture will actually read it.

Verified locally on `3bd06a5b`: `externalApiClient`, `externalApiContract`, and `listWorkbookDatasources` all green. Four authoring suites (`authorAction`, `authorCalc`, `authorSet`, `formatLabels`) fail with `Tableau Desktop session '12345' is …`, but they fail identically at the fork point `0124995e` without your changes — 19 failed / 3 passed either way — so they're pre-existing and need a live Desktop session, unrelated to this PR.

PR #633, opened 2026-07-24
@W-23567697 - Reuse list-datasources resolver mode for datasource identity

tonythebest:
## Summary
- add optional `resolveContentUrl` mode to `list-datasources` so callers can deterministically resolve exact `contentUrl -> LUID` through a single tool surface
- preserve no-oracle behavior in resolver mode by returning the same error for denied-vs-absent matches, and reject ambiguous parameter usage (`filter` + `resolveContentUrl`)
- keep `resolve-datasource-luid` available for compatibility, while documenting `list-datasources` resolver mode as the preferred path

## Test plan
- [x] `nvm use 22 && npm run test -- src/tools/web/datasources/listDatasources.test.ts src/tools/web/datasources/resolveDatasourceLuid.test.ts`
- [x] `nvm use 22 && npm run lint -- src/tools/web/datasources/listDatasources.ts src/tools/web/datasources/listDatasources.test.ts src/tools/web/datasources/resolveDatasourceLuid.ts src/config.ts src/scripts/createClaudeMcpBundleManifest.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/scripts/createClaudeMcpBundleManifest.ts
+++ b/src/scripts/createClaudeMcpBundleManifest.ts
@@ -679,7 +679,7 @@ const envVars = {
     type: 'boolean',
     title: 'Enable insight-cards MCP tools',
     description:
-      'Registers the datasource-context insight tools (generate-insight-cards, resolve-datasource-luid). Disabled by default; set to "true" to enable them.',
+      'Registers insight-card tooling (generate-insight-cards plus datasource resolver paths including list-datasources resolveContentUrl and resolve-datasource-luid). Disabled by default; set to "true" to enable.',
     required: false,
     sensitive: false,
   },
--- a/src/tools/web/datasources/listDatasources.ts
+++ b/src/tools/web/datasources/listDatasources.ts
@@ -2,17 +2,20 @@ import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 import { Ok } from 'ts-results-es';
 import { z } from 'zod';
 
+import { ArgsValidationError } from '../../../errors/mcpToolError.js';
 import { BoundedContext } from '../../../overridableConfig.js';
 import { useRestApi } from '../../../restApiInstance.js';
 import { DataSource } from '../../../sdks/tableau/types/dataSource.js';
 import { WebMcpServer } from '../../../server.web.js';
 import { paginate } from '../../../utils/paginate.js';
 import { genericFilterDescription } from '../genericFilterDescription.js';
+import { resourceAccessChecker } from '../resourceAccessChecker.js';
 import { ConstrainedResult, WebTool } from '../tool.js';
 import { parseAndValidateDatasourcesFilterString } from './datasourcesFilterUtils.js';
 
 const paramsSchema = {
   filter: z.string().optional(),
+  resolveContentUrl: z.string().nonempty().optional(),
   pageSize: z.number().gt(0).optional(),
   limit: z.number().gt(0).optional(),
 };
@@ -23,6 +26,8 @@ export const getListDatasourcesTool = (server: WebMcpServer): WebTool<typeof par
     name: 'list-datasources',
     description: `
   Retrieves a list of published data sources from a specified Tableau site using the Tableau REST API. Supports optional filtering via field:operator:value expressions (e.g., name:eq:Views) for precise and flexible data source discovery. Use this tool when a user requests to list, search, or filter Tableau data sources on a site.
+
+  Resolver mode: pass "resolveContentUrl" to deterministically resolve one datasource by exact, case-sensitive contentUrl and return a single-entry array containing the canonical datasource LUID.
 
   **Supported Filter Fields and Operators**
   | Field                  | Operators                                 |
@@ -70,6 +75,8 @@ export const getListDatasourcesTool = (server: WebMcpServer): WebTool<typeof par
       filter: "createdAt:gt:2023-01-01T00:00:00Z"
   - List data sources with the name "Project Views" in the "Finance" project and created after January 1, 2023:
       filter: "name:eq:Project Views,projectName:eq:Finance,createdAt:gt:2023-01-01T00:00:00Z"
+  - Resolve a datasource LUID from exact contentUrl (single-tool mode):
+      resolveContentUrl: "GUS-Work"
   `,
     paramsSchema,
     annotations: {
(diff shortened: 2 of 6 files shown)


src/tools/web/datasources/listDatasources.ts
```diff
@@ -118,10 +144,51 @@ export const getListDatasourcesTool = (server: WebMcpServer): WebTool<typeof par
...
+            const exact = datasources.filter(
+              (datasource) => datasource.contentUrl === resolveContentUrl,
+            );
+            if (exact.length === 0) {
+              return new ArgsValidationError(
+                `No datasource matched contentUrl "${resolveContentUrl}"`,
+              ).toErr();
+            }
+
+            if (exact.length > 1) {
```
tonythebest: **High — bounded-context existence oracle for multiple matches.** This branches on `exact.length > 1` before checking whether any candidate is visible to the caller. Zero matches and one denied match return the generic no-match error, but multiple denied resources return `Multiple datasources matched`, revealing their existence. Could we run `isDatasourceAllowed` across exact candidates first, then decide zero/one/multiple from the allowed set? Please add all-denied, mixed allowed/denied, and multiple-allowed tests.

src/scripts/createClaudeMcpBundleManifest.ts
```diff
@@ -679,7 +679,7 @@ const envVars = {
     type: 'boolean',
     title: 'Enable insight-cards MCP tools',
     description:
-      'Registers the datasource-context insight tools (generate-insight-cards, resolve-datasource-luid). Disabled by default; set to "true" to enable them.',
+      'Registers insight-card tooling (generate-insight-cards plus datasource resolver paths including list-datasources resolveContentUrl and resolve-datasource-luid). Disabled by default; set to "true" to enable.',
```
tonythebest, line 682: **Feature-gate contract mismatch.** This shipped manifest says `INSIGHTS_TOOLS_ENABLED` gates `list-datasources resolveContentUrl`, but `list-datasources` is always registered and resolver mode never checks the flag. Should this mode be staged (enforce the flag) or generally available (remove this claim and the matching `config.ts` comment)? Either way, please add a test that pins the chosen behavior.

src/tools/web/datasources/listDatasources.ts
```diff
@@ -79,17 +86,36 @@ export const getListDatasourcesTool = (server: WebMcpServer): WebTool<typeof par
...
+            return new ArgsValidationError(
+              'Pass either "filter" or "resolveContentUrl", not both in the same request.',
+            ).toErr();
+          }
+
           const datasources = await useRestApi({
             ...extra,
             jwtScopes: listDatasourcesTool.requiredApiScopes,
             callback: async (restApi) => {
+              if (resolveContentUrl) {
```
tonythebest, line 127: **Resolver mode accepts but silently ignores `pageSize` and `limit`.** Since this branch hardcodes page size/page number, callers can provide pagination arguments that are logged but have no effect. Could we reject those alongside `filter`, or explicitly define and test their resolver-mode semantics?

src/tools/web/datasources/listDatasources.ts
```diff
@@ -79,17 +86,36 @@ export const getListDatasourcesTool = (server: WebMcpServer): WebTool<typeof par
...
+                const response = await restApi.datasourcesMethods.listDatasources({
+                  siteId: restApi.siteId,
+                  filter: `contentUrl:eq:${resolveContentUrl}`,
```
tonythebest: **Raw filter-value interpolation.** `resolveContentUrl` is inserted directly into Tableau's filter grammar. Commas/colons and similar structural characters can alter the server-side filter and cause valid exact lookups to disappear before the local post-filter. Could we use the canonical filter-value escaping/validation and add special-character tests for both resolver paths?

src/tools/web/datasources/resolveDatasourceLuid.ts
```diff
@@ -21,7 +21,7 @@ export const getResolveDatasourceLuidTool = (
     server,
     name: 'resolve-datasource-luid',
     description:
-      'Resolve a published datasource LUID by exact, case-sensitive datasource contentUrl match.',
+      'Resolve a published datasource LUID by exact, case-sensitive datasource contentUrl match. Prefer list-datasources with resolveContentUrl for new callers.',
```
tonythebest, line 25: Keeping this tool temporarily is reasonable for compatibility, but the PR does not reduce tool count until it is removed. Could we link a deprecation/removal follow-up with a migration window? Please also document that the old tool returns compact `{id,name,contentUrl}` while the new mode returns a full single-entry `DataSource[]`, so consumers need a result-contract migration.

tonythebest (commented):
The consolidation direction makes sense, but these bounded-context and public-contract issues need fixing before approval. The Salesforce CLA check is also currently failing because the commit author email is not associated with a GitHub user. Focused tests/lint are green; detailed review: `docs/reviews/review-tony-PR-633-list-datasource-resolver-mode.md`.

src/tools/web/datasources/listDatasources.ts
tonythebest: Fixed in latest push: resolver now computes cardinality over allowed exact matches (not raw matches), so denied-only cases collapse to the same no-match error. Added tests for all-denied, mixed allowed/denied, and multiple-allowed cases.

src/scripts/createClaudeMcpBundleManifest.ts
tonythebest: Aligned the contract with implementation in latest push: resolver mode on list-datasources is now explicitly gated by INSIGHTS_TOOLS_ENABLED (returns a validation error when disabled). Added a focused test that pins this behavior.

src/tools/web/datasources/listDatasources.ts
tonythebest: Added explicit resolver-mode input validation: resolveContentUrl now rejects pageSize/limit rather than silently ignoring them. Covered by a dedicated unit test.

src/tools/web/datasources/listDatasources.ts
tonythebest: Addressed in latest push by validating resolver filters through parseAndValidateDatasourcesFilterString before REST call construction, and added special-character tests for both resolver paths (list-datasources and resolve-datasource-luid).

src/tools/web/datasources/resolveDatasourceLuid.ts
tonythebest: Agreed. Kept resolve-datasource-luid for short-term compatibility, updated guidance to prefer list-datasources(resolveContentUrl), and added migration notes in local docs. I will track the actual removal in a follow-up sprint story once selected/created so tool-count reduction is explicit and time-bound.

tonythebest:
Short status: all previously raised review threads are addressed in the latest push.

- Commit: `e53ffcf`
- No-oracle cardinality is now evaluated over allowed exact matches
- Resolver mode is now gated by `INSIGHTS_TOOLS_ENABLED`
- Resolver mode now rejects `pageSize`/`limit` arguments
- Resolver filter input is validated via datasource filter parsing
- Compatibility/migration note is included while callers transition off `resolve-datasource-luid`

I also replied inline on each thread with exact fix mapping.

tonythebest:
Closing — superseded. W-23567697 (this PR's story) is dropped (Never): monolith #60876 (merged to main 2026-07-24) emits the datasource server LUID directly on `GET /v0/workbook/datasources`, so a `list-datasources` resolver mode is no longer needed. The standalone `resolve-datasource-luid` tool is instead removed outright in #711 (W-23567579).

PR #624, opened 2026-07-23
docs: ask PR authors to write plain titles and state the decision first

mattcfilbert:
## Decision — a rule we now keep

Anyone writing a PR title, PR body, commit subject, or review comment writes it plainly, and opens the PR body with the decision the reviewer must accept. This adds a `Decision` section to the PR template and a short writing rule to `CLAUDE.md`. It sets a house style; it does not change any code.

The prompt: a reviewer told us our recent PR titles were unreadable ("what does `feat(remove-field): name-based self-feed via shared helper — fail-closed, no cache reuse` mean?") and that our descriptions never said whether a change set a new rule or was a one-off. Orwell's plain-writing rules fix the words; a decision-first template fixes the missing decision.

## Scope

- `.github/pull_request_template.md` — a title hint plus a `Decision` section that asks the author to state whether the change is a rule we keep or a one-off.
- `CLAUDE.md` — a short "Writing PRs, commits, and comments for humans" section.

No behavior change, no dependency change, no version bump.

## Reviewer decision

Approve = we ask PR authors to write plain titles and state the decision first. Everything else in the template is unchanged.

Posted by MattGPT (automation) on Matt Filbert's behalf.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #635, opened 2026-07-24
feat(binder): confident-bind synth stack — waterfall/measure-by-dim/acronym/series-split + honest-error and readback fixes

mattcfilbert:
## What to approve

The confident-bind synth stack: a set of binder and tool-surface changes that make plain chart asks bind in one call and fail honestly when they can't. It is the branch we've been driving the live eval against (the a2td external-studio lane). Approving it takes these changes into `feature/desktop` as one reviewed unit.

## What it changes (the load-bearing pieces)

- **Confident binds** for the P&L waterfall, bare "measure by dimension" asks, acronym-named measures (MAU/DAU), and trend series-split by a spare dimension — each renders in one bind call instead of a propose→hand-author round trip.
- **Auto-apply on the first bind call** so a confident bind renders immediately (the speed lever).
- **Read top-N and a context filter from the ask** on the confident path; **route a derived-metric ask to author-calc before binding.**
- **Honest errors**: six tools that reported failure as success now return real errors; viewpoints inject after the dashboard exists; summary-data empty/no-data results are terminal so the agent stops retrying.
- **Tool-surface budget kept green**: bind-template's description and the shared proposal schema were trimmed so the served dynamic-authoring surface stays under the 30k auto-deferral cliff — no budget or cap was raised.

## Is this a rule we keep or a one-off?

A rule: **a plain chart ask should bind in one confident call, and a tool must return an honest error rather than a false success.** New binder work should preserve both — and keep the served tool surface under its byte budget (trim wording, never raise the cap).

## State

- Full suite green: **5227 passing / 1 skipped**, `build:desktop` clean, tsc + eslint clean.
- Behaviorally proven live against a running Studio via the a2td eval lane: e1 (revenue-by-region) and m1 (P&L waterfall) both graded PASS on answer quality, single confident bind.
- Content-audited: carries 9 of the open desktop PRs' work (#629, #628, #620, #619, #618, #617, #616, #614, #610); only #624 (a docs-only PR) is not folded in.


mattcfilbert:
**Review follow-up — two findings addressed** (posted by MattGPT on Matt's behalf).

A review of this stack raised two items; both are now fixed in `bf3bdc0e`.

1. The bind-recovery machine could block a legitimately-corrected second proposal on a repeated ask and steer the agent onto freehand building — the one path the confident-bind proofs never exercise. It now judges each proposal by its signature: a genuinely new correction advances recovery and is allowed (fails open), a previously-seen signature is treated as loop evidence and blocks, and distinct corrections are bounded at 8. Self-correction gets through; only real thrash or a runaway is stopped. The machine still does its job.

2. The gross-margin formula was inlined into the always-on instructions, duplicating the knowledge doc the calc-first route already points at. Removed the literal formula; the route still says author the calc first and read knowledge for the formula, so the instruction string shrinks and the formula is earned by the read.

State after the fix: full suite 5229 pass, `build:desktop` clean, tsc + eslint clean. No bind parameter removed; the recovery machine is not gutted; the pinned instructions snapshot is re-baselined to the real generated value.

One residual to watch, not a blocker: the recovery change is covered by unit tests but has not been confirmed with a live propose → corrected-retry → block trace against a running Studio. Worth a live check before relying on that path heavily.

tableaukyler (approved):

PR #638, opened 2026-07-24
feat(desktop): add export-worksheet-image and export-dashboard-image tools

myu404:
## Summary

Two Desktop MCP tools that render one sheet as an image, siblings to `get-*-xml` / `apply-workbook`:

- **`export-worksheet-image`** and **`export-dashboard-image`** — resolve the target by **name or id**, then call the Desktop loopback routes `GET /v0/workbook/worksheets/{id}/image` and `.../dashboards/{id}/image`.

`mimeType` is an enum: **`image/png`** (default) or **`image/svg+xml`**. Any other value is rejected at the tool boundary.

## How a response becomes a tool result

1. **Inline** (default) — base64 image block, labelled from the response's `effectiveMimeType` (the *actual* rendered format, not the request; falls back to `image/png` if absent).
2. **SVG** — also emits the markup as a text block for clients that don't render SVG image blocks.
3. **`filePath`** — writes to the given absolute path, returns the path (no inline bytes). Relative / `..`-bearing paths are rejected server-side.
4. **Over the inline cap** — spills to a cache file, returns its path.

## Bounds

- **`inlineImageMaxBytes`** (default 1 MiB, env `INLINE_IMAGE_MAX_BYTES`) caps inline size. SVG is counted at ~2× decoded size because it is dual-emitted.
- **`imageExportTimeoutMs`** (default 30s, env `IMAGE_EXPORT_TIMEOUT_MS`) bounds the render call only. On timeout the tool returns a dedicated 504-style error naming the likely cause (a Desktop modal blocking the render); client cancellation is distinguished from a timeout.

## Write semantics & errors

Because `filePath` overwrites an absolute path and the over-cap path writes a cache file, both tools declare `readOnlyHint: false, destructiveHint: true, idempotentHint: false`. When the endpoint returns an RFC-9457 Problem, its `code` is surfaced to the model (previously dropped). Builds too old to serve these routes return an honest 404 ("do not retry"); a 409 is surfaced as viz-not-ready.

## Testing

- `exportSheetImage.test.ts` (21): inline/SVG/filePath/over-cap branches, `effectiveMimeType` labelling, deadline timeout vs. cancellation, HTTP 500 Problem-code surfacing, 409, not-found selector, too-new 404, and `filePath` boundary rejection.
- `config.desktop.test.ts`: both new env knobs (default / override / malformed).
- Contract test pins the two image routes as absent from the 0.1.0 fixture.
- Full unit suite, lint, tsc, and desktop registration/byte-budget checks pass.


src/tools/desktop/external-api/exportWorksheetImage.ts
```diff
@@ -0,0 +1,93 @@
...
+): DesktopTool<typeof paramsSchema> => {
+  const exportWorksheetImage = new DesktopTool({
+    server,
+    name: 'export-worksheet-image',
+    title,
+    description: 'Render one worksheet as an image.',
+    paramsSchema,
+    annotations: {
+      title,
+      readOnlyHint: true,
```
mattcfilbert: Please flip this to `readOnlyHint: false` (same line in `exportDashboardImage.ts:36`).

There are two write paths here. `filePath` asks Desktop to write to any absolute path the model picks, and it will overwrite an existing file — which also makes `destructiveHint: false` optimistic. And the over-cap branch calls `writeFileSync` into the cache dir (`exportSheetImageResult.ts:87`).

The repo already treats exactly that as disqualifying: `workbook/applyWorkbook.ts:46` carries `readOnlyHint: false, // writes cache files and updates workbook`, and `cache/writeCachedXml.ts:40` the same. Per the [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#tool-annotations) `readOnlyHint` means the tool does not modify its environment, and clients use it to decide whether to run a tool without prompting — so this is more than bookkeeping.

src/desktop/externalApi/externalApiClient.ts
```diff
@@ -263,6 +276,22 @@ export class ExternalApiClient {
     );
   }
 
+  async exportWorksheetImage(
```
mattcfilbert, line 280: Please give the image call its own deadline.

`request()` resolves the signal as `options.signal ?? AbortSignal.timeout(this.timeoutMs)` (line 385), so the 60s `DEFAULT_TIMEOUT_MS` only applies when nobody passes a signal — and on this path somebody always does. The executor takes a required `signal` (`externalApiToolExecutor.ts:337-353`) and the read harness hands it `extra.signal` (`externalApiReadHarness.ts:40`). `extra.signal` is the MCP request-handler signal: it fires on client cancellation, not on a clock. Net effect is that an image export has no upper bound.

That's usually harmless on a read. It isn't here — we probed these routes on Desktop `main.26.0723.2250` on 2026-07-24 and the first image call after Desktop starts blocks behind a modal dialog and never returns. Wrapping the image call in `AbortSignal.any([signal, AbortSignal.timeout(...)])` turns an unbounded hang into something the model can actually report. What do you think?

src/tools/desktop/external-api/exportSheetImageResult.ts
```diff
@@ -0,0 +1,137 @@
...
+      ],
+    };
+  }
+
+  // (2) Under the cap: inline image block.
+  if (mimeType === 'image/svg+xml') {
+    return {
+      isError: false,
+      content: [
+        { type: 'text', text: Buffer.from(imageBase64, 'base64').toString('utf-8') },
```
mattcfilbert: Please check what came back before committing to the SVG branch.

`mimeType` here is `effectiveMimeType`, which `resolveImageExportQuery` (lines 127-135) derives from the request argument alone — nothing from the response ever touches it. The verified `/v0` contract from monolith #60585 says an unsupported `mimeType` **silently falls back to `image/png`** and the response does not echo the effective format (the host-side `image/svg+xml` comparison is also case-sensitive).

So when Desktop declines SVG for a given sheet, this line decodes PNG bytes as UTF-8 and ships the binary into the model's context, and line 111 labels the image block `image/svg+xml`.

The contract's own guidance is to check the decoded magic bytes when the format matters, and here it does. A sniff for `\x89PNG` (or a leading `<svg` / `<?xml`) before choosing dual-emit would cover it.

src/tools/desktop/external-api/exportSheetImage.test.ts
```diff
@@ -0,0 +1,352 @@
...
+      expect(result.isError).toBe(true);
+      invariant(result.content[0].type === 'text');
+      expect(result.content[0].text).toContain('neither image bytes nor a file path');
+      expect(vi.mocked(writeFileSync)).not.toHaveBeenCalled();
+    } finally {
+      await harness.close();
+    }
+  });
+
+  it('surfaces a 409 viz-not-ready as a command error (not a too-new endpoint)', async () => {
```
mattcfilbert, line 375: Please add the 500 alongside this one.

The suite covers 409, the too-new 404, an unresolved selector and the neither-bytes envelope — good coverage of what the mock can reach. What's missing is the shape the real routes return on the build we have. We probed both endpoints on Desktop `main.26.0723.2250` (which ships monolith #60585) on 2026-07-24: every sheet type we tried — blank sheet, extension sheet, and a native slope chart — returned HTTP 500 with error `0x8F2A4D91` (inner `0xBAD0A3FA`) and raised a blocking modal in Desktop. Roughly 15 repros across 3 instances, on an active Creator seat, so it isn't license gating.

The description frames the degraded mode as "older Desktop → honest 404", but present-but-failing is the state of every build that actually has these routes today. A 500 does survive the pipeline honestly, with one wrinkle worth a test: `mapClientError` (`externalApiToolExecutor.ts:561-569`) keeps the Problem `code` in `error.code` but only `detail ?? title` reaches `message`, and `formatDesktopCommandExecutionError` (`errors/mcpToolError.ts:276-292`) prints `message` plus a `tableau-error-code` this path never sets — so a 500 carrying a code but no `detail` reads to the model as `External Client API problem (500)` with the code dropped.

Related and cheap while you're here: `exportSheetImageResult.ts:70` tells the model "This is unexpected — retry". Against an endpoint that raises a modal on failure, a retrying agent wedges Desktop once per attempt. Could that lead with the `filePath` alternative and drop the retry?

mattcfilbert (commented):
Nice shape overall, Michael — resolving the sheet by name-or-id off the inventory before touching the image route is the right call, and it structurally dodges the unknown-id defect on the host side (a bad id never reaches the endpoint, so you never see the nil-GUID 500). The SVG dual-emit and the cap-forces-a-file downgrade are both thoughtful, and the contract test that asserts the routes are *absent* from the 0.1.0 fixture is a good forcing function.

Four things below that I think are worth fixing before this merges. The common thread on two of them: we probed these endpoints live on Tableau Desktop `main.26.0723.2250` on 2026-07-24, and they behave worse than the "older Desktop returns an honest 404" story in the description. Details in the inline comments.

src/tools/desktop/external-api/exportWorksheetImage.ts
myu404: 🤖 MichaelGPT's response on behalf of Michael Yu

Fixed. Both tools now declare `readOnlyHint: false, destructiveHint: true, idempotentHint: false`. There are two write paths — a caller-supplied `filePath` has Desktop overwrite an arbitrary absolute path, and an over-cap inline image is written to a uniquely-named cache file — so read-only was wrong and the optimistic `destructiveHint: false` didn't reflect the `filePath` overwrite. This matches the precedent in `applyWorkbook.ts` / `writeCachedXml.ts`.

src/desktop/externalApi/externalApiClient.ts
myu404: 🤖 MichaelGPT's response on behalf of Michael Yu

Added. New config `imageExportTimeoutMs` (env `IMAGE_EXPORT_TIMEOUT_MS`, default 30000). The image call runs under `AbortSignal.any([requestSignal, AbortSignal.timeout(...)])` scoped to the image render only (the sheet-list call still rides the request signal). On a timeout it returns a dedicated `ImageExportTimeoutError` (504) that names the likely modal-dialog cause and says not to blindly retry — this converts the observed unbounded Desktop modal hang into a reportable error. Timeout is distinguished from client cancellation so a cancel isn't mislabeled as a timeout.

src/tools/desktop/external-api/exportSheetImageResult.ts
myu404: 🤖 MichaelGPT's response on behalf of Michael Yu

Fixed. Desktop silently falls back to PNG when it declines an SVG render and doesn't echo the format, so keying the block label off the requested type mislabeled a PNG as `image/svg+xml` (and decoded PNG-as-UTF8 for the text block). The result now sniffs the decoded magic bytes and keys the footprint, extension, block label, and dual-emit on the actual format. Follow-up: I'm adding a commit that consumes an `effectiveMimeType` response field directly once Desktop emits it, so the label comes from the server's declared rendered format rather than byte inspection.

src/tools/desktop/external-api/exportSheetImage.test.ts
myu404: 🤖 MichaelGPT's response on behalf of Michael Yu

Fixed. The Problem `code` was being dropped from model-facing output — `mapClientError` now sets `tableau-error-code` on the command error so `formatDesktopCommandExecutionError` surfaces it. Added a unit test for an HTTP 500 Problem (`code: 0x8F2A4D91`, `detail: 'Render failed'`) asserting both the code and detail reach the model. Also reworded the neither-bytes message to lead with the `filePath` alternative and drop the retry suggestion, since retrying against a modal-raising endpoint just re-wedges Desktop.

mattcfilbert (commented):
## MattGPT review (AI-assisted review posted on Matt Filbert's behalf; Matt approved before posting)

Solid pair of tools — the response-branch handling (inline / SVG text block / filePath / over-cap spill to cache) is thorough, and labelling the image block from `effectiveMimeType` rather than the request is exactly right. Three things to address before merge, two questions for your call.

### Please
- **`src/desktop/inlineImageCap.ts:16` — `imageByteLength` is exported but nothing calls it** (one grep hit in the whole tree: its own definition). Either wire it in, drop it, or if it's for tests, move it into an `exportedForTesting` object so the public surface only carries what the server imports.
- **Missing `npm run version:patch`.** The branch still carries the merge-base version (2.52.2 vs 2.60.2 on `feature/desktop`). Merging won't regress anything since `package.json` isn't in the diff, but the convention is a patch bump per PR — two new tools warrant one.
- **`src/config.desktop.ts` — two new env-parsed knobs with no test.** `config.desktop.test.ts` already has the 3-test precedent for `inlineXmlMaxBytes`; `inlineImageMaxBytes` deserves the same (default, override, malformed value).

### Questions
- **`destructiveHint: true` on both tools** — every sibling desktop tool sets `false`, including the cache-writing ones. The overwrite-capable `filePath` arg makes `true` defensible, but the default render path is read-only and per-tool hints can't split the difference — some clients will approval-gate every render. Worth a deliberate call either way; a one-line comment on the hint would stop the next reader from "fixing" it.
- **`filePath: z.string().optional()` has no boundary refinement.** The PR body says relative and `..`-bearing paths are rejected, and the mock models the 400, but no test exercises it and the tool boundary passes the string through raw. A `.refine()` with a message naming the requirement (absolute path) would let the model self-correct without a round-trip to Desktop — or at least a test pinning the 400 translation.

Nits are in Matt's local notes if you want them (SVG size reporting, a duplicated deadline block, a restated query type) — none blocking.

myu404:
🤖 MichaelGPT's response on behalf of Michael Yu

Addressed the second round. Pushed in `15876378`.

**Please**

- **`imageByteLength` dead export** — dropped. Confirmed zero callers in the tree before removing it; the server never imported it.
- **`version:patch`** — bumped to **2.60.3**. Merged `feature/desktop` into the branch first so the base was the real 2.60.2, not the stale merge-base version, so the patch lands above the base rather than regressing it.
- **Untested env knobs** — added the default / override / malformed-fallback trio for **both** new knobs, not just `inlineImageMaxBytes`: `imageExportTimeoutMs` was in the same boat. Follows the existing `inlineXmlMaxBytes` precedent in `config.desktop.test.ts`.

**Questions**

- **`destructiveHint: true`** — keeping it, now with a one-line comment on each tool so the next reader doesn't "fix" it. The overwrite-capable `filePath` is the reason: it can clobber an arbitrary absolute file the model picks, which is genuinely destructive. The read-only default render path can't split the difference in a per-tool hint, and I'd rather over-declare (a client may approval-gate a render) than under-declare a real overwrite. The closest sibling precedent is the cache-writing tools declaring `readOnlyHint: false`.
- **`filePath` boundary** — kept the schema string raw and added a test pinning the server's 400 instead of a client `.refine()`. A client-side refinement would have to re-implement the server's absolute-path rule (including the Windows drive-letter form), and a mismatch risks rejecting a path Desktop would actually accept. The test asserts a relative or `..`-bearing path is rejected and `writeFileSync` is never called.

**Nits** — folded in: deduped the duplicated per-tool render-deadline into one shared `exportSheetImageWithDeadline` helper. (SVG size reporting and the restated query type were already handled in the earlier round.)

One heads-up for the diff: the two tools are full-profile-only, so they don't touch the default served (dynamic-authoring) surface — that total is unchanged. They do push the opt-in `full` tools/list surface just past 46k; I raised that test's cap to the honest measurement with a comment. Crossing 46k on the opt-in full profile is benign (those clients auto-defer tool schemas, same as the web surface already does well past 46k); the lean served surface stays under the cliff.

mattcfilbert (approved):
lgtm

PR #640, opened 2026-07-25
list-available-fields: drop column-ref-only fields from slim verbosity

nawartamawi:
## What

`verbosity: "slim"` on `list-available-fields` now returns only the field-picking essentials per field:

```
{ caption, localName, type, role, datatype?, isAggregated? }
```

Three keys are dropped from each slim field:

- `columnInstanceName`
- `derivation`
- `typePivot`

The now-unused `typePivot()` helper is removed with them.

## Why

Those three keys were **column-ref reconstruction ingredients** — they let a caller rebuild `[datasource].[derivation:LocalName:typePivot]`. Slim callers (the viz-canvas insight-bundle flow) never reconstruct column refs; they pick fields by caption/localName/type/role and hand the localName back. The dropped keys carried no signal for that path and dominated the per-field byte count on wide datasources, so slim payloads shrink materially with no loss to any slim consumer.

Full verbosity (the default) is unchanged — it still emits the table + `column_ref`.

## Test

- `listAvailableFields.test.ts` updated: slim schema drops the three keys; the slim `toMatchObject` assertions drop them; new `toBeUndefined` assertions pin that they are absent; the aggregated-calc test is retargeted to assert `isAggregated` (the flag it actually still needs) rather than the removed derivation metadata.
- `npx tsc -p tsconfig.json --noEmit` clean, `npm run lint` clean, `npx vitest run listAvailableFields.test.ts` → 15/15 pass.

### DIFF
--- a/src/tools/desktop/fields/listAvailableFields.test.ts
+++ b/src/tools/desktop/fields/listAvailableFields.test.ts
@@ -335,9 +335,7 @@ describe('listAvailableFieldsTool', () => {
       role: 'dimension',
       datatype: 'string',
     });
-    // Slim carries only the field-picking essentials: the full column_ref and the
-    // column-ref-reconstruction ingredients (columnInstanceName/derivation/typePivot)
-    // are omitted.
+    // Slim omits column_ref and the columnInstanceName/derivation/typePivot fields.
     const first = groupFields[0] as Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


nawartamawi:
closing in favor of  https://github.com/tableau/tableau-mcp/pull/703

Issue #269, opened 2026-03-18
Docs for TELEMETRY_PROVIDER_CONFIG out of sync with code


anyoung-tableau:
The code block under [TELEMETRY_PROVIDER_CONFIG](https://tableau.github.io/tableau-mcp/docs/configuration/mcp-config/env-vars#telemetry_provider) for the `TelemetryProvider` class is missing the `recordHistogram` method. Consider linking to the code instead of including it in the docs.

PR #643, opened 2026-07-25
feat(desktop): v11 release — binder fixes, one-call derived metrics, slim tool surface

mattcfilbert:
## Decision

This is the release candidate for the next Studio MCP bundle, now at **2.59.0** with the reviewed stack folded in (Matt's call, 2026-07-27). Approving it means shipping the v11 binder fix stack, the one-call derived-metric route, the slim default tool surface, set-action authoring, author-set empty mode, and the scoped filter-limit preflight to Studio.

## Folded since the 2.58.0 freeze (all seven reviewed, fixed, and re-verified)

#655 set-action authoring · #657 author-set empty mode + mode-independent marker readback · #651 drilldown knowledge · #658 errors-as-modals catalog · #653 /v0 health cap + typed abort · #650 byte-exact sort pins · #654 filter-limit rule scoped to top-level groupfilters (riders dropped).

**Re-verification on the folded tip (run firsthand 2026-07-27):** full suite 368 files / 6243 passed / 332 skipped at final tip `a26da612`; tsc clean; lint clean; fresh desktop build at `29ba6c62`; live smoke h6 PASS 96 / e1 PASS 88 / m6 PASS 96 (worktree mode, TAS @ !90 head f873a88). e1's 88 vs its 92-96 record is single-leg judge variance, above floor.

**Paired final review complete (Opus breadth + Sol depth), both rounds of findings fixed on-branch:** `32cb6e38` closed the fold-created author-set P1 (top-n count positivity, datasource-scoped readback, empty-mode arg rejection); `a26da612` closed the breadth P2s (restored tool descriptions incl. the target_worksheet war-story fact, honest `applied:false` on sheet reuse, knowledge doc reconciled to the bind-first gate, ratchets re-measured at 29,181/44,663). Breadth verdict SHIP WITH FIXES → fixes shipped. Named non-gating residuals: classify.ts depth pass, template-duplicate byte compare, live set-authoring probe.

**Post-fold paired review (Opus breadth + Sol depth) caught one P1 the fold itself created and it is fixed at `32cb6e38`:** #654 dropped the author-set positivity guard as covered by #657, whose check is presence-only — so top-n `count='0'` escaped both guards. Now: positive-integer literal enforced (parameter refs marked write-blind), readback scoped to the target datasource (kills a cross-datasource false success), empty mode rejects contradictory top-n args. Honest caveat: the smoke trio does not exercise author-set; the fix is unit-covered, and a live set-authoring probe is the named residual.

## What's in it (all live-verified unless noted)

- **Binder fix stack**: s8 grain fix (first-ever s8 passes, 93 ×2), m6 proposal defaults + propose-path context measures (FAIL 60 → 94/95), waterfall presentation re-stamp, temporal/synonym classes, speed campaign.
- **One-call derived metrics** (new today): margin/ratio/growth asks bind in ONE bind-template(auto_apply:true) call with the calc inline. Live: h6 PASS 96 @ 9.0s median, legs at exactly 1 tool call; e1 92-96 and m6 94-95 smokes green on the same bundle.
- **Honest receipts** (new today): bound-without-apply receipts lead with NOT APPLIED; zero-row readbacks no longer ship a terminal done; calc-comment apostrophes no longer disable the fabrication scanner.
- **Freeze-review P1 fix**: field narrowing ranked business synonyms above ask-named fields — on wide schemas the named field got evicted and Call 2 was unconstructable. Named fields now hold the top tier with slot-kind representation reserved.
- **Combined-variant fix**: the slim tools/list override was hiding every web tool on the shared McpServer; it now installs only when the desktop server owns its McpServer (caught by e2e once a lint failure stopped masking it).
- **Slim surface**: default tools/list 33,022 → 28.8k bytes, honest ratchets.

## Test focus for the bundle re-pin

1. A derived-metric ask ("show me gross margin %") — expect one fast bind, no field-inspection preamble, answer names the formula.
2. Combined build lists desktop + web tools.
3. m6-shaped ambiguous top-N asks — proposal carries a recommended default.
4. Waterfall: gain/loss colors + labels.

## Known risks

- Route-prose changes are historically live-fragile; h6/e1/m6 smokes ran on the final route prose (receipts in comments), but watch early Studio sessions.
- m6 judge variance and e5 wall drift remain open watch items.
- Raw XML tools are opt-in (TOOL_PROFILE=full) on the default profile.
- Two lockstep-core files changed hashes: classify.ts AND src/desktop/templates/fieldReferenceRewriter.ts — agent-to-tableau-desktop needs a re-sync of both after this merges (the cross-repo byte match is a manual step).
- A bind-first gate now refuses list-available-fields and get-worksheet-xml until an authoring attempt (speed campaign); the discovery-first knowledge doc is being reconciled to describe it.

Ask-named fields are never evicted by prompt narrowing; unapplied bound receipts must say nothing landed; the desktop server never owns tools/list on a shared McpServer.


mattcfilbert:
Live release evidence (build 9df4cc86, external-studio lane, pid-matched readback):

- **s7 (Miller's office-map steel thread): PASS-SLOW 91 — the first pass ever recorded on either stack** (shipped extension: FAIL 33; our pre-release brain: FAIL 42). The scoped-data carve-out plus the real committed stage did it. Slow (229.7s) — speed is the next campaign.
- **e1: PASS 88 at 9.0s** — under the 10s product bar for the first time.
- h6 gross margin: PASS-SLOW 95-96, and the one-call calcs[] fusion works (route shrank from 6 calls to 4; the remaining calls are knowledge lookups).
- g1 (the fabrication guard for the carve-out): PASS — 'sales in the West region' over data with no region field still refuses correctly.
- m1 waterfall: PASS 92-94.
- **Known regression, documented:** m6 (ambiguous 'top customers') now asks and stops instead of building a stated default (FAIL vs the old board's PASS 78 — which itself was the agent defying a stop instruction). Two prose attempts to fix this regressed e1 live (0/5 vs 7/7 — bisected and reverted); the durable fix is product-shaped: the bind proposal should carry a canonical-default recommendation. Queued.

Every number above is from live runs against this exact head; suites 6114 passed / 0 failed, lint and lockstep clean.

mattcfilbert:
Update (build 73d88725): the m6 known-regression is FIXED — structurally, not with prose. A ranking-shaped proposal now carries a recommended default (exactly-one revenue-like measure, top_n 10, never auto-applied), and the live smoke converts m6 from FAIL 8 to **PASS-SLOW 83**, with e1 holding at PASS 92 / 8.8s and h6 at PASS-SLOW 95. Also folded: census tranche 1 — the served default tool surface shrank 33,022 → 28,553 bytes per tools/list ($schema strip, annotations.title dedupe, list-knowledge-resources moved to the full profile) with every byte ratchet lowered to the new honest numbers. Suites 6119 passed / 0 failed. No known regressions remain on this head.

mattcfilbert:
Speed round 2 folded (head bf214450), live-verified:

| song | before tonight | now |
|---|---|---|
| e1 | PASS ~88 @ 7-9s | PASS 90 @ 7.7-9.4s (holds) |
| m6 | FAIL 8 (built nothing) | **PASS 80 @ 9.2s — one call, under the 10s bar** |
| h6 | PASS-SLOW @ 23.4s | PASS-SLOW 95 @ 16.8s |
| e2 | PASS-SLOW @ 32-41s | PASS-SLOW 88 @ 22.3s |
| s7 | FAIL both stacks | PASS-SLOW 80-91 @ ~220s (its long-persistence shape; run-to-run variance is real and documented) |
| e5 / g1 | PASS | PASS-SLOW 96 / 90 (guards hold) |

What round 2 added: a decisive knowledge search returns the module body (2 calls → 1); a recommended ranking default auto-applies with a stated-choice disclosure (m6's one-call conversion); and the bind-first gate — orientation reads return a deterministic redirect until the session's first authoring attempt, then unlock for repair (Matt's ruling: gate, not cut — s7's repair path depends on these tools post-bind). The profile-cut branch was built, measured, and retired.

Suites 6150 passed / 0 failed; lint, tsc, lockstep clean. a2td side: judge structure-extraction (#269) and labeled model-variant runs (#270) merged. Known watch-items, named: m6's profit-trap caveat is judge/agent variance (FAIL 55 vs PASS 80 same build); e5 drifted to 14.9s this run — both queued for the N>1 variance instrument.

mattcfilbert:
Probe-review wave 1 folded (bf214450..c81b6cfe), live-verified at sonnet-5 low effort:

- **bind-template unfilled-encoding steer now emits a resolved column ref** (or a bounded candidate list) instead of a `<field>` placeholder. Live: the s8 leg that previously spiraled 232s/10 calls through the XML lane now finishes in 9.8s/1 call.
- **Canonical ranking default binds up to 3 context measures (profit/margin first) to the tooltip** and the applied receipt tells the agent to quote notable values; plus a one-line unconverted-sum caveat when a currency/unit dimension is off-view. Live: m6 80 → 93/94/96; e1 best-ever PASS 96 @ 7.1s.
- **pareto-chart + deviation-spine-chart computed-sorts parameterized** (they hardcoded Superstore fields; the rewriter silently dropped them on real data) and the rewriter now surfaces dropped optional sorts as receipt warnings.

Held back on `claude/stamped-template-polish`: the same sort fix for magnitude-simple-bar/ranking-ordered-column and the waterfall label/gain-loss defaults — those three templates carry W60 render stamps, so their bytes wait for the factory golden-parity re-stamp flow.

Suite 6,165 green; Sol review run on the fold (no P0/P1; one P2 hardening applied in c81b6cfe). Smoke legs: e1 96/91/80, m6 93/94/96, m1 93/93, g1 92, s8 FAIL 89 (deterministic wrong-measure floor — the binder picks player-grain `Goals` over `Goals For`; tracked separately, the floor is correct).

*(posted by MattGPT — Claude-driven automation on Matt's behalf)*

mattcfilbert:
**Wave-2 evidence (c81b6cfe → b1ee2e7f, sonnet-5 low, all legs live on a bundle built from this branch):**

- **s8 first passes ever: PASS 93 ×2** (was FAIL 89-94). Grain-aware disambiguation: the binder now reads each field's remote table from metadata-record parent-name; when one ask token names measures from different tables ('goals' → per-player [Goals] vs per-team [Goals For]), a unique compound name wins, else a co-tabled dimension decides, else the proposal carries grain labels and the model chooses. 4/4 legs bound goals_for onto size AND color. Two a2td harness fixes rode along (#272 caption↔local-name floor normalization, #273 tooltip-marker correction backed by golden-XML evidence). Wall variance 9.9-295s remains — the agent sometimes chases unplottable ISO codes; tracked.
- **m6 propose-path hole closed: FAIL 60 → 94 / PASS 95 @ 9.9s.** When the agent confirms via an explicit Call-2 proposal instead of the internal auto-default, applied_default (and the profit/margin context measures) silently vanished — the revenue leader's −75k profit went unmentioned again. Offline A/B proved classify output byte-identical pre/post wave-2, so the route choice is agent variance; both routes now attach the context measures.
- **h6 hedge dead: 6 → 4 calls, scores 95-97 held.** calcs[] formulas resolve loose field references deterministically (exact → plural; synonyms are candidates-only in errors), quote/]]-escape-aware; percent-like dividing calcs get p0% format. The remaining wall (27-50s) is prompt-side hedging (search-knowledge + list-available-fields), not tool mechanics.
- **No regressions:** e1 PASS 96 @ 8.8s, m1 PASS 93.

Suite 6,184 green, tsc clean, lockstep regenerated. Each branch Sol-reviewed with fix rounds before fold (3 P1s fixed on the calcs path, 1 fast-path-regression P1 fixed on the grain path).

mattcfilbert:
**Addendum (b1ee2e7f → 90391889):** the held waterfall presentation defaults are now IN this PR, stamp re-earned rather than forged — gain/loss boolean color (teal/red palette), SUM labels, parameterized computed-sort. Verified on two live m1 legs on a bundle built from the restamp branch (PASS 91 @ 6.8s one-call, PASS 93): judge confirmed the running total to the exact net income, boolean gain/loss color on marks, correct sort, and the earlier 'value labels are off' dock is gone. The evidence block states plainly this round is structural+values verification; the pixel scores stand on the 2026-07-06/07 gate for the unchanged geometry. The magnitude-simple-bar / ranking-ordered-column sort-param changes stay held on claude/stamped-template-polish pending their own verification. Suite 6,184 green on the fold.

mattcfilbert:
**Review by MattGPT — automation posting on Matt Filbert's behalf.**

Freeze review. Two defects. The first should block the freeze.

---

### P1 — field narrowing can drop the field the user named, leaving Call 2 with nothing to bind

`src/desktop/binder/classify.ts:1830`

```ts
const synonymCandidate = synonymCandidates.has(f);
const tier = synonymCandidate ? 3 : relevant ? 2 : compatible ? 1 : 0;
```

Business-synonym candidates now rank above the tier for fields the ask names. `fieldMatchesCaptionPattern` matches `sales`, `revenue`, or `amount` as a whole token anywhere in a field name, so on a wide schema every `... Amount` measure becomes a rank-3 candidate and fills the 20-field cap.

Measured on a 27-field schema (22 measures named `... Amount` / `... Sales`, 5 dimensions), ask `bar chart of revenue by Customer Segment`:

| | fields exposed | `Customer Segment` | dimensions |
|---|---|---|---|
| `origin/feature/desktop` | 20 | present | 4 |
| this branch | 20 | **gone** | **0** |

`buildCall2Contract` builds `compatible_field_names` from `llm_input.fields` through `fieldFitsProposedSlot` (`src/tools/desktop/binder/bindTemplate.ts:807`). With no dimensions left, the required `categorical` slot comes back empty, while `proposal_requirements.field_selection` still says "choose one exact compatible_field_names value; do not rename or infer a field." There is no valid Call 2.

It compounds with the new fail-closed gate at `classify.ts:3241`: more than one synonym candidate forces the propose path, and that is the same condition that causes the eviction.

`src/desktop/binder/field-narrowing.test.ts:6` still states the contract this breaks — "rank-1: fields whose name/caption tokens overlap the ask survive (even past N)". The test passes because its filler fields are named `Filler Meas N`, so none match a synonym pattern, and Superstore-scale schemas never reach the cap. That is why the live receipts did not catch it.

Fix: rank synonym candidates below `relevant`, or seed the kept set with named fields before the tier sort.

---

### P2 — an empty summary readback still ships the "done" marker

`src/tools/desktop/binder/bindTemplate.ts:1672` and `:1701`

`needsFollowUp` is computed at line 1672. `readAppliedSummaryRows` is awaited at line 1701, after it. `summary_rows_error` therefore never reaches the terminal decision, so a bind whose live readback returned zero rows still returns `nextAction{kind:'done'}` and "no further tool calls needed". `bindTemplate.test.ts:2226` pins `applied: true` for that case and does not check `nextAction`.

`capSummaryRows` already tells a successful fetch of zero rows ("empty readback") apart from a failed fetch, so this is real evidence, not a dark endpoint. The file's own comment says "A tool that cannot verify what it wrote must not report success" — here it can verify and drops the negative answer. The tool description asks the model to fetch when rows are missing, but that prose competes with a marker the same file calls "the durable machine contract".

Fix: await `readAppliedSummaryRows` before `needsFollowUp`, and let an empty readback suppress `doneNextAction`. It should not fail the bind.

mattcfilbert:
**Review by MattGPT — automation posting on Matt Filbert's behalf.** Follow-up to the comment above.

Two more, both the same shape as the P2 there: work gets dropped and the caller still reports success. Both are new in this PR. Neither is as serious as the P1.

---

### P2 — one apostrophe in a comment turns off field-reference checking for the rest of the formula

`src/tools/desktop/data-source/authorCalcCore.ts:230-241`

The new scanner tracks quote state:

```ts
if (quote) {
  if (char === quote && formula[index - 1] !== '\\') quote = undefined;
  continue;
}
if (char === "'" || char === '"') { quote = char; continue; }
if (char !== '[') continue;
```

It does not know about `//` or `/* */`, which Tableau formulas allow. So in:

```
IF [Sales] > 0 // don't count returns
THEN [Profit Ratio] END
```

`[Sales]` resolves, the apostrophe in `don't` opens a string that never closes, and every reference after it is skipped by both `resolveCaptionReferences` and `resolveLooseFormulaReferences`. A string ending in a literal backslash does the same thing, since the code tests one preceding character instead of counting them.

Skipping is silent. `resolveLooseFormulaReferences:194-223` only sets an error when it reaches a token, so a field name that does not exist is written through instead of raising `ArgsValidationError` — the fabrication guard `resolveLooseReferences: true` was added for. `authorCalculationsInWorkbook:100` then checks the readback with `hasColumnNameAndCaption`, which compares the column name and caption and never looks at the formula, so the calc comes back `Ok` and `bindTemplate.ts:2065` publishes it in `authored_calcs`. A calc that is red in Desktop reports as authored.

The base branch used `formula.replace(/\[([^\]]+)\]/g, …)` with no quote state, so this is new.

Fix: skip comments before the quote check, and count trailing backslashes rather than testing one.

### P2 — `inject-template` never reads the warnings channel this PR added

`src/desktop/templates/injectTemplateCore.ts` now returns `warnings` from `rewriteFieldReferencesWithDiagnostics` (for example, "computed-sort dropped: … did not resolve"). `bind-template` reads it at `bindTemplate.ts:1531` and lets it block the `done` marker.

`src/tools/desktop/template/injectTemplate.ts:205-209` checks `result.ok`, writes `result.xml`, and ignores `result.warnings`. Line 217 forwards only the unrelated `explicitBind.warnings`. So the same dropped sort is reported through one tool and swallowed by the other.

Fix: merge `result.warnings` into `explicitTemplateWarnings` before line 217.

mattcfilbert:

## Andy-lens release review — breadth pass on 29ba6c62

Read-only pass over everything **except** authorSet / authorAction / nonPositiveFilterCount / the byte ratchets, which a second reviewer took in depth (their P1 is already fixed in 32cb6e38). The release is sound and the shape is right: the binder work is deterministic and well commented, the bind-first gate is properly tested, and stripping tool names out of the parameter descriptions is exactly the right call. Five things to settle before merge — all coherence, none a live defect. I ran the knowledge, route, route-state, tool and cannot-know-gate suites firsthand: 127 passed.

### Please (address before merge)

**1. `bind-template`'s description went telegraphic while every other tool in this release went clearer.**

The rest of the diff does the right thing: `'Workbook path from resolve-field or list-available-fields'` becomes `'Cached workbook path from field resolution'`, `'Template id from list-templates'` becomes `'ID of an existing template.'`. Names of other tools out, plain nouns in. Good.

But `bind-template` now reads `'Bind. applied_default=>state as default+offer change. Quote summary_rows; fetch now if absent/cut/error.'`, and its `session` param is `'PID; omit if one/pinned.'` while the identical param on the neighbouring tools now reads `'Desktop process ID; omit to use the pinned or only running instance.'`. Same payload, same parameter, two registers.

Two problems. The description no longer says what the tool does — "Bind." is not enough to pick it, and the release headline is one-call derived metrics, which the description dropped. And the rest is response-handling advice the model cannot use until after it calls; it already ships in the result body as `guidance`. Suggest keeping the plain first sentence and dropping the duplicated response prose.

**2. `target_worksheet`'s description regressed the defect the comment above it records.**

`src/tools/desktop/binder/bindTemplate.ts:60-69`. The comment still says this parameter, undescribed, "cost 299 repeat binds and 2,562 seconds: with no way to learn that it means 'edit THIS sheet'". The description it guards went from `'Existing sheet to rebuild in place (see list-worksheets). Omit and a NEW sheet is created.'` to `'Rebuild; omit to create.'`, which never says the value is an existing worksheet name. Drop the tool reference, keep the noun — or delete the comment, because it no longer describes the code.

**3. The body names one lockstep file; the manifest changed two.**

`lockstep.hashes.json` has new hashes for both `src/desktop/binder/classify.ts` and `src/desktop/templates/fieldReferenceRewriter.ts` (+115 lines). The body's known-risks list names only `classify.ts`. `AGENTS.md` in this same diff now states that the gate "cannot see the consumer repo, so the cross-repo byte match itself is a manual step" — a manual step with an incomplete list is how `agent-to-tableau-desktop` drifts. Please name both files in the re-sync note.

**4. A new gate refuses two read tools, and the body doesn't mention it.**

`src/tools/desktop/tool.ts:167-186` plus `firstAuthoringAttempt` in route-state. `list-available-fields` and `get-worksheet-xml` now return `BIND_FIRST_ORIENTATION_REDIRECT` instead of data until a mutating tool has been attempted in the session. The implementation is fine and the tests are real.

The problem is coherence. Nothing in "What's in it" or "What future work must preserve" mentions it, so approving this release quietly approves a contract change on two read tools. And it contradicts knowledge shipped in the same commit: `resources/desktop/knowledge/personalization/discovery-first-authoring.md` still says to "Prefer the lightweight inventory calls before anything heavy: `list-available-fields`" and still lists "Blind build — applying a viz without first checking what the workbook contains" as the top mistake, while the route table now says never to call those two tools to orient. One of the two has to move, and the enforced one is the route table.

**5. `reusedSheetResult` sets `applied: authoredCalcs.length > 0`.**

No sheet is applied on the reuse path; `applied` turns true only because calcs were authored. On `AppliedFastPathResult` the same field means the chart is on screen. `reused:true, applied:true` is the shape that invites a terminal "done". The receipt and guidance are honest — they say no second copy was created and offer the rebuild — so it is just the flag. A separate `authored:true`, with `applied:false` on reuse, would say the true thing.

### Questions (your call)

- **Hidden round trip on every applied bind.** `readAppliedSummaryRows` runs on the success path with a 2 s timeout. h6 measures 9.0 s with "legs at exactly 1 tool call" — one agent call, two Desktop round trips. Is the 2 s worst case inside the e5 wall drift you already list as a watch item, or could it surface there?
- **A third `bindTemplate()` per Call 2.** The dry re-classify that recovers `recommended` for context measures. The intent is right — the agent's route shouldn't change the answer. Is a full re-classify the cheapest way to get it, given `BindRecoveryProposalContext.recommended` already carries the recommendation into Call 2?
- **`target_worksheet` escapes the recovery gate outright** (`bindTemplate.ts:~469` returns before the record is read). The gate exists because 55% of production binds re-bind the same worksheet, and naming a target is what a looping agent adds next. Is there a second guard?
- **`recommended` is hand-declared twice and pinned to literals.** `route-state.ts:72-73` types `top_n: 10` and the reason string as literal types, duplicating a shape classify.ts owns. Changing the default from 10 becomes a type change in two files.

### nits

- `execute-tableau-command`'s `command` lost the `namespace:command` format hint. Dropping the tool name is right; the format is the part that lets the model build a valid value. `'Command ID in namespace:command form.'` keeps both.
- `annotations.title` is now optional repo-wide (`src/tools/tool.ts`), which loosens the contract for the web tools too.
- The new author-action / author-set params ship with `.describe('')`, which emits `"description": ""` — bytes for no information. Breadth note only; that file had a separate reviewer.
- `capSummaryRows`' `cappedColumns` is never capped despite the name, so a wide-column readback reports `'oversize readback'`.

### Verdict

**Ship with fixes.** Nothing found here is a live defect. Items 3 and 4 are the ones I'd want changed before merge, because both are about what a reviewer is being asked to approve rather than about the code.

tableaukyler (approved):

mattcfilbert:
> [!NOTE]
> This review was written by **MattGPT** (Matt's review automation, run on Claude Opus 5) and posted on Matt's behalf for public discussion. It applies the review standards distilled from this repo's review history — not a human reviewer's verdict. Reviewed at head `b23f48d2`.

Sound stack, and the shape is right. Two calls worth crediting: the `bind-template` annotations comment that removes the old `readOnlyHint`/`idempotentHint` claim — naming the retry incentive it was creating is the right level of explanation — and clean `exportedForTesting` hygiene: every new export has a real product consumer, which on a stack this size does not happen by accident.

### Please (address before merge)

- **[package.json] The version moves backward.** The tip is `2.59.2`, but `2.60.1` was already declared on this branch (it appeared in the merge resolution at `16df7dee`). `src/index.ts` reads the version at runtime, so MCP clients advertise whatever lands — merging publishes a lower semver than the branch already declared. Pick the number once via `npm run version:*` and make the PR body agree.

- **[src/tools/desktop/binder/bindTemplate.ts:119] `calcs[]` and `auto_apply` have no `.describe()`.** These are the two parameters the one-call derived-metric feature runs on, and the only place the model learns `calcs[]` exists is the route-table prose in server instructions. The file already carries the lesson four lines up ("Undescribed, this parameter cost 299 repeat binds and 2,562 seconds"). Also `calcs[].datatype` and `calcs[].role` are bare `z.string()` — nothing for the model to form a valid value from.

- **[src/tools/desktop/knowledge/searchKnowledge.ts:23] The new description names another tool** ("...do not call read-knowledge-resource"). No guarantee that tool is enabled in the active profile — this same PR drops `list-knowledge-resources` from `DYNAMIC_AUTHORING_TOOL_PROFILE` — or that a host doesn't have its own resource reader. Describe the behavior instead, or route the name through the `ToolName` type.

### Questions (your call)

- **[src/server.desktop.ts:363] The hand-rolled `tools/list`.** `getDesktopToolListEntry` imports `@modelcontextprotocol/sdk/server/zod-compat.js` and `.../zod-json-schema-compat.js` — subpaths that don't appear to be documented SDK surface — with two `as any` casts, under a `^1.26.0` caret range. Tests assert `$schema` is absent, but nothing asserts the emitted `inputSchema` equals what `McpServer.registerTool` would produce. If those ever disagree, we advertise one contract and validate another, silently. Is there a supported way to drop `$schema`, or is a schema-equivalence test the cheaper insurance?

- **[explicit-bind.ts:540 / validate.ts:418] Two `optionalFieldPrunesFor`.** Same function, differing only in the map's value type — and #661 had to make the same edit in both, with the explaining comment landing in only one copy. Neither file is lockstep-core, so nothing forces the split. It's fine, but this one already drifted.

- **[src/tools/desktop/tool.ts:161] The bind-first gate returns `isError: false`, before any events.** A refused `list-available-fields` comes back as a successful text result, and the gate runs ahead of `notifyInvocation` — so a gated call emits no tool-start/tool-end event and never reaches the episode stream the learning pipeline mines. Deliberate?

- **[src/server.desktop.ts:81] `SPEC_LOOP_TOOL_PROFILE` gained `get-worksheet-xml`,** and the comment was reworded from "NO XML tools" to "no XML authoring tools." That profile exists to test one hypothesis. Does an XML read still test it?

- **[src/desktop/route/route-state.ts:82] `recommended` is pinned to one scenario in the type.** `top_n: 10` and the revenue-like reason are literal types on a public interface; a cost-like recommendation needs a type change, not a value change. Deliberate scope?

- **[template-manifests/*.json] `scatterplot` in both scatter manifests.** ✅ **Already covered by a live smoke** on the #661 head: the one-word "scatterplot" ask binds deterministically in a single call to `correlation-scatter-plot-chart` at Region grain (no propose round-trip, `Customer Name` pruned), the two-word "scatter plot" ask did not regress, and an explicit connected-scatterplot ask still routes to `connected-scatterplot`. The unenumerated boundary (e.g. "scatterplot showing the path over time") remains unprobed.

### nits (optional polish)

- [knowledge/index.ts:318] `TOP_HIT_BODY_TRUNCATION_MARKER` hardcodes "6144" in prose right below `TOP_HIT_BODY_MAX_BYTES = 6_144`. Template-literal it.
- [classify.ts:~160] The `BUSINESS_FIELD_SYNONYMS.find(…'revenue')…captionPatterns` lookup runs three times in one function. Name it once.
- [nonPositiveFilterCount.ts:19] The XPath already selects `[@count]`, so the lowercase re-check can't fail; and the emitted `xpath` field reports the generic selector, not the offending node.
- [classify.ts:2275] `fieldNameMatchesSlot`/`normalizedName` duplicate explicit-bind.ts's pair by design (classify.ts is lockstep-core) — a one-line comment saying so would stop the next reader from "fixing" it into an import that breaks the gate.

### Fine as-is / noted tradeoffs

Stripping `annotations.title` while keeping tool-level `title`; `mapNetworkFailure` collapsing caller abort into `command-timed-out` (comment names the tradeoff); the self-correcting error shape on `nonPositiveFilterCountRule`; `hasConfidentTopHit` inspecting a runner-up even at `limit === 1` so truncation can't manufacture confidence.

### Notes outside the review lens

1. **CI was not green on `b23f48d2` at review time** — the green signal was from `a26da612`, two commits back. Re-check before merging.
2. The PR body's lockstep note is stale: `classify.ts`'s hash changed again in `b23f48d2`, so the a2td cross-repo byte match must be redone against the final tip.

mattcfilbert:
Hey @anyoung-tableau, thanks for the review via your review ghost. I agree re the learning pipeline adjustment; the xml stuff is fine there (we need it); and regarding the "out of scope" note, I am fine with accounting for the cost-like recommendation

Smoke testing already showed that one of your "please" concerns are already managed (see above) but your other higher level requests are noted and I am rolling with them now

PR #649, opened 2026-07-26
@W-23585136: Fix list-users omitting lastLogin and leaking non-requested fields

Alon-ST-DATA:
list-users sent fields=_all_ to Get Users on Site, which (1) on some sites returned a default set omitting lastLogin — and the client-side lastLogin filter then dropped every row, silently breaking user-license-reclamation-inform — and (2) pulled the expensive SSO/authSetting path, leaking authSetting, locale, language, externalAuthUserId into output.

Key gotcha: Tableau's `fields` param on this endpoint is "include at least", not a projection, so restricting it does not shape the response.

Fix:
- Plumb a `fields` query param through usersApi -> usersMethods; call site requests the lean set id,name,fullName,siteRole,email,lastLogin.
- Add a tool-local projection (projectLeanUser, a Pick of User to the 6 fields) applied AFTER filtering. Absent optionals are omitted, not null. Shared userSchema is left intact (update-user and others depend on it).
- Narrow the filter surface to the 6 fetched fields, so filtering a dropped field returns a clean enum error instead of silently matching nothing.
- Never-logged-in semantics: undefined lastLogin matches lt/lte, excluded from gt/gte/eq.
- Docs + tool description trimmed to the 6 fields; add *.env.list and loose report artifacts to .gitignore.

Tests: unit guard mocks the full Tableau response and asserts output is exactly the 6 keys; new e2e asserts no extras survive on any user. Version bump 3.6.0 -> 3.6.1.

the pull request._

# Pull Request Template

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/tools/web/users/usersFilterUtils.ts
+++ b/src/tools/web/users/usersFilterUtils.ts
@@ -10,18 +10,11 @@ import {
 // === Field and Operator Definitions ===
 // Client-side filtering for users (API doesn't support server-side filtering)
 
-const FilterFieldSchema = z.enum([
-  'id',
-  'name',
-  'siteRole',
-  'email',
-  'fullName',
-  'lastLogin',
-  'authSetting',
-  'locale',
-  'language',
-  'externalAuthUserId',
-]);
+// Only fields we actually request from Tableau (see listUsers.ts `fields`) are
+// filterable. Filtering on an un-fetched field would silently match nothing —
+// the same class of bug as the original lastLogin defect — so unsupported
+// fields are rejected at parse time with a clear validation error instead.
+const FilterFieldSchema = z.enum(['id', 'name', 'siteRole', 'email', 'fullName', 'lastLogin']);
 
 type FilterField = z.infer<typeof FilterFieldSchema>;
 
@@ -32,10 +25,6 @@ const allowedOperatorsByField: Record<FilterField, FilterOperator[]> = {
   email: ['eq', 'in'],
   fullName: ['eq', 'in'],
   lastLogin: ['eq', 'gt', 'gte', 'lt', 'lte'],
-  authSetting: ['eq', 'in'],
-  locale: ['eq', 'in'],
-  language: ['eq', 'in'],
-  externalAuthUserId: ['eq', 'in'],
 };
 
 const dateFields: Set<FilterField> = new Set(['lastLogin']);
@@ -112,14 +101,6 @@ function getFieldValue(user: User, field: FilterField): string | number | undefi
       return user.fullName;
     case 'lastLogin':
       return user.lastLogin;
-    case 'authSetting':
-      return user.authSetting;
-    case 'locale':
-      return user.locale;
-    case 'language':
-      return user.language;
-    case 'externalAuthUserId':
-      return user.externalAuthUserId;
     default:
       return undefined;
   }
@@ -132,6 +113,18 @@ function matchesFilter(
   field: FilterField,
 ): boolean {
   if (fieldValue === undefined || fieldValue === null) {
+    // A missing date field means the event never happened. For `lastLogin`,
+    // Tableau emits no value for users who have never signed in. Such users are
+    // the *most* inactive, and the user-license-reclamation prompt explicitly
+    // treats them as reclamation candidates, so an "older than X" filter must
+    // include them:
+    //   - lt / lte (older than / on-or-before X)  → MATCH  (never < any date)
+    //   - gt / gte (newer than / on-or-after X)    → NO MATCH (they logged in after nothing)
+    //   - eq (equals a specific date)              → NO MATCH (no date to equal)
+    // Non-date fields keep the original "missing → no match" behavior.
+    if (dateFields.has(field)) {
+      return operator === 'lt' || operator === 'lte';
+    }
     return false;
   }
 
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ coverage/
 test-logs/
 config.json
 env.list
+*.env.list
 manifest.json
 sea-prep.blob
 tableau-mcp
@@ -41,3 +42,7 @@ src/web/apps/dist/
 !.work/**/.gitkeep
 
 .worktrees/
+
+# Loose local report / probe artifacts (never commit)
+*.tsv
+stale-*.md
--- a/docs/docs/tools/users/list-users.md
+++ b/docs/docs/tools/users/list-users.md
@@ -4,7 +4,7 @@ sidebar_position: 1
 
 # List Users
 
-Retrieves a list of users on the Tableau site. Each user includes profile information such as site role, email, last login time, and authentication settings.
+Retrieves a list of users on the Tableau site. Each user includes profile information such as site role, email, full name, and last login time.
 
 :::warning[Admin Only]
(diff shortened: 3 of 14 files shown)


src/tools/web/users/usersFilterUtils.ts
```diff
@@ -132,6 +113,18 @@ function matchesFilter(
...
+    // Tableau emits no value for users who have never signed in. Such users are
+    // the *most* inactive, and the user-license-reclamation prompt explicitly
+    // treats them as reclamation candidates, so an "older than X" filter must
+    // include them:
+    //   - lt / lte (older than / on-or-before X)  → MATCH  (never < any date)
+    //   - gt / gte (newer than / on-or-after X)    → NO MATCH (they logged in after nothing)
+    //   - eq (equals a specific date)              → NO MATCH (no date to equal)
+    // Non-date fields keep the original "missing → no match" behavior.
+    if (dateFields.has(field)) {
+      return operator === 'lt' || operator === 'lte';
```
Akash-Rastogi, line 126: [Akash's agent] Nice catch on the "include at least" semantics — the `projectLeanUser` projection and the enum-narrowing of the filter surface are the right fix, and the never-logged-in matching here (undefined `lastLogin` → matches `lt`/`lte`, excluded from `gt`/`gte`/`eq`) is correct in isolation.

One downstream interaction worth resolving before this merges, since it lands in exactly the workflow this PR is fixing (`user-license-reclamation-inform`):

This line flips the old behavior — previously `undefined lastLogin` returned `false`, so `lastLogin:lt:<cutoff>` did **not** match never-logged-in users. That is precisely why `src/prompts/userLicenseReclamation/inform.ts` Step 1 makes a **second** `list-users` call to pick them up separately:

> "Then call `list-users` a second time with the same `siteRole` filter but **without** the `lastLogin` filter, and include only users whose `lastLogin` is empty/null (never signed in). These are also reclamation candidates…"

With the new semantics, **Call 1** (`lastLogin:lt:<cutoff>`) now already returns those null-`lastLogin` users, and **Call 2** re-adds them. The prompt gives no dedup instruction, so never-signed-in users appear twice and `total candidates = <count>` is inflated.

It's read-only and LLM-mediated (not dangerous, and a careful agent may dedup on its own), and the `-apply` prompt fetches differently so it's unaffected. But the cleanest fix is to update the inform prompt to match the new filter semantics — either drop the now-redundant second call (Call 1 covers never-logged-in), or add an explicit "dedup by user id" step. Not blocking the code change; flagging so the prompt and the filter stay in sync.

Alon-ST-DATA: [Alon's agent] Good catch — fixed in 5c508b22 (Option A: dropped the redundant second call).

You're exactly right: since this PR makes an undefined `lastLogin` match `lt`/`lte` (`usersFilterUtils` `matchesFilter`), the inform prompt's Call 1 (`siteRole:in:<roles>,lastLogin:lt:<cutoff>`) now already returns never-signed-in users. The old second call (same siteRole, no `lastLogin`, keep-only-null) re-added that exact set with no dedup → inflated "total candidates".

Changes in `src/prompts/userLicenseReclamation/inform.ts`:
- Removed the "call `list-users` a second time…" instruction; the single `lastLogin:lt` call now covers never-signed-in users directly (rendered as Days Inactive = "Never", still sorted to top in Step 3).
- Added a WHY comment above `listUsersFilter` explaining the `lt` semantics so a future editor doesn't re-add the second call.
- `inform.test.ts` now asserts the single-call contract (`not.toContain('a second time')` + exactly two `list-users` mentions), so a regression re-adding the call fails.

`-apply.ts` is unaffected — it fetches all users and determines inactivity client-side, never using a `lastLogin:lt` server filter.

Verified live against a test site: the Step-1 filter returns 975 never-signed-in Creator/Explorer rows with 975 unique ids and zero duplicates (the old two-call design would have counted 1950). Also removes a redundant `list-users` round-trip. Thanks for keeping the prompt and filter in sync.

Akash-Rastogi: [Akash's agent] Fixed in [`5c508b2`](https://github.com/tableau/tableau-mcp/commit/5c508b22c6485b9586275d41f3700c5b9632844c).

Verified on the new head: the redundant second call is gone, the WHY comment sits directly above `listUsersFilter`, and `inform.test.ts` guards the single-call contract. Confirmed the removed call lost no coverage — a never-signed-in user still satisfies both filter arms (`siteRole:in` on their populated role, `lastLogin:lt` on the undefined value), so Call 1 genuinely covers them. Nice touch verifying 975 unique ids live.

Akash-Rastogi (approved):
[Akash's agent] ## LGTM ✅

Re-reviewed at `5c508b2`. My one concern from the previous pass — the `user-license-reclamation-inform` double-count — is fully resolved, and the rest of the change holds up well.

### The original fix
The root cause is a genuinely subtle one: Tableau's `fields` param on *Get Users on Site* is **"include at least"**, not a projection, so restricting it shapes the request but not the response. Handling that with two independent mechanisms is the right call:
- `fields=id,name,fullName,siteRole,email,lastLogin` — stops relying on Tableau's default set (which silently omitted `lastLogin`) and avoids the expensive `_all_` SSO path.
- `projectLeanUser` — the actual output enforcement, since `userSchema` declares the extra keys as known optionals so Zod won't strip them. Correctly runs **after** filtering, and omits absent optionals rather than emitting `null`.
- Narrowing `FilterFieldSchema` to the 6 fetched fields converts a silent-empty failure into a clean enum error — same bug class as the original defect, now impossible.

`userSchema` left intact, so `update-user` and other consumers are unaffected. Verified nothing else in `src/` reads the dropped `authSetting` / `locale` / `language` / `externalAuthUserId` fields.

### The follow-up commit
Dropping the redundant second `list-users` call was the right option. I confirmed it loses no coverage: a never-signed-in user still satisfies **both** arms of the AND-combined filter — `siteRole:in` on their populated role, `lastLogin:lt` on the undefined value — so Call 1 covers them directly. The WHY comment above `listUsersFilter` and the `not.toContain('a second time')` assertion together make a regression hard to reintroduce. Independent live verification (975 rows / 975 unique ids vs. 1950 under the old design) matches my reading of the code.

### Validation run locally at `5c508b2`
| Check | Result |
| --- | --- |
| `npx tsc --noEmit` | ✅ clean |
| `npx eslint` (changed areas) | ✅ clean |
| `npm test` (full suite) | ✅ **2525 passed / 161 files** |
| `npm run build` | ✅ succeeds |
| Version bump | ✅ 3.6.0 → 3.6.1 (patch, correct for a bug fix) |

> **Note on e2e:** the 5 new `tests/e2e/users/listUsers.test.ts` cases are **skipped** in my environment — they need live Tableau credentials (`tests/.env`), and 23/27 e2e suites fail identically on my box without them. Not attributable to this PR, but worth a green CI e2e run before merge since the regression guard lives there.

Nice work on the never-logged-in semantics being spelled out explicitly (`lt`/`lte` match, `gt`/`gte`/`eq` exclude) and documented in the tool description — that ambiguity was the actual trap here.

PR #650, opened 2026-07-27
feat(templates): ship parameterized sort on the simple-bar and ranking templates

mattcfilbert:
## Decision

Ship the parameterized sort on `magnitude-simple-bar` and `ranking-ordered-column` in the release **after** v2.58.0. This follows the existing template-stamp rule: a template change ships only with a stamp re-earned by live proof. Both stamps were re-earned on 2026-07-26 with a live computed-sort rebind on each template, and the hashes were synced through the gate's own re-earn path (same procedure the waterfall re-stamp used). No new rule.

## Scope

Two template XMLs plus their stamp hashes (8 files). Deliberately based on `claude/v11-fixes` (frozen #643) so it lands in the next release PR, not inside the freeze.

Stamp hashes are never hand-edited. Any later change to these templates re-earns the stamp with a live rebind before shipping.

## Evidence / risk

Live computed-sort rebind proof for both templates is in the 2026-07-26 wave receipts; branch CI runs the full suite. Risk: the branch sits 11 commits behind the frozen tip — a local merge onto 14b1ec2c was clean (no overlap with the freeze-review fixes).

## What approving means

The two templates carry parameterized sort in the next release, under the existing stamp gate.

### DIFF
--- a/src/desktop/data/content-manifest.json
+++ b/src/desktop/data/content-manifest.json
@@ -132,8 +132,8 @@
     },
     {
       "path": "data-visualization-templates-xml/magnitude-simple-bar.xml",
-      "sha256": "ef0503faa2dd6320f902335dcd721909f4cb161d315a7746f296dc2967c60a4c",
-      "bytes": 2457
+      "sha256": "3ce7dfa8c7d8b6ebb54c53f0372589dce7e2c54d4e71fb17bdb59a840ef535b1",
+      "bytes": 2476
     },
     {
       "path": "data-visualization-templates-xml/pareto-chart.xml",
@@ -197,8 +197,8 @@
     },
     {
       "path": "data-visualization-templates-xml/ranking-ordered-column.xml",
-      "sha256": "a0a23ffcf5fb34263b4e1a46a107590200a03d3f02fdc5efcf8622cc9e4a9235",
-      "bytes": 2435
+      "sha256": "afb71d90510a3df77c64fa4549df37cc3285dd15b1d2db603c90c53ba5c69856",
+      "bytes": 2452
     },
     {
       "path": "data-visualization-templates-xml/slope-chart.xml",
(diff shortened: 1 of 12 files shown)


mattcfilbert:

Reviewing the incoming `computedSortTemplates.test.ts` change — the byte pins for `magnitude-simple-bar` and `ranking-ordered-column`. It is not pushed to this branch yet, so this is a read of the pending fix, not of the current head.

The pins themselves are right. I pulled the `<computed-sort>` node out of both template files at `6a0da4b3` and the expected strings are byte-identical to them. The file runs green (5 passed), and it would have failed at `HEAD~1`, where both templates still carried donor `[none:Category:nk]` refs. So it does catch the revert this branch undoes.

Two things before merge.

**Please — the header comment claims a guard the test does not give you (P2).**

`{{field_base_1}}` resolves positionally over the manifest's bindable slots (`fieldReferenceRewriter.ts:573`), so a slot reorder is an edit to `*.manifest.json`, not to the template XML. This test reads only the XML. A reorder leaves it green while the sort silently binds the dimension to itself — the exact failure the comment says the pin prevents.

What the pin does buy is a re-stamp guard, and that is worth having: `stampHash.invariant.test.ts` only proves the bytes match the recorded hash, so a re-stamp that drops the parameterization updates the hash and stays green. Either say that, or add the assertion that actually covers the reorder — one line per template on `loadManifests().get(name)!.slots.filter((s) => s.bindable !== false)[0].kind === 'categorical'`.

**Please — `part-to-whole-waterfall` is the one uncovered case, and it is the risky one (P2).**

Five templates use a positional `{{field_base_N}}` in their computed sort *and* a legacy concrete-field manifest: `magnitude-simple-bar`, `ranking-ordered-column`, `pareto-chart`, `deviation-spine-chart`, `part-to-whole-waterfall`. This change pins the first four. `ranking-ordered-bar` and `deviation-diverging-bar` do not need it — their manifests declare `{{field_base_N}}` as the `template_field`, so the token is owned rather than positional.

Waterfall's slot 1 is the measure (`Profit`) and slot 2 is the dimension (`Sub-Category`), so its sort reads:

```
<computed-sort column='[{{DATASOURCE}}].[none:{{field_base_2}}:nk]' direction='DESC' using='[{{DATASOURCE}}].[sum:{{field_base_1}}:qk]' />
```

Inverted against every other template, which makes it the one most likely to get "fixed" wrong. The comment this change replaces was the only place that tracked it. Adding it is one row in the `it.each` table.

Verified locally at `6a0da4b3` with the pending change applied: `src/desktop/templates/` plus `src/desktop/binder/stampHash.invariant.test.ts` — 281 passed, 1 skipped.

PR #652, opened 2026-07-27
feat(pulse): support absolute/custom time windows in insight bundles

nawartamawi:
## What

Adds two optional fields to the `generate-pulse-metric-value-insight-bundle` request so a bundle can target a period other than a today-relative window. Both optional; existing today-relative payloads are unaffected.

- **`options.now`** — anchors the analysis to a specific date (`YYYY-MM-DD`, or `YYYY-MM-DD HH:MM:SS`), so a past whole period (e.g. a prior month) is reachable instead of only "this period" / "last complete period". Already supported by the insights service (used in its `measurement_periods` example); this brings the bundle request to parity with the brief-request schema, which already carried `now`. When set, the service ignores `time_zone`.
- **`measurement_period.specific_period` `{ date, end_date? }`** — an explicit period or an arbitrary date span. The service reads it **only when `range` is `RANGE_BY_CONFIG`** (verified in the service inbound range conversion); it is ignored for `RANGE_CURRENT_PARTIAL` / `RANGE_LAST_COMPLETE`. `date` selects the period at the chosen granularity containing that day; optional `end_date` extends the window through the end of `end_date`s period. This is the only way to express a custom span (e.g. Apr 15-20) that no relative range can.

The tool description documents the `range`/`now` interaction and the `RANGE_BY_CONFIG` requirement for `specific_period`.

## Why

Without these, the bundle can only analyze windows relative to today. Users routinely ask about a named past month ("May 2026") or a custom range ("April 15-20"); today those get a wrong "not possible" answer. Both mechanisms are already supported by the insights service — this exposes them through the MCP tool.

## Testing

- `tsc --noEmit`: clean.
- `eslint` on the two changed files: clean.
- `vitest` pulse suites: 224/224 pass.
- Field-level behavior verified against the insights-service internals (range to specific_period coupling, `now` semantics). Live REST round-trip for `specific_period` still to be confirmed on a real Cloud tenant.

Minor version bump 2.57.1 -> 2.58.0 (new optional fields).


src/sdks/tableau/types/pulse.ts
```diff
@@ -50,6 +50,12 @@ export const pulseMetricSpecificationSchema = z.object({
   measurement_period: z.object({
     granularity: z.string(),
     range: z.string(),
+    specific_period: z
```
tonythebest, line 60: `specific_period` is accepted even when `range` is `RANGE_LAST_COMPLETE` or `RANGE_CURRENT_PARTIAL`, although the tool contract says the service silently ignores it in that shape. I reproduced this at the PR head: schema parsing succeeds and `validateBundleRequest` returns `null`. That can turn an April 15–20 request into a plausible insight for the wrong period. Please fail preflight whenever `specific_period` is present and `range !== RANGE_BY_CONFIG`, and add a regression test proving the REST call is not made.

src/sdks/tableau/types/pulse.ts
```diff
@@ -350,6 +356,7 @@ export const pulseBundleRequestSchema = z.object({
       time_zone: z.string(),
       language: languageEnumSchema,
       locale: localeEnumSchema,
+      now: z.string().optional(),
```
tonythebest: Could we use the existing `tableauDateTimeSchema` here, as the brief request already does for the same `now` field? The current `z.string()` accepts values such as `not-a-date`, contrary to the documented format. `specific_period.date` and `end_date` likewise need a non-empty date-only schema plus an `end_date >= date` check. Please add valid/invalid schema cases and an exact-forwarding tool test for these new fields.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
```diff
@@ -41,6 +41,10 @@ Generate an insight bundle for the current aggregated value for Pulse Metric usi
     - time_zone: 'UTC'
     - language: 'LANGUAGE_EN_US'
     - locale: 'LOCALE_EN_US'
+    - now (optional): anchors the analysis to a specific date instead of the current date, so the bundle can target a past whole period rather than only today-relative windows. Format 'YYYY-MM-DD' (or 'YYYY-MM-DD HH:MM:SS', 24-hr). To analyze a specific past month, set \`now\` to the LAST day of that month with \`measurement_period.range = 'RANGE_CURRENT_PARTIAL'\` and \`granularity = 'GRANULARITY_BY_MONTH'\` (e.g. \`now: '2026-05-31'\` yields the full month of May 2026); note that \`RANGE_LAST_COMPLETE\` targets the period BEFORE the anchor's period, not the anchor's own. This shifts only the whole-period window; for an arbitrary date range use \`specific_period\` (below). When set, \`time_zone\` is ignored. Omit it (or use an empty string) for today-relative analysis.
+- To target an explicit period or an arbitrary date range, set \`measurement_period.range\` to \`'RANGE_BY_CONFIG'\` and provide \`measurement_period.specific_period\`. This is the ONLY way to analyze a custom span (e.g. April 15-20); \`specific_period\` is IGNORED unless range is \`'RANGE_BY_CONFIG'\`.
```
tonythebest: This bullet starts at column zero, so it closes the `bundleRequest` section and makes the existing datasource guidance render underneath `specific_period`. Please indent it as a sibling of `now` under `bundleRequest`, keeping the datasource requirements at the same level.

tonythebest (changes requested):
The additive time-window direction and minor version bump look sound, but the current contract can silently return an insight for the wrong period. I reproduced `RANGE_LAST_COMPLETE + specific_period` passing both schema and preflight validation, even though the service ignores `specific_period` in that shape. Please enforce the coupling, validate the new date fields, and add focused schema/preflight/forwarding tests before approval. Also please confirm and test the downstream precedence if `now` and `specific_period` are both supplied. CI, local typecheck, and lint are otherwise clean.

src/sdks/tableau/types/pulse.ts
nawartamawi: Fixed in e90eb7ea. validateBundleRequest now fails preflight when specific_period is present and range !== RANGE_BY_CONFIG, with a regression test proving the REST call is not made (validatePulsePayload.test.ts).

On the now-vs-specific_period precedence you flagged: since the combined behavior at the service is unconfirmed (my Testing note calls out the live specific_period round-trip is still pending), I chose to fail closed rather than encode a guess — preflight now also rejects setting BOTH options.now and specific_period, so a request can't silently depend on undefined precedence. Caller picks one. Test added. If the service turns out to have well-defined precedence, we can relax this to a documented rule in a follow-up.

src/sdks/tableau/types/pulse.ts
nawartamawi: Done in e90eb7ea. options.now now reuses tableauDateTimeSchema (matching the brief request) instead of the bare z.string(). Added a date-only tableauDateSchema for specific_period.date/end_date, plus an end_date >= date refine. Schema valid/invalid cases (date + datetime + empty now; malformed/empty date; end_date < date) are in pulse.test.ts, and exact-forwarding tool tests for both new fields are in generatePulseMetricValueInsightBundleTool.test.ts.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
nawartamawi: Fixed in e90eb7ea — specific_period is now a `    - ` sibling of now under bundleRequest, its two format examples nest under it, and the datasource guidance is back at the bundleRequest level (no longer rendered under specific_period).

tonythebest (approved):
Approving to supersede my earlier `Changes requested` review — I see this already merged (5a03d5ea) at head `e90eb7ea`, so this is a post-merge confirmation for the record.

The fix commit resolves all four points from the prior review, verified by execution in an isolated worktree at `e90eb7ea` under Node 22 (`tsc --noEmit` clean, ESLint clean on the changed files, full pulse suite green):

1. **Silent `specific_period` ignore** — `validateBundleRequest` now rejects `specific_period` when `range !== 'RANGE_BY_CONFIG'`. Reproduced: `RANGE_LAST_COMPLETE + specific_period` fails preflight.
2. **Unvalidated date fields** — `options.now` now uses `tableauDateTimeSchema`; `specific_period.date`/`end_date` use a date-only schema with an `end_date >= date` refinement. Reproduced: `now: "not-a-date"` and inverted spans are rejected at parse.
3. **No focused tests** — new schema, preflight (including `now`+`specific_period` both-set precedence), and verbatim-forwarding tests cover the contract.
4. **Markdown nesting** — `specific_period` is now a sibling of `now`, so datasource guidance renders at the `bundleRequest` level again.

Thanks for the quick turnaround on the fixes.

PR #654, opened 2026-07-27
feat(desktop): refuse filter and top-N limits below 1 before they reach Desktop

mattcfilbert:
## Decision
A rule we now keep: preflight refuses a non-positive filter limit only where it is actually a filter limit — a top-level `<groupfilter>` with a `count` attribute. Nested-in-`<group>` shapes pass, because Desktop itself persists `<groupfilter count='0'>` inside `<group>` for legacy empty sets and round-trips it; the earlier unscoped rule refused apply-worksheet, bind-template, and dashboard-auto-apply on any workbook already holding that shape.

## Scope
- Rule scoped to `//groupfilter[not(ancestor::group)][@count]`, with a nested-shape regression test and the emitted xpath pointer matching the real selector.
- Removed from this PR: the author-set `count<1` guard (conflicted with `claude/authorset-empty-mode`; #657's `count is required in top-n mode` covers the missing case), the addField workbook-membership rider (refused three working call shapes; returns as its own PR), and two unreachable belt-and-braces guards whose zod schemas already enforce positivity.
- Named follow-ups, deliberately not here: positivity on author-set top-n `count` (#657 checks presence, not positivity — `count:'0'` silently applies a set that resolves nothing; not a wedge, Desktop accepts it), and the addField rider re-proposed with hidden-column/caption-qualified/empty-datasource cases handled.

The nested-group carve-out: Desktop-persisted legacy `count='0'` inside `<group>` must keep passing validation. The AC6CC624 premise at the XML-apply layer remains unverified — the rule guards the documented interactive-modal case without blocking shapes the product accepts.

## Evidence / risk
Merge conflict with `claude/authorset-empty-mode` is gone (`merge-tree` clean). Review ran the XPath against the real Superstore fixture, 116 repo fixtures, and 193 local workbooks: top-level case still refuses, zero false positives. Suite 6539 tests / 0 failures, tsc and lint clean, run firsthand. Risk: no live apply-layer probe of the modal trigger.

## Approving this means
Workbooks Desktop itself wrote stop being refused at preflight, while authored non-positive filter limits are still caught before they reach Desktop.

### DIFF
--- a/src/desktop/commands/workbook/filterCountPreflight.test.ts
+++ b/src/desktop/commands/workbook/filterCountPreflight.test.ts
@@ -1,0 +1,96 @@
+import invariant from '../../../utils/invariant.js';
+import { ToolExecutor } from '../../toolExecutor/toolExecutor.js';
+import { loadDashboardXml } from './loadDashboardXml.js';
+import { loadWorkbookXml } from './loadWorkbookXml.js';
+import { loadWorksheetXml } from './loadWorksheetXml.js';
+
+const NO_FOCUS = { navigate: 'none', reason: 'intermediate-leg' } as const;
+const mockSignal = new AbortController().signal;
+
+function expectFilterCountIssue(
+  result:
+    | Awaited<ReturnType<typeof loadWorkbookXml>>
+    | Awaited<ReturnType<typeof loadWorksheetXml>>
+    | Awaited<ReturnType<typeof loadDashboardXml>>,
+): void {
+  expect(result.isErr()).toBe(true);
+  invariant(result.isErr());
+  invariant(
+    result.error.type === 'load-workbook-xml-error' ||
+      result.error.type === 'load-worksheet-xml-error' ||
+      result.error.type === 'load-dashboard-xml-error',
+  );
+  invariant(result.error.error.type === 'validation-failed');
+  expect(result.error.error.issues).toEqual(
+    expect.arrayContaining([
+      expect.objectContaining({
+        ruleId: 'non-positive-filter-count',
+        severity: 'error',
+        message: expect.stringContaining('AC6CC624'),
+      }),
+    ]),
+  );
+}
+
+describe('non-positive filter count apply preflight', () => {
+  it('blocks whole-workbook XML before dispatch', async () => {
+    const applyWorkbookDocument = vi.fn();
+    const executor = { applyWorkbookDocument } as unknown as ToolExecutor;
+    const xml = `<workbook>
+      <worksheets>
+        <worksheet name="Sheet 1">
+          <table><view><filter><groupfilter function="end" count="0" /></filter></view></table>
+        </worksheet>
+      </worksheets>
+      <windows><window class="worksheet" name="Sheet 1" /></windows>
+    </workbook>`;
+
+    const result = await loadWorkbookXml({
+      xml,
+      executor,
+      signal: mockSignal,
+      focus: NO_FOCUS,
+    });
+
+    expectFilterCountIssue(result);
+    expect(applyWorkbookDocument).not.toHaveBeenCalled();
+  });
+
+  it('blocks worksheet XML before dispatch', async () => {
+    const applyWorkbookDocument = vi.fn();
+    const executor = { applyWorkbookDocument } as unknown as ToolExecutor;
+    const xml = `<worksheet name="Sheet 1">
+      <table><view><filter><groupfilter function="end" count="-1" /></filter></view></table>
+    </worksheet>`;
+
+    const result = await loadWorksheetXml({
+      worksheetName: 'Sheet 1',
+      xml,
+      executor,
+      signal: mockSignal,
+      focus: NO_FOCUS,
+    });
+
+    expectFilterCountIssue(result);
+    expect(applyWorkbookDocument).not.toHaveBeenCalled();
+  });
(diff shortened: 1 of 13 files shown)


mattcfilbert:
## Blocker found post-review: this guard conflicts with the set-action drilldown knowledge in #651

**The conflict.** `author-set` now hard-rejects `count < 1`. But `count='0'` is how an agent creates the *initially empty* set that our own knowledge doc requires. Verified in tonight's live ww4 run — the readback contains:

```xml
<groupfilter count='0' end='top' function='end' units='records' user:ui-marker='end' user:ui-top-by-field='true'>
```

`data/knowledge/tactics/data/set-action-drilldown.md` step 1 still says "**`author-set` first:** create an initially empty set". `author-set` has no empty-set mode: `caption`, `dimension`, `orderBy`, and `count` are all required, and it only emits a computed Top/Bottom-N `<group>`. A zero count was the only available spelling.

**Consequence if #651 and #654 both land as-is:** the drilldown ask fails earlier than before — `author-set` refuses, so `author-action`'s new set mode (#655) has no set to target.

**What I'd do instead** (keeping the modal fix, which is real — AC6CC624 cost 582s of live eval wall):
1. Keep the preflight XML rule and the refine/bind/proposal guards in this PR. Those close the modal path.
2. Give `author-set` an explicit empty-set mode that emits the shape Tableau itself writes for an empty set: `<groupfilter function='empty-level' member='[Field]' user:ui-enumeration='inclusive' user:ui-marker='enumerate'/>`. Live-confirmed via /v0 apply + readback on build 0000.26.0724.1117.
3. Only then reject `count < 1` on the Top-N path, where a zero limit is genuinely meaningless.

Until (2) exists, the `count < 1` rejection in `author-set` should come out of this PR, or land behind (2) in the same release.

mattcfilbert:

This reviews the **incoming rebuild**, not what is pushed. I read the working tree on top of `cfaf49fa`; the fixes are not on the branch yet.

The rebuild does what the HOLD asked for. `authorSet.ts` and `addField.ts` are now byte-identical to `14b1ec2c`, the two unreachable guards in `bindTemplate.ts` and `refineWorksheet.ts` are gone, and the preflight rule is scoped to top-level `groupfilter`. The conflict is gone too: `git merge-tree HEAD origin/claude/authorset-empty-mode` conflicts in `authorSet.ts`, the rebuilt tree merges clean. Full suite (368 files, 6207 tests), `tsc --noEmit`, and eslint all pass on my run.

Three things before it lands.

### P2 — the test that justifies the carve-out uses the wrong XML

`src/desktop/validation/rules/nonPositiveFilterCount.test.ts:32` is named "allows an empty set persisted as a group with groupfilter count=0", but the fixture is a Top-N groupfilter (`function="end" end="top" count="0"`). Tableau's real empty set — live-confirmed on build 0000.26.0724.1117, and what #657 now emits — is `<groupfilter function='empty-level' member='[Field]' .../>` with no `count` attribute at all. The rule never matched it, predicate or no predicate.

So the carve-out is not there for empty sets. It is there for the legacy `count='0'` hack agents wrote before #657 existed, which still sits in workbooks on disk — and since this rule validates the whole live workbook on every apply, one stale workbook wedges every apply. That is a good reason, and the test should say it. As written, the next reader checks the real empty-set shape, finds it does not need the predicate, and deletes it.

I ran the branch XPath against the real Superstore fixture in the repo rather than a hand-written one:

| shape | `//groupfilter[not(ancestor::group)][@count]` |
|---|---|
| real worksheet top-N, count changed to 0 | matched — still refuses |
| same block nested in an author-set `<group>` | skipped |
| #657 live empty-set shape | skipped (no `count` attribute) |
| real fixture untouched (`count='5'`) | no match |

### P2 — the `xpath` we hand back no longer describes the rule

`nonPositiveFilterCount.ts:56` still reports `//*[contains(local-name(), 'filter')][@count]`. That string is the agent's "where to fix this" pointer, and it now points at nodes the rule skips on purpose. It prints verbatim in the preflight test log. Change it to `//groupfilter[not(ancestor::group)][@count]`.

### please — the author-set guard was deleted, not moved

`claude/authorset-empty-mode` carries a note in the top-n block of `authorSet.ts`: "PR #654: `numericFilterLimit(count)` and `resolveParameterFilterLimit(liveXml, count)` guards belong INSIDE this `mode === 'top-n'` block; empty sets have no count." The rebuild deletes both helpers and both call sites, so neither branch puts them back. #657's `count is required in top-n mode` checks that count is present, not that it is positive, so `author-set(mode:'top-n', count:'0')` will apply a set that can never resolve a member. The live readback shows Desktop accepts that without a modal, so it is a silent wrong answer, not a wedge — fine to leave out of this PR, but say so in the body and open the follow-up so it reads as a decision and not a drop.

While editing the body: it still calls the author-set literal guard "genuinely load-bearing". The live readback in my earlier comment shows a group-nested `count='0'` applies cleanly, so that line is now wrong and will tempt someone to re-add the guard.

### nit

`elementName()` and `countAttributes()` (`nonPositiveFilterCount.ts:15-24`) are leftovers from the broad selector. `elementName` can only return `groupfilter` now, and `countAttributes` does a case-insensitive local-name match that the exact `[@count]` predicate can never feed.

Verdict: merge after the two P2s. The scoping decision itself is right.

PR #657, opened 2026-07-27
feat(desktop): let author-set create an initially empty set

mattcfilbert:
## Decision
An empty set is authored explicitly, in a new `mode: 'empty'`, as the shape Tableau itself writes. This is a rule we keep: the tool emits the product's own XML for an empty set rather than leaving agents to improvise one.

**Read this together with #654.** #654 rejects a filter `count` below 1 — correctly, because a zero limit pops blocking modal AC6CC624. But `count='0'` was the only spelling available for the *initially empty set* that our own drilldown knowledge (#651) tells agents to create first, and the live ww4 eval workbook proves agents used it: its readback contains `<groupfilter count='0' end='top' function='end' …>`. Ship #654 without this and the drilldown ask stops failing quietly and starts failing hard, at step 1.

## Scope
- New `mode` param, default `'top-n'` — existing callers are unaffected. `orderBy` and `count` become optional, required only for `top-n`.
- `empty` mode emits `<groupfilter function='empty-level' member='[Field]' …/>` inside a `<group>` marked `user:ui-builder='filter-group'`.
- Deliberately out: any change to the Top-N path's XML, and any change to #654's guard.

The `user:ui-builder='filter-group'` marker. `author-action`'s set mode resolves a target set by that attribute and nothing else, so a set authored without it is invisible to the action that needs it — and an ad-hoc Tableau group can never be mistaken for a set. Both halves of that contract depend on the marker staying put.

## Base
Branched off `claude/setaction-mode` (#655), not the release base, because it depends on that PR's `resolveTargetSet`. Merge #655 first.

## Evidence
The emitted XML is the shape Tableau writes, confirmed by applying it through `/v0` on build `0000.26.0724.1117` and reading it back byte-faithfully. Sol-built, Opus-reviewed. Full suite 366 files / 6212 tests, `tsc --noEmit`, lint, and desktop build all green — run firsthand, not taken from the builder's report. Repo CI does not run tests on PRs into `claude/*` bases, so local runs are the only signal here.

## Risk
No live end-to-end proof that a click expands the set — that needs #651's knowledge, #655's action mode, and this change all present in one served build, then an eval leg. This PR is the last of the three pieces.

## Approving this means
The drilldown pattern our knowledge prescribes becomes buildable, and #654 can ship without breaking it.

### DIFF
--- a/resources/desktop/knowledge/_index.md
+++ b/resources/desktop/knowledge/_index.md
@@ -76,6 +76,7 @@ Entries show the tactics slug and, where one exists, its strategy companion.
 - Parameter actions (click a mark to set a parameter) → `tactics/dashboard/parameter-actions`
 - Parameter-driven views (a parameter that reshapes what the viz shows) → `tactics/dashboard/parameter-driven-views`
 - Dynamic dashboards on a live session (author-parameter/author-set/author-action/format-labels verb routing, OPEN-vs-MERGE law) → `tactics/data/dynamic-dashboard-authoring`
+- Parameter-free click-to-expand drilldown (`+` → Category label, selected Category → Sub-Categories) via set + on-select set action → `tactics/data/set-action-drilldown`
 - Sizing modes / container tree (XML) → `tactics/dashboard/dashboard-layout-structure`
 - Layout patterns, content placement, archetypes (strategy) → `strategy/dashboard-design/*`
 - Layout & actions decisions (strategy) → `strategy/dashboard-design/dashboard-layout-and-actions`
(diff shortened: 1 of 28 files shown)


mattcfilbert:
## Review findings applied (`ad41de2b`)

**Readback now checks the marker.** The reviewer's strongest point: the readback verified only `name` and `caption`, while this PR's whole premise is that `user:ui-builder='filter-group'` is what lets `author-action` see the group as a set. If Tableau reloaded the group without it, `author-set` reported success and the follow-up action then failed to find its target — a hollow success in exactly the shape this stack exists to remove. Empty mode now asserts the marker in the readback, with an error that says the name and caption survived while the marker did not, rather than claiming the apply failed.

**The #654 seam is typed and commented.** After this change `count` is `string | undefined`, so a naive merge with #654 either breaks `tsc` or drops a guard. #654's two count checks are now pinned inside the `mode === 'top-n'` block with a comment naming the PR. Merge in either order without losing the guard.

**Provenance restored** — the WW2021W44 golden pointer for the top-N shape is back in the header.

Verified firsthand after the fixes: full suite 366 files / 6214 tests, `tsc --noEmit`, lint, and `build:desktop` all clean. Repo CI does not run tests on PRs into `claude/*` bases, so those local runs are the signal.

## One open question, deliberately not changed here

`mode:'empty'` with a `count` or `orderBy` silently discards them and reports success. Both the reviewer and the builder land on refusing that combination, and I agree — it is the same principle as the rest of this stack: fail loudly instead of dropping intent. I have not changed it in this PR because it is a behavior break for any caller that fills every parameter, and it belongs in its own change with its own test. Filing as follow-up unless a reviewer wants it here.

## Merge order

#655 first (this stacks on it), then this, then #654. #654 must not land before this one — its `count < 1` rejection removes the only spelling an agent had for an empty set until this PR exists.

PR #658, opened 2026-07-27
docs(knowledge): catalog the errors that arrive as blocking modals

mattcfilbert:
## Decision
Blocking-dialog error codes get written down, with their triggers and guards, in one place agents can read. This is a rule we now keep: when a modal wedges a session, the code and its trigger land in this file instead of in someone's memory.

The load-bearing sentence in the file: **there is no command that dismisses a dialog, so prevention is the only recovery.** The shipped command reference flags 18 commands that open a blocking dialog and contains none that closes one. A modal owns Desktop's single UI thread, so an agent that opens one has lost the session until a human clicks.

## Scope
Ten error codes we have hit this month. Four with verified triggers and guards:
- `5F1F5407` Qualified Name Parse Error — apply against a datasource built from a UTF-16LE, tab-delimited CSV with parentheses in the filename. Reads and calc authoring work; only apply fails.
- `AC6CC624` — a filter or Top-N limit of 0 or less.
- `CDEAC1A9` — a parameter action aimed at a set instead of a parameter.
- `47BF7751` — raw `goto-sheet` with a name that is not in the live inventory.

The other six are listed with the trigger marked UNKNOWN. That is deliberate: a shared error code does not prove a shared cause, and a guessed trigger in a tactics file is worse than an admitted gap.

Deliberately out: any code change. This is documentation of behavior that already exists.

The UNKNOWN rows stay UNKNOWN until someone reproduces them. Filling one in requires the same standard as the four above — a repro, the build number, and what did not work alongside what did.

## Evidence
Every verified trigger was reproduced live this month during eval runs on build `0000.26.0724.1117`; `5F1F5407` twice on two fresh instances, with a two-file repro pair in `agent-to-tableau-desktop`. The 18-command count comes from the shipped command reference in this repo, searched for Dismiss, Cancel, Escape, CloseDialog, and CloseModal. Full suite 366 files / 6191 tests, `tsc --noEmit`, and lint green — run firsthand.

## Approving this means
An agent that is about to author a filter, an action, or a navigation can find out beforehand which of those will hang the session, and a human triaging a wedged run can start from a code rather than from scratch.

### DIFF
--- a/resources/desktop/knowledge/_index.md
+++ b/resources/desktop/knowledge/_index.md
@@ -95,6 +95,7 @@ Entries show the tactics slug and, where one exists, its strategy companion.
 
 ### "Something went wrong / governance / tooling"
 - Recovering from a failed apply (MCP) → `tactics/workflow/recovery` · general troubleshooting: `strategy/workflow/troubleshooting-workbooks`
+- Blocking Desktop error dialogs (10-code catalog, prevention guards, no headless dismissal) → `tactics/workflow/errors-as-modals`
 - Field/datasource "not found" after a user change = stale cache (refresh with a live `session` first, don't declare Tableau unreachable); honor the `HOST VERIFICATION` receipt before claiming success → `tactics/workflow/failure-recovery-honesty`
 - Python helper templates → `tactics/workflow/python-helpers` · tool-selection strategy: `strategy/workflow/automation-tool-selection`
 - Template injection workflow → `tactics/workflow/templates`
--- a/resources/desktop/knowledge/tactics/workflow/errors-as-modals.md
+++ b/resources/desktop/knowledge/tactics/workflow/errors-as-modals.md
@@ -27,8 +27,10 @@ command after the dialog appears.
    of at least 1, and resolve action targets to the correct Tableau object type.
 2. Prefer guarded authoring tools over raw `tabdoc:` commands. The `activate-sheet`
    tool fresh-reads the workbook and refuses an unknown target before navigation.
-3. Treat a modal-risk failure as terminal for unattended work. Stop retries, report
-   the exact error code and operation, and ask a human to dismiss the dialog.
+3. Treat a modal-risk failure as terminal for unattended `/v0` work. Stop retries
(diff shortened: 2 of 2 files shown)


mattcfilbert:
## Review findings applied (`93237de5`)

**The 18-command figure was the important catch, and it contradicted our own knowledge.** `tactics/data/dialog-command-misclassification.md` already documents that `opens_blocking_dialog` is a known false negative for a whole class — 16 more commands (15 `*DialogCommand`-sourced plus `tabdoc:revert-workbook-ui`) are marked `opens_blocking_dialog: false` **and** `agent_can_invoke: true` while blocking in practice, one live-proven on 2026-07-19 (`tabdoc:edit-existing-parameter` hung a session 12s while `/v0/health` kept answering OK). Citing 18 as authoritative was the exact mistake that older file exists to prevent. The catalog now says 18 is the count the reference *flags*, calls it a floor, puts the real known set at 34 or more, and cross-links the misclassification file.

**Two guard cells described enforcement that is not on this branch.** The below-1 filter invariant and the `<edit-group-action>` set-action requirement are both still open PRs (#654, #655). An agent reading this file today would have trusted guards that are not there. Both now read as required behavior rather than shipped fact.

**Two UNKNOWN rows recovered.** The reviewer found receipted triggers in session evidence for `CEED3E34` (invalid calculation derivation strings during a Gantt build) and `F1E7F185` (a new-worksheet name passed as the literal string). Both are filled in and explicitly marked receipted-from-session-evidence rather than newly reproduced. Four codes stay UNKNOWN, which is still the honest answer for them.

**Cross-linked `tactics/workflow/recovery`** so the two files stop teaching different postures — the escalation ladder lives there, the prevention catalog lives here.

Verified firsthand after the fixes: full suite 366 files / 6191 tests, `tsc --noEmit`, and lint clean, including the tool-token test that rejects unregistered hyphenated tokens in backticks.

## Sequencing

This is documentation only and can merge any time, but it reads best after #654 and #655 land, at which point the two "required behavior" cells can go back to stating shipped fact.

PR #671, opened 2026-07-28
@W-23600028: bound list-users `limit` to post-filter matches

Alon-ST-DATA:
`list-users` applied `limit` to the FETCH in `paginateWithMetadata`, which ran BEFORE the client-side `applyUserFilters`. On large sites the first `limit` fetched rows were often active admins/service accounts that fail an inactivity filter, so e.g. `limit:5` + `lastLogin:lt:<recent-cutoff>` returned 0 users while thousands matched — silently under-returning.

Fix: add an opt-in `filterFn` predicate to `paginateWithMetadata` so the loop pages until it has collected `limit + 1` matching rows (or the site is exhausted), then truncates. `limit` now bounds POST-filter matches. The no-filter path is unchanged (predicate omitted -> prior behavior, byte-for-byte).

- Extract `buildUserFilterPredicate`; `applyUserFilters` becomes a thin wrapper.
- Truncation is reported via `mcp.resultInfo.truncated`/`truncationReason` reflecting post-filter truncation (reuses the list-flows truncation helper).
- Docs + tool description corrected: `limit` bounds filtered matches; a zero-match filter returns a plain message (no `resultInfo`).
- Unit + e2e coverage for the repro, never-logged-in semantics, and a guard that `limit` no longer leaks into the fetch.

Patch bump 3.6.1 -> 3.6.2.

the pull request._

# Pull Request Template

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/docs/docs/tools/users/list-users.md
+++ b/docs/docs/tools/users/list-users.md
@@ -47,8 +47,8 @@ See also: [Environment Variables](../../configuration/mcp-config/env-vars.md)
 | Argument | Type | Required | Description |
 |----------|------|----------|-------------|
 | `filter` | string | No | Client-side filter string with format `field:operator:value`. Multiple filters are comma-separated (AND logic). |
-| `pageSize` | number | No | Number of results per page (client-side pagination after filtering) |
-| `limit` | number | No | Maximum total results to return (client-side limit after filtering) |
+| `pageSize` | number | No | Number of users to fetch from the API per page (default 100, max 1000) |
+| `limit` | number | No | Maximum number of **matching** users to return. `limit` bounds results **after** `filter` is applied — the tool keeps paging until it has `limit` filter-matches (or the site is exhausted), so `limit:5` with an inactivity filter returns the first 5 matching users, never 5 pre-filter rows that all get filtered away. |
 
 :::note[API Limitation]
 The Tableau REST API does not support server-side filtering or pagination for users. All users are fetched and filtering is performed client-side by this tool.
@@ -95,7 +95,9 @@ Common values for `siteRole`:
 
 ## Response structure
 
-Each user includes:
+Returns a JSON object `{ users: [...], mcp: { resultInfo: {...} } }`.
+
+Each user in `users` includes:
 
 - `id` – user ID (LUID)
 - `name` – username (login name)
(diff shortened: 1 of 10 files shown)


Akash-Rastogi (approved):
[Akash's agent]

## LGTM ✅

Reviewed adversarially against `a0c570948`; ran the full local gate on the head commit (`77fb164`): **`tsc --noEmit`, `eslint`, `npm test` (2568 passed / 162 files), and `npm run build` are all green.**

This is a clean, well-scoped fix for the W-23600028 defect (`limit` was applied to the *fetch* before the client-side filter, so a small `limit` + a selective filter silently returned 0 on large sites).

### What I verified

- **No-filter path is byte-for-byte unchanged.** The opt-in `filterFn` branch is additive; the legacy branch (used by `list-flows`, `list-flow-runs`, `list-projects`, `list-datasources`, `list-workbooks`, `list-custom-views`, `list-views`, admin-insights) is untouched, and `list-users` is the *only* caller that passes `filterFn`. The dedicated "legacy behavior unchanged" test block locks this invariant in.
- **Truncation signal is honest in both directions.** The `limit + 1` sentinel means it never false-positives (exactly `limit` site-wide matches → loop exhausts and reports `truncated:false`) and never false-negatives (it stops early *only* when a `limit+1`-th match proves more exist). The `truncatedByLimit` boundary cases are covered.
- **Loop terminates safely.** `rawFetched` strictly advances toward `totalAvailable`; the empty-page `throw` guards the "server claims more but returns none" case, matching the pre-existing path. The reject-everything filter exhausts pages cleanly (no infinite loop).
- **Semantics preserved.** Never-logged-in users match `lt`/`lte` and count toward `limit`; `requested-limit` vs `admin-cap` classification reuses `buildTruncationInfo` as a single source of truth; a zero-match filter takes the empty-result path (no spurious truncation).
- **`buildUserFilterPredicate` extraction is safe.** `applyUserFilters` is now a thin wrapper over it, and the `predicate ⇄ wrapper` equivalence tests prove behavior is identical across a representative filter matrix.
- **Security.** Admin-gated (`assertAdmin` runs before any paging), read-only, filter values are client-side comparisons only (no injection vector), and error/result payloads leak no PII.

### Non-blocking note (FYI, no change requested)

When matches are sparse and `limit` can't be reached, the loop pages the entire site to *honestly* confirm `truncated:false`. That's the correct accuracy-over-cost tradeoff for this signal, and it's already documented in the tool description, the docs page, and the e2e test comments — so this is behaving as intended. 👍

PR #676, opened 2026-07-29
feat(desktop): execute-authoring-plan reports an intent digest of each step's effects

mattcfilbert:
Adds an intent digest to execute-authoring-plan: after a plan runs, the result carries a per-step record of what the plan actually did, derived from a command-effect registry. The rule this sets, and the one to keep: the digest never claims an effect the code cannot observe — unclassified commands are refused before any write, unobservable host evidence is reported as unobservable, and assertions that can never be observed (datasource-exists, field-binding derivation) are refused at compile time. worksheet-exists and dashboard-exists now assert the sheet was created by this plan, diffed against a pre-plan baseline — not merely that a sheet by that name exists.

Review trail: two Opus 5 review rounds (first: FIX-FIRST with 6 findings, all fixed and re-verified at the code sites; second: delta-review confirmed the six and added the datasource-exists unpassability, fixed by compile refusal). Validation, run firsthand three times across the rounds — full suite 6462 passing / 331 skipped across 379 files, tsc clean, lint clean. Version 2.62.0 → 2.63.0 (MINOR, new tool capability).

Known nit accepted: mutationClass currently has only test consumers; it stays because the digest schema is the product surface the next packet (refusal-text rewrites) builds on.

Approving this means: the plan tool's receipts stay honest by construction, and future effect kinds must come with an observable check or a refusal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.62.0",
+  "version": "2.63.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/validation/readback-verify.ts
+++ b/src/desktop/validation/readback-verify.ts
@@ -72,18 +72,6 @@ export interface WorksheetSignature {
   declaredInstances: Set<string>;
 }
 
-export interface FilterSignatureExpectation {
-  column: string;
-  members: string[];
-  mode: 'include' | 'exclude';
-  function?: string;
-}
-
-export interface FilterSignatureMatch {
-  matched: boolean;
-  observed: FilterSignature[];
-}
-
 function isRecord(value: unknown): value is XmlRecord {
   return !!value && typeof value === 'object' && !Array.isArray(value);
 }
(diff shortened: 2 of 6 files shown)


mattcfilbert:
Parking this as a draft. Treating it as a likely failed experiment rather than work in progress.

Three reasons, in order of weight:

1. It targets `claude/authoring-plan-spike`, and `execute-authoring-plan` exists on no other branch — zero files on `feature/desktop` and zero on `claude/v12-fixes`, our release branch. Merging this changes nothing that ships, and the spike has no open PR into the release.

2. The lane has never completed a live transaction. Both r7 kill-test runs refused at preflight, so no step of any plan has executed against a real Desktop. This PR adds roughly 1,950 lines of receipt machinery to a path that has not yet succeeded once end to end.

3. A review found it derives `kind:'encoding'` for `x`/`y`, which is structurally unsatisfiable, so it raises `IncompleteOperationError` on plans that actually succeeded. A fix round has run but has not been re-reviewed.

Worth keeping the branch: the honesty design is sound, and the effect-registry idea — refuse unclassified commands before any write, report unobservable host evidence as unobservable rather than as success — should be reused if the plan lane is ever revived. What it needs first is one plan executing a single step live.

Posted by MattGPT on Matt's behalf.

PR #677, opened 2026-07-29
feat(desktop): census refusals name the nearest working door

mattcfilbert:
Rewrites the 11 capability-census reason strings so a refused agent reads what is missing and the exact tool or command shape that works today. This sets the rule for refusal text: name the door, in plain words, in at most two sentences — and never claim more than the shape asserts. Two factual corrections included: worksheet lifecycle (create, name, activate all work via commands; field placement still has no command) and dashboard composition (new-dashboard plus add-sheet-to-dashboard). Strings only — no entry or detection changes.

Review trail: Opus 5 review, six findings applied (FIXED-LOD idiom for the date-window door, zoneId claim dropped, worksheet-lifecycle headline inverted, arg-shape consistency, redundancy cut). Validation firsthand: full suite 6433 passing, tsc clean, lint clean. One known gap the review surfaced, left for a follow-up: no test pins the reason content (only a no-newline check) — a reason-content ratchet would make these strings regression-proof.

Approving this means refusals become teaching moments; future census entries follow the same two-sentence door-naming shape.

### DIFF
--- a/src/tools/desktop/commands/authoringCapabilityCensus.test.ts
+++ b/src/tools/desktop/commands/authoringCapabilityCensus.test.ts
@@ -1,3 +1,4 @@
+import { knownCommands } from '../../../desktop/commandRegistry.js';
 import { checkAuthoringCapabilityCensus } from './authoringCapabilityCensus.js';
 
 const SAFE_PLAN = { steps: [{ command: 'tabdoc:save' }] };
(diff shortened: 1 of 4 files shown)


mattcfilbert:
Retargeting back to `claude/authoring-plan-spike`. I moved this to the release branch earlier by mistake.

`src/tools/desktop/commands/authoringCapabilityCensus.ts` does not exist on `claude/v12-fixes` or on `feature/desktop` — the capability census is spike-only, the same as `executeAuthoringPlan.ts`. Retargeting a PR does not rebase its commits, so pointing this at the release wave would have dragged 32 files of unrelated spike divergence into it.

Same standing as #676: the work is sound, but it lands on a lane that is not in the shipping wave. It should follow the spike, if and when the spike is proposed for release.

Posted by MattGPT on Matt's behalf.

mattcfilbert:
Also parking this as a draft, alongside #676.

Both sit on the same lane. The capability census and `execute-authoring-plan` exist only on `claude/authoring-plan-spike` — neither file is on `claude/v12-fixes` or `feature/desktop` — and the spike has no open PR proposing it for release. So this improves refusal text that no user currently reads.

The work itself holds up and is worth keeping: every refusal now names the nearest door with its real argument shape, two of the old reasons were factually wrong (worksheet lifecycle and dashboard composition both work today), and there is now a ratchet so a refusal can never name a command absent from the registry. That ratchet is the durable part — it makes the error class impossible rather than corrected once, and it is worth porting wherever refusal text ships from.

Un-draft this if and when the spike is proposed for a release wave.

Posted by MattGPT on Matt's behalf.

PR #681, opened 2026-07-29
@W-23545107: Adding file mode to view image and custom view image tools

yyk16293:
# Pull Request Template

## Description
Adding file mode to view image and custom view image tools so that tools return file urls.

## Motivation and Context
Currently, tools return base64 encoded data for images. For large images, base64 encoded data results in huge amount of data transfer and also fills up context window super fast causing inaccurate responses affecting overall experience. 

Putting this feature under feature flag **view-file-mode**

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/views/getCustomViewImage.ts
```diff
@@ -81,7 +85,7 @@ export const getGetCustomViewImageTool = (
             return new CustomViewNotAllowedError(isAllowedResult.message).toErr();
           }
 
-          const result = await useRestApi({
+          const imageResult = await useRestApi({
```
jarhun88, line 88: btw whats going to happen if the user specifies SVG as their param type? will it still work

yyk16293: Yes, image is uploaded to S3 and image url is returned.

src/tools/web/convertViewImageToToolResult.ts
```diff
@@ -1,5 +1,61 @@
...
+ * `image_url`; the presigned URL qualifies (its auth lives in the query string
+ * and Slack fetches it server-side). Hosts that don't understand the key ignore
+ * it, so this is inert for non-Slack clients. SVG is intentionally excluded:
+ * Slack `image` blocks do not support SVG, so emitting one would fail to render.
+ */
+export function convertViewImageUrlToToolResult(
+  url: string,
+  format: 'PNG' | 'SVG' | undefined = 'PNG',
+): CallToolResult {
+  const resolvedFormat = format ?? 'PNG';
```
jarhun88, line 28: what is the point of this param? doesnt format take care of it

src/tools/web/uploadImageToS3.ts
```diff
@@ -0,0 +1,174 @@
+import { randomUUID } from 'node:crypto';
```
jarhun88, line 1: if we are really short on time we can do this, but i feel like it makes more sense to have these imports only for the tabhf repo so we dont have to force all users to install aws deps to use the app when they wont be using it. Similar to the telemetry and feature flag interface and implementation patterns. thoughts?

src/tools/web/uploadImageToS3.ts
```diff
@@ -0,0 +1,174 @@
...
+ * initialized or loaded into memory there. (Note: in the bundled production
+ * build the SDK bytes are still inlined into the output; the dynamic import
+ * defers runtime initialization, not the packaged size.)
+ *
+ * The AWS SDK v3 Node HTTP handler enables keep-alive by default, so a single
+ * cached client reuses connections across uploads.
+ */
+let s3BundlePromise:
+  | Promise<{
+      client: any;
```
jarhun88: define whats inside the promise as a type, so that it can be reused across this file

src/tools/web/uploadImageToS3.ts
```diff
@@ -0,0 +1,174 @@
...
+  client: any;
+
+  PutObjectCommand: any;
+
+  GetObjectCommand: any;
+
+  getSignedUrl: any;
+}> {
+  if (!s3BundlePromise) {
+    s3BundlePromise = (async () => {
```
jarhun88: i think it will be abit easier to read if you move the async function as an internal helper function.

s3BundlePromise = await getS3BundleInternal()

src/tools/web/uploadImageToS3.ts
yyk16293: We will do as followup.

src/tools/web/convertViewImageToToolResult.ts
yyk16293: Do as followup

jarhun88 (approved):

PR #683, opened 2026-07-29
@W-23633781 - Make tableau-mcp-desktop serve both profiles (desktop stdio + web http)

tonythebest:
## Summary

The shipped **`tableau-mcp-desktop`** binary now serves **both** profiles from one artifact, selected by `TRANSPORT` — `stdio` → desktop authoring (unchanged), `http` → web/insights. This lets `tab-agent-south` spawn the web profile from the **same installed binary** it already launches for desktop authoring, so Chiron's **B1** needs **no new release artifact** and **no `upload-binaries.yml` / Southard packaging change**.

> Reworked from the earlier approach (a separate `bundled` SEA variant, per the discussion below): reusing the existing `tableau-mcp-desktop` name keeps the release workflow and Southard packaging untouched, and drops the extra variant.

## Changes
- **`src/index.desktop.ts`** — dispatch on `TRANSPORT`: `stdio` → `getDesktopConfig()` + `DesktopMcpServer` (the unchanged desktop authoring path); `http` → `getConfig()` + `startExpressServer` (mirrors `index.ts`'s http case). `stdio` behaviour is unchanged for existing desktop/npm consumers; `http` is additive.
- **`src/transportProfile.ts`** — `resolveTransportProfile(TRANSPORT)` returns `desktop`/`web` and **throws** on anything else, so a typo can't fail open to the OAuth‑disabled HTTP path. Unit‑tested in `src/transportProfile.test.ts`.

The separate `bundled` SEA variant (and its `variants.ts`/`buildSea.ts`/`build.ts`/`prepareVariantPackage.ts`/`package.json` additions) is dropped as redundant.

## Companion change (separate MR)
`tab-agent-south` drops the `_looks_desktop_only` discovery guard so the now‑web‑capable `tableau-mcp-desktop` is accepted for the web profile (the guard existed only to reject the *old* stdio‑only desktop SEA). It's a separate MR, gated on this shipping — an old desktop binary + guard‑removed backend soft‑degrades (Increment C) rather than crashes.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `transportProfile.test.ts` — dispatch unit tests (unset / stdio / http / invalid) pass (4/4)
- [x] `tsx src/scripts/build.ts --variant desktop` → builds `build/index.desktop.js` + stages the desktop‑data allowlist
- [x] eslint clean
- [x] **Both profiles from the desktop binary (local):** `stdio` → 31 desktop authoring tools; `http` → boots the web/insights server (`tableau-mcp v2.47.0 streamable HTTP server available…`); and driven through `tab-agent-south` + the real UI it returned **real Pulse insight cards** for GUS‑Work.
- [ ] released‑SEA http smoke — CI/reviewer

## Notes
- Base is **`feature/desktop`**.
- **"One binary":** the installer keeps vendoring the single `tableau-mcp-desktop` (now dual‑profile). `tableau-mcp`'s other variants (`default` web, `combined`) are unchanged and remain for their own consumers.
- **Naming:** `tableau-mcp-desktop` serving `http` is a mild misnomer, kept intentionally so the release workflow + Southard packaging need **no** change (a rename is exactly what would force those edits). Open to a follow‑up rename if preferred.

### DIFF
--- a/src/index.desktop.ts
+++ b/src/index.desktop.ts
@@ -1,15 +1,38 @@
+#!/usr/bin/env node
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
 import { SetLevelRequestSchema } from '@modelcontextprotocol/sdk/types.js';
 import dotenv from 'dotenv';
 
+import pkg from '../package.json';
 import { getDesktopConfig } from './config.desktop.js';
+import { getConfig } from './config.js';
+import { initializeFeatureGate } from './features/init.js';
+import { getTableauServerInfo } from './getTableauServerInfo.js';
 import { FileLogger, setFileLogger } from './logging/fileLogger.js';
 import { log } from './logging/logger.js';
 import { isNotificationLevel, notifier, setNotificationLevel } from './logging/notification.js';
+import { RestApi } from './sdks/tableau/restApi.js';
 import { DesktopMcpServer } from './server.desktop.js';
+import { startExpressServer } from './server/express.js';
+import { resolveTransportProfile } from './transportProfile.js';
 
-async function startServer(): Promise<void> {
-  dotenv.config();
+const serverVersion = pkg.version;
+
+// The shipped tableau-mcp-desktop binary now serves BOTH profiles from one artifact,
+// selected purely by TRANSPORT — so tab-agent-south can spawn the web/insights profile
+// from the very same installed binary it launches for desktop authoring, without a
+// second bundled artifact (Chiron insights "B1"):
+//
+//   TRANSPORT=stdio (default) -> desktop authoring profile: DesktopMcpServer over
+//     stdio, driving the local Tableau Desktop via the External Client API. No
+//     Tableau Server credentials.
+//   TRANSPORT=http            -> web/insights profile: the Express server calling
+//     Tableau Server/Cloud REST (OAuth or passthrough wg_session auth), mirroring the
+//     http case of src/index.ts.
+//
+// stdio behaviour is unchanged for existing desktop consumers; http is additive.
+
+async function startDesktopProfile(): Promise<void> {
   const config = getDesktopConfig();
 
   const notificationLevel = isNotificationLevel(config.defaultNotificationLevel)
@@ -22,10 +45,6 @@ async function startServer(): Promise<void> {
         fileNamePrefix: 'desktop-mcp-',
       }),
     );
-  }
-
-  if (config.transport !== 'stdio') {
-    throw new Error('Transport must be stdio for Desktop server');
   }
 
   const server = new DesktopMcpServer();
@@ -41,6 +60,85 @@ async function startServer(): Promise<void> {
 
   setNotificationLevel(server.mcpServer, notificationLevel);
   notifier.info(server.mcpServer, `${server.name} v${server.version} running on stdio`);
+}
+
+async function startWebProfile(): Promise<void> {
+  const config = getConfig();
+
+  RestApi.host = config.server;
+
+  // Initialize feature gate provider
+  initializeFeatureGate();
+
+  // Start fetching server info immediately but don't block the port from opening.
+  // Any failure here is fatal and logged explicitly -- no silent failures. The port
+  // opens first so health checks can succeed, then we await this before declaring ready.
+  const serverInfoReady = getTableauServerInfo(config.server).catch((error) => {
+    log({
+      message: 'Fatal error initializing server info',
+      level: 'error',
+      logger: 'startup',
+      data: error,
+    });
+    process.exit(1);
+  });
+
+  const notificationLevel = isNotificationLevel(config.defaultNotificationLevel)
+    ? config.defaultNotificationLevel
+    : 'debug';
+  if (config.loggers.has('fileLogger')) {
+    setFileLogger(new FileLogger({ logDirectory: config.fileLoggerDirectory }));
+  }
+
+  const { url } = await startExpressServer({
+    basePath: 'tableau-mcp',
+    config,
+    logLevel: notificationLevel,
+  });
+
+  // Port is now open. Wait for server info before logging the ready message.
+  await serverInfoReady;
+
+  if (!config.oauth.enabled) {
+    log({
+      message:
+        '⚠️ TRANSPORT is "http" but OAuth is disabled! Your MCP server may not be protected from unauthorized access. Non-OAuth HTTP usage is intended only for testing/prototyping or deployments that are licensed and approved for user-based licensing (UBL). For general multi-user HTTP deployments, prefer OAuth. By explicitly disabling OAuth with DANGEROUSLY_DISABLE_OAUTH="true", confirm this matches your Tableau licensing/security guidance.',
+      level: 'info',
+      logger: 'startup',
+    });
+  }
+
+  log({
+    message: `tableau-mcp v${serverVersion} ${config.disableSessionManagement ? 'stateless ' : ''}streamable HTTP server available at ${url}`,
+    level: 'info',
+    logger: 'startup',
+  });
+
+  if (config.disableLogMasking) {
+    log({ message: '⚠️ Log masking is disabled!', level: 'info', logger: 'startup' });
+  }
+
+  if (config.breakGlassDisableGlobally) {
+    log({
+      message:
+        '⚠️ BREAK_GLASS_DISABLE_GLOBALLY is enabled! This means that the MCP server will be disabled globally and will return errors to all users!',
+      level: 'info',
+      logger: 'startup',
+    });
+  }
+}
+
+async function startServer(): Promise<void> {
+  dotenv.config();
+  // Profile is selected (and TRANSPORT validated) before any profile-specific config is
+  // loaded: getDesktopConfig() requires stdio, getConfig() (web) requires SERVER/auth, so
+  // we must not construct the wrong one — and an invalid transport must not fail open to
+  // the HTTP path (see resolveTransportProfile).
+  if (resolveTransportProfile(process.env.TRANSPORT) === 'desktop') {
+    await startDesktopProfile();
+  } else {
+    await startWebProfile();
+  }
 }
 
 startServer().catch((error) => {
--- a/src/transportProfile.test.ts
+++ b/src/transportProfile.test.ts
@@ -1,0 +1,21 @@
+import { resolveTransportProfile } from './transportProfile.js';
+
+describe('resolveTransportProfile', () => {
+  it('defaults to the desktop profile when TRANSPORT is unset', () => {
+    expect(resolveTransportProfile(undefined)).toBe('desktop');
+  });
+
+  it('maps stdio to the desktop authoring profile', () => {
+    expect(resolveTransportProfile('stdio')).toBe('desktop');
+  });
+
+  it('maps http to the web/insights profile', () => {
+    expect(resolveTransportProfile('http')).toBe('web');
+  });
+
+  it('throws on an invalid transport instead of failing open to the web path', () => {
+    expect(() => resolveTransportProfile('htpp')).toThrow(/Unsupported TRANSPORT/);
+    expect(() => resolveTransportProfile('sse')).toThrow(/expected "stdio"/);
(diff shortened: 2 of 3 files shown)


tonythebest:
## Local E2E — both profiles verified against the real binary

Built the `bundled` variant and tested it end to end locally:

**Web profile (http):** drove `tab-agent-south`'s real bundled-spawn path (`plan_bundled_web_mcp` + `WebMcpSupervisor`) with **no PAT in the environment**, pointed at `build/index.bundled.js`. The bundled binary booted its web profile — `tableau-mcp v2.47.0 streamable HTTP server available at …/tableau-mcp` — with only a placeholder base-auth identity, and the token-injecting proxy came up in front of it. (The base PAT is never signed in under `ENABLE_PASSTHROUGH_AUTH`; every request authenticates from the injected `wg_session`.)

**Desktop profile (stdio):** `initialize` + `tools/list` over stdio on the same binary returned **31 desktop authoring tools** (`bind-template`, `apply-worksheet`, `author-calc`, …), confirming the stdio branch dispatches to the desktop profile.

**Build/checks:** `tsc --noEmit` clean; `build --variant bundled` produces `build/index.bundled.js` and stages the desktop-data allowlist (11 entries); `buildSea.test.ts` / `seaSmoke.test.ts` / `content-manifest-staging.test.ts` pass (14/14); eslint clean.

Not yet exercised locally: a tools call *through* the proxy with a live `wg_session` (the passthrough path is unchanged from the default binary).

src/scripts/build.ts
```diff
@@ -64,7 +64,7 @@ const globalValues: Record<GlobalIdentifierName, string> = {
     console.log(`⚠️ ${warning.text}`);
   }
 
-  if (variant === 'desktop' || variant === 'combined') {
+  if (variant === 'desktop' || variant === 'combined' || variant === 'bundled') {
```
tableaukyler: why do we need a 3rd type? Combined and bundled are practically synonyms. I would expect "combined" to do what this is doing

tableaukyler: ah looks like combined == everything. bundled == serve different tools depending on profile

tonythebest: Good catch — they aren't synonyms.

`combined` (`index.combined.ts`) fuses both toolsets into one `McpServer` over **stdio only** (`getConfig()` requires `SERVER`, and it throws on non-stdio). It can't serve the web/insights HTTP profile, and it can't boot desktop authoring without Server creds.

`bundled` (`index.bundled.ts`) is a transport-switched dispatcher: `TRANSPORT=stdio` → desktop authoring (`getDesktopConfig()`, no Server); `http` → web/insights over Express. Same binary, one profile per process.

Reusing `combined` would redefine an already-published variant (union-over-stdio → transport dispatch) and drop its `SERVER` requirement — breaking for existing `combined` consumers. Happy to rename `bundled` to something clearer (`dual-profile`?) if the concern is the name rather than the extra type.

Separately: did you/Matt/Nawar land on single-process dual-transport vs profile-switched? Still open from the Slack thread and would change this entrypoint's shape.

src/scripts/buildSea.ts
```diff
@@ -40,6 +40,15 @@ const seaVariants = {
     binaryBase: 'tableau-mcp-desktop',
     assetDirs: DESKTOP_ASSET_DIRS,
   },
+  // One binary, both profiles (B1): TRANSPORT=stdio serves the desktop authoring
+  // profile, TRANSPORT=http serves the web/insights profile. It embeds the desktop
+  // assets (read via node:sea in the stdio profile) just like the desktop variant.
+  bundled: {
```
tonythebest: This adds `bundled` to the default SEA build, but `upload-binaries.yml` still copies/smokes/archives only `default` and `desktop`, so every release builds this artifact and discards it.

Southard still publishes/packages under `tableau-mcp-desktop`, which its own discovery guard rejects — so the shipped-binary / discovery / AUTO ACs for W-23633781 stay unmet.

Before marking Ready: land or link (a) `upload-binaries.yml` copy+smoke+archive for both platforms, (b) `buildSea.test.ts` assertions so the omission can't pass CI, and (c) the coordinated Southard publish/package/manifest change that keeps the basename non-`desktop` (e.g. `tableau-mcp-bundled`).

src/index.bundled.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+  }
+}
+
+async function startServer(): Promise<void> {
+  dotenv.config();
+  // Profile is selected by TRANSPORT before any profile-specific config is loaded:
+  // getDesktopConfig() requires stdio, getConfig() (web) requires SERVER/auth, so we
+  // must not construct the wrong one. Default (stdio) is the desktop authoring profile.
+  const transport = process.env.TRANSPORT ?? 'stdio';
```
tonythebest: Could this validate `TRANSPORT` and accept only `stdio` or `http`?

Today every typo enters the web branch, while `getConfig()` may normalize that same value back to `stdio`. With inherited server/PAT config that can start the HTTP listener unexpectedly; without it, Desktop startup regresses to a hard failure where the prior desktop entrypoint would normalize to stdio.

src/scripts/buildSea.ts
```diff
@@ -40,6 +40,15 @@ const seaVariants = {
...
+    buildVariant: 'bundled',
+    entry: 'build/index.bundled.js',
+    binaryBase: 'tableau-mcp-bundled',
+    assetDirs: DESKTOP_ASSET_DIRS,
```
tonythebest: The bundled SEA embeds only the desktop asset dirs. Web MCP App resources live under `build/web/apps/dist` and are read with ordinary filesystem I/O, so those resource reads have no path inside the self-contained SEA.

Chiron's three insight tools aren't app tools, so this doesn't block the narrow B1 demo — but the broader "serves both profiles" claim overstates the SEA web surface. Prefer either SEA-aware web-app assets/read logic, or an explicit reduced contract (non-app insight tools only) plus a SEA resource-read smoke.

src/index.bundled.ts
```diff
@@ -0,0 +1,152 @@
...
+    log({
+      message:
+        '⚠️ BREAK_GLASS_DISABLE_GLOBALLY is enabled! This means that the MCP server will be disabled globally and will return errors to all users!',
+      level: 'info',
+      logger: 'startup',
+    });
+  }
+}
+
+async function startServer(): Promise<void> {
```
tonythebest: New side-effecting entrypoint with no dispatch tests, and `seaSmoke` forces `TRANSPORT=stdio`, so CI never exercises the HTTP/Chiron branch.

Please add coverage for unset / stdio / http / invalid dispatch, plus an HTTP bundled-SEA startup smoke (passthrough-safe config is enough; live `wg_session` tool call can stay a separate integration check).

tonythebest (commented):
Self-review follow-up after the Studio thread + Kyler’s combined-vs-bundled question. Direction is still gated on process model (single-process dual-transport vs profile-switched) and on shipping/consuming the artifact end-to-end.

src/index.bundled.ts
tonythebest: Done in the latest push (`5109662`): extracted `resolveBundledProfile(process.env.TRANSPORT)` into `src/bundledProfile.ts` — it returns `desktop` for `stdio`/unset, `web` for `http`, and **throws** on anything else, so a typo no longer fails open to the OAuth‑disabled HTTP path (nor regresses desktop to a hard failure). Unit‑tested in `src/bundledProfile.test.ts`.

src/index.bundled.ts
tonythebest: Added `src/bundledProfile.test.ts` covering unset / stdio / http / invalid dispatch (4 tests, passing). The full HTTP bundled‑SEA startup smoke belongs in the release workflow's `seaSmoke` step (which currently forces `TRANSPORT=stdio`) — tracked with the release‑workflow follow‑up below; the live `wg_session` tool call stays a separate integration check.

src/scripts/buildSea.ts
tonythebest: Acknowledged — this is the "shipped" gap. `build:sea` now builds the `bundled` binary, but wiring `upload-binaries.yml` to copy/smoke/archive it (and the coordinated Southard change to publish/package/launch `tableau-mcp-bundled` without a `desktop` basename) is a follow‑up: the workflow‑file edit is blocked by a local Actions‑file guard, so the exact linux/windows snippet is in the PR description for manual apply. I'll narrow the story's "shipped/enablement" AC to that follow‑up rather than claim it in this PR.

src/scripts/buildSea.ts
tonythebest: Correct — the bundled SEA embeds only `DESKTOP_ASSET_DIRS`, matching the `default` (web) SEA variant, which also doesn't embed `build/web/apps/dist` (those MCP App HTML resources are read via ordinary fs). Chiron's three insight tools aren't App tools, so the narrow B1 path is unaffected. Making the SEA web‑App resources self‑contained is a broader change that applies to the `default` web SEA too, so I'll track it as a follow‑up rather than fold it into B1.

tonythebest:
**Reworked (head `489ea07`) per the "why a 3rd variant / can we reuse tableau-mcp-desktop" discussion.**

Instead of a separate `bundled` SEA variant, the existing **`tableau-mcp-desktop`** binary now dispatches on `TRANSPORT` (`stdio` → desktop authoring, `http` → web/insights). Net effect:
- No new release artifact → **no `upload-binaries.yml` change** (it already ships `tableau-mcp-desktop`).
- **No Southard packaging rename** (it already vendors `tableau-mcp-desktop`).
- The `bundled` variant + its build-script additions are removed.

The `TRANSPORT` validation (M1) now lives in `src/transportProfile.ts` / `resolveTransportProfile` with tests in `src/transportProfile.test.ts` (renamed from `bundledProfile`). The SEA-web-app-asset note still stands and is unchanged (same as the pre-existing `default` web SEA). The one downstream change is dropping tab-agent-south's `_looks_desktop_only` guard — separate MR, gated on this shipping.

src/scripts/build.ts
tonythebest: Update: I reworked this to **drop the separate variant entirely** — the existing `tableau-mcp-desktop` binary is now the dual-profile dispatcher (`stdio` → desktop, `http` → web). So there's no 3rd variant anymore: `combined` (union-of-both-toolsets over stdio) is untouched, and `desktop` simply gains the http/web profile. Bonus: reusing the name means no `upload-binaries.yml` or packaging change.

src/scripts/buildSea.ts
tonythebest: Resolved by the rework: reusing `tableau-mcp-desktop` (now dual-profile) means the release workflow already ships it and Southard already vendors it — **no `upload-binaries.yml` change and no packaging rename**. The only remaining downstream change is dropping tab-agent-south's `_looks_desktop_only` discovery guard (separate MR), so the shipped-binary / discovery / AUTO ACs are reachable without touching the release path.

src/index.desktop.ts
```diff
@@ -1,15 +1,38 @@
...
+import { RestApi } from './sdks/tableau/restApi.js';
 import { DesktopMcpServer } from './server.desktop.js';
+import { startExpressServer } from './server/express.js';
+import { resolveTransportProfile } from './transportProfile.js';
 
-async function startServer(): Promise<void> {
-  dotenv.config();
+const serverVersion = pkg.version;
+
+// The shipped tableau-mcp-desktop binary now serves BOTH profiles from one artifact,
```
tableaukyler, line 21: I think this is perfect for now, but before we merge `feature/desktop`, I think we should be updating the main index.ts to serve both profiles. That'll require more work, and we also need to move the knowledge/templates out of the binary before doing so, but ya. LGTM for now

tableaukyler (approved):

src/index.desktop.ts
tonythebest: Thanks — agreed. Captured as a follow-up so it's not lost: **W-23651728** (update mainline `index.ts` to dispatch both profiles + move the knowledge/templates out of the SEA, before `feature/desktop` → `main`). Keeping this PR scoped to the desktop-binary dispatch; resolving.

PR #691, opened 2026-07-31
feat(desktop): handle External Client API 0.2.0 async dispatch (202 poll)

lauren-jacksonSFDC:
## Decision

The External Client API (Athena V0) moved to **unified async dispatch** at apiVersion **0.2.0**: every command enqueues, waits a ~250 ms server sync window, and either returns a terminal Operation at **200** or **202 + `Location`** requiring the client to poll `GET /v0/operations/{id}`. The MCP client gated success on `res.ok` (true for 202), so writes reported **false success**, workbook-XML reads returned **JSON as XML**, and inventory reads **silent-emptied** — all on a read that merely overflowed the window.

This teaches the desktop transport to speak the async world. **Polling lives entirely in `ExternalApiClient`** (the only layer holding the raw `Response`/`Location`/`Retry-After`), so the async seam stays invisible to the executor, consumers, and `tools/list` (**no new tool** — the byte budget has ~4 bytes of headroom).

This is a **new rule the codebase now keeps**, not a one-off: branch on **HTTP 202 status**, never on `apiVersion` (mixed-fleet safe).

## What changed

- **Client polling:** on 202, follow `Location` to a terminal Operation, honoring `Retry-After` and an overall poll deadline (`pollDeadlineMs`, default 300 s — deliberately > the 60 s per-fetch timeout). For invokeCommand, the terminal poll body **carries `result`**, so it is unwrapped. New typed `ExternalApiError` variants: `read-overflowed`, `operation-pending` (503), `awaiting-user`, `operation-expired` (404), `poll-timeout`.
- **Executor terminality:** `buildCommandStatus` treats only `SUCCEEDED`/`FAILED`/`CANCELLED` as terminal; a non-terminal state is reported `running`/`queued`, **never `completed`**. Dead `state === 'error'` check removed.
- **Read honest-error:** a typed read that 202s returns `read-overflowed` (never JSON-as-XML / silent-empty); a 503 precursor overflow returns `operation-pending`. Typed-read poll-payload retrieval is deferred to the Desktop-side follow-up (the wire is the uniform Operation envelope); until then an overflowed typed read fails honestly rather than guessing.
- **Consumer readback poll:** `verifyPostApplyWorksheetReadback` now polls 8×/250 ms (mirroring `refineWorksheet`) so a race against the async settle no longer misreports a durable apply as a dropped node.
- **Contract intake (live 0.2.0):** fixture re-captured from a running Desktop build (`info.version` 0.2.0). Added the `operations`/`operationById` routes, `operationListSchema`, `updatedAt` on the Operation envelope, and `operation-pending` + `invalid-query-parameter` to `PROBLEM_CODES`. `operation-not-found` is **deliberately not** in `PROBLEM_CODES` — it is a real 404 code the client tolerates, but the live spec's `Problem.code` enum omits it (worth an upstream flag). The envelope parse schema stays looser than the spec's required set on purpose (fail-open); the contract test asserts it never demands a field the spec doesn't.
- **Mock:** `MockOverride.headers`, `/v0/operations/{id}` + `/v0/operations` poll routes, and a `setOperation` scenario helper.

## Verification

Green on Linux CI (build 22.7.5 + 24.x). Locally (Windows) the pre-existing suite has ~80 environment-dependent failures (hash-stamp / fs-path / temp-dir tests) unrelated to this change: **80 failed on the clean base vs. 80 with this branch, plus the new passing async tests — zero new failures.**

New coverage: 202→poll→SUCCEEDED (unwrapping invokeCommand result), 202-read→`read-overflowed`, 503→`operation-pending`, poll 404→`operation-expired`, poll-timeout, `Retry-After` spacing, terminal-in-window does exactly one request, non-terminal never reported completed, CANCELLED→failed, and the readback-poll race. Also exercised by hand against a live 0.2.0 Desktop build (long apply → 202 → poll → terminal; large read → `read-overflowed`; `tabui:*` invokeCommand → 202 → poll → result).

### DIFF
--- a/src/desktop/externalApi/externalApiClient.ts
+++ b/src/desktop/externalApi/externalApiClient.ts
@@ -49,6 +49,8 @@ export type ExternalApiClientOptions = {
   fetchFn?: typeof fetch;
   /** Global ceiling for each request; health remains capped at its shorter route budget. */
   timeoutMs?: number;
+  /** Overall wall-clock ceiling for a 202→poll loop, separate from the per-fetch timeout. */
+  pollDeadlineMs?: number;
 };
 
 export type WorkbookDocument = {
@@ -78,6 +80,24 @@ export type ImageExportQuery = {
 const DEFAULT_TIMEOUT_MS = 60_000;
 const DEFAULT_HEALTH_TIMEOUT_MS = 10_000;
 
+// Bounds the whole 202→poll loop. Deliberately larger than the per-fetch timeout so async dispatch
+// does not regress the max operation length it exists to enable.
+const DEFAULT_POLL_DEADLINE_MS = 300_000;
+const DEFAULT_RETRY_AFTER_SECONDS = 1;
+const HTTP_ACCEPTED = 202;
+const HTTP_SERVICE_UNAVAILABLE = 503;
+
+// Wire states are UPPER_SNAKE_CASE; unknown values count as non-terminal per the spec.
+const TERMINAL_STATES = new Set(['SUCCEEDED', 'FAILED', 'CANCELLED']);
+
+function isTerminalState(state: string | undefined): boolean {
+  return state !== undefined && TERMINAL_STATES.has(state.toUpperCase());
+}
+
+const HEADER_LOCATION = 'location';
+const HEADER_RETRY_AFTER = 'retry-after';
+const HEADER_OPERATION_ID = 'x-tableau-operation-id';
+
 /**
  * Typed client for a single Tableau Desktop External Client API instance.
  *
@@ -89,11 +109,13 @@ export class ExternalApiClient {
   private readonly instance: ExternalApiInstance;
   private readonly fetchFn: typeof fetch;
   private readonly timeoutMs: number | undefined;
+  private readonly pollDeadlineMs: number;
 
   constructor(instance: ExternalApiInstance, options: ExternalApiClientOptions = {}) {
     this.instance = instance;
     this.fetchFn = options.fetchFn ?? fetch;
     this.timeoutMs = options.timeoutMs;
+    this.pollDeadlineMs = options.pollDeadlineMs ?? DEFAULT_POLL_DEADLINE_MS;
   }
 
   get baseUrl(): string {
@@ -135,7 +157,7 @@ export class ExternalApiClient {
       contentType: 'application/xml',
       body: xml,
     });
-    return this.parseEnvelope(response);
+    return this.parseEnvelope(response, signal);
   }
 
   async applyWorksheetDocument(
@@ -198,10 +220,9 @@ export class ExternalApiClient {
     const response = await this.request('POST', EXTERNAL_API_ROUTES.invokeCommand, {
       signal,
       contentType: 'application/json',
-      // Field name `parameters` is a best-guess from the PR contract — see report.
       body: JSON.stringify({ namespace, command, parameters: params }),
     });
-    return this.parseEnvelope(response);
+    return this.parseEnvelope(response, signal);
   }
 
   async fetchOpenApi(signal?: AbortSignal): Promise<Result<unknown, ExternalApiError>> {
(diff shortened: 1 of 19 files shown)


src/desktop/externalApi/externalApiClient.ts
```diff
@@ -301,14 +322,71 @@ export class ExternalApiClient {
...
+      const polled = await this.request('GET', location, { signal });
+      if (polled.isErr()) {
+        return Err(polled.error);
+      }
+
+      const res = polled.value;
+      if (res.status === 404) {
+        return Err({ type: 'operation-expired', operationId });
+      }
+      if (!res.ok) {
```
lauren-jacksonSFDC, line 428: ✨ _Lauren's code review bot:_ ✨

A 401 while polling replays the already accepted mutation.

**Trigger:** Desktop accepts a command or workbook update with 202, then the poll returns 401, so `withRescan` reruns the operation closure and sends the original POST again.

**Why this matters:** The command can execute twice, or a workbook update can overwrite edits made after the first POST.

src/desktop/externalApi/externalApiClient.ts
```diff
@@ -327,6 +405,10 @@ export class ExternalApiClient {
     }
 
     const res = response.value;
+    const overflow = readOverflowError(res);
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

Async workbook validation is misclassified as a read overflow instead of being polled.

**Trigger:** `/workbook/document:validate` returns its documented 202 response, but `parseJson` routes it through `readOverflowError` and never follows the `Location` operation.

**Why this matters:** Valid slow validations fail with unrelated advice to narrow a read.

src/desktop/externalApi/externalApiClient.ts
```diff
@@ -410,6 +496,65 @@ function composeWithTimeout(signal: AbortSignal | undefined, timeoutMs: number):
...
+// A typed read's payload is unreachable by polling, so an overflow (202) is a terminal error here,
+// not a poll trigger like it is on the write path.
+function readOverflowError(res: Response): ExternalApiError | undefined {
+  if (res.status === HTTP_ACCEPTED) {
+    return {
+      type: 'read-overflowed',
+      operationId: res.headers.get(HEADER_OPERATION_ID) ?? undefined,
+    };
+  }
+  if (res.status === HTTP_SERVICE_UNAVAILABLE) {
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

Classifying every typed-endpoint 503 as operation-pending discards genuine service-unavailable errors.

**Trigger:** Desktop returns the documented 503 because the API is disabled or shutting down, but this branch runs before `mapErrorResponse` and replaces the Problem body with `operation-pending`.

**Why this matters:** Callers receive a misleading retry instruction and lose the real failure code and details.

src/desktop/externalApi/externalApiToolExecutor.ts
```diff
@@ -517,8 +517,8 @@ function buildCommandStatus(
   outcome: RawOutcome,
   { namespace, command }: { namespace: string; command: string },
 ): Result<GetCommandStatusResponse, ExecuteCommandError> {
-  const state = outcome.state?.toLowerCase();
-  const failed = state === 'failed' || state === 'error' || outcome.envelopeError !== undefined;
+  const state = outcome.state?.toUpperCase();
+  const failed = state === 'FAILED' || state === 'CANCELLED' || outcome.envelopeError !== undefined;
```
lauren-jacksonSFDC, line 570: ✨ _Lauren's code review bot:_ ✨

Legacy `ERROR` operation states are no longer recognized as failures.

**Trigger:** A supported 0.1.0 instance returns `state: "ERROR"` without an error object, so `failed` remains false and `buildCommandStatus` returns `Ok` with status `running`.

**Why this matters:** A failed command can be reported as still active, allowing higher-level callers to treat dispatch as successful.

src/desktop/externalApi/externalApiClient.ts
lauren-jacksonSFDC: Fixed (caa… follow-up commit). `readOverflowError` now only maps a 503 to `operation-pending` when the Problem `code` actually says so; a genuine `api-disabled`/shutting-down 503 falls through to `mapErrorResponse` and keeps its real code. Added a test asserting a 503 `api-disabled` surfaces as `{type:'problem', code:'api-disabled'}`, not a retry instruction.

src/desktop/externalApi/externalApiClient.ts
lauren-jacksonSFDC: Real mislabel, deferring the fix (not this PR). `validateWorkbookDocument` uses `parseJson`, so its documented 202 becomes `read-overflowed` with narrow-the-read advice — wrong for a write-shaped route. The correct fix is to poll it like apply/invokeCommand, but the polled validate body's shape isn't the same as its in-window 200 (`ValidationResult`), and confirming that depends on the in-flight monolith read-payload work (W-23640925). Current behavior fails safe (a genuine error, not a false success). Tracked in the refactor plan; not building it unverified here.

src/desktop/externalApi/externalApiClient.ts
lauren-jacksonSFDC: Confirmed real — this is open question B6 in the plan. `withRescan`/`withClientRescan` rerun the whole operation closure on 401, so a 401 on the poll leg replays the original POST → possible double-apply. The safe fix (rescan the poll GET in place with a fresh token, without re-POSTing) is a change to the rescan-wrapper design that I can't verify without a live mid-poll 401 in the harness. Deferring rather than shipping an unverified retry change; documented as B6.

src/desktop/externalApi/externalApiToolExecutor.ts
lauren-jacksonSFDC: Disagree — no change. Verified against merged monolith `Operation.h`: the 0.2.0 `OperationState` enum has no `ERROR` (states are QUEUED/RUNNING/AWAITING_USER/SUCCEEDED/FAILED/CANCELLED), and the header comment mandates "clients must tolerate unknown values and treat them as non-terminal." So mapping a phantom `ERROR` to failed would violate the contract — an unknown state is correctly reported `running`, not swallowed as success (a genuine failure always carries `state:FAILED` or an `error` object, both already handled). Removing the dead `state==='error'` check was intentional.

mattcfilbert (approved):
Posted by MattGPT (Matt's agent) on Matt's behalf.

Approving. Reviewed through Andy's lens by an Opus pass focused on the async correctness, which is the load-bearing risk here and holds up:

- 202 detection branches on **HTTP status** (`response.value.status === HTTP_ACCEPTED`, client.ts:332), not `apiVersion` — so it works against a mixed fleet.
- Polling follows `Location`, honors `Retry-After`, bounds the loop with a deadline check at the top, and terminates on SUCCEEDED/FAILED/CANCELLED, AWAITING_USER, poll-404 (`operation-expired`), and deadline (`poll-timeout`).
- A 202 read returns `read-overflowed` — never JSON-as-XML or a silent empty body. `buildCommandStatus` reports non-terminal as running/queued, never completed, which guards the old false-success-on-202 bug.
- An abort during a poll throws from the delay helper and is caught at the tool boundary, so it becomes an error result, not a hang. Covered by the executor abort test and the poll-readback abort test. CI green on both Node matrices.

Two optional robustness notes, neither blocking:

1. **`Location` is treated as relative to `baseUrl`** (client.ts:360,491). An absolute `Location` (allowed by the spec) would build a bad URL and fail as a network error. Your live 0.2.0 server emits relative, so this is latent — worth a one-line assumption comment or an absolute-URL guard at the seam.
2. **The inter-poll `delay(retryAfterMs)` is not clamped to the remaining deadline** (client.ts:358). A `Retry-After` larger than the poll budget overshoots by up to one interval and leaves a live timer when the tool-level deadline wins. Bounded in practice by the loopback server's small Retry-After; clamp to `deadline - Date.now()` if you want the budget exact.

Note, not a finding: no `package.json` bump, but `version-bump-check.yml` runs only on PRs into `main`, so it does not apply to a `feature/desktop` PR — the branch bumps once at release time.

mattcfilbert (approved):
Posted by MattGPT (Matt's agent) on Matt's behalf.

Re-approving at `54c3c5e`. Lauren, the updates close the gap Will called out: overflowed JSON reads now poll and unwrap the terminal `result`, document reads unwrap `result.document`, and a failed read keeps Desktop's error instead of turning into a schema error. The abort signal also reaches the per-sheet apply poll. CI is green on both Node versions.

One non-blocking follow-up:

- `src/desktop/externalApi/externalApiClient.ts:232` — please pass `signal` to `parseJson`. The captured 0.2.0 contract allows `202` for workbook validation, and `parseJson` now polls that response. This is the one call that drops the caller's abort signal, so canceling a slow validation can leave it polling until the 300-second deadline. `return this.parseJson(response, validationResultSchema, signal)` plus a 202-validation abort test should cover it.

The operation is read-only and the poll is bounded, so I don't think this needs to hold the merge. The async result path otherwise holds up.

PR #692, opened 2026-07-31
refactor(desktop): share one readback poll across apply-verify call-sites

lauren-jacksonSFDC:
## Decision

The post-apply readback race — a re-read taken right after an apply can see PRE-apply XML, because Desktop settles asynchronously after it reports the command terminal — was handled inconsistently. `loadWorksheetXml`/`refineWorksheet` polled the readback (8×/250ms); the four datasource authoring tools (`author-set`, `author-action`, `author-calc`, `format-labels`) did a **single** read and **hard-failed** the tool if the just-applied change wasn't visible yet. That's a false-negative: a durable apply reported as "did not apply".

This extracts one shared `pollReadback` primitive and routes all six read-modify-apply-verify call-sites through it, so the async-settle protection is uniform and a future change touches one place instead of six. This is a **rule the codebase now keeps**: post-apply verification polls; it does not trust a single immediate read.

## What changed

- **New `src/desktop/commands/workbook/pollReadback.ts`** — the retry-until-settled loop + a fake-timer-safe `sleep` + the `READBACK_POLL_*` constants, previously duplicated. Returns the last read as `{settled: false}` (rather than erroring) when the budget is spent, so each caller keeps its own "landed" predicate and its own durable-miss handling.
- **Converged 6 call-sites:**
  - `refineWorksheet` — de-duped onto the shared primitive (behavior identical; deleted its local `sleep` + inline loop).
  - `loadWorksheetXml` `verifyPostApplyWorksheetReadback` — now polls past an error-severity finding that may be a pre-apply race.
  - **`authorSet`, `authorAction`, `authorCalcCore`, `formatLabels`** — the fix: their single-read readback now polls before declaring a durable miss.
- **Tests:** new `pollReadback.test.ts` (5 cases, fake-timer). All existing worksheet/datasource suites stay green (138 tests across 13 files).

Net −8 lines despite the new module + test — the de-duplication pays for itself.

## Check-in order

This is **slice A** of a larger desktop External-Client-API cleanup, and lands **first**:
1. **This PR** (readback-poll gateway) → merge into `feature/desktop`.
2. The async-dispatch PR (#691) then rebases onto the refreshed `feature/desktop`. Expect one conflict in `loadWorksheetXml.verifyPostApplyWorksheetReadback` (both touch it); this PR's shared-poll version supersedes #691's inline poll.
3. Later slices (typed `:action` routes, grouped dispatchers, `invokeCommand` backdoor removal) follow as the monolith ships the P0/P1 command routes.

## Draft — why

Opening as draft to run CI on Linux (local Windows suite has ~80 pre-existing env-only failures unrelated to this change). No new behavior for the agent; this is a correctness + de-duplication refactor.

Note: the six datasource *miss-path* tests now exercise the full poll budget with the real timer (~1.8s each), since they await the tool without advancing fake timers. Correct but slower; a follow-up could retrofit `vi.useFakeTimers()` as `refineWorksheet`'s tests do.

### DIFF
--- a/src/desktop/commands/workbook/loadWorksheetXml.ts
+++ b/src/desktop/commands/workbook/loadWorksheetXml.ts
@@ -23,6 +23,7 @@ import { withApplyLock } from './applyMutex.js';
 import { getWorkbookXml } from './getWorkbookXml.js';
 import { getWorksheetFragment } from './getWorksheetXml.js';
 import { applyWorkbookText } from './loadWorkbookXml.js';
+import { pollReadback } from './pollReadback.js';
 
 export type LoadWorksheetXmlError =
   | { type: 'invalid-xml' }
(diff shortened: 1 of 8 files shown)


lauren-jacksonSFDC:
Folding this into #691. The live smoke test showed slice A can't stand alone on feature/desktop — datasource authoring does a read-modify-apply-**readback**, and on 0.2.0 those reads intermittently 202; the 202 handling lives in #691. Cherry-picked the pollReadback commit onto the #691 branch so both ship together.

src/desktop/commands/workbook/loadWorksheetXml.ts
```diff
@@ -85,21 +86,30 @@ export async function verifyPostApplyWorksheetReadback(
...
 ): Promise<PostApplyWorksheetReadbackVerification> {
   try {
-    const reread = await getWorksheetFragment({ worksheetName, executor, signal });
-    if (reread.isErr()) {
+    // The apply lands asynchronously, so a re-read that looks like it dropped a node might just be
+    // the pre-apply worksheet — poll rather than trust the first read.
+    const polled = await pollReadback({
+      read: () => getWorksheetFragment({ worksheetName, executor, signal }),
+      settled: (fragment) =>
+        !verifyWorksheetReadback(intendedXml, fragment).some((f) => f.severity === 'error'),
```
lauren-jacksonSFDC, line 94: âœ¨ _Lauren's code review bot:_ âœ¨

Sort-only mismatches stop readback polling on the first stale read.

**Trigger:** A sort-only async apply returns pre-apply XML on the first read; because sort differences are warnings, the settled predicate succeeds and records a failed sort verification before the new sort can appear.

**Why this matters:** Valid sort edits can be reported as failed even though Desktop applies them successfully moments later.

PR #693, opened 2026-07-31
refactor(desktop): remove legacy check-for-user-changes tool + relocate executor response types

lauren-jacksonSFDC:
## Decision

The External Client API is the only Desktop transport, so the legacy Agent API is dead — this removes it. Pure removal + relocation; no behavior change to any surviving tool on this transport.

## What changed

**Legacy tool + types (commit 1):**
- Deleted `check-for-user-changes` (tool + test, factory/toolName/profile entries, and the now-single-purpose `excluded` Set in `_getToolsToRegister`).
- Relocated the still-used executor response types (`GetCommandStatusResponse`, `ExecuteCommandResponseError`) from `sdks/desktop/agentApi/types.ts` → `src/desktop/externalApi/executorResponseTypes.ts`; deleted the dead Agent-API-only schemas; removed `sdks/desktop/agentApi/` entirely.

**The events gate (commit 2):**
- Removed `getEvents` from the `ToolExecutor` base class + the External executor + `GetEventsArgs`/`GetEventsResponse`/`getEventsResponseSchema`.
- Removed the mid-operation user-edit "events gate" from `bindTemplate`, `dashboardAutoApply`, and `deleteWorksheet`. On the External transport this gate was **already inert** — `getEvents` always errored, so the anchor was never set and the gate always skipped. Removing it is behavior-preserving here.
- Dropped the now-dead `eventSequence` field from the sheet-reuse memory (`route-state.ts` + bindTemplate's reuse/rebuild logic); reuse-if-remembered is now unconditional (it always was on this transport).
- −505 lines net.

**Note (does the operations poll replace this?):** No — operations polling tracks *our own* async command lifecycle; the events gate detected *human* edits mid-apply. But that gate never functioned on this transport (no events endpoint), so nothing running was lost. A real "human edited mid-apply" guard would need a new signal (a Desktop events route or a doc-version check) and is out of scope.

## Verification

tsc clean, full `npm run lint` clean. All directly-affected suites pass (248 tests across the 7 touched files). Whole-area failure count is identical to `feature/desktop` (80 pre-existing Windows-CRLF/env failures on both — `stampHash`, `discovery`, `knowledge*`, `episode*`, etc., green on Linux).


mattcfilbert (approved):
Posted by MattGPT (Matt's agent) on Matt's behalf.

Approving. Reviewed through Andy's lens by an Opus pass that checked the dead-code claim by enumeration, not assumption: `git grep` at the PR head for every removed symbol — the tool name, `getCheckForUserChangesTool`, `getEvents`, `GetEventsFailedError`, `eventSequence`, `eventsAnchor` — returns zero hits across src, resources, docs, and scripts. The enum, the factory list, and the desktop server's exclude set are all updated consistently, no profile array hardcodes the name, and the two surviving types are relocated with both consumers repointed. Removal is behavior-preserving: `ExternalApiToolExecutor` is the only executor, `getEvents` always errored on that transport, so the events gate always skipped anyway. Tests for the removed tool are deleted and the gate's test blocks scrubbed. CI green on both Node matrices.

One nit, optional: `src/tools/desktop/worksheet/deleteWorksheet.test.ts:236` — the test-local `refusalSchema` enum still lists `'user-changed-workbook'`, which the production `DeleteWorksheetRefusalReason` union dropped in this PR. Harmless (permissive enum, no assertion depends on it), just a stale value the sweep missed.

lauren-jacksonSFDC:
Thanks Matt (and MattGPT) 🙏

Fixed the nit: dropped the stale `'user-changed-workbook'` value from the test-local `refusalSchema` enum in `deleteWorksheet.test.ts` so it matches the production `DeleteWorksheetRefusalReason` union (`'dashboard-referenced' | 'last-worksheet'`). Folded into the same commit that dropped the value.

Also rebased onto the latest `feature/desktop` (it had moved — async-dispatch #691, per-sheet apply #690, sqlproxy #705). The only conflict was in `externalApiToolExecutor.test.ts`, where #691 added the async-dispatch (202) `describe` block next to the old `getEvents` block this PR removes — resolved by keeping the async-dispatch tests and dropping `getEvents`. `tsc` clean; touched-file tests green.

PR #694, opened 2026-07-31
feat(desktop): add insight-authoring templates for Pulse insight types (W-23651207)

nawartamawi:
## Decision

Add two purpose-built worksheet templates so the agent can author a viz directly from a Pulse insight, selected by insight type via `bind-template`/`inject-template`.

## What changed

- **`insight_unusualchange`** — daily (Day-Trunc) time-series line with a linear trendline + confidence bands over a relative-date window. Maps to Pulse **Unusual Change** / Current Trend. Slots: `event_date`, `measure`.
- **`insight_barchart`** — horizontal bars of an aggregated measure across a categorical dimension, sorted DESC. Maps to Pulse **Top/Bottom Drivers/Contributors**. Slots: `category`, `measure`.

Both are templatized from hand-authored live worksheets. Manifests set `fast_path_eligible` and register the templates in `template-manifests.index.json` + `content-manifest.json` (regenerated via `buildTemplateManifests.ts`) so the binder can discover and select them.

## Testing

- Manifests pass `validateManifest` (runtime gate).
- Placeholder substitution produces well-formed XML for both templates.
- Verified end-to-end live: agent selects the template by insight type via explicit `proposal.template` and applies against a published datasource.

Work item: W-23651207

### DIFF
--- a/src/desktop/data/content-manifest.json
+++ b/src/desktop/data/content-manifest.json
@@ -2,9 +2,9 @@
   "_generated": true,
   "_generator": "src/scripts/buildTemplateManifests.ts",
   "_warning": "GENERATED FILE — do not hand-edit. Re-run `npx tsx src/scripts/buildTemplateManifests.ts`.",
-  "content_version": "2.60.3+content.2026-07-31",
+  "content_version": "2.60.3+content.2026-08-03",
   "schema_version": "2",
-  "generated": "2026-07-31",
+  "generated": "2026-08-03",
   "engine_compat": {
     "server_min": "2.60.3",
     "node": ">=22.7.5"
@@ -114,6 +114,16 @@
       "path": "data-visualization-templates-xml/gantt-timeline-chart.xml",
       "sha256": "61b09b5f43eae3f5bfdd90c9de66abc1b5eba1ff9c04141738b05749851cac6a",
       "bytes": 1939
+    },
+    {
+      "path": "data-visualization-templates-xml/insight_barchart.xml",
+      "sha256": "bf6522c2dde81ffeaea143e5400329fb5066262f07c4cfe7561402eae744dab5",
+      "bytes": 2378
+    },
+    {
+      "path": "data-visualization-templates-xml/insight_unusualchange.xml",
+      "sha256": "e4f4ccf812f6fd73880e192ef1fd96461ccc90e77acefc897a2e670fc2e1f69d",
+      "bytes": 2203
     },
     {
       "path": "data-visualization-templates-xml/kpi-text.xml",
--- a/src/desktop/data/content-manifest.json
+++ b/src/desktop/data/content-manifest.json
@@ -2,11 +2,11 @@
   "_generated": true,
   "_generator": "src/scripts/buildTemplateManifests.ts",
   "_warning": "GENERATED FILE — do not hand-edit. Re-run `npx tsx src/scripts/buildTemplateManifests.ts`.",
-  "content_version": "2.59.1+content.2026-07-27",
+  "content_version": "2.60.3+content.2026-07-31",
   "schema_version": "2",
-  "generated": "2026-07-27",
+  "generated": "2026-07-31",
   "engine_compat": {
-    "server_min": "2.59.1",
+    "server_min": "2.60.3",
     "node": ">=22.7.5"
   },
   "resources": [
@@ -252,8 +252,8 @@
     },
     {
       "path": "template-manifests.index.json",
-      "sha256": "9a72ff66a3a4b9720cf2309c2a7793129d8b6bee120ad202316a8b1d214f5255",
-      "bytes": 283278
+      "sha256": "03bde655ff89182d17b9d936c9e3c349193ec95d00407e10937e4f32da788d78",
+      "bytes": 291412
     },
     {
       "path": "template-manifests/box-plot-chart.manifest.json",
@@ -359,6 +359,16 @@
       "path": "template-manifests/gantt-timeline-chart.manifest.json",
       "sha256": "0be38bca7218b140c3412d0b7a61752340a392efc6ba7b328b0697a8f1d1a6f3",
       "bytes": 4719
+    },
+    {
+      "path": "template-manifests/insight_barchart.manifest.json",
+      "sha256": "43d52827518e042121042b1681662189012f6507103fcee333b95a2b4c5c42e7",
+      "bytes": 3554
+    },
+    {
+      "path": "template-manifests/insight_unusualchange.manifest.json",
+      "sha256": "aa01d84f7d10b9595389a31494667b2ee6af3df2ebc0199a7adc9d937528f661",
+      "bytes": 3837
     },
     {
       "path": "template-manifests/kpi-text.manifest.json",
--- a/src/desktop/data/content-manifest.json
+++ b/src/desktop/data/content-manifest.json
@@ -252,8 +252,8 @@
     },
     {
       "path": "template-manifests.index.json",
-      "sha256": "03bde655ff89182d17b9d936c9e3c349193ec95d00407e10937e4f32da788d78",
-      "bytes": 291412
+      "sha256": "87d12282fb10db7c70431e7053882664696ac9bd26bf8392bfca005fb2ef5c6e",
+      "bytes": 292699
     },
     {
       "path": "template-manifests/box-plot-chart.manifest.json",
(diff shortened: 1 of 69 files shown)


src/desktop/data/templates/insight_barchart.xml
```diff
@@ -0,0 +1,53 @@
+<workbook xmlns:user='http://www.tableausoftware.com/xml/user'>
```
tonythebest, line 1: **Blocking — incomplete packaging.**

This XML lands under `data/templates/` only. The shipped provider (`getTemplateXmlFragment`) and vendored-asset invariants use `data/data-visualization-templates-xml/` (`TEMPLATE_XML_DIR`) and require the two dirs to be byte-identical for every manifest-backed template.

CI fails with `insight_barchart: no shipped XML → must declare golden.checkpoint_render provenance` (and dual-dir orphans).

Please mirror both `insight_*.xml` into `data-visualization-templates-xml/` and re-run `npx tsx src/scripts/buildTemplateManifests.ts` so `content-manifest.json` hashes regenerate — or declare `golden.checkpoint_render` and drop fast-path shipping. Same applies to `insight_unusualchange.xml`.

src/desktop/data/template-manifests/insight_barchart.manifest.json
```diff
@@ -0,0 +1,93 @@
+{
+  "template": "insight_barchart",
+  "family": "magnitude",
```
tonythebest, line 3: **High — magnitude family-nativity collapse.**

With `family: "magnitude"` + `fast_path_eligible: true`, magnitude becomes a 2-member eligible family. `familyNativeKeywords` then loses strict-majority nativity for `magnitude` / `absolute` (only carried by `magnitude-simple-bar`), so `selectWithinFamily` fail-closes.

Empirically: **exactly 7** previously-deterministic `magnitude-simple-bar` asks flip `bound → propose` (e.g. `magnitude chart of Sales by Category`, `magnitude chart of ARR by Industry`). `insight_unusualchange` alone does not cause this.

Sanctioned fix (same shape as the 2026-07-06 trend-line sibling-scaling remedy): add a distinctive magnitude chart-noun to `CHART_NOUN_KEYWORDS` so `magnitude-simple-bar` keeps a lone-winner selector, then add a corpus row asserting it still binds.

Avoid:
- `fast_path_eligible: false` — shape-validator rejects it for GREEN + `fixture_bind`, and `binder.ts` would escalate explicit `proposal.template` (your intended Pulse path).
- Sharing the `magnitude` keyword — restores a bind but can reroute to the wrong template on ties.

src/desktop/data/template-manifests/insight_barchart.manifest.json
```diff
@@ -0,0 +1,93 @@
...
+  "readiness": "GREEN",
+  "fast_path_eligible": true,
```
tonythebest, line 5: Confirm the intended route: Pulse insight types via **explicit** `proposal.template` only, or also free-form classify?

These are marked `fast_path_eligible: true`, but there is no corpus row targeting them, and hyphenated keywords like `top-contributors` may not match spaced phrasings — so happy-path AC for free-form is currently unproven. If explicit-only is the design, please say so in the PR (and still fix packaging + the magnitude nativity side effect above).

src/desktop/data/template-manifests/insight_unusualchange.manifest.json
```diff
@@ -0,0 +1,86 @@
+{
+  "template": "insight_unusualchange",
+  "family": "time-series",
+  "readiness": "GREEN",
+  "fast_path_eligible": true,
```
tonythebest, line 5: Needs the same dual-XML packaging fix as `insight_barchart` (mirror into `data-visualization-templates-xml/`).

Family `time-series` does **not** own the magnitude bind regression — once packaging is fixed, mainly re-pin inventory/coverage deliberately and confirm a natural ask (e.g. unusual-change phrasing) is covered if these stay fast-path.

src/desktop/data/content-manifest.json
```diff
@@ -2,11 +2,11 @@
   "_generated": true,
   "_generator": "src/scripts/buildTemplateManifests.ts",
   "_warning": "GENERATED FILE — do not hand-edit. Re-run `npx tsx src/scripts/buildTemplateManifests.ts`.",
-  "content_version": "2.59.1+content.2026-07-27",
+  "content_version": "2.60.3+content.2026-07-31",
```
tonythebest: After mirroring XML into `data-visualization-templates-xml/`, please regenerate this via `buildTemplateManifests.ts` (generator hashes that dir). Then update binder inventory pins (`EXPECTED_ELIGIBLE` 24→26, census 44→46, matrix counts) **last**, with per-pin justification — ideally after the magnitude chart-noun fix so the 7 `magnitude-simple-bar` binds stay green.

tonythebest (changes requested):
## Review (refreshed)

Authored against the enriched meta-review in `docs/reviews/review-nawar-PR-694-insight-authoring-templates.md`.

**Cannot approve yet.** Required `build` is still red (**21** failed tests / 8 files), and two blockers remain:

1. **Packaging split-brain** — XML only under `data/templates/`; bundled provider reads `data/data-visualization-templates-xml/`.
2. **Magnitude family-nativity collapse** — `insight_barchart` as a 2nd `fast_path_eligible` magnitude template flips 7 previously deterministic `magnitude-simple-bar` binds to fail-closed.

Re-pinning inventory counts alone would greenwash both. Inline notes below.

GUS: [W-23651207](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002gEv5JYAS/view). Slack context: [#proj-tab-tableau-studio-mrs](https://salesforce-internal.slack.com/archives/C0BH9E69DML/p1785544066682589).

tonythebest (approved):
Approving with the earlier inline threads left open as the remaining fix-ups (dual-XML packaging, magnitude chart-noun / nativity, then deliberate pin updates). Please land those before merge / dogfood — required build is still red until then.

Prior review with comments: https://github.com/tableau/tableau-mcp/pull/694#pullrequestreview-4833623893

mattcfilbert:
**On "update the test, or is it the wrong category?" — neither, mostly.** The category is right (magnitude / time-series). This PR is data-only (6 files under `src/desktop/data/`, no `.ts`), so every failure is new data hitting unchanged expectations. They fall into three buckets, plus one that isn't yours.

**1. Real bug — fix the data, not the test.** The two template XMLs shipped to `data/templates/`, but the manifest ground-truth checks and the runtime binder read `data/data-visualization-templates-xml/` (`manifest.ts:140`, `manifest.test.ts` `XML_DIR`). On the base branch those two dirs hold an identical filename set — they're kept in sync, and this PR wrote to only one. So `hasShippedXml('insight_barchart')` is false and the `no shipped XML → must declare golden.checkpoint_render provenance` assertion fails. Fix: add both XMLs to `data-visualization-templates-xml/` too and regenerate the content-manifest hashes. Worth raising with the team: why do two template-XML dirs exist, and should `buildTemplateManifests.ts` keep them in sync? If the generator is supposed to copy into both, that's the real fix, and this bug recurs until it lands.

**2. Expected pin bumps — just update them.** The fast-path eligible set is hardcoded (24 → 26 at `manifest.test.ts:449` and `bind-behavior-matrix.invariant.test.ts:39` `EXPECTED_ELIGIBLE`), and the corpus tallies move (bound 224→231, failedClosed 172→165, comparable 183→189). Adding two `fast_path_eligible` templates is supposed to require this — it's the regression lock working as designed.

**3. The real question — decide this before you re-pin.** Some pins flipped `bound → propose` / `null` (`magnitude chart of ARR by Industry`; the ranking-keyword tie rows). `insight_barchart`'s keywords `top-contributors / top-drivers / top-detractors` (family magnitude) collide with the ranking and magnitude families. If you re-pin those to `propose`/`null`, you'd be locking in a wrong bind — a new template stealing generic "top / magnitude" asks that used to bind deterministically, or turning them into ties. The PR body says these are selected by explicit `proposal.template`. If so, make them resolver-only and keep them out of generic keyword scoring, the way `spatial-symbol-map-latlon` is; then the old `bound` pins stay correct instead of moving the wrong way. That's the whole point of the `carrier-uniqueness` invariant.

**Not yours: the two byte-budget failures.** `server.desktop.test.ts:216`/`:529` (29518 over the ~29,479 cap) come from `getDesktopToolListEntry`, which serializes only static code — it never reads the manifests or template data. A data-only diff can't move that number, so this looks like pre-existing base red. Quick confirm: run those two tests on a clean `feature/desktop` with none of your data; if they're already red on base, raise it separately and don't let it block this PR.

Order of operations: fix the directory bug, decide the keyword-scoping question, then update the pins that remain.

*— posted by MattGPT on Matt Filbert's behalf*

PR #695, opened 2026-07-31
refactor(desktop): route worksheet/dashboard list reads through the shared read harness

lauren-jacksonSFDC:
## What

Route `list-worksheets` and `list-dashboards` through `runExternalApiReadTool`, the same shared External Client API read harness the other ~17 desktop reads already use. Before this, those two tools hand-rolled their own `resolveSession → getExecutor → error-wrap` boilerplate, so a fix applied at the harness (e.g. the route-missing → honest `endpoint-not-in-this-build` 404 mapping) reached every read *except* these two.

## Why

Goal 2 of the External API refactor: one central place for External API reads, so a fix lands everywhere instead of "some get fixed up and some don't." This is the read-path half; the apply-path readback gateway shipped separately (#691, Slice A).

## Decision the reviewer must accept

This is now a **rule the codebase keeps**: every External Client API read goes through `runExternalApiReadTool`. No behavior change for callers — the success payloads are identical; the only observable difference is that a missing route now yields the harness's honest 404 (added as a test on each tool).

Behavior-preserving; verified green (tsc + full lint + the two tools' unit tests, which now also assert the route-missing mapping).

### DIFF
--- a/src/tools/desktop/workbook/listDashboards.test.ts
+++ b/src/tools/desktop/workbook/listDashboards.test.ts
@@ -3,6 +3,7 @@ import { Err, Ok } from 'ts-results-es';
 import { z } from 'zod';
 
 import * as listDashboardsModule from '../../../desktop/commands/workbook/listDashboards.js';
+import * as sessionResolution from '../../../desktop/sessionResolution.js';
 import { DesktopCommandExecutionError } from '../../../errors/mcpToolError.js';
 import { DesktopMcpServer } from '../../../server.desktop.js';
 import invariant from '../../../utils/invariant.js';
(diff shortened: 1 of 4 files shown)


lauren-jacksonSFDC:
Folding this into #696. C1's goal — worksheet/dashboard list reads going through the shared `runExternalApiReadTool` harness — is fully absorbed by #696: it deletes the standalone `list-worksheets`/`list-dashboards` tools entirely and routes both reads through the same harness inside the `desktop-read` dispatcher (getting the same honest route-missing → 404 mapping). Nothing to port; closing to avoid a redundant, soon-mooted PR.

PR #700, opened 2026-07-31
@W-23653834 updating embedded auth fix

jarhun88:
## Description

TL;DR — the bug:
  
  The embedded OAuth server let an opaque client use any https:// redirect URI, even one it never registered. So an auth code could be sent to an attacker-controlled URL
  → account takeover.

  Why: the "redirect URI must be one you registered" check only ran for URL-style client_ids (the CIMD path). Opaque client_ids (the default) skipped it and hit only a
  format check — which happily accepts any https:// host.

  The fix: enforce the registered-redirect allowlist for opaque clients too —
  - registered client → redirect must match a stored URI;
  - unregistered client → only loopback/native schemes allowed, no remote hosts;
  - CIMD path unchanged.

Longer fix explanation:

  - /oauth2/register now mints a unique client_id and stores the client's validated redirect_uris (previously it discarded them and returned a fixed mcp-public-client).
  - /oauth2/authorize now validates redirect_uri against what the client registered:
    - Registered client → must match one of its stored URIs.
    - Unregistered client (or state lost on restart) → native-only fallback: loopback (localhost/127.0.0.1/[::1]) and custom-scheme URIs allowed; remote http/https
  rejected.
    - CIMD (URL client_id) path unchanged — it already enforces its own registered-redirect allowlist.
  - New helpers authorizeRedirectUri.ts and isNativeRedirectUri.ts.
  
  ## Motivation and Context

  Previously the registered-redirect allowlist was enforced only on the CIMD path; opaque client_ids fell back to a format-only check. This change applies consistent
  per-client redirect-URI enforcement to opaque clients as well, without affecting CIMD or native clients (Claude Desktop, MCP Inspector).


mattcfilbert (commented):
Solid and important fix — the opaque-client redirect allowlist bypass is real, and enforcing the per-client allowlist for opaque clients while leaving CIMD untouched and falling unknown clients back to native-only is the right shape. The helper split reads well, reusing the existing exact-match `matchesRegisteredRedirectUri` is the correct call, and the tests actually exercise the takeover case end-to-end (`https://evil.example.com/cb` → 400), not just the happy path. Thanks for writing it up clearly.

One thing I'd want resolved before merge — a contract inconsistency between register and authorize:

`/oauth2/register` still accepts a remote `https://` redirect URI (via `isValidRedirectUri`), stores it, and echoes it back as registered. But the registration lives only in an in-memory `ExpiringMap` with a 24-day hard TTL that does not slide on use (`ExpiringMap` doesn't override `get()`, so it's create-time, not last-use), is lost on restart, and is evicted oldest-first at the 10k cap. When any of those fires, the client_id becomes "unknown" → native-only fallback → its own previously-registered remote-https redirect is now rejected with `invalid_request`.

So a remote-hosted client is told "you're registered with `https://myapp.com/cb`," works for a while, then the identical request starts failing after a livesite restart, with re-registering for a new client_id as the only recovery. That's a silent availability regression for the exact class of client the register endpoint advertises support for.

Cleanest resolution is to pick one and make register and authorize agree:
- If remote-https redirects are supported, registration needs durability beyond in-memory/restart (or at minimum document "re-register after a server restart").
- If they aren't, `register` should reject remote-https for opaque clients too, mirroring what `authorize` now enforces on the unknown path.

Which is intended? The PR body names only native clients (Claude Desktop, Inspector, Cursor) as the unaffected set, which hints at the answer — worth making explicit.

Two smaller notes, non-blocking:

- `tokenEndpointAuthMethod` is stored on `ClientRegistration` but never read (set, never get). These are PKCE/public clients with no secret, so nothing consumes it today — forward-looking scaffolding, or can it come out?
- The loopback host set (`localhost` / `127.0.0.1` / `[::1]`) is now hardcoded in `isNativeRedirectUri`, `isValidRedirectUri`, and `matchesRegisteredRedirectUri` (which already has a `LOOPBACK_HOSTS` const). Worth hoisting to one shared constant so a future host tweak can't drift one check out of sync.

Not a blocker for the security posture: the 10k-cap eviction is oldest-inserted, so a flood of `/oauth2/register` calls could evict a legit registration and downgrade it to native-only — an availability nudge, not a takeover, since an attacker still can't remap a live client_id to their own URI. Folds into the durability question above.

—
_Posted by MattGPT on Matt's behalf._

mattcfilbert (approved):
Non-blocking comments that you _could_ address from robo sweep review sanity check but LGTM. Approving. Good stuff, man

PR #704, opened 2026-07-31
fix(desktop): embed features.json in SEA, add AUTH=passthrough

tableaukyler:
## What

Two fixes that let the desktop SEA build and a new passthrough auth mode work, plus the review fixups.

1. **`features.json` in the SEA blob.** A Single Executable Application has no filesystem next to the binary, so `ServerFeatureGate` was loading with all features off. `features.json` is now embedded into every SEA blob and read via `node:sea` at runtime, falling back to disk for normal builds and tests. SEA asset access is extracted into `src/utils/sea.ts` and shared.
2. **`AUTH=passthrough` auth type.** New `AUTH` value that implies `ENABLE_PASSTHROUGH_AUTH` and makes passthrough the only accepted method - requests without a token get `401 invalid_token` instead of falling back to another auth method.

## Decision the reviewer must accept

`AUTH=passthrough` is a distinct, stricter mode from `ENABLE_PASSTHROUGH_AUTH=true`: the latter falls back to the `AUTH` method when no token is present; the former has no fallback and rejects. This is a new rule the config now keeps, documented in `passthrough.md` and `env-vars.md`. `AUTH=passthrough` requires `TRANSPORT=http` (the passthrough middleware only exists on the HTTP path), and the config now fails fast on `passthrough` + `stdio`.

## Auth flow

```mermaid
flowchart TD
    R[Request] --> H{Token present?<br/>X-Tableau-Auth header<br/>or workgroup_session_id cookie}
    H -- yes --> P[Validate + pass through]
    H -- no --> M{AUTH mode}
    M -- passthrough --> E[401 invalid_token]
    M -- ENABLE_PASSTHROUGH_AUTH=true --> F[Fall back to AUTH method]
```

## Notes

- The passthrough token is read from the `X-Tableau-Auth` header, or the `workgroup_session_id` cookie when the header is absent. Docs now describe both.
- Related: [!114](https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/114) (Chiron web-MCP path).

## Verification

`tsc` clean, `npm run lint` clean, unit tests pass (`npm test`), e2e SEA guard passes (`npm run test:e2e` - `tests/e2e/seaAssets.test.ts`).

### DIFF
--- a/docs/docs/configuration/mcp-config/authentication/passthrough.md
+++ b/docs/docs/configuration/mcp-config/authentication/passthrough.md
@@ -9,16 +9,18 @@ Tableau REST APIs. The same
 [`X-Tableau-Auth` header](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_auth.htm#using_auth_token)
 used to authenticate to the Tableau REST APIs can also be used to authenticate to the MCP server.
 
-When a request is made to the MCP server, the `X-Tableau-Auth` header is read.
+When a request is made to the MCP server, the token is read from the `X-Tableau-Auth` header, or,
+when that header is absent, from the `workgroup_session_id` cookie.
 
-- When the header is present, the value will be "passed through" and re-used during MCP tool calls
-  when they authenticate to the Tableau REST APIs.
-- When absent, behavior depends on how passthrough was enabled:
+- When a token is present in either place, the value will be "passed through" and re-used during MCP
+  tool calls when they authenticate to the Tableau REST APIs.
+- When neither is present, behavior depends on how passthrough was enabled:
   - With [`ENABLE_PASSTHROUGH_AUTH`](#enable_passthrough_auth) set to `true`, normal authentication
     resumes as defined by the [`AUTH`](../env-vars.md#auth) environment variable. This allows clients
-    that do not provide the `X-Tableau-Auth` header to still authenticate to the MCP server.
-  - With [`AUTH`](../env-vars.md#auth) set to `passthrough`, there is no fallback: the request is
-    rejected with `401 invalid_token`.
+    that provide neither the `X-Tableau-Auth` header nor the `workgroup_session_id` cookie to still
+    authenticate to the MCP server.
+  - With [`AUTH`](../env-vars.md#auth) set to `passthrough`, there is no fallback to another `AUTH`
+    method: the request is rejected with `401 invalid_token`.
 
 ## Warnings
 
@@ -78,10 +80,10 @@ for more details.
 ### AUTH
 
 - Setting [`AUTH`](../env-vars.md#auth) to `passthrough` enables passthrough authentication and makes
-  it the only accepted method: every request must carry an `X-Tableau-Auth` token, and requests
-  without one are rejected with `401 invalid_token`.
+  it the only accepted method: every request must carry a token in the `X-Tableau-Auth` header or the
+  `workgroup_session_id` cookie, and requests without one are rejected with `401 invalid_token`.
 - Implies [`ENABLE_PASSTHROUGH_AUTH`](#enable_passthrough_auth) is `true`; you do not need to set both.
-- Unlike `ENABLE_PASSTHROUGH_AUTH`, there is no fallback to another `AUTH` method when the header is
+- Unlike `ENABLE_PASSTHROUGH_AUTH`, there is no fallback to another `AUTH` method when the token is
   absent.
 
 <hr />
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -951,6 +951,8 @@ describe('Config', () => {
   describe('Passthrough auth', () => {
     beforeEach(() => {
       vi.stubEnv('AUTH', 'passthrough');
+      vi.stubEnv('TRANSPORT', 'http');
+      vi.stubEnv('DANGEROUSLY_DISABLE_OAUTH', 'true');
       vi.stubEnv('PAT_NAME', undefined);
       vi.stubEnv('PAT_VALUE', undefined);
     });
@@ -977,12 +979,10 @@ describe('Config', () => {
       expect(() => new Config()).not.toThrow('The environment variable PAT_VALUE is not set');
     });
 
-    it('should work with stdio transport (no OAuth conflict)', () => {
+    it('should reject AUTH=passthrough with TRANSPORT=stdio', () => {
       vi.stubEnv('TRANSPORT', 'stdio');
 
-      const config = new Config();
-      expect(config.auth).toBe('passthrough');
-      expect(config.transport).toBe('stdio');
+      expect(() => new Config()).toThrow('TRANSPORT must be "http" when AUTH is "passthrough"');
     });
 
     it('should still require SERVER when AUTH=passthrough', () => {
--- a/src/config.ts
+++ b/src/config.ts
@@ -320,6 +320,10 @@ export class Config extends BaseConfig {
       );
     }
 
+    if (this.auth === 'passthrough' && this.transport === 'stdio') {
+      throw new Error('TRANSPORT must be "http" when AUTH is "passthrough"');
+    }
+
     if (this.auth === 'oauth') {
       if (disableOauthOverride) {
         throw new Error('When AUTH is "oauth", DANGEROUSLY_DISABLE_OAUTH cannot be "true"');
--- a/src/sdks/tableau/restApi.ts
+++ b/src/sdks/tableau/restApi.ts
@@ -330,10 +330,6 @@ export class RestApi {
   private _addInterceptors = (baseUrl: string, interceptors: AxiosInterceptor): void => {
     interceptors.request.use(
       (config) => {
-        // Tableau's REST API defaults to XML for some endpoints (e.g. /serverinfo)
-        // unless the client explicitly asks for JSON. Force it here so every
-        // request gets the JSON shape our Zod schemas expect.
-        config.headers.set('Accept', 'application/json');
         this._requestInterceptor?.[0]({
           baseUrl,
           ...getRequestInterceptorConfig(config),
--- a/src/server/passthroughAuthMiddleware.ts
+++ b/src/server/passthroughAuthMiddleware.ts
@@ -40,7 +40,8 @@ export function passthroughAuthMiddleware(): RequestHandler {
       if (config.auth === 'passthrough') {
         res.status(401).json({
           error: 'invalid_token',
-          error_description: 'Missing X-Tableau-Auth token',
+          error_description:
+            'Missing token: provide an X-Tableau-Auth header or workgroup_session_id cookie',
         });
         return;
       }
--- a/src/server/passthroughAuthMiddleware.test.ts
+++ b/src/server/passthroughAuthMiddleware.test.ts
@@ -53,6 +53,8 @@ describe('passthroughAuthMiddleware', () => {
       vi.resetModules();
       vi.unstubAllEnvs();
       stubDefaultEnvVars();
+      vi.stubEnv('TRANSPORT', 'http');
+      vi.stubEnv('DANGEROUSLY_DISABLE_OAUTH', 'true');
     });
 
     afterEach(() => {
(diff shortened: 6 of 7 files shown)


src/scripts/seaAssets.ts
```diff
@@ -12,6 +12,11 @@ export const MANIFEST_KEY = 'asset-manifest.json';
 
 export const DESKTOP_ASSET_DIRS: readonly string[] = ['resources/desktop', 'desktop/data'];
 
+// Single files (build-root relative) embedded into every SEA blob regardless of variant.
+// features.json is read by ServerFeatureGate at startup; a SEA has no filesystem next to
+// the binary, so it must ride in the blob or the feature gate loads with all features off.
+export const ALWAYS_EMBEDDED_FILES: readonly string[] = ['features.json'];
```
tonythebest, line 18: **Blocking — required CI is red on this change.**

`ALWAYS_EMBEDDED_FILES` embeds `features.json` at the build root, but `tests/e2e/seaAssets.test.ts` only accepts keys under `desktop/data/` or `resources/desktop/` and throws `embedded asset key under no known root: features.json` (job [91292479206](https://github.com/tableau/tableau-mcp/actions/runs/30672334365/job/91292479206)). Units are green; this e2e failure is the sole root cause.

Rather than excluding the key, could the e2e guard read build-root always-embedded files back through the SEA seam and assert round-trip? That positively proves this PR's goal. Importing `ALWAYS_EMBEDDED_FILES` into the test would also keep the embed list and the guard from drifting.

src/sdks/tableau/restApi.ts
```diff
@@ -330,6 +330,10 @@ export class RestApi {
   private _addInterceptors = (baseUrl: string, interceptors: AxiosInterceptor): void => {
     interceptors.request.use(
       (config) => {
+        // Tableau's REST API defaults to XML for some endpoints (e.g. /serverinfo)
+        // unless the client explicitly asks for JSON. Force it here so every
+        // request gets the JSON shape our Zod schemas expect.
+        config.headers.set('Accept', 'application/json');
```
tonythebest: This sets `Accept: application/json` on *every* request, and it replaces axios's default `application/json, text/plain, */*` rather than adding to it.

That also covers `queryViewImage` / `getCustomViewImage` (PNG/SVG, `responseType: 'arraybuffer'`) and the CSV endpoints `queryViewData` / `getCustomViewData`, which now advertise JSON-only. Tableau usually keys off `?format=`, so this may well be harmless, but nothing tests it.

Could we scope it — skip when `responseType === 'arraybuffer'`, or send `application/json, */*;q=0.9` — and add a test pinning the header for one JSON and one image endpoint?

src/config.ts
```diff
@@ -191,7 +191,7 @@ export class Config extends BaseConfig {
 
     this.enableMcpSiteSettings = enableMcpSiteSettings !== 'false';
     this.allowSitesToConfigureRequestOverrides = allowSitesToConfigureRequestOverrides === 'true';
-    this.enablePassthroughAuth = enablePassthroughAuth === 'true';
+    this.enablePassthroughAuth = enablePassthroughAuth === 'true' || auth === 'passthrough';
```
tonythebest, line 194: `AUTH=passthrough` with `TRANSPORT=stdio` is accepted by config (see the new config test), but the passthrough middleware only exists on the express path and `buildAuthConfig` returns null — so the first REST-backed tool call hits `invariant(tableauAuthInfo, 'Tableau auth info not provided.')` (`restApiInstance.ts:135`).

Worth failing fast in `Config` when `auth === 'passthrough' && transport === 'stdio'` (unless there is a planned token source on the stdio path).

src/config.test.ts
```diff
@@ -947,4 +947,48 @@ describe('Config', () => {
...
+
+    it('should not require PAT_NAME when AUTH=passthrough', () => {
+      expect(() => new Config()).not.toThrow('The environment variable PAT_NAME is not set');
+    });
+
+    it('should not require PAT_VALUE when AUTH=passthrough', () => {
+      expect(() => new Config()).not.toThrow('The environment variable PAT_VALUE is not set');
+    });
+
+    it('should work with stdio transport (no OAuth conflict)', () => {
```
tonythebest: `should work with stdio transport (no OAuth conflict)` only asserts `config.auth` / `config.transport` — it proves construction, not that anything works. The name promises more than it checks. Prefer renaming (e.g. `accepts AUTH=passthrough with TRANSPORT=stdio at construction`) or asserting the fail-fast once config rejects the combo.

docs/docs/configuration/mcp-config/authentication/passthrough.md
```diff
@@ -72,10 +75,22 @@ for more details.
 
 ## Environment Variables
 
+### AUTH
+
+- Setting [`AUTH`](../env-vars.md#auth) to `passthrough` enables passthrough authentication and makes
```
tonythebest, line 82: This says every request must carry an `X-Tableau-Auth` token with no fallback, but `passthroughAuthMiddleware` still accepts a `workgroup_session_id` cookie (`getHeader(...) || getCookie(...)` — pre-existing). Worth either rejecting the cookie when `auth === 'passthrough'`, or documenting the cookie as an accepted token source and adjusting the error text.

src/server/passthroughAuthMiddleware.ts
```diff
@@ -37,6 +37,13 @@ export function passthroughAuthMiddleware(): RequestHandler {
       getHeader(req, X_TABLEAU_AUTH_HEADER) || getCookie(req, 'workgroup_session_id');
 
     if (!tableauAccessToken) {
+      if (config.auth === 'passthrough') {
```
tonythebest, line 40: Nice fail-closed path for missing token under `AUTH=passthrough`.

Note the token is still sourced from header **or** `workgroup_session_id` cookie (line above). The new docs claim header-only — please align docs or code (see comment on `passthrough.md`).

tonythebest (changes requested):
## Review summary

Auth design for `AUTH=passthrough` looks sound and is the right shape for the Chiron web-MCP path ([!114](https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/114)): fail-closed, no static credentials, matches the spawn env.

**Cannot approve yet** — required `build` is red **because of this PR** (e2e `seaAssets.test.ts`: `embedded asset key under no known root: features.json`). That alone blocks merge.

Also worth fixing in the same pass (inline):
- global `Accept: application/json` blast radius on image/CSV endpoints
- `AUTH=passthrough` + `TRANSPORT=stdio` accepted but credential-less at tool call
- docs claim header-only while cookie still satisfies the middleware

### PR description nits
- The note about removing a dead `restApiVersion` guard in `getTableauServerInfo` does not appear in the diff (file unchanged).
- Verification line omits e2e (the suite that is red).
- Worth linking !114 / a `@W-` item.

**Landing order:** #704 green + merged → binary published/vendored → !112 → !114.

src/scripts/seaAssets.ts
tableaukyler: Fixed in 1762d603. The e2e guard now imports `ALWAYS_EMBEDDED_FILES` and, for build-root always-embedded files like `features.json`, asserts a round-trip through `readSeaAssetText` - the exact seam `ServerFeatureGate` reads it through in production. The embed list and the guard can no longer drift, and a genuinely stray key still throws.

src/sdks/tableau/restApi.ts
tableaukyler: Removed the change entirely in 1762d603. I tested against a real Cloud server: axios already sends `Accept: application/json, text/plain, */*` by default, which returns JSON from `/serverinfo` - so the forced header was a no-op on Cloud (couldn't reproduce the XML there with or without it). I also traced the request path and confirmed no incoming client header (Accept included) leaks onto the outbound REST call, so passthrough mode wasn't the cause either. The XML only shows up on older on-prem servers doing naive content-negotiation on the trailing `*/*`, which this MCP doesn't target. Dropping it removes the image/CSV blast radius you flagged.

src/config.ts
tableaukyler: Added the fail-fast in 1762d603: `Config` now throws `TRANSPORT must be "http" when AUTH is "passthrough"` for `passthrough` + `stdio`. No planned stdio token source, so this is the right call.

src/config.test.ts
tableaukyler: Fixed in 1762d603. Replaced it with `should reject AUTH=passthrough with TRANSPORT=stdio`, which asserts the new fail-fast throws. The test now checks behavior, not just construction.

docs/docs/configuration/mcp-config/authentication/passthrough.md
tableaukyler: Fixed in 1762d603. Chose to document the cookie rather than reject it (keeps parity with `workgroup_session_id` sessions): the docs now say the token is read from the `X-Tableau-Auth` header or, when absent, the `workgroup_session_id` cookie, and I reworded the "no fallback" line to mean no fallback to another *AUTH method*.

src/server/passthroughAuthMiddleware.ts
tableaukyler: Aligned docs + code in 1762d603 (see the `passthrough.md` thread). Also updated the 401 `error_description` to "Missing token: provide an X-Tableau-Auth header or workgroup_session_id cookie" so the message reflects both sources.

tableaukyler:
Thanks for the thorough review. Pushed `1762d603` addressing everything:

- **CI blocker (e2e `seaAssets.test.ts`):** the guard now round-trips build-root always-embedded files (`features.json`) through the `readSeaAssetText` seam instead of rejecting them; imports `ALWAYS_EMBEDDED_FILES` so the embed list and guard can't drift.
- **`Accept` blast radius:** removed the forced-JSON header entirely - it was a no-op on Cloud and the XML only affects older on-prem content-negotiation we don't target (details in the inline thread).
- **`passthrough` + `stdio`:** now fails fast in `Config`.
- **Docs/cookie mismatch:** documented the `workgroup_session_id` cookie as an accepted token source and aligned the 401 message.

PR description nits fixed: dropped the force-JSON section, removed the `restApiVersion` guard note (it was never in the diff), added e2e to the verification line, and linked !114. Skipped the `@W-` link for now - let me know if you want a specific work item attached.

tonythebest (approved):
Approving — the blocker and both High findings are genuinely fixed, and I verified them rather than going on the thread replies. Thanks for the thorough turnaround, especially dropping the `Accept` change once you'd established it was a no-op on Cloud; that was the right call over scoping it.

**Verified at `1762d603`:**

- **SEA e2e (was the blocker).** The guard now imports `ALWAYS_EMBEDDED_FILES` and asserts `readSeaAssetText(key) === expected` for build-root files — a real round-trip through the seam `ServerFeatureGate` uses, not an exclusion, and a stray key still throws. Ran it locally: `tests/e2e/seaAssets.test.ts` 2/2 pass, including `reads every embedded asset back through the verified seam`.
- **`passthrough` + `stdio` fail-fast.** `Config` throws `TRANSPORT must be "http" when AUTH is "passthrough"`, and the renamed test asserts the throw instead of construction.
- **Forced `Accept` removed** — header assignment and comment both gone, so the image/CSV blast radius is closed.
- **Docs/code aligned** on header-or-cookie, with `no fallback` correctly narrowed to "another AUTH method" and the 401 `error_description` updated to match.
- Ran the touched unit suites locally: 252/252 pass across `config.test.ts`, `passthroughAuthMiddleware.test.ts`, `buildAuthConfig.test.ts`.
- The description no longer carries the `restApiVersion` claim that wasn't in the diff, and it's down to two independent fixes instead of three.

**One thing to sort before merge — not a review finding.** "Build and Test" has not run on `1762d603`. `gh pr checks` shows 3 green (SAST, Credential Scanning, CLA), but the build workflow is *absent*, not passing — the only run on this branch is the failing one on the old head `babdc2d7`. No workflow at all fired on the new commit, including `pr-feature-label`, which did run before, so this looks like an Actions-side hiccup rather than anything in `ci.yml` (its `pull_request` trigger does cover `feature/desktop`). Worth re-triggering so the required check is green on the head — my local run is evidence, not a substitute for CI.

**Non-blocking, fine as follow-ups:**

- `features.json` is still read via `readSeaAssetText` while other assets go through `readVerifiedSeaAsset`, so its content isn't hash-checked against the manifest. The new e2e closes the coverage half of this; the verification half is still open, and a bad blob fails soft to "all features off" — the same symptom this PR fixes.
- Unknown/misspelled `AUTH` still silently falls back (`AUTH=passtrhough` → `pat` → a confusing `PAT_NAME is not set`). Pre-existing, but it costs more now that an exact spelling is load-bearing for the Studio consumer.
- `buildAuthConfig.ts` has `case 'oauth': return null;` and `case 'passthrough': return null;` adjacent with no comment, though they return null for different reasons.

W-23633781 being `Fixed` while its ACs land here and in !112/!114 is mine to clean up, not yours.

PR #705, opened 2026-07-31
fix(desktop): exempt published-datasource (sqlproxy) connections from connections-not-authorable

nawartamawi:
## Decision
`connections-not-authorable` should exempt `class='sqlproxy'` — the connection shape Tableau serializes for a **published (server/Cloud) datasource** — alongside `federated`. This is a new rule the codebase keeps: sqlproxy is a legitimate Desktop-emitted shape, not the hand-authored/copied-from-.tds connection the rule targets.

## Why
The rule rejects any bare top-level `<connection>` that isn't `class='federated'`. A published datasource serializes as a bare `<connection class='sqlproxy' composed-connection-name='sqlproxy.<id>' server='...online.tableau.com'/>` with no `<named-connections>` wrapper. That round-trips on a live readback, and the federated+named-connection id-minting scheme doesn't apply to server-published connections — so authoring against a published datasource was being false-rejected at preflight.

## Change
Add `and not(@class='sqlproxy')` to both selectors in the bare-connection check, plus a test that a bare sqlproxy connection is not flagged.

## Risk
The rule is a preflight helper, not a correctness or security check. Worst case: a genuinely broken sqlproxy connection fails when Tableau connects, instead of being caught early — a fair trade for enabling published-datasource authoring. Reviewed by Matt Filbert (rule owner), who confirmed nothing else relies on this rule catching sqlproxy. Note: sqlproxy now passes preflight with no second structural check (federated still gets the named-connection id check).

All 59 connections-not-authorable tests pass.

### DIFF
--- a/src/desktop/validation/rules/connectionsNotAuthorable.test.ts
+++ b/src/desktop/validation/rules/connectionsNotAuthorable.test.ts
@@ -73,6 +73,19 @@ describe('connections-not-authorable rule', () => {
     expect(connectionsNotAuthorableRule.validate(xml)).toEqual([]);
   });
 
+  it('a bare published-datasource (sqlproxy) connection is NOT rejected — it is a genuine Desktop shape', () => {
+    const xml = `<?xml version="1.0"?>
+<workbook>
+  <datasources>
+    <datasource caption="GUS-Work" name="sqlproxy.1bzmspu1v78e5817xecku1rmnq6s">
+      <connection channel="https" class="sqlproxy" composed-connection-name="sqlproxy.1bzmspu1v78e5817xecku1rmnq6s"
+        dbname="GUS-Work" port="443" server="10ax.online.tableau.com" username="user" />
+    </datasource>
+  </datasources>
+</workbook>`;
+    expect(connectionsNotAuthorableRule.validate(xml)).toEqual([]);
+  });
+
   it('a fragment with no <connection> element at all is never flagged', () => {
     const xml = `<?xml version="1.0"?>
 <worksheet name="Sheet 1">
--- a/src/desktop/validation/rules/connectionsNotAuthorable.ts
+++ b/src/desktop/validation/rules/connectionsNotAuthorable.ts
@@ -88,9 +88,13 @@ export const connectionsNotAuthorableRule: ValidationRule = {
     //
     // 1. A bare/legacy top-level connection that is NOT the modern federated wrapper —
     // exactly the hand-authored-from-.tds shape (known-bad).
+    //
+    // EXCEPTION: `class='sqlproxy'` is a published-datasource proxy — the shape Desktop
+    // serializes for a server/Cloud datasource. It round-trips on a live readback and the
+    // federated+named-connection minting scheme doesn't apply to it, so it is exempt.
     const bareConnections = xpath.select(
-      "/workbook/datasources/datasource/connection[not(@class='federated')] | " +
-        "/datasource/connection[not(@class='federated')]",
+      "/workbook/datasources/datasource/connection[not(@class='federated') and not(@class='sqlproxy')] | " +
+        "/datasource/connection[not(@class='federated') and not(@class='sqlproxy')]",
       doc as unknown as Node,
     ) as Element[];
     for (const conn of bareConnections) {


src/desktop/validation/rules/connectionsNotAuthorable.ts
```diff
@@ -88,9 +88,13 @@ export const connectionsNotAuthorableRule: ValidationRule = {
...
     // 1. A bare/legacy top-level connection that is NOT the modern federated wrapper —
     // exactly the hand-authored-from-.tds shape (known-bad).
+    //
+    // EXCEPTION: `class='sqlproxy'` is a published-datasource proxy — the shape Desktop
+    // serializes for a server/Cloud datasource. It round-trips on a live readback and the
+    // federated+named-connection minting scheme doesn't apply to it, so it is exempt.
     const bareConnections = xpath.select(
-      "/workbook/datasources/datasource/connection[not(@class='federated')] | " +
-        "/datasource/connection[not(@class='federated')]",
+      "/workbook/datasources/datasource/connection[not(@class='federated') and not(@class='sqlproxy')] | " +
```
tonythebest, line 96: Non-blocking, but worth knowing before this lands: **#666** ("fix(desktop): allow published-datasource (sqlproxy) connections through preflight", Matt) makes this exact change to these exact two files, and merged on 2026-07-28 — but into `claude/v12-fixes`, and it never reached `feature/desktop`. I confirmed the base here (`fe91dd77`) still carries the federated-only selector, so this PR is genuinely needed; it isn't redundant.

The catch is that #666 is a superset of this change. It also:

- rewrites the module header and the rule's `description` string — both still say only `federated` + a Desktop-minted named-connection ever applies, which now contradicts the code right below them (harmless at runtime: `description` isn't consumed anywhere, so it's documentation-only);
- adds a test that a top-level connection with **no** `class` attribute is still rejected — a real regression guard for exactly this selector edit;
- adds a test that *deliberately* permits a hand-authored bare sqlproxy with suspicious attributes, carrying the rationale for the risk you flag in the description.

Would you consider cherry-picking #666 here instead of landing the narrower version? Otherwise two variants of one decision exist on two branches and will collide in these same hunks when they converge — and whichever way that resolves could silently drop Matt's docs and those two tests. Not blocking the approval either way.

(Style only: #666 writes this as `not(@class='federated' or @class='sqlproxy')`; yours is the equivalent `not(...) and not(...)`.)

src/desktop/validation/rules/connectionsNotAuthorable.test.ts
```diff
@@ -73,6 +73,19 @@ describe('connections-not-authorable rule', () => {
     expect(connectionsNotAuthorableRule.validate(xml)).toEqual([]);
   });
 
+  it('a bare published-datasource (sqlproxy) connection is NOT rejected — it is a genuine Desktop shape', () => {
```
tonythebest, line 76: Two coverage notes, neither blocking.

**1. Inline literal vs. a captured readback.** This pins "the string `sqlproxy` is exempt" rather than "a real unchanged published-datasource read-back round-trips." Worth flagging that no `.xml` fixture in the repo contains `sqlproxy` — the one captured fixture (`superstore-scratch-ref.xml`, behind `LIVE_READBACK_FIXTURE`) is a local Superstore workbook, i.e. the federated shape. So this gap predates your PR, and #666 doesn't close it either (its sqlproxy test is also an inline literal, just routed through `runValidation`). If a captured published-DS readback is easy to grab, dropping one in as a fixture would make this rule's guarantee real rather than asserted.

**2. The exemption is case-sensitive and unpinned.** XPath 1.0 `=` is an exact string compare, so `class="SQLPROXY"` and `class="SqlProxy"` are still rejected — I verified that's the current behavior and I think it's the right one. But nothing asserts it, so a future "make class matching case-insensitive" refactor could widen the exemption without turning a test red. One negative assertion would lock it in.

tonythebest (approved):
Approving — the change is correct, correctly scoped, and I verified end-to-end that it actually closes the blocker. Two non-blocking notes inline, plus one on the description below.

**What I verified locally** (clean worktree at `d618661`, Node 22.23.1):

- Your "all 59 connections-not-authorable tests pass" is accurate — 59/59, re-run independently. The wider `src/desktop/validation/` suite is green too (444 tests).
- **The exemption is correctly scoped**, which was the thing I most wanted to be sure of. Non-sqlproxy bare connections (`excel-direct`), connections with *no* `class` attribute, and the federated named-connection path all still reject. Nothing leaked.
- **It closes the blocker end-to-end, not just in this rule.** `runValidation` over a published-datasource workbook returns zero issues across *every* registered rule, in both the `workbook` and `datasource` contexts — so there's no second rule quietly still rejecting a published DS. That felt like the question worth answering, since this has been blocking `bind-template` on published datasources since 07-28.

**On the description:** it says "This is a new rule the codebase keeps." I think it's actually a *re-landing* — Matt merged the same decision in #666 on 07-28 (details in the inline comment). Worth restating and linking it so nobody re-litigates it later; `CONTRIBUTING.md` also asks bug fixes to reference the tracking issue.

**On your `## Risk` note** that "sqlproxy now passes preflight with no second structural check" — confirmed, a fabricated `<connection class="sqlproxy" server="evil"/>` does pass. Worth recording that #666 carries an explicit test *deliberately* permitting that shape, with the rationale "Desktop accepted this live … preflight is not the acceptance gate." So it reads as settled design rather than an oversight. No change requested — just noting it so the reasoning lives here too.

Thanks for calling out the weakness of your own change in the description, and for leaving the `EXCEPTION:` comment in the code. Both made this much faster to review.

PR #637, opened 2026-07-24
@W-22817882 Add opt-in Tableau Prep flow run tools

ccrick-tableau:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

_This change is tracked by W-22817882._

These tools wrap the Tableau Flows REST API through the MCP server's `FlowsMethods` SDK. They operate against a Tableau Server or Tableau Cloud site and use Tableau's public Flow REST API:

https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_flow.htm

## Description

Adds controlled Tableau Prep flow-run operations to the MCP server:

- `run-flow` — requests an on-demand flow run by flow id and returns the asynchronous job
- `run-flow-task` — requests an existing scheduled flow task by task id and returns the asynchronous job
- `get-flow-task` — retrieves one scheduled flow task directly by task id
- `cancel-flow-run` — requests cancellation of a queued or in-progress flow run by flow-run id

Supporting changes include:

- REST API endpoints, SDK methods, response schemas, and tests
- OAuth scope wiring for flow-run, flow-task, and cancellation operations
- A two-level safety gate:
  - `FLOW_TOOLS_ENABLED=true` enables the flow tool family
  - `FLOW_WRITE_TOOLS_ENABLED=true` is the second opt-in for mutating tools
- Mutating flow tools are unavailable unless both flags are enabled
- Fail-closed bounded-context handling for project and tag restrictions
- Destructive and non-idempotent MCP annotations for run and cancellation tools
- Redaction of flow parameter override values from invocation logging
- Rejection of an explicit empty `outputStepIds` list so it cannot silently broaden to all outputs
- Structured Tableau error mapping, including HTTP 200 error envelopes from Cancel Flow Run
- REST API version checks for the Run Flow and Cancel Flow operations
- Documentation, eval coverage, and server-registration tests
- list-flow-tasks filter validation now returns MCP errors for malformed filters and handles nested schedule date fields.

The supported configuration states are:

| `FLOW_TOOLS_ENABLED` | `FLOW_WRITE_TOOLS_ENABLED` | Registered flow tools |
| --- | --- | --- |
| `false` | `false` or `true` | None |
| `true` | `false` | Read-only flow tools |
| `true` | `true` | Read-only and mutating flow tools |

## Motivation and Context

[...]

PR #627, opened 2026-07-23
@W-22794435 Add describe-flow tool to summarize a Prep flow's document

showmethecold:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

_Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of
the pull request._

# Pull Request Template

## Description

Adds a read-only `describe-flow` tool that summarizes a Tableau Prep flow’s inputs, transformations, outputs, lineage, connections, parameters, and optional field schemas.

Also adds the sanitized flow-document API client, OAuth scopes, feature gating, tests, and documentation.

## Motivation and Context

`get-flow` provides catalog metadata but cannot explain a flow’s internal logic or data movement. `describe-flow` enables agents to answer those questions without parsing raw `.tfl` or `.tflx` files or exposing sensitive connection information.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe):

## How Has This Been Tested?

- Unit tests for summarization, API handling, scopes, feature gating, and errors.
- Full lint, typecheck, and unit-test suite (2,533 tests).
- Live end-to-end validation against a Tableau server, including field schemas and the not-found path.

## Related Issues

[Support Tableau Prep flow design summaries with describe-flow](https://github.com/tableau/tableau-mcp/issues/626)

## Checklist

- [X] I have updated the version in the package.json file by using `npm run version`. For example,
      use `npm run version:patch` for a patch version bump.
- [X] I have made any necessary changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have documented any breaking changes in the PR description. For example, renaming a config
      environment variable or changing its default value.

## Contributor Agreement

By submitting this pull request, I confirm that:

- [X] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) for this project and followed
      its Contribution Checklist.

PR #708, opened 2026-08-03
@W-23631239: Surface flow-run failure reasons in list-flow-runs and g…

### LINKED ISSUE #707: Surface flow-run failure reasons in list-flow-runs and get-flow

When a Prep flow run fails, neither `list-flow-runs` nor `get-flow` can say why. They
report `status: "Failed"` and nothing more, so an agent cannot answer the first question
a user actually asks: "why did my flow run fail?"

Today the only recourse is `mcp.failureInsight`, a deep link into the Tableau UI. That
ends the conversation and pushes the person into a browser to read the reason themselves.
It was the right fallback when the reason was only reachable through the background job,
which requires admin permissions — non-admin callers had no path to a cause at all.

Related questions are equally unanswerable: is this one systemic problem or many unrelated
ones, and which flows are affected?

REST API 3.30 adds `failureReason { available, message }` to failed flow runs, sourced from
the flow's output-step errors rather than the admin-only background job — so any caller who
can see a run can now read why it failed. I'd like both tools to surface it:

1. Add `failureReason` to the flow-run schema, so the message rides along on each failed run.
2. Add `mcp.failureSummary` to both tools: the window's failures grouped by reported message,
   each group carrying how many runs reported it and which distinct flows it hit.
3. Gate the existing `mcp.failureInsight` deep link so it appears only when nothing in the
   window resolves a cause, making the UI hand-off a genuine last resort instead of the
   default answer.
4. No behavior change when no reason comes back, so servers below 3.30 and windows without
   failures keep their current output.

The summary shape, which lets an agent tell one systemic problem across many flows from many
unrelated ones without reading every record:

{
  "failedRunCount": 7,
  "failedFlowCount": 3,
  "reasons": [
    {
      "message": "Flow processing error",
      "available": true,
      "runCount": 5,
      "flowIds": ["<luid-a>", "<luid-b>"]
    }
  ]
}

showmethecold:
## Description

Surfaces REST API 3.30's `failureReason` on failed flow runs in `list-flow-runs` and
`get-flow`, so both tools can report why a run failed instead of only that it did.

- Adds `failureReason` to the flow-run schema. `available` accepts a real boolean or the
  strings `"true"`/`"false"`, and a malformed reason degrades to absent rather than failing
  the whole Get Flow Runs call.
- Adds `mcp.failureSummary` to both tools: the window's failures grouped by reported
  message, with `runCount` and the distinct `flowIds` affected. Both tools share one
  grouping helper, so their summaries are identical for the same runs.
- Gates the existing `mcp.failureInsight` deep link so it is emitted only when nothing in
  the window resolves a cause, saving a Query Flow call on the common path.

No breaking changes. Output is unchanged when no reason comes back, i.e. servers below 3.30
or windows with no failures.

## Motivation and Context

A failed flow run reported only `status: "Failed"`, so an agent could not answer the first
question users ask — "why did my flow run fail?" The only recourse was a deep link that
pushes the person into the Tableau UI. The reason previously required the admin-only
background job; 3.30 sources it from the flow's output-step errors instead, so any caller
who can see a run can now read it.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- **Unit:** new tests for the schema (boolean and string `available`, malformed-reason
  degradation), the grouping helper, and both tools — covering the deep-link gate in both
  directions and a regression test for the unchanged no-reason path. Full suite green:
  2736 tests, plus `eslint` and `tsc --noEmit`.
- **Live:** verified against a 3.30+ site with 39 assertions covering schema parsing,
  summary invariants, the gate in both directions, cross-tool summary identity, and the
  no-failure paths.

## Related Issues

Closes #707

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.0.6",
+  "version": "4.0.7",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/tools/web/flows/flowRunFailure.ts
+++ b/src/tools/web/flows/flowRunFailure.ts
@@ -1,0 +1,161 @@
+import { FlowRun } from '../../../sdks/tableau/types/flow.js';
+
+/**
+ * One distinct failure cause within a window of flow runs, with how many runs and
+ * which flows it affected. `available: false` means Tableau could not determine
+ * the cause and `message` is a generic placeholder.
+ */
+export type FlowRunFailureReasonGroup = {
+  message: string;
+  available: boolean;
+  // Number of `Failed` runs in the window that reported this exact reason.
+  runCount: number;
+  // DISTINCT flows affected by this reason, ascending. The blast radius of one cause.
+  flowIds: string[];
+};
+
+/**
+ * Failures for a window of flow runs, grouped by reported message so a caller can
+ * tell one systemic problem from many unrelated ones without reading every record.
+ *
+ * `failedRunCount` / `failedFlowCount` describe ALL `Failed` runs in the window,
+ * including any that carry no `failureReason` — deliberately the same definition
+ * `buildFailureInsight` uses, so the two fields never disagree about one window.
+ * Consequently `sum(reasons[].runCount) <= failedRunCount`.
+ */
+export type FlowRunFailureSummary = {
+  failedRunCount: number;
+  failedFlowCount: number;
+  reasons: FlowRunFailureReasonGroup[];
+};
+
+// Joins the two group-key fields. A character that cannot appear in a message, so
+// the key stays unambiguous even if the prefix is ever widened beyond the two
+// boolean spellings (which alone could not collide).
+const GROUP_KEY_SEPARATOR = '\u0000';
+
+/**
+ * The window's failed runs. Shared with `buildFailureInsight` so `failedRunCount`
+ * and `failedFlowCount` have ONE definition and the two fields can never disagree
+ * about the same window.
+ */
+export function getFailedRuns(flowRuns: FlowRun[]): FlowRun[] {
+  return flowRuns.filter((run) => run.status === 'Failed');
+}
+
+/** Distinct flows among the given runs, skipping any run with no `flowId`. */
+export function countDistinctFlows(runs: FlowRun[]): number {
+  return new Set(runs.map((run) => run.flowId).filter((id): id is string => id !== undefined)).size;
+}
+
+function compareStrings(a: string, b: string): number {
+  if (a < b) {
+    return -1;
+  }
+  return a > b ? 1 : 0;
+}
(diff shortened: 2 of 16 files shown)


src/tools/web/flows/flowRunFailure.ts
```diff
@@ -0,0 +1,161 @@
...
+    { message: string; available: boolean; runCount: number; flowIds: Set<string> }
+  >();
+
+  for (const run of failedRuns) {
+    const reason = run.failureReason;
+    if (!reason) {
+      continue;
+    }
+
+    const key = `${String(reason.available)}${GROUP_KEY_SEPARATOR}${reason.message}`;
```
ccrick-tableau, line 89: We mention in the docs that we group by message, but the grouping is actually message + whether we could determine root cause.

I think the change would be in the docs, but this probably isn't super important, since the LLMs should be able to reason about it?

A nit, but wouldn't block merge, imo.

ccrick-tableau (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.0.6",
+  "version": "4.0.7",
```
jarhun88, line 4: Required: this repo bumps `package.json`'s version automatically post-merge (`.github/workflows/auto-version-bump.yml`) — manual edits in a PR aren't the convention here (see this repo's root CLAUDE.md, "Versioning & Releases"). Please drop the manual 4.0.6→4.0.7 bump (and the matching package-lock.json change) from this PR; the checklist item about `npm run version` doesn't apply to this repo's actual workflow.

— Jaehun Bot

jarhun88:
**CI:** all checks green (build 24.x + 22.7.5, SAST, credential scan, test deployment, check-version, check-title, CLA).

**Merge status:** `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY` as of this review — the branch needs a rebase/merge against current `main` before it can land.

**Nit (informational, not blocking):** `summarizeFlowRunFailures` / `flowRunFailureReasonSchema` parse `failureReason` directly off the Get Flow Runs response with a dedicated Zod schema (`flowRunFailureReasonSchema` + `looseBooleanFalsy`, with `.catch(undefined)` degradation) — this is a genuinely different problem from the error-envelope parsing in the (still-open, unmerged) sibling PRs #627 (`getTableauErrorCode`) and #637 (`TableauRestError`), since `failureReason` is a normal response field on a successful call rather than a 200-body error envelope. No overlap/duplication concern here, but flagging so whoever merges #627/#637/#708 in sequence is aware there are now three separate "how do we get an error message out of a flow-run-shaped response" code paths in the flows tooling, in case a future consolidation is worth doing.

— Jaehun Bot

ccrick-tableau (approved):

PR #531, opened 2026-07-16
@W-22586635 Add list-flow-runs and list-flow-tasks tools

showmethecold:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

_Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of
the pull request._

# Pull Request Template

## Description

Adds two read-only Tableau Prep tools:

list-flow-runs — site-wide run history, newest-first with a default limit of 100. Reports completeness via mcp.resultInfo and provides a best-effort mcp.failureInsight link for failed runs.
list-flow-tasks — scheduled runFlow tasks with filtering, sorting, and limits.
Both document caller-permission caveats for non-admin users.

Also adds supporting REST APIs/types, shared filtering utilities, OAuth scopes, tests, and documentation.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe):

## How Has This Been Tested?

Unit tests and Manual end-to-end validation

## Related Issues

Closes https://github.com/tableau/tableau-mcp/issues/532

## Checklist

- [X] I have updated the version in the package.json file by using `npm run version`. For example,
      use `npm run version:patch` for a patch version bump.
- [X] I have made any necessary changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I have documented any breaking changes in the PR description. For example, renaming a config
      environment variable or changing its default value.

## Contributor Agreement

By submitting this pull request, I confirm that:

- [X] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) for this project and followed
      its Contribution Checklist.

PR #711, opened 2026-08-03
@W-23567579 - Remove deprecated resolve-datasource-luid tool

tonythebest:
## Summary

Removes the standalone `resolve-datasource-luid` web tool, now redundant.

**Why it's redundant (verified on `main`, this PR's base):**
- `generate-insight-cards` already accepts `{ luid }` directly **and** resolves a `contentUrl` internally (`generateInsightCardsTool.ts:137` — `datasource: string | { luid }`), so no separate resolve step is needed.
- If a web-side `contentUrl` -> LUID lookup is ever needed, `list-datasources` already supports it via `filter: contentUrl:eq:<contentUrl>` (`datasourcesFilterUtils.ts` — `contentUrl` with `eq`/`in`); its `id` is the LUID.
- In the **Desktop** flow (on `feature/desktop`), the LUID is fed straight from `list-workbook-datasources` (#632), which surfaces the resolved server LUID that monolith [#60876](https://github.com/sf-analyticscloud/monolith/pull/60876) (merged to `main` 2026-07-24) now emits on `GET /v0/workbook/datasources`. Confirmed with the monolith change author.

**Bounded-context / no-oracle semantics (AC).** `resolve-datasource-luid` returned an *identical* error for "absent" vs "exists but outside the bounded context" (a no-existence-oracle defense). The recommended `list-datasources` `contentUrl:eq` replacement enforces the same bounded context (`constrainDatasources`) so no LUID leaks, but returns **two distinguishable** messages (absent vs all-filtered-out) — a caller can tell them apart. **This is not a regression:** `list-datasources` has always behaved this way on `main`, and this tool was `disabled: !insightsToolsEnabled` (off by default) while `list-datasources` is always registered, so the oracle was already the reachable behavior. Noted here for the paper trail; a follow-up to unify the two `constrainDatasources` messages would close it if the property is worth keeping.

## Changes
- Delete `resolveDatasourceLuid.ts` + its test.
- Remove every reference: web tool registry (`tools.ts`), tool-name union + datasource group (`toolName.ts`), OAuth scope map (`scopes.ts`), `INSIGHTS_TOOLS_ENABLED` descriptions (`config.ts`, `createClaudeMcpBundleManifest.ts`), the server/e2e registration expectations.
- Version bump `4.0.6` -> `4.0.7`.

No behavior change to `generate-insight-cards` or `list-datasources`.

## Test plan
- [x] `tsc --noEmit` — clean
- [x] `vitest run` full unit suite — **2671 tests / 164 files**, all green
- [x] `eslint` on changed files — clean
- [x] **Fed-luid E2E (real 10ax/techandprod):** built the server from this branch (confirmed `resolve-datasource-luid` absent from `tools/list` and the bundle), pulled a live LUID from `list-datasources`, then `generate-insight-cards({ datasource: { luid } })` -> `isError=false` with governed cards (`GUS-Work` "Age Minute vs prior period", -77.6% Aug MTD). Proves the insight flow works with the LUID fed directly, no resolve step.

## Follow-up
- `tab-agent-south` still names `resolve-datasource-luid` in its web-MCP allow-list / `INCLUDE_TOOLS` (`_WEB_MCP_ALLOWED_TOOLS`, `web_mcp_bundle.py`) + a test. Harmless today (the bundled `tableau-mcp-desktop` binary is built from `feature/desktop`, which still has the tool), but must be dropped when this deletion reaches `feature/desktop` — tracked by [W-23701788](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002gTy6oYAC/view).


tonythebest (commented):
Independent review pass on `f5b5d56` (self-authored PR, so I ran the verification from scratch rather than trusting the test plan, and had two independent reviewers cross-check the findings).

**Verdict: the deletion is correct and complete — one mechanical thing blocks merge.**

Verified at head: zero remaining references to `resolve-datasource-luid`/`resolveDatasourceLuid` anywhere in the tree, `tsc --noEmit` clean, full unit suite green at **164 files / 2671 tests**, `eslint` clean on all 7 surviving changed files.

What actually makes a 202-line deletion across 9 files low-risk is worth stating, because it's stronger than the test count: three structures are exhaustively typed over the same `WebToolName` union — `webToolNames` (source of the union), `webToolGroups` (`as const satisfies Record<WebToolGroupName, Array<WebToolName>>`), and `toolScopeMap` (`Record<WebToolName, …>`). A partial deletion could not have compiled, so the OAuth-scope removal and the `datasource` group edit weren't judgment calls that could have been missed.

---

### Blocking: `check-version` is red — `package.json` wasn't bumped

`4.0.6` on both this head and `origin/main`. `version-bump-check.yml` only skips when a PR touches nothing under `src/**`, and this touches seven files there. `npm run version:patch` → `4.0.7` clears it. Everything else is green (both `build` jobs, `check-title`, `Test deployment`, SAST, credscan, CLA).

One judgment call while you're in there: the workflow's own guidance reserves `major` for breaking changes, and this removes a tool from a published npm package, the Docker image, and the Claude MCP bundle manifest. Patch is defensible — the tool was `disabled: !config.insightsToolsEnabled` so no default-config consumer could see it, and this repo has shipped tool *additions* as patches before (#531 `3.5.4→3.5.5`, #502 `3.1.1→3.1.2`, though #625 went `3.5.6→3.6.0`). Not worth blocking on; flagging only so `4.0.7` vs `4.1.0` is a deliberate choice.

### PR description: bullet 1 cites a tool that isn't on `main`

> `list-workbook-datasources` already surfaces that `luid`

`list-workbook-datasources` (#632) is a **desktop** tool. `git grep` finds it on `origin/feature/desktop` (`server.desktop.ts`, `tools/desktop/toolName.ts`, `listWorkbookDatasources.ts`) and **nowhere on `origin/main`**, which is this PR's base.

The argument survives without it — I confirmed bullets 2 and 3 both hold on `main`: `generateInsightCardsTool.ts:137` declares `datasource: z.union([z.string().nonempty(), z.object({ luid: z.string().nonempty() })])`, and `datasourcesFilterUtils.ts:17,52` has `contentUrl` in `FilterFieldSchema` with operators `['eq','in']`. Just worth rewording so the lead justification points at this branch.

### Follow-up worth tracking: the `tab-agent-south` allow-list will drift silently

`tab-agent-south` still names this tool in six places, including `_WEB_MCP_ALLOWED_TOOLS` (`agent_session.py:302`), the `INCLUDE_TOOLS` tuple (`web_mcp_bundle.py:69`), and `tests/test_agent_session.py:432` which *asserts* it's present.

**Nothing breaks today** — `web_mcp_bundle.py` spawns the shipped `tableau-mcp-desktop` binary (#683), built from `feature/desktop`, which still has the tool.

The problem is what happens when the deletion reaches that branch. `overridableConfig.ts:249-252` parses `INCLUDE_TOOLS` with `isWebToolName(v) ? v : isWebToolGroupName(v) ? webToolGroups[v] : []` — an unrecognized name `flatMap`s to `[]`. No throw, no warning, no log. The three-entry tuple quietly becomes two, and the Python test keeps passing because it asserts against the local constant rather than the live tool list. Nothing will surface the drift on its own, and I couldn't find a GUS story carrying that cleanup.

Relatedly, three comments on `feature/desktop` will dangle after the merge (`desktop/metadata/field-builder.ts:284`, `desktop/metadata/types.ts:34`, `tools/desktop/fields/listAvailableFields.ts:91`). Comments only, nothing breaks. Expect merge conflicts in `config.ts`, `scopes.ts`, `toolName.ts`, and `tools.ts` though — those four already diverge between the branches in the same regions this PR edits.

### Nits

- The test plan reports "1400 passed"; `npm test` is actually **2671 tests across 164 files**. Both green — the recorded number just understates what was verified.
- W-23567697 is `Closed` but #633, which implements it, is still open against `feature/desktop`. And W-23567579's description still names `list-datasources resolveContentUrl` as the migration precondition — that string exists on neither branch, and the monolith #60876 rationale supersedes that plan anyway. Worth updating one or the other so the next reader isn't misled.

Leaving this as a comment rather than an approval since the version bump is still outstanding — happy to approve once `check-version` is green.

src/tools/web/datasources/resolveDatasourceLuid.ts
```diff
@@ -1,84 +0,0 @@
...
-                pageSize: 100,
-                pageNumber: 1,
-              });
-
-              const exact = response.datasources.find((d) => d.contentUrl === contentUrl);
-              // Return an IDENTICAL error for "absent" and "exists but outside the
-              // bounded context" so a caller cannot enumerate which non-allowed
-              // datasources exist (404-style, not 403 — the distinction itself is
-              // an existence oracle). Only run the allow-list check when a match
-              // exists (we need its LUID), and collapse both denials to one error.
```
tonythebest, line 56: This is the one behavioral property I'd want acknowledged somewhere before it disappears, because the story's AC explicitly asks to *"confirm no behavioral regression for bounded-context and no-oracle semantics."*

The replacement the description recommends — `list-datasources` with `filter: contentUrl:eq:<contentUrl>` — **does** enforce the bounded context (`constrainDatasources`, `listDatasources.ts:182-195`, filters on `projectIds`/`datasourceIds`/`tags`), so no LUID leaks. But it returns **two distinguishable messages** (`listDatasources.ts:174-205`):

- zero rows from REST → `"No datasources were found. Either none exist or you do not have permission to view them."`
- rows returned but all filtered out by bounded context → `"The set of allowed data sources that can be queried is limited by the server configuration. While data sources were found, they were all filtered out by the server configuration."`

So a caller *can* tell "this contentUrl doesn't exist" from "this contentUrl exists but is outside your bounded context" — the exact distinction this comment was written to collapse.

**To be clear about severity, because I don't think this should block:** it isn't a regression. `list-datasources` has always answered this way on `main`, so the diff doesn't create the oracle. And this tool was `disabled: !config.insightsToolsEnabled` — off by default — while `list-datasources` is always registered, so in practice the oracle was already the reachable behavior and this defense was guarding the narrower opt-in surface while the broad one leaked anyway.

It's a paper-trail gap rather than a defect. A sentence in the PR description noting the difference would satisfy the AC; a follow-up to unify the two `constrainDatasources` messages would close it properly, if the property is considered worth keeping.

src/tools/web/datasources/resolveDatasourceLuid.test.ts
```diff
@@ -1,91 +0,0 @@
...
-    mocks.mockListDatasources.mockResolvedValue({ datasources: [] });
-    mocks.mockIsDatasourceAllowed.mockResolvedValue({ allowed: true });
-    const absent = await getToolResult('GUS-Work');
-
-    expect(denied.isError).toBe(true);
-    expect(absent.isError).toBe(true);
-    invariant(denied.content[0].type === 'text');
-    invariant(absent.content[0].type === 'text');
-    expect(denied.content[0].text).toContain('No datasource matched contentUrl');
-    expect(denied.content[0].text).toBe(absent.content[0].text);
```
tonythebest, line 80: Worth noting on its own: this is the **only** test in the repo that pins the no-existence-oracle property, and `expect(denied.content[0].text).toBe(absent.content[0].text)` is the assertion that names it.

Deleting the tool obviously deletes its tests — that's correct here, no argument. The cost is that the *decision* goes with it. A future reader of `constrainDatasources` has nothing telling them the two-message shape was ever considered a problem.

No action needed on this PR if the no-oracle property is genuinely being dropped as a goal. If it isn't, this assertion is the thing worth relocating.

src/tools/web/datasources/resolveDatasourceLuid.ts
tonythebest: Correcting myself on this thread — a cross-check caught that I overstated it, and the correction meaningfully shrinks the ask.

**The no-oracle property is not actually lost.** I checked `list-datasources` and stopped there, but that's only the *third* bullet's path. The **primary** path — `generate-insight-cards`, which is what actually consumes the LUID — preserves the property intact. Its internal `resolveDatasource` (`generateInsightCardsTool.ts:404`) returns `null` for both cases:

- not found → `null` at `:458`
- found but outside the bounded context → `null` at `:468`

with a comment at `:460-461` making the intent explicit in almost the same words as the tool being deleted here:

> `// Withhold identity for a datasource outside the bounded context. Returning`
> `// null (same as "not found") keeps absent vs not-allowed indistinguishable`
> `// upstream, so this can't be used to enumerate non-allowed datasources.`

It also gates on the LUID *before* any identity lookup, so a disallowed datasource never reaches `queryDatasource`, `readMetadata`, or the Insight Service at all.

So the deleted logic's near-duplicate lives on in the path that matters, and my framing above — "the replacement doesn't preserve the property" — was wrong as a general statement. It's true only of `list-datasources` + `contentUrl:eq:`.

**What that leaves is much smaller:** not "restore a lost security property," just "the PR body recommends the one path that leaks, when the path callers should actually use doesn't." Pointing bullet 3 at `generate-insight-cards` instead would resolve it, and I'd consider the story's *"confirm no behavioral regression for … no-oracle semantics"* AC satisfied by the code above rather than needing new work.

Sorry for the noise — better to correct it here than leave the stronger claim standing.

tonythebest:
Two addenda to my review above, both from cross-checking it. One sharpens the `tab-agent-south` follow-up; one corrects a claim I got wrong.

### The `INCLUDE_TOOLS` silent-drop fails *open*, not closed

I said the three-entry allow-list "quietly becomes two." True here, but I understated the general failure mode. `server.web.ts:136` applies the include filter only when the list is non-empty:

```ts
if (includeTools.length > 0 && !includeTools.includes(tool.name)) continue;
```

Combined with `overridableConfig.ts:249-252` `flatMap`ping unknown names to `[]`, a deployment whose `INCLUDE_TOOLS` consisted **solely** of removed or misspelled names doesn't end up with zero tools — it ends up with **every** tool. The allow-list silently becomes a no-op and the full surface registers.

That inverts the failure mode from "quietly loses a capability" to "quietly loses the restriction." Not reachable from the known consumer — `tab-agent-south`'s tuple has three entries and only one is going away, so it degrades to two rather than to empty — but it's the reason the follow-up story is worth filing rather than shrugging at, and worth a hardening note in it. A parse-time warning on unrecognized names would cover both this and the ordinary drop.

### Correction to my `feature/desktop` claims — I was reading a stale ref

I listed three dangling comments and a −88 divergence. Those were computed against `origin/feature/desktop` at `12ae10e7`, which turned out to be frozen: this clone has a single-branch fetch refspec, so `git fetch --prune` advances `origin/main` and silently leaves everything else alone. The real head is `60d4266b`, roughly 30 PRs ahead.

Re-verified against the real head:

- **Two** dangling comments, not three: `src/desktop/metadata/field-builder.ts:302` and `src/desktop/metadata/types.ts:36`. The two `listAvailableFields` sites have already been cleaned up.
- Divergence in the four files this PR edits is **+8 / −90**, not −88. Conflict expectation is unchanged.
- On bullet 1 of the description: `list-workbook-datasources` **does** surface the `luid` at that head (`listWorkbookDatasources.ts:23,62,68-69`, published non-federated only, field omitted when the API returns `null`). So that bullet is substantively right and only pointed at the wrong branch — a wording fix, not a factual error. My original comment was harsher than it needed to be.

Neither of these changes the recommendation: the version bump remains the only thing blocking.

tonythebest:
Thanks for the thorough pass — addressed:

- **check-version (blocking):** bumped `4.0.6` -> `4.0.7` (`279199f8`). Patch per your lean — the tool was `disabled: !insightsToolsEnabled` so no default-config consumer saw it.
- **Description bullet 1:** reworded — the on-`main` justification now points at `generate-insight-cards` `{ luid }`/`contentUrl` + `list-datasources` `contentUrl:eq`; `list-workbook-datasources` (#632) is now called out explicitly as the `feature/desktop` Desktop-flow source, not on `main`.
- **No-oracle AC:** added a "Bounded-context / no-oracle semantics" paragraph to the description (details in the two inline replies).
- **Test count:** corrected to **2671 / 164**; also added a real **fed-luid E2E** against 10ax — `generate-insight-cards({ datasource: { luid } })` -> `isError=false` with governed cards, proving the insight flow works with the LUID fed directly.
- **tab-agent-south drift:** filed [W-23701788](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002gTy6oYAC/view) to drop the tool from `_WEB_MCP_ALLOWED_TOOLS`/`INCLUDE_TOOLS` + test when the deletion reaches `feature/desktop`.
- **#633 / W-23567697:** closed #633 (W-23567697 dropped to Never; #60876 supersedes the resolver-mode plan).

CI should go green on the version bump — ready for re-review.

src/tools/web/datasources/resolveDatasourceLuid.ts
tonythebest: Added a **Bounded-context / no-oracle semantics (AC)** paragraph to the PR description capturing exactly this: the `list-datasources` `contentUrl:eq` replacement enforces the same bounded context (`constrainDatasources`, no LUID leak) but returns two distinguishable messages — so it's a paper-trail gap, not a regression. The tool was `disabled` by default while `list-datasources` was always registered, so the oracle was already the reachable behavior on `main`. Left the two-message unification as an optional follow-up if the property is worth keeping.

src/tools/web/datasources/resolveDatasourceLuid.test.ts
tonythebest: Agreed — the no-oracle property is being dropped as a goal for this surface (the tool was opt-in, and the always-registered `list-datasources` path already distinguished the two messages), so the assertion goes with the tool. Captured the decision in the PR description's no-oracle paragraph so the rationale isn't lost.

nawartamawi (approved):

src/scripts/createClaudeMcpBundleManifest.ts
```diff
@@ -653,7 +653,7 @@ const envVars = {
     type: 'boolean',
     title: 'Enable insight-cards MCP tools',
     description:
-      'Registers the datasource-context insight tools (generate-insight-cards, resolve-datasource-luid). Disabled by default; set to "true" to enable them.',
+      'Registers the datasource-context insight tools (generate-insight-cards). Disabled by default; set to "true" to enable them.',
```
nawartamawi, line 656: should we remove generate-insight-cards too ? since it's not used

tonythebest: Good call — filed a follow-up to remove it: [W-23714833](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002gWqxLYAS/view). The Chiron UI generates cards via `generate-pulse-metric-value-insight-bundle` (InsightBundlesProvider), so `generate-insight-cards` does look unused — the story confirms no live consumer, then removes it on both `main` and `feature/desktop` (plus the tab-agent-south allow-list). Keeping it out of this PR since #711 is scoped to `resolve-datasource-luid`. Resolving.

PR #712, opened 2026-08-04
refactor(desktop): trust per-item /document bare fragments and route refine-worksheet through per-sheet apply

myu404:
## Decision

The External Client API's per-item `/document` routes now return a **bare** `<worksheet>`/`<dashboard>` fragment on GET and replace that item **in place** on POST. This PR aligns the desktop client with that contract: it trusts the per-item route's bare-fragment shape instead of slicing a single item out of a whole-`<workbook>` envelope, and it routes `refine-worksheet` through the same in-place per-sheet apply that `apply-worksheet` uses. This is a rule the codebase now keeps, not a one-off.

## What changed (four commits)

1. **`feat(desktop): route refine-worksheet through the per-sheet apply`** — `refine-worksheet` edits a worksheet it has already fetched, so it now replaces that sheet in place (`requireExistingSheet: true`) via the per-sheet `/document` route, instead of upserting it into the whole workbook. A worksheet that no longer resolves now surfaces an error rather than being silently re-created. (`refine-worksheet` is the only refine-family tool — there is no `refine-dashboard`/`refine-storyboard`.)

2. **`refactor(desktop): drop client-side read-path fragment slicing (revert #601)`** — the GET routes now return bare fragments, so the `getWorksheetFragment`/`getDashboardFragment` wrappers and the `worksheetDocumentToFragment` helper are removed and their six callers point back at `getWorksheetXml`/`getDashboardXml`. Behavior-preserving: the read path already returned a bare fragment on both transports, and the removed re-slice was a no-op whose null→error branch duplicated the gateway's own resolution error.

3. **`refactor(desktop): consume the bare storyboard document fragment directly`** — the same for the storyboard `/document` route: `get-storyboard-xml` returns the document as-is (guarding only that it parses to a `<dashboard>`), removing the last caller of `dashboardDocumentToFragment` and the helper itself. The mock server's storyboard fixture is corrected to the bare-fragment shape.

4. **`test(desktop): align worksheet and dashboard mock /document fixtures to bare fragments`** — flatten the mock server's remaining whole-`<workbook>` worksheet and dashboard document fixtures to bare fragments, matching the live shape and the read path that no longer slices. Test-only.

## Relationship to #601

#601 added client-side fragment extraction because, at the time, the per-item `/document` routes returned a whole `<workbook>` scoped to the requested item — the client had to slice out the single fragment. The backend now returns bare fragments directly, making that slicing a no-op. Commits 2–4 revert #601's **read-path** extraction accordingly (2–3 in the client, 4 in the test mock). The whole-workbook **upsert (create) path** that #601 also introduced is unrelated to the read-path slicing and is kept.

## Validation

`tsc --noEmit` clean · `eslint` clean · full unit suite green (6441 passed / 330 skipped).

### DIFF
--- a/src/desktop/commands/workbook/getDashboardXml.test.ts
+++ b/src/desktop/commands/workbook/getDashboardXml.test.ts
@@ -2,7 +2,7 @@ import { Err, Ok } from 'ts-results-es';
 
 import { ExternalApiToolExecutor } from '../../externalApi/externalApiToolExecutor.js';
 import { ExecuteCommandError } from '../../toolExecutor/toolExecutor.js';
-import { getDashboardFragment, getDashboardXml } from './getDashboardXml.js';
+import { getDashboardXml } from './getDashboardXml.js';
 
 describe('getDashboardXml', () => {
   it('resolves dashboard name to id and fetches the per-item document', async () => {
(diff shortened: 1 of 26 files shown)


lauren-jacksonSFDC (approved):

mattcfilbert (approved):
Approving. Clean cutover — the removed `worksheetDocumentToFragment` / `dashboardDocumentToFragment` helpers have no other callers, `extractSheetXml` / `extractDashboardXml` stay for their real users, and the `refine-worksheet` rewire through `requireExistingSheet: true` comes with a real test proving a vanished sheet now errors instead of being silently re-created. Decision-first PR body, too.

One thing to confirm before or right after merge — not blocking:

The two removed `*ToFragment` helpers were shape-tolerant: they sliced a fragment out of a whole-`<workbook>` document *and* passed a bare fragment through unchanged. The new read path returns the document verbatim, so this is a hard cutover to the 0.2.0 bare-fragment shape with no version floor. On a Desktop still serving the old 0.1.x whole-`<workbook>` shape (what #601 was written against), `get-worksheet-xml` would hand back a whole workbook and the downstream canonical-name gate would reject it with the "cached file holds a whole `<workbook>` document" message.

`buildCommandStatus` already version-gates the sibling 0.1.x → 0.2.0 change via `supportsOperationResult(apiVersion)`, and `apiVersion` is on the instance — so the boundary is known and used one function over. If the connected Desktop is guaranteed 0.2.0+ (pinned-client lockstep), this is a non-issue — worth a line in the PR body saying we require 0.2.0+. If a 0.1.x build can still be reached, consider gating on that same boundary or keeping the tolerant slice as a fallback.

Small consistency nit: `get-storyboard-xml` kept a `parseXML(...).dashboard` guard on the raw document, but the worksheet and dashboard read paths now return the XML with no parse check. If that guard earns its keep for storyboards, it earns it for the other two — and it'd fail loudly instead of leaking a `<workbook>` to the agent.

—
_Posted by MattGPT on Matt's behalf._

mattcfilbert (approved):
lgtm

PR #714, opened 2026-08-04
feat(desktop): .tbm-native template engine + command-reference migration

mmmnorthmark:
## Summary

Desktop template-engine work stream: makes bundled Tableau `.tbm` bookmarks the
canonical, metadata-optional source for worksheet templates, activates the
aggregation-compatibility guard on the inferred path, and migrates the command
reference to the codegen nested schema.

The raw diff is large because it ships **data assets**, not just code:
- **133 `.tbm` bookmark templates** — the drop-in template library.
- **`tableau-desktop-commands-reference.json`** refreshed to the nested codegen schema (~47k-line generated artifact).
- Actual reviewable code is **~6.6k insertions across 62 `.ts` files**.

## What changed

**Template read pipeline (.tbm-first)**
- `.tbm` bookmarks are now the canonical template source; `listTemplateNames()` /
  the intelligence provider surface every bookmark plus every manifest-backed
  `.xml`, hiding only raw orphans.
- Metadata-optional slot inference: infer from all reference sites (not just
  rows/cols/mark), LOD + display optionality, communicative role and table-calc
  semantics as first-class slot facts.
- Root `<simple-id>` injection on emitted worksheets/windows; facet-splice gating
  so the greedy structural XML fallback only runs when no manifest slots are supplied.

**Tool surface**
- New `build-worksheets-from-templates` (metadata-optional, dual-mode) and a
  clarified `apply-worksheet` upsert seam.
- Removed 3 self-contained legacy tools (propose/validate/list-xml).
- `list-templates` routed through the metadata-optional resolver.

**Aggregation-compatibility guard**
- Activated on the pure-inferred `.tbm` path (`inferSlots`, `bookmarkTemplate`,
  `binder/validate`, `binder/explicit-bind`).

**Command reference migration**
- Consumers migrated to the nested schema (`serialized`/`summary`/`classification`/
  `invocation`); reference refreshed. Reference-omitted pane-invoked executables
  (e.g. `generate-viz-from-notional-spec`) are unioned back into the known-command
  guard so the fail-closed check doesn't reject real verbs.

**Docs**
- `dashboard-xml-guide.md` moved into `knowledge/tactics/dashboard/`; readers
  repointed. Added `template-read-pipeline` + `template-tool-surface-redesign`
  proposals under `docs/proposed/`.

## Testing

Full desktop unit suite green locally (216 files, 3231 passed, 1 skipped, 0 failed)
under Node v24.14.1.

## Notes for reviewers

[...]

PR #716, opened 2026-08-04
feat(templates): add repository discovery with one-shot apply

mattcfilbert:
Decision: template lookup and construction stay read-only. A caller may change Tableau only by applying a fresh, one-shot artifact that creates a new worksheet. This is the rule every caller must keep, not a TAS-only exception.

MCP provides caller-neutral utilities for discovery, binding, and apply while leaving template ownership with the repository. It scans `Bookmarks` and `Tableau Agent/templates`, compiles `.tbm` files, and applies through the External Client API. It does not use Tableau's native `open-bookmark` command. TAS owns its separate conversation policy.

Future changes must preserve the build/apply split, new-worksheet-only targets, session-bound one-shot artifacts, live checks immediately before apply, and stop-on-uncertain-write behavior.

## What changed

- discover protected, overridable, custom, and repository bookmark templates without copying user files
- return bounded catalog results with provenance, slot details, and honest eligibility blockers
- build selected worksheets as non-mutating artifacts
- compare the full source workbook byte-for-byte immediately before apply
- consume stale, failed, and uncertain artifacts instead of replaying them
- verify the new worksheet and window after apply
- block existing-sheet replacement through this path
- support caller-neutral guidance for one named chart or up to three sequential brainstorming views
- continue through ordinary authoring when no template fits
- repair the shipped `spatial-symbol-map` manifest against its active six-slot `.tbm`
- accept only Tableau's implicit generated Latitude, Longitude, and Geometry fields without fabricating datasource declarations
- fix two invalid bundled Profit Ratio calculations
- keep all 133 templates discoverable while marking 24 unresolved table-calculation templates ineligible for pass 1

## Evidence

- `scripts/agent-check`: all 6 checks passed
- final Desktop suite: 3,417 passed, 1 skipped
- final cross-repo adversarial review: no P0-P2 findings
- live natural-language run: a non-geocodable business Region field was rejected before construction; the caller then built and host-verified the nearest sound bubble view on a fresh worksheet
- dead Clipboard datasource regression: apply stops before the whole-workbook write and returns repair guidance

## Known limits

- The External Client API has no workbook revision or `If-Match` condition for its final POST. The byte-exact check closes stale artifacts before dispatch, but an edit racing the final read-to-write window can still be overwritten.
- This branch is stacked on draft PR #714. GitHub cannot target its fork branch, so this PR remains based on `feature/desktop`; do not merge it before #714 lands and this branch is rebased.
- The 24 blocked table-calculation templates need a later addressing-field rewrite before pass-1 apply.
- Template discovery is intentionally uncached so user files appear at runtime; a protected-corpus scan currently costs about 1.8 seconds.
- Several older whole-workbook writers still bypass the shared validated apply path. This change does not widen that existing gap.

Approving this draft means accepting the caller-neutral catalog and one-shot new-worksheet apply boundary. It does not approve the stacked base or dictate any caller's conversation policy.

— MattGPT, updated with Matt Filbert's approval


mattcfilbert:
Decision: template artifacts are caller-neutral and one-shot. The MCP checks the source workbook byte-for-byte immediately before apply; stale, failed, and uncertain artifacts are consumed rather than replayed. The External Client API still cannot make the final whole-workbook POST conditional on a revision, so an edit that races that write remains a host limitation.

The generated symbol-map path now uses the active six-slot .tbm contract and accepts only the Tableau-generated Latitude, Longitude, and Geometry fields that do not need datasource declarations. A named chart can build one fresh sheet; an open analytical request can build up to three sequential sheets. If no template fits, the caller can continue with ordinary authoring.

Canonical checks are green, including 3,417 Desktop tests with 1 skipped. In a natural live run, the MCP correctly rejected a business Region field that Tableau could not geocode; TAS then built and host-verified the nearest sound bubble view on a fresh worksheet.

— MattGPT

mattcfilbert:
New rule: apply each live template artifact before building another one in the same Desktop session. The tool response and route instructions now say this directly; offline builds do not make the live-session claim. This keeps one valid artifact at a time and does not add a confirmation step.

Evidence: `scripts/agent-check` passed all six checks (5,599 general tests and 3,418 Desktop tests, with one skipped in each run). The cross-repo review found no P0–P2 issues. The remaining check is the live conversation after the local stack restarts.

— MattGPT

mattcfilbert:
New rule: template construction now runs the same worksheet preflight as apply before it creates an artifact. Error findings reject the candidate early; warnings still pass, and this adds no user confirmation step. The caller is told to choose another template instead of rebuilding the same template and field mapping.

Why: in the live three-view test, `deviation-diverging-bar` created an artifact that apply rejected for `redundant-color-encoding`. The agent then rebuilt the identical candidate and got the same guaranteed failure. This change moves that deterministic rejection to construction, where the existing one-alternative fallback can switch immediately.

Evidence: TDD regression, 37/37 focused tests, 78/78 relevant review tests, clean adversarial review, and all six `scripts/agent-check` checks passed (5,601 general tests; 3,420 Desktop tests). The remaining proof is one live rerun through TAS.

— MattGPT

mattcfilbert:
Intentional redundant color no longer blocks a valid template.

The validator now keeps `redundant-color-encoding` as a warning. Real validation errors still stop construction before an artifact exists. A successful template apply returns bounded warning rule IDs so TAS can make the event searchable without exposing validation details or trusting model-written telemetry.

Live proof: `deviation-diverging-bar` built and applied in Sample - Superstore with Sub-Category, Profit, and Sales. Host verification passed, and the apply returned `redundant-color-encoding` as a warning.

Validation:

- 5,603 general tests and 3,422 Desktop tests passed
- Lint, typecheck, Desktop build, and the lockstep gate passed
- Final cross-repo review found no remaining P0-P2 issue

TAS records the corresponding LangSmith flag in `browser-clients/tab-agent-south!127`.

— MattGPT

mattcfilbert:
THIS IS REPLACED

PR #721, opened 2026-08-05
feat(desktop): add bookmark templates for worksheet authoring

mattcfilbert:
## Decision

Use Tableau MCP as the common helper layer for worksheet bookmark templates. Repositories own the `.tbm` files; MCP discovers, binds, validates, and applies them; each agent owns its conversation with the user.

This is the rule future work must keep. We do not use Tableau's native `open-bookmark` command, and MCP does not become the source of template content. Approving this draft means accepting that boundary, the prepare-then-apply worksheet flow, and the repository discovery order described below.

This draft supersedes #716. Leave #716 unmerged.

## What changed

- scan the protected seed, `Tableau Agent/templates`, and the repository `Bookmarks` folder at runtime
- compile worksheet bookmarks into datasource-neutral template workbooks and derive slot metadata from each bookmark
- return bounded catalog results with provenance and pass-1 eligibility
- prepare a worksheet as a session-bound, one-use artifact before writing the workbook
- reject stale plans, existing-sheet replacement, source drift, uncertain replay, and unsafe template output
- verify the worksheet and window after apply, with clear recovery guidance when the write succeeded but verification did not
- preserve the current `feature/desktop` tool set and per-sheet apply path
- use exact catalog lookup for common named bar, line, scatter, and bubble requests; use broad discovery for open analytical intent
- let basic bar and column templates inherit Tableau's automatic mark color unless the template sets a color intentionally
- record template fallback and apply outcomes for trace review
- replace an existing dashboard's primary layout and viewpoints in one workbook apply, so Desktop never sees an invalid half-updated dashboard
- allow a safe dashboard edit when unrelated validation errors already exist elsewhere in the workbook, while still blocking any new error and every broken target-dashboard reference

## Behavior to preserve

- A direct, sound chart request builds without a conversational confirmation gate.
- An explicit request not to change the workbook stays read-only.
- Open analytical intent may produce up to three fresh worksheets so an author can compare ideas in Tableau.
- Template artifacts create fresh worksheets only. They never replace an existing worksheet or window.
- Each artifact is applied before another is built; stale or uncertain artifacts are never replayed.
- Repository template metadata is untrusted data, not agent instruction.
- TAS and third-party agents use the same MCP utilities while keeping their own interaction policy.
- Existing dashboard updates rebuild the primary layout. Tableau keeps any existing phone and tablet layouts because this tool does not accept a device-layout specification.

## Verification

- `scripts/agent-check`: all 6 checks passed
- general unit suite: 6,795 passed, 330 skipped
- Desktop suite: 4,554 passed, 330 skipped
- fresh-context dashboard review: no P0-P2 findings
- live Desktop proof: rebuilt an existing Superstore dashboard in one document apply; the primary layout contained only the requested worksheet, Tableau preserved its device layouts, and error `2805CF18` did not recur

## Known limits

- The External Client API cannot make its final document POST conditional on a workbook revision. The byte-exact pre-dispatch check closes stale plans, but an edit racing the final read-to-write window can still be overwritten.
- Twenty-six bookmark templates with unresolved pass-1 hazards remain discoverable but cannot be applied through this path yet.
- Proof-stamped legacy XML copies still carry their historical explicit gray mark color. The current dynamic path resolves the `.tbm` files and inherits Tableau automatic color; changing the legacy copies requires fresh live render proof and new stamps.
- Repository discovery is intentionally uncached so new user templates appear at runtime.

Posted by MattGPT with Matt Filbert's approval.

### DIFF
--- a/lockstep.hashes.json
+++ b/lockstep.hashes.json
@@ -2,10 +2,10 @@
   "src/desktop/binder/calc-derivation.ts": "5105d466e3c31019affa91320edf3047259301ca5cd27aeb5883ba1faa251098",
   "src/desktop/binder/classify.ts": "191eb4d5592b4dac44c9ee9afbaf79b566c424ff3b86f3c5ef1381b1e5b3fdba",
   "src/desktop/binder/escape.ts": "3c79bcf77c33fad64975e95ea57df5fb3e416008633e15dc7a585018fbc92d3f",
-  "src/desktop/binder/manifest-types.ts": "22740b978fdf7f1f463aa9ccd94f1e2ae5d383e610bf9063bff37e1624d173f8",
+  "src/desktop/binder/manifest-types.ts": "bdc917927aa1e513520df1a393b7b3b3c12c705715e15948ed590044287e7f59",
   "src/desktop/binder/manifest-validation.ts": "fc46c7b874cb2bd5958a2ed40848d18f6b3654e2d8559cff90fb841ca8fdd8d6",
   "src/desktop/binder/stringTemporal.ts": "8b4808dcc410453fed10041d2cb8908f9728a0dd94353bf0c6bc31a295a67c9b",
   "src/desktop/binder/waterfall.ts": "a157e6c94e88dd6ff2ecb5d4977c6b62f6366e338a1e7491788fe78ee2fa2fae",
-  "src/desktop/derivations.ts": "d95ec48b7d9989e2489b73f25f102e7b3b6766129601c4fa89392b6aa1dec093",
-  "src/desktop/templates/fieldReferenceRewriter.ts": "ea78b827e2ba236f30c3bef30f6f9ba871768666395d92616ed5c3617b352786"
+  "src/desktop/derivations.ts": "6d94cf346139ece1592d42cdd0b184376279a176dcc55975e386e2627b773450",
+  "src/desktop/templates/fieldReferenceRewriter.ts": "40068068cff3d622e0d05abdf4c8d5725b64d10351dfbd9806e20537e1bb850e"
 }
--- a/resources/desktop/knowledge/personalization/discovery-first-authoring.md
+++ b/resources/desktop/knowledge/personalization/discovery-first-authoring.md
@@ -3,7 +3,7 @@
 ## Scope Check
 
 - Primary audience: Tableau users building a new viz with the agent (and SEs assisting them)
-- Authoring outcome improved: The agent starts chart asks with the schema-aware binder, then uses focused discovery for repair, edits, or inventory instead of fabricating fields.
+- Authoring outcome improved: The agent grounds chart asks in the template catalog and live fields, then builds a guarded worksheet artifact and applies it without fabricating fields.
 - In-scope reason: Directly improves how the agent turns an ambiguous "build me a viz" into a viable, correct Tableau viz grounded in the real workbook.
 - Out-of-scope risk: Not a project-scoping or requirements-gathering framework, and not a dashboard/whole-workbook assembly flow - this is scoped to a single non-trivial build-a-viz turn.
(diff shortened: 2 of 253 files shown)


lauren-jacksonSFDC:

**On the External Client API usage — parts of this branch look authored against a pre-#691/#712 snapshot**

Three findings share one root cause: this branch reintroduces assumptions that #691 and #712 specifically removed.

**1. `waitForDesktopSettle` resurrects the single-read-after-apply pattern #691 deleted (the important one).** The new post-apply *window* verification in `loadWorksheetXmlViaExternalApi` does one fixed `waitForDesktopSettle(500ms)` then one `getWorkbookXml` (`loadWorksheetXml.ts` ~L599–627, helper L84). #691's commit message is explicit: *"A single readback immediately after apply can race that settle and see the PRE-apply XML… Poll instead of reading"* — and it replaced a `setTimeout` single-read with `pollReadback` (8×250ms, abortable). The body verification directly above this block already uses `pollReadback`; the new window block doesn't. Worse, the window read is on the *whole-workbook* route while the body poll settles the *per-sheet* route, and when body verification is `status:'skipped'` it returns `ok:true`, so the window block can fire on the 500ms delay alone. A false miss returns `readback-failed` → `mutationOutcome:'possibly-dispatched'` ("apply MAY have landed, go inspect Tableau") on a one-shot artifact. This should use `pollReadback` keyed on the window fingerprint, not a fixed sleep.

**2. `worksheetDocumentToFragment` is re-added (`sheets.ts:185`) — the helper #712 deleted.** #712 ("Remove the … worksheetDocumentToFragment helper that #601 added") removed it because the per-item `/document` GET returns a bare `<worksheet>` fragment, so there's nothing to slice out of a whole-workbook envelope. It's dead here (no prod caller; the read path that would use it is untouched), so no behavior change today — but it re-encodes the stale whole-workbook-envelope assumption, and left in place it invites someone to wire it back into a re-slice that would reduce a bare fragment to nothing. Please delete it. (The sibling new export `extractLastWorksheetArtifact` *is* used by the template build path — that one's fine.)

**3. The pre-dispatch guard hashes the raw whole workbook (`targetWorksheetState.ts:576`) — a fresh instance of the same "live doc is a frozen snapshot" assumption.** `workbookSha256 = sha256(workbookXml)` over the raw workbook with no volatile-attribute stripping — even though the window fingerprint in the same file strips `@_active`/`@_maximized`/`simple-id` (L166–168) because those are volatile. Since `worksheetApplyStatesEqual` requires whole-workbook byte-equality, any change *anywhere* between build and apply — an unrelated sheet edit, or just switching the active sheet — trips `preview-state-changed` and burns the one-shot artifact. For a net-new worksheet the correctness checks are already precise: `target`/`targetWindow=='present'` catches name collision, `dependenciesSha256` catches referenced-field drift. The raw whole-workbook hash only adds spurious rejections. It fails safe (rejects, never corrupts) and I get it's an optimistic-concurrency stand-in for the missing revision/etag, so this is usability not correctness — but I'd drop the `workbookSha256` conjunct (or strip volatile attrs like the window fingerprint does).

**Minor, for the description:** the "one workbook apply" dashboard win only applies when the dashboard window already exists — a net-new dashboard still takes the two-apply path (`buildAndApplyDashboard` / `applyDashboardWithViewpoints` re-read + second `applyWorkbookText`). Worth a Known-limits line.

(I confirmed the artifact path *should* use whole-workbook upsert rather than the per-sheet route — that route can't create a net-new sheet and artifacts are always net-new, so that wiring is correct.)

mattcfilbert:
Superseded by #723. This draft carried too much stale Desktop code from the original spike. #723 starts from the current feature/desktop branch and reapplies only the TBM template runtime, list/build/apply tools, and puppetry-compatible bind-template. Closing this draft so review stays on the branch we intend to ship.

Posted by MattGPT with Matt Filbert's approval.

PR #722, opened 2026-08-06
feat(desktop): add create-worksheet-blank / create-dashboard-blank / create-storyboard-blank tools

myu404:
## Decision

Add three MCP desktop tools — **`create-worksheet-blank`**, **`create-dashboard-blank`**, **`create-storyboard-blank`** — that create a new empty sheet in the open workbook via the server's per-kind blank-sheet routes (`POST /v0/workbook/{worksheets,dashboards,storyboards}:new`). This is a new capability, not a change to existing behavior.

These are deliberately separate from the `apply-*` / `build-and-apply-*` tools: the routes create a blank sheet with a server-generated name and accept **no** XML, so they cannot be folded into the templated/raw-XML apply path. `apply-*` edits an existing sheet's XML in place; `create-*-blank` makes a new empty one.

## What changed

- **Client** (`externalApiClient.ts`, `types.ts`): three `createBlank*` methods + a `buildNewSheetRoute` helper. The server has no typed result to return for these routes, so it responds with a generic Operation envelope — sync or 202-polled — same as every other write route. Parsed via the existing `parseEnvelope` path; there is no created-sheet `id`/`name` in the response.
- **Executor** (`externalApiToolExecutor.ts`): three methods routed through the shared `applyDocument` helper, the same write-and-fold-to-command-status pipeline `undo`/`redo` use.
- **Tools** (`external-api/create*Blank.ts`): one file per tool, mirroring the existing per-tool convention. Each takes an optional 0-based `index` tab position. Since the response carries no created-sheet identity, success is reported the same way `undo`/`redo` do — a two-branch message keyed on command status (e.g. "Created a blank worksheet." vs. "Requested a new blank worksheet; Desktop is still creating it.").
- **Registration**: added to `toolName.ts`, `tools.ts`, and the lean dynamic-authoring tool profile.
- **Tests + mock**: client, executor, and tool-level tests; mock server returns a bare Operation envelope from the `:new` routes and no longer rejects a request body (the live server ignores it, since these routes never read one).

## The `index` tab position

Optional 0-based integer. Forwarded as `?index=k` **only when supplied** — `index=0` is sent (insert before the first tab) and is contract-distinct from omitting it, which appends at the end. A value `>=` the current tab count also appends.

## Compatibility

The routes require a Desktop build that serves the `:new` endpoints. Until then, each tool surfaces a clear "endpoint not in this build" message via the existing route-missing path (the same pattern the undo/redo tools shipped with). The generated openapi contract spec-list is intentionally left unchanged until a build that serves these routes is captured.

## Validation

`tsc --noEmit` clean · `eslint` clean · full unit suite green (6435 passed / 330 skipped).


mattcfilbert (approved):
**Andy-lens review: ship it.**

Solid, and the last commit fixed the thing that mattered. Realigning the client from a bare `CreatedSheets` body to `parseEnvelope` matches how the live server actually answers these routes — sync or 202-polled, same as every other write — and routing the executor through the shared `applyDocument` pipeline is the right reuse. The `state: succeeded` → `completed` mapping is exercised end to end: the happy-path test drives the real mock server through the client and asserts "Created a blank worksheet.", so the two-branch success message keys on a real status, not a guess. Dead `CreatedSheets` / `createdSheetsSchema` fully removed, no orphan. Registration is alphabetized and consistent across `toolName.ts` / `tools.ts` / the profile. Good separation call keeping these distinct from `apply-*`: the routes take no XML and mint a server-named sheet, so folding them into the apply path would be wrong.

One question, your call:

- Since the `:new` response carries no created-sheet id or name, the agent can't reference what it just made — its next move (apply/refine/activate) needs a name it has to go re-list to discover. That's the server contract, not something to fix here, so it's fine as-is. But it's worth a line in the tool description telling the model to list sheets to find the new one, so it doesn't stall guessing. Minor.

`index=0` is correctly sent and contract-distinct from omitting it (append) — nice that the mock and a test both pin that. Compatibility handling via the existing route-missing path is the same honest pattern undo/redo shipped with.

Heads up, not blocking on my end: GitHub shows this as conflicting with `feature/desktop` — needs a rebase before it can merge.

Approving.

— MattGPT

myu404:
Closing because #759 merged with these tools.

PR #725, opened 2026-08-06
feat(desktop): add delete-sheet, rename-sheet, sort-worksheet + worksheet logical-table read tools, drop delete-worksheet

lauren-jacksonSFDC:
## What

Adds five Tableau Desktop External Client API tools, rewires `activate-sheet` onto the `:goToSheet` route, and removes the hand-written `delete-worksheet` tool — reconciled against the live 0.2.4 OpenAPI spec.

New tools:
- `list-worksheet-logical-tables` — `GET .../worksheets/{id}/logicalTables` (id + caption).
- `get-worksheet-underlying-data` — `GET .../worksheets/{id}/logicalTables/{logicalTableId}/data` → `SummaryData` (row-level underlying data for one logical table).
- `delete-sheet` — `POST .../{worksheets|dashboards|storyboards}/{id}:delete` (body-less). Deletes a worksheet, dashboard, or storyboard by name or id.
- `rename-sheet` — `POST .../{worksheets|dashboards|storyboards}/{id}:rename`. Renames a worksheet, dashboard, or storyboard by name or id.
- `sort-worksheet` — `POST .../worksheets/{id}:sort`. Applies or clears a field sort on a worksheet.

Changed:
- `activate-sheet` — same tool name and result surface, but its dispatch now goes through `POST /v0/workbook:goToSheet` (resolve name/id from the live inventory → `:goToSheet` by id) instead of a whole-workbook XML read + `tabdoc:goto-sheet`. Drops the XML read and its two XML-only behaviors (the already-active short-circuit and `previousSheet`) — the inventory API carries no active-sheet signal. The core `activateSheetValidated` fn (still XML-based) stays for the post-apply focus path in `applyFocus.ts`.

Removed:
- `delete-worksheet` — the hand-written whole-workbook XML-surgery tool. `delete-sheet` replaces it via the `:delete` route and also covers dashboards and storyboards.

## The decisions to accept

**Use the API's routes instead of hand-editing workbook XML.** `delete-worksheet` is dropped for `delete-sheet`; `activate-sheet` is re-backed by `:goToSheet`. `delete-sheet` keeps one client-side guard: it refuses to delete the workbook's **last remaining worksheet** (the server silently no-ops that case, so the guard turns a mute no-op into a clear refusal). The old delete tool's dashboard-referenced-worksheet refusal is not carried over — the command owns that policy now.

**Profiles:** the two logical-table reads, `delete-sheet`, `rename-sheet`, `sort-worksheet`, and the re-backed `activate-sheet` are on the default dynamic-authoring profile.

## Spec parity

- Recaptured `__fixtures__/externalClientApi-openapi.json` to live 0.2.4 (no hand-edits), including Operation `blockingWindows`/`WindowInfo`.
- Added `logical-table-not-found` to `PROBLEM_CODES` and the `LogicalTableItem`/`LogicalTableList`/`WindowInfo` schemas + all new route consts to the contract test's parity lists.
- Sort body (`fieldName` required; `direction` `asc`/`desc`; `sortType` `data-source-order`/`alpha`; `clearSort`) matches the live `SortRequest` schema.
- Fixed `columnsToIncludeByFieldName` on both `get-worksheet-underlying-data` and the pre-existing `get-summary-data`: the spec says this param is repeated per field name, never comma-split (a field name may contain a comma). Both now emit repeated query pairs.

## Tests

- `deleteRenameSortLogicalTableTools.test.ts` covers `delete-sheet`/`rename-sheet`/`sort-worksheet` routing per sheet kind, the last-worksheet refusal, the sort body, and the logical-table + underlying-data reads.
- `activateSheet.test.ts` covers the rewired tool (inventory resolve → `:goToSheet` by id, dashboard target, unknown-sheet error) and keeps the `activateSheetValidated` core-fn tests for the focus path.
- Client tests for every route incl. the repeated-param encoding and the write-route bodies.
- Contract test green against the 0.2.4 fixture; `server.desktop.test.ts` counts/byte-budgets re-measured.


mattcfilbert (approved):
lgtm

mattcfilbert (approved):
lgtm

PR #723, opened 2026-08-06
feat(templates): use TBM files for worksheet templates

mattcfilbert:
## Decision

This PR makes `.tbm` files the source of truth for worksheet templates. Every caller can use `list-templates` → `build-worksheets-from-templates` → `apply-worksheet`; `bind-template` remains available for puppetry and bring-your-own agents.

Templates ship as a versioned content pack instead of inside the MCP binary. TAS or a hosted deployment mounts that pack; MCP discovers, validates, binds, and applies it.

Approving this PR accepts that boundary and supersedes the template implementation in draft PR #721.

## Scope

- Reapplies the template work onto current `feature/desktop`; current Desktop behavior remains authoritative outside the template slice.
- Removes static manifests, generated catalogs, legacy XML copies, and the old intelligence pack.
- Scans protected templates, `<repo>/Bookmarks`, and `<repo>/Tableau Agent/templates`.
- Does not use Tableau's native `open-bookmark` command.
- Keeps template interaction policy out of MCP.

## Rules future work must preserve

- Normal authoring has no phrase, turn, or prior-list gate.
- `list-templates` returns structural facts, including semantic roles.
- Build is read-only and returns an artifact; apply checks target drift and Desktop instance before mutation.
- A failed pre-dispatch attempt may choose another template; an uncertain write is never replayed.
- `bind-template` remains for puppetry and bring-your-own-agent compatibility.
- Content packs contain only `.tbm` files plus transport integrity data, never manifests, credentials, or local paths.
- User templates override shipped content; one invalid user template cannot poison the catalog.

## What changed

- Added runtime TBM discovery, slot inference, compilation, and list/build/apply tools.
- Added a separate deterministic content-pack producer.
- Preserved puppet compatibility through a runtime projection.
- Removed the static manifest, catalog, XML, and intelligence systems.
- Added Bookmarks discovery and external-tier precedence.
- Added artifact coexistence, drift and Desktop-instance checks, and post-apply verification.
- Added fail-closed pack security checks and scrubbed the source corpus.

## Evidence

- `scripts/agent-check`: all six phases green.
- 5,606 general tests and 3,306 Desktop tests passed.
- 59 content-pack security tests and six TBM corpus invariants passed.
- Fresh security review: pass.
- The 133-template pack builds byte-for-byte deterministically: `54b5df0fec04dc207fbfbeabcc0a50dfe6f56e7cbe50b02555899140d74b9385`.

## Remaining work

- Run a live Desktop apply probe before moving the draft to ready.
[...]

PR #726, opened 2026-08-06
fix(validation): allow relative-date filter references

mattcfilbert:
## Decision

The `invalid-column-instance-pivot` rule now accepts `[none:<date>:qk]` only when a relative-date filter uses it, including that filter's matching `<slices>` entry.

**This is a rule we now keep:** relative-date filter references are valid Tableau XML even without a general-purpose quantitative column-instance declaration. The same reference on a shelf or in another context still fails validation.

## Scope

- Recognizes relative-date filter references within the same worksheet.
- Allows the matching slice entry Tableau writes for that filter.
- Does not normalize or rewrite the template XML.
- Does not exempt other `none:...:qk` references.

## Why

PR #723's artifact path assembles a whole workbook and runs the full workbook validator before apply. That exposed an old validator assumption: every undeclared `none:...:qk` reference was treated as fabricated. Desktop accepts this shape for relative-date filters, and Nawar's repro applied and verified after disabling only this rule.

Keep this exception tied to the relative-date filter and its slice. Do not turn it into a broad date-field exemption.

## Evidence

- Added a regression for an undeclared relative-date filter plus its matching slice.
- Added a negative regression proving the same reference still fails on a shelf.
- Red before the fix: the relative-date case failed with `invalid-column-instance-pivot`.
- `scripts/agent-check`: all six checks passed.
  - 5,609 general tests.
  - 3,309 desktop tests.
  - Lint, typecheck, desktop build, and lockstep checks passed.

## Reviewer decision

Approving this means relative-date template filters can pass preflight without weakening the guard for fabricated column-instance references.

### DIFF
--- a/src/desktop/validation/rules/invalidColumnInstancePivot.test.ts
+++ b/src/desktop/validation/rules/invalidColumnInstancePivot.test.ts
@@ -30,6 +30,30 @@ describe('invalid-column-instance-pivot rule', () => {
     expect(rule.validate(xml)).toHaveLength(0);
   });
 
+  it('accepts a relative-date filter and its matching slice without a column-instance declaration', () => {
+    const xml = `<worksheet name="Relative Date">
+      <table><view>
+        <filter class="relative-date" column="[DS].[none:Close Date:qk]" first-period="-30" last-period="0" period-type-v2="day" />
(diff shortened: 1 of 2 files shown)


nawartamawi (approved):
tested locally with my insights templates, working well.

PR #732, opened 2026-08-07
fix(desktop): use verified authoring paths by default

mattcfilbert:
## Decision

The default Studio profile now has one dashboard mutation path: `run-dashboard-batch`. Template selection and artifact building stay separate. The batch applies 0–6 prepared worksheet artifacts in order, then composes one dashboard. Full-profile puppet and bring-your-own-agent callers keep `compose-dashboard` and the phased dashboard tools.

Future dashboard work should preserve this rule: validate everything knowable before the first write, stop after the first failed or uncertain step, and never replay or roll back a partial batch blindly.

## What changed

- Replaced the competing default dashboard paths with one bounded batch.
- Derives new worksheet names from artifact titles; `existingWorksheetNames` is only for sheets already live in the workbook.
- Reserves and preflights every artifact, existing sheet, Desktop instance, and target state before writing. An invalid second artifact now causes zero workbook changes.
- Caps the complete receipt below 8 KiB so the SDK cannot truncate it mid-JSON.
- Verifies every worksheet apply and the final dashboard, including same-name replacement.
- Keeps direct compose and phased dashboard tools in the full profile for puppet/BYOA callers.
- Keeps the current embedded TBM corpus as a temporary compatibility fallback until separate content-pack delivery lands.
- Retains the earlier fixes in this PR for fresh-sheet `get-summary-data`, explicit bubble binding, and Desktop-written connection names.

No tool is deleted or renamed. This does not add granular dashboard editing or a generic MCP queue.

## Evidence

- `scripts/agent-check`: all five checks green; 369 files and 5,607 tests passed.
- Live two-artifact batch: both worksheets applied and passed readback; the dashboard composed with exactly those derived sheet names.
- Live failure probe: a valid artifact followed by a missing artifact returned a retry-safe preflight failure with no worksheet or dashboard written.
- Live replacement probe: the same dashboard was replaced twice; readback showed one dashboard with the requested sheets.
- Worst successful six-artifact receipt: 8,038 bytes, below the 8 KiB cap.
- Fresh adversarial review found no remaining P0–P2 issue.

## Risk

The v0 External Client API has no ETag/`If-Match`. These checks prevent avoidable partial work and stale writes visible before dispatch, but only a future conditional-write API can close the race with a simultaneous edit from another process.

Approving this PR accepts one bounded default dashboard path while preserving the full puppet/BYOA surface.

— MattGPT


mattcfilbert:
MattGPT update after the paired TAS live test:

- `dashboard-auto-apply` now returns the top three template candidates and all 20 relevant fields for each unresolved view. It no longer repeats the unused proposal schema or floods the client with the full catalog.
- The live proposal stayed valid JSON at about 10–15 KB instead of being truncated mid-response.
- With the TAS bounded-recovery branch, the same natural three-view dashboard request completed in 61.2 seconds, down from 173 seconds. Readback confirmed all three sheets plus the dashboard.
- PR #732 includes the current `feature/desktop` tip, including #726's relative-date fix and #733's validation performance fix.
- Full local MCP gate is green: 5,606 tests, lint, typecheck, desktop build, and lockstep.

Paired TAS MR: https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/130

mattcfilbert:
Shrinking down 130 drastically in response to feedback from @mmmnorthmark and reconfiguring the auto-apply for dashes in this PR

mattcfilbert:
Final gating decision: `run-dashboard-batch` adds no caller-side turn gate. The MCP reserves and preflights every artifact and existing sheet before the first write, then applies sequentially and stops on the first failed or uncertain step. A live invalid-second-artifact probe made zero workbook changes. Cross-process races still require future ETag/`If-Match` support. — MattGPT

mattcfilbert:
Adversarial live smoke found and fixed one more preflight gap: using the same name for a worksheet and dashboard let the worksheet land before Desktop rejected the dashboard with D6BB802B. The batch now rejects that collision before any write. The same matrix also passed hybrid existing/new composition, consumed-artifact replay, duplicate names, stale targets, and cross-session misuse. Full local gate: 369 files / 5,608 tests. — MattGPT

lauren-jacksonSFDC (approved):

PR #733, opened 2026-08-07
fix(validation): make qualified-name-brackets linear, not O(n²)

nawartamawi:
## Decision

`qualified-name-brackets` collected nodes with `xpath.select('//@* | //text()')`. The `//@*` descendant global-attribute axis is evaluated **O(n²)** in node count by the `xpath` package, so on a large workbook the call spins a CPU core and never returns. It runs on every apply via `runValidation(_, 'workbook')`, so any apply against a large datasource hangs.

This PR swaps that select for a **linear iterative DOM walk** over the same node set (all attributes + all text nodes). No behavior change — same nodes visited, same issues reported.

## Repro / evidence

- Applying against a live sqlproxy datasource (a work-item extract: ~155 fields, ~63k attribute+text nodes, 1.5MB workbook) hangs the apply preflight.
- Timed all validation rules on that workbook: 33 finish under 220ms; `qualified-name-brackets` alone blocks past 90s and never returns.
- Confirmed clean O(n²) scaling on `//@*`: 400 / 800 / 1600 nodes = 51 / 210 / 938 ms.
- A small datasource (a few hundred nodes) stays sub-second, which is why this only surfaces on realistic/large datasources.
- After the fix: **~85ms** on the same workbook, same output (0 issues).

It's the only rule using that global-axis select.

## Tests

- The rule's 12 unit tests pass unchanged (behavior identical).
- Removed the now-unused `xpath` import.

### DIFF
--- a/src/desktop/validation/rules/qualifiedNameBrackets.ts
+++ b/src/desktop/validation/rules/qualifiedNameBrackets.ts
@@ -24,12 +24,12 @@
  * false-rejects valid content (e.g. a bundled template).
  */
 import { DOMParser } from '@xmldom/xmldom';
-import * as xpath from 'xpath';
 
 import type { ValidationIssue, ValidationRule } from '../types.js';
 
 const TEXT_NODE = 3;
 const ATTRIBUTE_NODE = 2;
+const ELEMENT_NODE = 1;
 
 /**
  * Attributes / elements that carry FREE TEXT (calc formulas, captions, rich-text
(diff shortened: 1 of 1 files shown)


mattcfilbert:
test failure will be addressed by my PR @nawartamawi . approving and merging

mattcfilbert:
MattGPT review: no blocking issues found.

The iterative DOM walk preserves the relevant node set and removes the quadratic XPath behavior. In a local shape probe, the old scan grew from 34 ms at 400 elements to 514 ms at 1,600; the new scan stayed below 1 ms. The existing rule tests pass.

The current red check is the known Vitest `onTaskUpdate` timeout after all 5,560 tests passed, not a product-test failure. Non-blocking follow-up: add one large synthetic XML regression test so this performance characteristic stays protected.

mattcfilbert (approved):
good pr. hell yeah

PR #734, opened 2026-08-07
feat(desktop): add insights bar + line chart templates

nawartamawi:
Two templates for the Pulse insights viz path:

- `insights_bar_chart` — dimension × measure (top/bottom contributors, drivers, detractors, risky-monopoly).
- `insights_line_chart` — measure over a date, with an optional relative-date filter slot (trend/PoP/unusual-change).

Both carry the generic `Sample - Superstore` placeholder datasource and rebind to the live datasource at bind time, matching the other content-pack templates. Included in the template content pack (135 total).

Base: `feature/desktop` (the template system + content-pack builder live there, not on main).

Draft — paired with the tab-agent-south deterministic-insight path and tab-agent-south-ui apply_insight frame (internal GitLab MRs).


mattcfilbert:
Reviewed the full template path end-to-end (Andy lens). The change is sound and fail-closed at the right seams (unknown slot, missing datasource, unresolved token residue); three things before merge.

**Please**

- **Lint is red on both build jobs** — 3 prettier errors (`binder.ts:470`, `bindTemplate.ts:1960/1965`), all `--fix`-able. `tsc` passes; lint is the only failing step. Since a `feature/desktop` version bump publishes to Artifactory, green CI is release-gating here.
- **No version bump** — `package.json` is still `2.60.5`. A new tool param + two templates + a new escape path is backwards-compatible new surface → `npm run version:minor`. Heads-up that `version-bump-check.yml` only gates PRs to `main`, so it won't flag this on `feature/desktop` — it's manual here.
- **`skip_validation` is an LLM-visible tool param** (`bindTemplate.ts:115`), gated only by its description. Any model turn can pass `skip_validation:true` to bypass the apply-time preflight, and the backend can't strip it (its fast path keys on `auto_apply`, not `skip_validation`). The gate has to live here — enforce caller identity server-side, or accept the exposure explicitly.

**nit** — the auto_apply path double-escapes `template_parameters`: `sanitizeTemplateParameters` (`binder.ts:287`) escapes, then `injectTemplateCore.ts:334` escapes the same values again. Harmless for the shipped ASC/DESC/ISO-date tokens (no XML metacharacters), and it mirrors the pre-existing title double-escape already documented at `bindTemplate.ts:1246` — but a future `{{PARAM}}` carrying `&`/`<`/`'` would ship doubly-escaped. Escape once, or decode-for-compare symmetrically.

For reviewers reading the three insight PRs together: the deterministic path is `build_viz` (UI) → picker (backend) → `bind-template` with `skip_validation` (this PR). This PR is sound; the two blockers in that path are elsewhere — the backend inherits a lean `TOOL_PROFILE` that omits `bind-template` (I verified `bind-template` is not in `DYNAMIC_AUTHORING_TOOL_PROFILE` on this head), and the UI hangs on a dropped frame.

_Posted by MattGPT, automation acting for Matt Filbert._

mattcfilbert (approved):
Approved by MattGPT (automation acting for Matt Filbert, authorized in-session).

Re-review delta since the prior round is resolved. CI is green on the merge ref after folding in the byte-budget fix: bind-template's now server-gated `skip_validation` had its description dropped (an LLM turn can never honor it — it needs `allowSkipValidation` from the trusted deterministic spawner), and the full-surface ratchet moved 48958 → 49076 with a sign-off note. bind-template is not in the served dynamic-authoring profile, so the 46K auto-deferral cliff is untouched (served surface unchanged at ~29380). Full unit suite (5663 tests) green locally on the merge tree.

PR #739, opened 2026-08-07
fix(execute-tableau-command): verify file connections

mattcfilbert:
## Decision

One rule for this command: a root `tabui:data-catalog-connect-to-file` call must not report success unless a new workbook datasource appears. Leaf connections and every other command keep their current behavior.

Approving this PR means root file connections stop returning false success and return the datasource they actually added.

## Description

The command previously trusted Desktop's success envelope even when the workbook gained no datasource. This change reads the datasource inventory before and after the command. It returns the verified datasource on success, refuses before dispatch when it cannot establish a baseline, and reports unknown state without automatic retry when readback fails.

This adds no new tool or generic postcondition framework.

## Motivation and Context

A caller could receive "Command executed successfully" while the requested datasource was absent from workbook state. Later field and worksheet operations then failed far from the real cause.

Verification requires an explicit `IsLeafConnection: false`; omitted and leaf-mode calls are deliberately unchanged.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- Added regression coverage for verified success, no-op false success, unavailable baseline, failed post-command readback, and the registry's rewritten parameter name.
- Live-tested the built MCP with Tableau Desktop: the response returned the new datasource identity, and its fields were immediately discoverable.
- Ran `scripts/agent-check`: lint, typecheck, 369 test files / 5,619 tests, desktop build, and lockstep checks all pass.

## Related Issues

None.

## Checklist

- [ ] Version bump — no bump on the feature branch
- [x] Documentation updated — no documentation change needed
- [x] Tests added or updated
- [x] Unit tests pass locally
- [x] No breaking changes


lauren-jacksonSFDC:

mattcfilbert:
**MattGPT** — automated, posted on Matt Filbert's behalf.

**Root cause: a stale branch, not the datasource logic.** CI failed because this branch predated two changes on `feature/desktop`; merging base in fixes both.

1. #730 rewrote `emittedToolReferences.test.ts` to walk the whole `src/tools/desktop` tree and check every kebab-case string literal against the tool registry and the `NON_TOOL_VOCABULARY` allowlist. This tool's file added five tokens that weren't on the list, so the guard failed once the PR was tested merged into base. The branch's own copy of the guard was the older fixed-list version, which is why it passed locally.
2. The branch bumped the version to 2.60.6 while base moved to 2.60.5, so `package.json` and its lockfile conflicted on merge.

**The fix:**
- Added the four real non-tool tokens to `NON_TOOL_VOCABULARY`, sorted: `command-postcondition` (error type), `data-catalog-connect-to-file` (command id), `is-leaf-connection` and `is-leaf-connection-ui` (arg keys).
- Reworded the postcondition error prose to drop the fifth token, `no-op`. It is an ordinary word, and allowlisting it would hide a future real `no-op` tool-name typo. "The command may have been a no-op" became "The command may have had no effect".
- Dropped the version bump (base is 2.60.5; feature-branch PRs do not bump — `version-bump-check` runs only on PRs to `main`).
- Added two tests: a leaf connection and a connection with no leaf flag both skip the datasource check, so the postcondition fires only for root connections.

The datasource-verification logic is unchanged. CI is green on both Node matrix legs.

**Two follow-ups worth a look, out of scope here:** the added-datasource match keys on id → luid → name|caption, so a rename or a duplicate name could be misread as added or missed; and the new `verification` field is not held to the 16 KiB result byte budget the `result` field uses. Both depend on whether Desktop always returns stable ids and whether a response-size limit applies — worth confirming before relying on them.

PR #740, opened 2026-08-07
feat(telemetry): measure Desktop authoring work

mattcfilbert:
## Decision

Desktop authoring telemetry stays opt-in. It records counts, timings, fixed outcomes, serialized sizes, and hashed request IDs. It does not record workbook content, tool payloads, names, paths, routes, errors, or schemas.

## Scope

This extends the existing episode event stream. It measures:

- each tool call and result size;
- each logical External Client API operation, including one event across polling or a discovery rescan;
- the tool schema size registered at startup;
- the final outcome of run-dashboard-batch.

It does not add a telemetry service, change the tool surface, or change authoring behavior. Future telemetry must keep the same bounded fields and must never delay a Desktop result.

## Evidence

- scripts/agent-check: all five checks green
- 369 test files and 5,644 tests passed
- Desktop build and lockstep check passed
- Regression coverage includes 202 polling, 401 rescan, a telemetry sink that never settles, safe request correlation, and every dashboard batch outcome class
- Independent adversarial review found no remaining blockers

Version: 2.60.7.

Approving this PR means we can measure real Desktop authoring cost and failure shape without widening the agent tool surface or collecting user content.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.60.5",
+  "version": "2.60.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.60.5",
+      "version": "2.60.7",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.60.5",
+  "version": "2.60.7",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/episode-events.test.ts
+++ b/src/desktop/episode-events.test.ts
@@ -77,6 +77,61 @@ describe('episode event writer', () => {
     expect(events[0].ts).toEqual(expect.any(String));
     expect(readdirSync(dir)[0]).toMatch(/^episodes-.*\.jsonl$/);
   });
+
+  it('writes only bounded metadata for startup, Desktop RPC, and batch events', async () => {
+    const dir = tmpDir();
+    const cfg = config({ episodeEventsEnabled: true, episodeEventsDirectory: dir });
+
+    await emitEpisodeEvent(cfg, {
+      type: 'tool_schemas_registered',
+      profile: 'dynamic-authoring',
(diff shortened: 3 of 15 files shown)


mattcfilbert:
Andy-lens rereview found two real gaps; both are fixed in ea441a20:\n\n- The common Desktop tool wrapper awaited episode-file writes. A slow or stuck telemetry sink could delay a completed tool call. Telemetry is now observational, with regressions for sinks that never settle.\n- The startup schema metric counted only Desktop tools but did not say so. The event now declares surface: desktop, so combined-server data cannot be read as the full shared surface.\n\nFresh scripts/agent-check: all five checks green; 369 test files and 5,646 tests passed.\n\n_Posted by MattGPT for Matt Filbert._

tableaukyler:
**KylerGPT review — LGTM, no blocking findings.**

This adds opt-in Desktop authoring telemetry: a `desktop_rpc` observer around `withRescan` (one logical event per Desktop op, including a 401 rescan), a `tool_schemas_registered` startup event, and a `batch_apply` outcome event. It also flips the episode-event emits from `await` to fire-and-forget `void`.

What I checked and liked:
- The fire-and-forget path is genuinely non-blocking. `withRescan` wraps delivery in `try { ... } finally { void delivery?.catch(() => undefined) }`, so a telemetry sink that throws or never settles can't change or delay the Desktop result. The two "never settles" / "callback failure" tests pin exactly that.
- The event union is bounded metadata only — counts, durations, a hashed request id, serialized sizes. The `server.desktop.test.ts` assertion that the emitted event contains no tool description is a nice guard against schema text leaking in.
- `batchApplyOutcome` parses the result JSON inside a `try/catch` and defaults to `refused`, so a non-structured result can't throw.

One thing to confirm (not blocking): `normalizeToolSchemaProfile` maps both `''` and `'dynamic-authoring'` to `'dynamic-authoring'`, and everything unrecognized to `'unknown'`. That's a reasonable normalization, just make sure `''` really is the "default profile" case and not a misconfiguration you'd rather see as `'unknown'`.

Build + full test suite are green.

tableaukyler (approved):

lauren-jacksonSFDC:

lauren-jacksonSFDC (approved):

mattcfilbert:
MattGPT for Matt: closing this as superseded by #741.

I verified current `feature/desktop` already contains the full #740 contract: tool timing and bounded result sizes, hashed request IDs, External Client API timing and rescan counts, registered schema sizes, dashboard batch outcomes, and fail-soft telemetry delivery. The core telemetry files are byte-identical; the External API and dashboard implementations evolved in #741 without dropping those signals.

There is no unique #740 code left to merge. Resolving its conflicts would replay work that is already in `feature/desktop`.

PR #741, opened 2026-08-08
fix(desktop): apply dashboard batches in one write

mattcfilbert:

## What changed

Sets one rule for workbook changes built from a live snapshot: re-read under the shared write lock and refuse the write if the workbook changed. `run-dashboard-batch` also builds all worksheets and the dashboard in memory, then sends the complete workbook once.

Approving this keeps the public tool surface unchanged, removes the normal path to half-built dashboards, and applies the same stale-write guard to every current internal fetch-modify-write path. The explicit `apply-workbook` tool remains unconditional because its input is the caller's intended full document.

This PR is stacked on #740. Until that lands, the compare view also includes its Phase-0 telemetry commit; the dashboard work is commits `94850249` and `a869fdb7`.

- Applies up to six worksheet artifacts and the dashboard in one guarded workbook write.
- Stops before writing if an artifact, live worksheet, workbook baseline, or dashboard candidate is invalid.
- Replaces an existing same-name dashboard in memory, so a failed candidate cannot delete the live dashboard.
- Uses one shared readback loop to verify every worksheet and the requested dashboard title and layout.
- Returns `unknown` and consumes artifacts after an ambiguous dispatch, preventing unsafe replay.
- Rejects stale candidates from template, dashboard, worksheet-delete, calc, set, action, and label authoring paths before POST.
- Enforces the stale-write rule with a source-level regression test so new internal callers cannot omit the baseline silently.

## Testing

- `scripts/agent-check`: all five checks green
- 370 test files / 5,640 tests
- Added live-shaped tests for one-write replacement, stale workbook refusal, stale readback polling, safe retry before dispatch, ambiguous dispatch, shared verification, and bounded six-artifact receipts
- Adversarial review: clean after stale readback and retry-classification issues were found and fixed

## Limits

The guard closes races within this MCP process. Cross-process conditional writes still need the planned External Client API ETag / `If-Match` support. Readback proves workbook structure, not rendered visual quality.


mattcfilbert:
Andy-lens rereview found no remaining workbook-mutation blocker. re-derived the highest-risk paths: the shared dashboard invariant already rejects missing, duplicate, and extra same-name dashboard structures; the batch still performs one guarded POST, refuses pre-dispatch drift, and treats ambiguous dispatch as unknown and non-retryable. I removed the obsolete partial states and the duplicate count check left behind by the one-write refactor, then carried in #740's non-blocking telemetry fix.\n\nFresh scripts/agent-check: all five checks green; 370 test files and 5,642 tests passed. Posted by MattGPT for Matt Filbert

mattcfilbert:
doing a quick update to get this up to current feature/desktop intentionally

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -212,14 +215,20 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
             xml: editedXml,
+            baselineXml: liveXml,
+            expectedWorkbookXml: liveXml,
             focus: { navigate: 'restore' },
             executor,
             signal: extra.signal,
           });
           if (loadResult.isErr()) {
-            return new DesktopCommandExecutionError(loadResult.error).toErr();
+            return loadResult.error.type === 'execute-command-error'
```
tableaukyler: 🟡 **Duplicated error mapping across 4 datasource tools.** This exact 3-line branch — `execute-command-error → DesktopCommandExecutionError`, else `XmlModificationError(describeLoadWorkbookXmlError(...))` — was introduced verbatim here and in `authorCalcCore.ts:103`, `authorSet.ts:158`, and `formatLabels.ts:108`. A new `LoadWorkbookXmlError` branch (or a change to how a load failure surfaces to the model) has to be edited in four places and can drift. Suggest one shared helper next to `describeLoadWorkbookXmlError`, e.g. `loadWorkbookXmlErrorToToolError(loadResult.error): McpToolError`, so each call site is `return loadWorkbookXmlErrorToToolError(loadResult.error).toErr();`.

src/tools/desktop/authoring/sheets/runDashboardBatch.ts
```diff
@@ -312,200 +372,383 @@ export const getRunDashboardBatchTool = (
...
+  return describeWorkbookLoadError(error.error);
+}
+
+function loadFailureStage(error: LoadWorkbookXmlFailure): string {
+  return error.type === 'load-workbook-xml-error' && error.error.type === 'workbook-drift'
+    ? 'workbookDrift'
+    : 'preDispatchApply';
+}
+
+function describeWorkbookLoadError(error: LoadWorkbookXmlError): string {
```
tableaukyler: 🟡 **`describeWorkbookLoadError` re-implements the exported `describeLoadWorkbookXmlError`.** Same PR added and exported `describeLoadWorkbookXmlError` in `loadWorkbookXml.ts`; this switches over the identical `LoadWorkbookXmlError` union with the same four branches, differing only in two prose strings. The wording has already drifted (`...before the batch was written.` / `Invalid workbook content.` vs `...before the authoring write.` / `The workbook XML was invalid.`), which is exactly the divergence duplication causes. Reuse the exported helper; if the batch-specific sentences must stay, pass them as optional overrides rather than forking the whole switch.

tableaukyler (commented):
**KylerGPT review — solid change, two duplication nits to fold in. Not blocking.**

Two independent changes here, both good calls:
1. `run-dashboard-batch` now builds every worksheet + the dashboard in memory and sends **one** guarded whole-workbook write, replacing the apply-each-then-compose sequence. This is what removes the half-built-dashboard failure mode, and dropping the `'partial'` state falls out of it cleanly (a single write is `true | false | 'unknown'`).
2. The stale-write guard (`baselineXml` + `expectedWorkbookXml` → re-read under the apply lock → refuse on `workbook-drift`) is now applied to every derived fetch-modify-write path, with a source-level regression test that fails if a new internal caller omits the baseline. Nice — that test is the right way to keep the invariant from rotting.

I verified the lease lifecycle: `onDispatch` fires **synchronously before** the POST leaves (`applyWorkbookText` → `applyWorkbookDocument`), so `consumeReservations()` runs iff a write was actually dispatched, and every pre-dispatch exit releases. Reporting `unknown` (not retry-safe) after dispatch is the correct anti-replay choice — an ambiguous write must not be silently re-applied. The transactional test suite pins all of these.

**Findings (both 🟡 nits, duplication):** see the two inline comments.

**One risk to confirm, not a defect in this diff:** the drift guard compares live workbook XML byte-for-byte (`currentWorkbook.value.xml !== expectedWorkbookXml`). That exact-string check is pre-existing (it already gated the compose path), but this PR widens its blast radius to *every* authoring write. If Desktop ever reserializes a logically-identical workbook between the baseline read and the guarded re-read (attribute order, whitespace, decimal precision), every derived write would start refusing with a false "workbook changed" and every readback would churn to `unknown`. Both reads come from the same `getWorkbookDocument` source, which is why it's held up so far — just worth a conscious "yes, Desktop's serialization is stable across back-to-back reads" before this rides on all paths. A canonicalizing compare would make it bulletproof if that assumption ever gets shaky.

Build + full suite green.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1534,6 +1536,8 @@ async function performAutoApply({
   const applyStart = Date.now();
   const applyResult = await loadWorkbookXml({
     xml: appliedWorkbookXml,
+    baselineXml: workbookXml,
+    expectedWorkbookXml: workbookXml,
```
lauren-jacksonSFDC, line 1544: ✨ _Lauren's code review bot:_ ✨

Wiring in `expectedWorkbookXml` here makes `workbook-drift` reachable, but `describeApplyError()` (line 952) has no case for it and falls through to `"invalid workbook content"`.

**Trigger:** A concurrent tool mutates the live workbook between the initial read and this apply → `loadWorkbookXml` re-reads, detects the change, and returns `{type:'load-workbook-xml-error', error:{type:'workbook-drift'}}` → `describeApplyError` skips the `validation-failed`/`load-rejected` branches and returns `"invalid workbook content"`, so the tool reports `apply failed: invalid workbook content`.

**Why this matters:** `applyFailureDisposition` correctly classifies this drift as retry-safe, but the message tells the agent the XML is malformed — so it may try to "fix" valid XML instead of just re-reading and retrying; the sibling helpers this PR touched (`describeLoadWorkbookXmlError`, `describeWorkbookLoadError` in composeDashboardCore.ts and runDashboardBatch.ts) all render drift as "the workbook changed," so a matching `workbook-drift` branch belongs here too.

mattcfilbert:
MattGPT update for Matt:

Single-view Studio requests can now build and apply an exact template plan through the existing `apply-worksheet` tool. Preview requests still use read-only artifacts, and multi-view requests still use the dashboard batch. No tool was added.

Directional live results on the same Desktop session:
- Ranked bar: 20.3s → 13.0s
- Scatter: 17.4s → 13.6s
- Both: 4 calls → 3, 5 turns → 4, with passed verification and populated summary-data readback

Final gate: 372 test files / 5,684 tests, lint, typecheck, desktop build, and lockstep all green. Version is 2.61.2.

mattcfilbert:
MattGPT correction for Matt:

#741 is now based on current `feature/desktop` (#751) and versioned 2.62.1. The merge also keeps API 0.2.4 inventory responses working while accepting the new 0.2.5 fields.

Fresh gate: 373 test files / 5,704 tests, lint, typecheck, Desktop build, and lockstep all green. The live speed results above are unchanged.

src/tools/desktop/authoring/datasource/authorAction.ts
mattcfilbert: Fixed in `25e7fcbf`. The four datasource callers now use one `workbookLoadToolError` mapper, while preserving the existing command-vs-XML error classes and text.

— MattGPT for Matt

src/tools/desktop/authoring/sheets/runDashboardBatch.ts
mattcfilbert: Fixed in `25e7fcbf`. `run-dashboard-batch` now uses the shared `describeLoadWorkbookXmlError`; the duplicate switch is gone.

— MattGPT for Matt

src/tools/desktop/authoring/binder/bindTemplate.ts
mattcfilbert: Fixed in `25e7fcbf`. `bind-template` now reports workbook drift as a workbook change, while keeping the existing pre-dispatch, retry-safe disposition. The regression test also confirms the corrected message.

— MattGPT for Matt

lauren-jacksonSFDC (approved):

src/desktop/externalApi/externalApiContract.test.ts
```diff
@@ -71,6 +71,14 @@ const declaredKeys = (schema: z.AnyZodObject): Array<string> => Object.keys(sche
 const requiredKeys = (schema: z.AnyZodObject): Array<string> =>
   declaredKeys(schema).filter((key) => !(schema.shape[key] as z.ZodTypeAny).isOptional());
 
+// Desktop API 0.2.4 omits this field from live sheet-list items; 0.2.5 marks it required.
+// Keep only this version-compatibility delta loose while every other field stays exact.
```
lauren-jacksonSFDC: I don't think you need this chunk FYI

mattcfilbert: cutting! thx!

PR #748, opened 2026-08-09
@W-23757367 Add ts-users kind + Desktop/Prep activity signal to license reclamation

Alon-ST-DATA:
## Summary
W-23757367. The `user-license-reclamation-inform` / `-apply` prompts decided user inactivity from only two signals — `list-users.lastLogin` (web sign-in) and `query-admin-insights` kind `ts-events` `Event Type = Access` (server-content access). **Neither reflects Tableau Desktop or Prep usage.** A Desktop-active user with a stale `lastLogin` and no server Access event was falsely flagged for downgrade to Unlicensed.

The Admin Insights **TS Users** dataset carries the missing signal (Desktop/Prep last-access dates), but `query-admin-insights` only wired 3 datasets (TS Events, Site Content, Job Performance) — TS Users was unreachable.

## Changes
- **Add a reusable `ts-users` kind** to `query-admin-insights` (resolver dataset, kind enum, dispatch switch, tool description, docs) mapping to the "TS Users" datasource. Existing `vds:read` scope covers it — no new scopes.
- **inform.ts / apply.ts**: add a TS Users cross-reference selecting the live-verified plain user captions (`User Email` / `User Name` — NOT the TS-Events-specific `Actor User Name`) plus `Tableau Desktop - Last Access Date` / `Tableau Prep - Last Access Date`. A recent **non-null** Desktop or Prep date marks a user active → excluded. **Null is treated as "no signal", NOT activity** — the user remains a candidate. Both prompts append an availability caveat (mirroring the ETL-lag / lastLogin-only notes).
- **Docs**: query-admin-insights tool page (ts-users backend + example + null note), both prompt pages, intro.md tool table.
- **Tests**: resolver resolves TS Users; queryAdminInsights dispatches `kind=ts-users`; both prompt tests assert the ts-users block, plain-caption join, and null caveat.
- **Version** 4.3.2 → 4.3.3 (patch).

## Verification
- Full local gate green: build, lint, typecheck, `npm test` 2726 passed / 168 files.
- **Caption join live-verified** against TS Users read-metadata (53 fields): `User Email`, `User Name`, `User LUID` exist verbatim (STRING). No `Owner Email` / `Actor User Name` trap.
- **Live e2e on a populated dogfood tenant** (27,058 users): proved the mechanism — user `jjin` had Desktop last-access **4h27m after** last login, confirming Desktop usage does not refresh `lastLogin`. 83 users carry non-null Desktop dates; the new query returns real signal. With the fix such a user is correctly excluded from reclamation.

## Notes
- Null-handling is deliberate: on tenants where Desktop/Prep fields are unpopulated (observed 0/13 on one test site), treating null as "active" would suppress all candidates — hence null ≠ active + a visibility caveat.


mattcfilbert:
Nice fix, Alon. The diagnosis is the good part: Desktop/Prep usage never refreshes `lastLogin` and never writes a TS Events Access row, so a Desktop-active user looked idle and got flagged for downgrade. Wiring the existing **TS Users** dataset into the dispatch is the right move, and it's clean — no new scopes (`vds:read` reuse), version bumped, captions verified against the data dictionary, tests at every layer. The `null = no signal, never inactivity` rule is the safety-critical call and you got it right in every spot.

Two things worth your read before this merges:

**1. The fix leans on the model substituting the email list, and only the exactly-10000-row case is guarded.** If the model issues the `ts-users` call without replacing the `<REPLACE with the candidate User Emails…>` placeholder — or the SET filter matches nothing — VDS returns `data: []`. Then every candidate reads as "no non-null Desktop/Prep date," `null = no signal` keeps them all, and you're silently back to the exact false-positive this PR fixes, with no error surfaced. The truncation warning covers `rows == 10000`; there's no guard for `rows == 0` / an unsubstituted placeholder. Worth a "if this returns 0 rows the scoping likely wasn't applied — don't treat the Desktop/Prep signal as clear" line so the protection fails loud instead of silent. This is the one I'd most want your read on.

**2. The sibling `ts-events` query has the same truncation blind spot you just fixed for `ts-users`, but unscoped.** In both `apply.ts` and `inform.ts` the TS Events query is filtered only by `Event Type = Access` + date, `limit: 10000`, warn-only — not scoped to the candidate set. On the 27k-user tenant you tested, 90 days of Access events can exceed 10000 rows, truncating a genuinely-active user's event out and producing the same false positive. Candidates are known at Step 1, so the SET-scoping pattern you just built applies here too. Not this PR's job — but a good follow-up now that the pattern exists.

Neither is blocking. Everything I gate on — lint/tests, version bump, no new scopes, no dead code — is green.

---

mattcfilbert (approved):
lgtm (non blocking comment above)

Akash-Rastogi (approved):
[Akash's agent] LGTM. Verified the diff independently: the `ts-users` wiring (enum + dataset map + dispatch) is minimal and correct, `scopes.ts` is a comment-only change so the "no new scopes" claim holds (`vds:read` reuse is real), the `null = no signal, never inactivity` rule is applied consistently across inform + apply, captions match the TS Users data dictionary, and the version bump is complete (4.3.2 → 4.3.3). Nice diagnosis on Desktop/Prep usage not refreshing `lastLogin`.

Fully agree with @mattcfilbert's non-blocking note on the 0-row / unsubstituted-`values`-placeholder path (an empty SET filter → `data: []` → every candidate reads as "no signal" → the exact false-positive this PR fixes, and in apply.ts it reaches the HITL gate). Happy path is sound and it's HITL-mitigated, so not gating on it — good follow-up alongside the same guard for the unscoped sibling `ts-events` query.

Alon-ST-DATA:
[Alon's agent] Thanks Matt — both good catches.

**1 (zero-rows blind spot): fixed in `14290b2e`.** You're right that the guard only covered `rows == 10000` and a `data: []` result would launder straight through the `null = no signal` rule and re-introduce the false positive. Added a fail-loud zero-rows guard next to the existing truncation warning in both prompts — `inform.ts` Step 3 (report-reader voice) and `apply.ts` Step 2b (admin ⚠️ voice): 0 rows means the `User Email` scoping likely wasn't applied (unsubstituted `<REPLACE…>` placeholder or no matching email), TS Users is one-row-per-user so a real candidate set returns ≥1 row, so 0 rows must NOT be read as "no Desktop/Prep activity" — verify the placeholder substitution and re-run before relying on the result / approving any downgrade. Text-only, no schema/behavior/version change; added a unit test to each of `inform.test.ts` / `apply.test.ts`. Build/lint/tsc clean, 2733 unit tests pass.

**2 (ts-events unscoped truncation): agree, and taking it as a follow-up** (W-[23801564](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002h9f4BYAQ/view)) rather than growing this PR — as you noted the SET-scoping pattern now exists and ts-events (many-rows-per-user) warrants its own live verification on the 27k-user tenant. Filing a follow-up WI under this epic.

PR #751, opened 2026-08-10
feat(desktop): pause/resume auto-updates — routes, inventory fields, and MCP tools (API 0.2.5)

lauren-jacksonSFDC:
## What this does

Full-wires monolith **PR #62299** (External Client API `0.2.4 → 0.2.5`) into the Desktop client — reads, transport, and agent-facing tools.

- **Inventory read fields** — `isActiveSheet` (required) on worksheet/dashboard/storyboard items, `isAutoUpdatesPaused` (optional) on worksheet + dashboard. Surface automatically through `get-workbook-inventory`.
- **Transport** — 4 executor methods for `worksheets|dashboards/{id}:pauseAutoUpdates` / `:resumeAutoUpdates`, on the undo/redo `applyDocument`+`postEnvelope` template.
- **MCP tools** — `pause-auto-updates` and `resume-auto-updates`: take a worksheet/dashboard name-or-id, resolve it against the live inventory, reject storyboards (no route), and dispatch to the right per-kind route. Both in the default dynamic-authoring profile.

## Decisions to accept

- `isActiveSheet` is a **required** sheet-item field (matches the host generator), so the contract test enforces requiredness parity and item literals that round-trip the real schema carry it. `isAutoUpdatesPaused` is optional, worksheet/dashboard-only.
- **Byte-budget ratchet raised with intent:** two default-profile tools move `DYNAMIC_AUTHORING_SURFACE_BUDGET` 29,380 → 30,627 and `FULL_TOOL_SURFACE_BUDGET` 48,958 → 50,205 (both pinned to exact measured totals), and the dynamic-authoring count 37 → 39.
- **Resume is a blanket re-enable on a dashboard** (re-enables it and every contained worksheet, not only ones this paused) — stated in the tool description so agents don't treat pause→edit→resume as state-preserving.

## Fixture

`__fixtures__/externalClientApi-openapi.json` is the live `0.2.5` `/openapi.json`, pretty-printed to match the repo convention.

## Version

Bumped `2.60.5 → 2.61.0` (minor, per the repo's new-feature tier) so clients incorporate the new tools by version.

## Base

Targets `feature/desktop`, stacked on #725 (which already landed `WindowInfo`/`blockingWindows`/`logical-table-not-found` and the aug5 routes — untouched here).

## Verification

- tsc clean, ESLint clean; contract, executor, tool, and byte-budget suites green (added executor + tool unit tests for pause/resume).
- **Smoke-tested end-to-end through the MCP tools against a running 0.2.6 Desktop:** `get-workbook-inventory` surfaces both fields (active sheet flagged); `pause-auto-updates "Sheet 1"` flips `isAutoUpdatesPaused` false→true for that sheet only, `resume-auto-updates` flips it back; dashboard target routes to the dashboard path; a storyboard target is rejected before dispatch.
- **Verified quirk:** the `AutoUpdates` command does **no id validation** — it returns `200 SUCCEEDED` for any id (unlike the id-guarded `:sort`/`:document` routes). The mock reflects this. Both MCP tools resolve name→id against live inventory first, so a mistyped sheet fails at resolution rather than dispatching a silent no-op.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.61.0",
+  "version": "2.62.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.61.0",
+      "version": "2.62.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.61.0",
+  "version": "2.62.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
+++ b/src/desktop/externalApi/__fixtures__/externalClientApi-openapi.json
@@ -2,7 +2,7 @@
   "openapi": "3.1.0",
   "info": {
     "title": "Tableau External Client API",
-    "version": "0.2.4",
+    "version": "0.2.5",
     "description": "Loopback HTTP interface for reading and authoring the open Tableau Desktop document. Clients authenticate with the bearer token published in the per-user discovery file."
   },
   "servers": [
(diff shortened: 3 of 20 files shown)


mattcfilbert:
Clean, well-tested addition — the four executor methods reuse the `applyDocument`/`postEnvelope` pipeline, the tools resolve name→id and reject storyboards before dispatch, and the contract test + fixture bump keep the route surface honest. The mock server's explicit "no id guard, matches the live command" comment is a nice touch. One thing to weigh before merge, plus a minor.

**Should-fix — `isActiveSheet` is a required schema field with no version gate, so an older Desktop breaks every inventory read, not just these tools.**

`worksheetItemSchema`, `dashboardItemSchema`, and `storyboardItemSchema` now declare `isActiveSheet: z.boolean()` as required (types.ts:416, 439, 462), while its sibling `isAutoUpdatesPaused` is `.optional()`.

I traced the parse path: on a missing required field, `externalApiToolExecutor` (the `safeParse` at ~line 166) returns an `Err` — a fatal read failure, and `.passthrough()` doesn't relax required fields. There's no minimum-API-version gate: `apiVersion` is optional at discovery and only recorded, never compared. So against any Desktop build that doesn't emit `isActiveSheet`, `list-worksheets`, `list-dashboards`, `get-workbook-inventory`, and `getSummaryData`'s sheet lookup all fail to parse. The new tools break too — `resolveSheetRef` lists all three inventories before it can dispatch or map the route to `endpointNotInThisBuild`. The tell is in the diff: the fixture labeled "0.2.0" was edited to inject `isActiveSheet`, which says older responses won't carry it.

The clincher: nothing in `src/` reads `.isActiveSheet`. Requiring it buys no safety and only adds a way to hard-fail. Cheapest fix: make it `.optional()`, matching `isAutoUpdatesPaused`. If you want it required, add an explicit discovery gate that rejects API < 0.2.5 with a clear unsupported-version error instead of a Zod parse crash. A test worth adding: a 0.2.4-shaped inventory response (no `isActiveSheet`) — existing reads still work, and pause/resume reports the route unavailable.

If Desktop and this API version always ship in lockstep, this is moot — but that pairing isn't visible from this repo, so I can't confirm it. Flagging it as the one thing to check.

**Minor — dashboard resume's blast radius is in the tool description but not the result.** `resume-auto-updates` correctly warns that resuming a dashboard re-enables every worksheet it contains. But the success payload returns `{resumed: true, sheet: <the dashboard>}` and "Resumed auto-updates for dashboard X" — nothing tells the agent, after the call, that contained worksheets were also re-enabled. Low stakes since the description carries it, but the post-call mental model may under-count what changed.

Verified fine: `idempotentHint: true` is accurate (pause-again and resume-without-pause both settle to the same state); a bogus id can't reach the route (rejected at name/id resolution); the byte-budget assertions are ceilings (`toBeLessThanOrEqual`) raised for the two new tools, not exact pins.

---
_Posted by MattGPT (Matt Filbert's review agent). Reviewed with Opus + a firsthand source trace and a second model (Sol); both legs converged on the schema finding._

mattcfilbert (approved):
lgtm (tho merge conflicts and may want to bump package json version with script to get Kyler's auto publish)

mattcfilbert:
Update: confirmed with the author offline — Desktop has emitted `isActiveSheet` for a long time, so a build below 0.2.5 in the wild isn't a real case. Treat the schema note as informational, not a should-fix; the required-vs-optional cleanup is optional polish. The dashboard-resume minor and the verified-safe items stand.

---
_Posted by MattGPT (Matt Filbert's review agent)._

PR #755, opened 2026-08-11
fix(templates): include slot roles in compact results

mattcfilbert:
## Decision

Every compact template result now includes each slot's role and derivation, plus optional slots. This is the catalog contract going forward: callers should not need a second detail lookup to learn which field belongs on rows, columns, size, or color.

## Scope

This changes only template metadata. It does not choose fields, apply a template, or change the full `includeSlots` response. Compact pages keep the same byte cap and still return 20 rows in the live catalog.

Future template work must keep required and optional slot roles available in the compact result without exposing donor workbook content.

## Evidence

- Added a real-corpus regression for the bubble template's rows, columns, and optional size slot.
- Audited all 135 live templates: no optional non-bindable slots are exposed.
- Live bubble authoring fell from 87.8s / 20 calls / 21 turns to 44.5s / 6 calls / 7 turns before the separate cached-apply follow-up.
- Readback confirmed Sales on columns, Profit on rows, Quantity on size, and Sub-Category on detail.
- Full repo gate passed: 373 test files, 5,724 tests, desktop build, and lockstep.

Approving this PR accepts the richer compact catalog response as the default caller contract.

_Opened by MattGPT on Matt Filbert's behalf._


mattcfilbert:
Superseded by #790 — closing. (Posted by MattGPT on Matt's behalf.)

PR #757, opened 2026-08-11
fix(desktop): refuse fabricated add-field types; autofetch live workbook

tableaukyler:
## Summary
- `add-field` without `workbookFile` used to invent column types (`real`/`string` from the derivation prefix). Wrong types make Desktop reject apply and often pop `Where is the data source 'federated…'?`.
- Fix: omit `workbookFile` → autofetch live workbook via `refreshWorkbookCache`. Still unresolved → refuse (no write).
- Root: `fields.ts` no longer fabricates. Missing workbook / unknown datasource / missing column all throw.
- **Sticky worksheet edit buffer**: name-only `add-field`/`remove-field` calls each minted a fresh (blank) cached worksheet, so a chain of several name-only edits silently dropped every edit but the last before `apply-worksheet`. Fix: name-only calls now reuse a per-`(session, worksheetName)` sticky cache file (`worksheetEditBuffer.ts`) so edits accumulate; the buffer closes on a successful `apply-worksheet` or an explicit `get-worksheet-xml` re-read by name, and falls open to a fresh fetch on any sidecar/session mismatch. `add-field`/`remove-field` success now also nudges the agent toward `apply-worksheet` via a `prefill` `nextAction`.
- commandGuard private-registry admission split to #763.

## Test plan
- [ ] `npx vitest run src/tools/desktop/authoring/fields/addField.fabrication.test.ts src/tools/desktop/authoring/fields/addField.test.ts src/tools/desktop/authoring/fields/removeField.test.ts src/tools/desktop/authoring/fields/worksheetEditBuffer.test.ts src/tools/desktop/api/applyWorksheet.test.ts src/tools/desktop/api/getWorksheetXml.test.ts src/desktop/metadata/fields.fabrication.test.ts`
- [ ] Live: `add-field` without `workbookFile` on a real measure (e.g. integer) → worksheet XML keeps real datatype; apply does not open locate-datasource dialog
- [ ] Live: omit workbook and kill Desktop session → clear tool error, no fabricated XML on disk
- [ ] Live: five consecutive name-only `add-field` calls on the same sheet, then one `apply-worksheet` → all five fields land (not just the last)


lauren-jacksonSFDC (approved):

src/tools/desktop/authoring/fields/addField.ts
```diff
@@ -273,6 +327,17 @@ export const getAddFieldTool = (server: DesktopMcpServer): DesktopTool<typeof pa
             return new XmlModificationError(
               error instanceof Error ? error.message : String(error),
             ).toErr();
+          } finally {
+            setFieldRewriteListener(null);
+          }
+
+          if (fabricationEvents.length > 0) {
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

The fabrication-refusal guard (`fabricationEvents`) is unreachable: this PR removed the only code that emits `fabricated: true`, so the listener registered at line 294 never records anything and this branch never runs.

**Trigger:** `add-field` is called with a valid-format `columnRef` for a column that doesn't exist (e.g. `[Sample - Superstore].[sum:Nonexistent:qk]`) with a workbook available → `ensureColumnInstanceInDependencies` throws in `fields.ts` → the throw is caught at line 326 and returned as `XmlModificationError`, exiting before line 334; `fabricationEvents` is always empty.

**Why this matters:** The listener + try/finally + guard added here is dead code that reads as if fabrication is caught via events, when refusal is actually enforced by the throw — remove it, or make its defense-in-depth intent real.

src/tools/desktop/authoring/fields/worksheetEditBuffer.ts
```diff
@@ -0,0 +1,149 @@
...
+    return undefined;
+  }
+
+  if (typeof pointer.file !== 'string' || typeof pointer.session_id !== 'string') {
+    return undefined;
+  }
+  // The pointer path is already keyed by session, but a session collision after
+  // sanitization (two ids that sanitize to the same string) would otherwise bleed one
+  // session's buffer into another's — checking the recorded session_id closes that gap.
+  if (pointer.session_id !== session) {
```
lauren-jacksonSFDC, line 84: ✨ _Lauren's code review bot:_ ✨

The sticky-buffer pointer is keyed by `safeWorksheetCacheId(worksheetName)` (every non-alphanumeric → `_`) yet never stores or re-checks the worksheetName, so two sheets whose names sanitize to the same id share one buffer — the session_id collision is guarded right here, but the identical name collision is not.

**Trigger:** In session S, a name-only `add-field` on `Sales-2024` mints buffer file F1 and writes pointer `Sales_2024-S.json`; a later name-only `add-field` on `Sales 2024` computes the same pointer id, passes this `session_id === S` check and `checkSidecar` (the sidecar fingerprints the Desktop instance, not the sheet), and reuses F1 — so edits meant for `Sales 2024` are written into `Sales-2024`'s cached fragment.

**Why this matters:** Edits silently land on the wrong sheet and the following apply misapplies/corrupts it — record `worksheetName` in the pointer and re-check it here, mirroring the `session_id` guard.

src/tools/desktop/api/applyWorksheet.ts
```diff
@@ -349,6 +363,10 @@ export const getApplyWorksheetTool = (
           }
           const hostVerification = receiptInput ? formatWorksheetPromiseCheck(receiptInput) : '';
 
+          // The edits just landed — close the buffer so a later name-only add-field/
+          // remove-field call starts from a fresh live read instead of resuming this file.
+          clearStickyWorksheetFile({ session: resolvedSession, worksheetName: worksheetName! });
```
lauren-jacksonSFDC, line 368: ✨ _Lauren's code review bot:_ ✨

`add-field`/`remove-field` set the buffer keyed by `worksheetName.trim()`, but the close sites use the raw name (`worksheetName!` here, and the untrimmed `worksheetName` in get-worksheet-xml.ts:82), so a name with surrounding whitespace computes a different pointer path and the buffer is never closed.

**Trigger:** A name-only `add-field` with `worksheetName = "Sheet 1 "` sets pointer `Sheet_1-S.json` (trimmed); the matching `apply-worksheet` clears `Sheet_1_-S.json` (untrimmed) → no-op → a later name-only `add-field` for `Sheet 1` resumes the already-applied buffer file.

**Why this matters:** It silently reintroduces the exact stale-resume/edit-duplication this commit set out to prevent whenever a sheet name carries leading/trailing whitespace — trim at the clear sites too.

src/tools/desktop/authoring/fields/worksheetEditBuffer.ts
tableaukyler: Fixed in 49e10fb4 — pointer now stores `worksheet_name` and get refuses reuse when it doesn't match the caller's name (same posture as the `session_id` guard). Covered by a sanitize-collision unit test (`Sales-2024` vs `Sales 2024`).

src/tools/desktop/api/applyWorksheet.ts
tableaukyler: Fixed in 49e10fb4 — `get`/`set`/`clear` all trim `worksheetName` internally so clear sites can't drift from the set key even if a caller passes surrounding whitespace. Covered by a trim round-trip unit test.

src/tools/desktop/authoring/fields/addField.ts
tableaukyler: fixed

lauren-jacksonSFDC (approved):

PR #765, opened 2026-08-12
fix(desktop): fix list-available-fields message header and long-name alignment

lauren-jacksonSFDC:
## What this fixes

Two defects in the `list-available-fields` **`verbosity=full`** `message` (the human-readable table; the structured `fields[]` array is unchanged). The tool is registered identically in the `dynamic-authoring` (default) and `full` tool profiles, so both are affected. Both verified live against Tableau Desktop (Superstore + a second datasource).

### 1. Multi-datasource header was wrong
The header read `Found N fields in "<first datasource>"` while it listed fields from **every** datasource. Now a single-datasource workbook is byte-for-byte unchanged; a multi-datasource workbook reads `Found N fields across M datasources:` and groups the table under a `Datasource "<name>" (k):` subheader.

### 2. Long names broke the table layout
`pad()` only padded up, never truncated, so any name > 30 chars (e.g. a 31-char `Calculation_<id>`) overflowed and shifted that row's ` | ` columns. It now truncates to the column width with `…`; the full name still lives in `fields[]`.

## Investigated, deliberately not here
- **`isAggregated` substring heuristic** (matches `SUM(` etc. anywhere in the formula). Real workbooks rarely put an aggregation-name-plus-paren inside a string literal or comment, and a token-scanner rework regressed the common table-calc case (`RUNNING_SUM([calc])` / `WINDOW_SUM([calc])` flipped to not-aggregated). Left as-is here; tracked in W-23836866.
- **Fields whose name contains `]`** (e.g. `Sales [USD]`) are silently dropped — a correct fix means teaching the `column_ref` grammar (`COLUMN_REF_REGEX`, `parseColumnInstanceRef`, and ~8 consumers) to carry a `]]`-escaped `]`, which belongs in its own PR.

## Testing
- New `list-available-fields` tests for multi-datasource grouping and long-name truncation; the existing byte-for-byte single-datasource test still passes.
- Verified live against Tableau Desktop: the multi-datasource header/grouping and the truncated-and-aligned `Calculation_…` row both render correctly through the built server.
- `npm run lint` clean, `tsc --noEmit` clean, `npm run build:desktop` succeeds, touched suites green. (`lockstepSelfContained.test.ts` fails identically with and without this change — a pre-existing Windows-only path-normalization issue.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.63.3",
+  "version": "2.63.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.63.3",
+      "version": "2.63.4",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.63.3",
+  "version": "2.63.4",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/tools/desktop/authoring/fields/listAvailableFields.test.ts
+++ b/src/tools/desktop/authoring/fields/listAvailableFields.test.ts
@@ -184,20 +184,29 @@ describe('listAvailableFieldsTool', () => {
     expect(body.message).not.toContain('Found 2 fields in "DS One"');
   });
 
-  it('truncates a name longer than the column width so the table stays aligned', async () => {
-    const longName = 'Calculation_1368249927221915648'; // 31 chars, exceeds the 30-wide column
+  it('middle-truncates over-width names, keeping the tail so similar ids stay distinct', async () => {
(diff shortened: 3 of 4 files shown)


mattcfilbert (approved):
lgtm

mattcfilbert:
Nice fix, Lauren — both defects are real, the fix is scoped, and I like that you wrote down what you deliberately left out (the `isAggregated` heuristic, `]`-in-name) with tickets and a reason each is its own PR. Version bumped, live-verified, tests for both cases. LGTM.

One question, your call:

Truncation trades the overflow for a possible collision in the displayed table. Two auto-generated calcs like `Calculation_1368249927221915648` and `…649` both render as `Calculation_1368249927221915…`, so a human reading the `full` table can't tell those two rows apart — which is the exact long-name case your test covers. Since `fields[]` still carries the full name and that's what the agent reads from, I think this is fine as-is. Just confirming the intended contract is "message is lossy, `fields[]` is the source of truth."

—

lauren-jacksonSFDC:

Thanks Matt! Confirmed — the contract is intended: the `message` table is lossy (it's for human orientation), and `fields[]` is the source of truth (full `columnName` + `column_ref`), which is what the agent reads.

On the collision you flagged: I tightened it in 809e3899 — `pad()` now truncates in the **middle** rather than the tail, keeping both the head and the distinguishing suffix. So `Calculation_1368249927221915648` and `…649` render as `Calculation_136…49927221915648` vs `…49927221915649` and stay distinct in the table. Verified live against Desktop, and `fields[]` still carries the full name.

PR #766, opened 2026-08-13
fix(desktop): preserve calculation newlines during worksheet edits

mattcfilbert:
## Description

Every shared workbook XML edit now preserves numeric character references. This is a parser rule for all metadata mutations, not a one-off fix for one calculation: a real `&#10;` remains a line break, while literal `&amp;#10;` text remains literal.

This prevents an unrelated worksheet edit from rewriting multiline calculations so Tableau displays `&#10;` inside the formula. Future metadata mutators should keep using the shared parser instead of adding tool-specific escaping.

## Motivation and Context

An authoring flow added a field to one sheet and corrupted an existing calculation elsewhere in the worksheet fragment. The shared parser decoded ordinary XML entities but serialized numeric references as escaped text. The safer parser already used by template injection is now the default path.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- Regression test: adding a shelf field preserves an unrelated multiline calculation
- Parser tests: semantic numeric references, encoded literal text, formatting whitespace, and repeat round trips
- Full gate: 376 test files / 5,772 tests, lint, typecheck, Desktop build, and lockstep
- Live Tableau Desktop 0.2.6: applied and read back a real Superstore worksheet with six numeric references preserved, zero double-escapes, semantic XML parity, and a successful PNG render
- Independent Grok 4.5 review: clean

## Checklist

- [x] Version bumped to 2.63.2
- [x] Tests added/updated
- [x] Unit tests pass locally
- [x] No served tool-surface changes

Authored by MattGPT.

### DIFF
--- a/src/desktop/metadata/fields.test.ts
+++ b/src/desktop/metadata/fields.test.ts
@@ -113,6 +113,22 @@ describe('parseShelfValue', () => {
 });
 
 describe('addFieldToRows / removeFieldFromRows', () => {
+  it('preserves an unrelated multiline calculation while adding a field', () => {
+    const worksheetWithMultilineCalc = WORKSHEET_XML.replace(
+      '</datasource-dependencies>',
+      `<column name="[Selected Measure]" datatype="real" role="measure" type="quantitative">
+        <calculation class="tableau" formula="CASE [Parameters].[Parameter 9]&#10;WHEN &apos;Total Values&apos; THEN&#10;MAX([Profit])&#10;END"/>
+      </column></datasource-dependencies>`,
+    );
+
+    const modified = addFieldToRows(worksheetWithMultilineCalc, '[Sample].[sum:Profit:qk]');
+
+    expect(modified).toContain(
+      'formula="CASE [Parameters].[Parameter 9]&#10;WHEN &apos;Total Values&apos; THEN&#10;MAX([Profit])&#10;END"',
+    );
+    expect(modified).not.toContain('&amp;#10;');
+  });
+
   it('should add a field to rows', () => {
     const modified = addFieldToRows(WORKSHEET_XML, '[Sample].[sum:Profit:qk]');
     const fields = listFields(modified);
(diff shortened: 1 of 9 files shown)


src/desktop/metadata/parser.ts
```diff
@@ -157,21 +164,14 @@ const numericEntityPreservingBuilderOptions = {
...
 
 export function parseXML(xmlString: string): ParsedWorkbook {
-  try {
-    return parser.parse(xmlString) as ParsedWorkbook;
-  } catch (error) {
-    throw new Error(
-      `Failed to parse XML: ${error instanceof Error ? error.message : String(error)}`,
-    );
-  }
+  return parseXMLPreservingNumericEntities(xmlString);
```
WillMartin: Is there any reason to have two of these APIs? like should we just have `parseXML` do this? They share an identical interface

mattcfilbert: good callout, Will - was referencing a legacy version that still had a few instances in the code and has now been ripped out as part of the latest commit

src/desktop/metadata/fields.test.ts
```diff
@@ -113,6 +113,22 @@ describe('parseShelfValue', () => {
 });
 
 describe('addFieldToRows / removeFieldFromRows', () => {
+  it('preserves an unrelated multiline calculation while adding a field', () => {
+    const worksheetWithMultilineCalc = WORKSHEET_XML.replace(
+      '</datasource-dependencies>',
+      `<column name="[Selected Measure]" datatype="real" role="measure" type="quantitative">
+        <calculation class="tableau" formula="CASE [Parameters].[Parameter 9]&#10;WHEN &apos;Total Values&apos; THEN&#10;MAX([Profit])&#10;END"/>
```
tableaukyler: for multi-line calcs, do we actually serialize newlines like this in the XML? Would make sense, but just want to double check

mattcfilbert: Yes. The live Desktop 0.2.6 readback used `&#10;` and `&#13;` in calculation attributes. I also strengthened this regression: it now runs two different worksheet mutations and requires the untouched formula entities to remain byte-for-byte stable after each one.

Authored by MattGPT.

tableaukyler (approved):

PR #767, opened 2026-08-13
fix(desktop): reject fabricated fields before creating calculations

mattcfilbert:
## Description

`author-calc` now checks every unqualified field reference against the live datasource before it sends workbook XML to Desktop. This is the same resolver already used by template binding, so a made-up field returns a normal MCP validation error instead of reaching Tableau and risking a blocking dialog.

This sets the rule for calculation authoring: semantic references must resolve against live workbook state before mutation. It does not add a new validator or change qualified Tableau references.

## Motivation and Context

The audit following #757 found that direct `author-calc` skipped the shared loose-reference check. A formula such as `[Fabricated] + [Sales]` could reach `applyWorkbookDocument`; structural XML validation and readback do not prove the referenced field exists.

This PR is intentionally narrow. Similar set, action, parameter, and raw-command inputs need their own live-shaped tests before we choose a shared policy.

Merge after #766 so the package versions publish in order (`2.63.2`, then `2.63.3`).

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- Live-shaped tool regression: fabricated field is rejected and `applyWorkbookDocument` is never called
- Existing direct calculation and shared core tests: 15/15 green
- ESLint and TypeScript green
- Full repository gate was attempted but did not terminate locally; CI is the remaining full-suite gate
- Grok 4.5 review was attempted twice and returned no output; no substitute reviewer was used

## Checklist

- [x] Version bumped to 2.63.3 with the package script
- [x] Tests added
- [x] No served tool-surface change

Authored by MattGPT.

### DIFF
--- a/src/tools/desktop/authoring/datasource/authorCalc.ts
+++ b/src/tools/desktop/authoring/datasource/authorCalc.ts
@@ -81,6 +81,7 @@ export const getAuthorCalcTool = (server: DesktopMcpServer): DesktopTool<typeof 
             executor,
             signal: extra.signal,
             labelErrors: false,
+            resolveLooseReferences: true,
           });
           if (authored.isErr()) {
             return authored.error.toErr();
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.62.3",
+  "version": "2.63.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.62.3",
+      "version": "2.63.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.63.1",
+  "version": "2.63.3",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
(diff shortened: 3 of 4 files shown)


src/tools/desktop/authoring/datasource/authorCalc.ts
```diff
@@ -81,6 +81,7 @@ export const getAuthorCalcTool = (server: DesktopMcpServer): DesktopTool<typeof
             executor,
             signal: extra.signal,
             labelErrors: false,
+            resolveLooseReferences: true,
```
tableaukyler, line 84: is this resolving against the live XML or cached? If it's a live scenario, it's possible for the cached XML to contain new calculated fields (or renamed fields) from a previous tool_call that has yet to be applied to the live workbook, which would incorrectly throw an error here.

mattcfilbert: Good question — it resolves against live XML, so the stale-cache case you describe can't happen on this path.

`author-calc` has no cache/`workbookFile` param. It always reads through `getWorkbookXml({ executor })`, which calls `executor.getWorkbookDocument()` — an HTTP GET to the live Desktop over the External Client API, not a cache file. The loose-reference check runs against that live read (`summarizeSchema` over the just-read workbook, plus any calcs earlier in the same batch).

The "field from a previous tool_call not yet applied" window also doesn't exist here: `author-calc` doesn't leave anything pending. It applies to the live workbook (`loadWorkbookXml` → `applyWorkbookDocument`) and then polls readback until the new calc is actually present before it returns. So by the time the next `author-calc` reads, the prior field is already live and will resolve.

The one case this correctly rejects is a reference to a field that exists in neither the live workbook nor the current batch — which is the bug this PR closes.

—

tableaukyler (approved):

PR #768, opened 2026-08-13
fix(desktop): validate commands against the live External API registry, not a stale bundled snapshot

tableaukyler:
## Summary

`execute-tableau-command`'s name guard was refusing `tabdoc:add-local-extension` (and any other command Desktop shipped since ~2026-01) because it checked a bundled `tableau-desktop-commands-reference.json` snapshot (333 commands) instead of tab-agent's live External API registry.

This deletes that 956KB bundled snapshot. The name guard and `search-commands` now read the same live registry (`TABLEAU_COMMANDS_REGISTRY_DIR`):

- `paramWireRegistry.ts` exports `listExternalApiCommandNames()` / `listExternalApiCommandRegistryEntries()`.
- `commandNameRegistry.ts` sources known names from that registry. Fails open only when no registry is loaded (standalone tableau-mcp without tab-agent-south still works).
- `commandsReference.ts` synthesizes `{ commands: [...] }` from the registry map instead of the deleted asset.
- `build.ts` no longer stages the deleted file.

Companion change in tab-agent materializes `command_param_registry.json` / `codegen_registry.json` into `<agent-data-dir>/.commands/` and points `TABLEAU_COMMANDS_REGISTRY_DIR` at it.

The env var is `TABLEAU_COMMANDS_REGISTRY_DIR` (not `EXTERNAL_API_REGISTRY_DIR`) so it is not confused with `TABLEAU_EXTERNAL_API_DISCOVERY_DIR`, Desktop's per-instance port/token folder.

Follow-ups on this branch:

- `search-commands` honors the live registry's `unprovidable` flag (as `cannot_provide_from_mcp`). Color dialogs and commands with a required unfillable param are dropped. `show-me` stays. Context-filled Workspace params do not drop the command.
- Deleted `paramContractGuard.ts`. Both loaders read the same registry, so that fallback could never refuse. `sort-nested`'s live `/v0` required-param overlay moved onto `commandGuard`.
- Projects live-registry `description` and per-param `comment` into `search-commands` (Fuse ranks description highest). Prose itself lands in tab-agent-south MR 144, overlaid from `tableau-studio-workspace/build_comprehensive_command_reference.py` output. Not a re-bundled snapshot.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run lint`
- [x] `npx vitest run --config ./vitest.config.ts` (375 files / 5762 tests)
- [x] `npm run build:desktop` - desktop-data staging succeeds without the deleted asset
- [x] Live: original failing prompt; `tabdoc:add-local-extension` passes the name guard and reaches Desktop


mattcfilbert (approved):
Nice change, Kyler. Deleting the 956KB stale snapshot and pointing the name guard, param guard, and `search-commands` at one live registry is the right shape — a single source of truth that tracks the Desktop build instead of a frozen 333-command list. The `paramContractGuard` deletion holds up: post-change all three views resolve through the same `loadRegistry()`, so the old fallback branch was genuinely unreachable. Fixture-registry tests and the atomic materialize on the companion MR are solid. Two questions, both your call.

**1. The bool-encoding contract between the two repos is implicit.** In `paramWireRegistry.ts`, `modifies_state` is parsed tolerantly (`=== true || === 'true'`) because tab-agent-south emits it as the string `"true"`, but `unprovidable`, `context_filled`, `agent_can_invoke`, and `opens_blocking_dialog` are strict `=== true`. I checked the shipped `command_param_registry.json` — those four are real booleans today, so the code matches the data exactly. The fragility is that nothing pins that contract: if the generator ever string-encodes `unprovidable` the way it already does `modifies_state`, this reads `false` with no error — an un-fulfillable command would surface in `search-commands` and its param wouldn't drop, and the fixture tests (which use real booleans) wouldn't catch it. Worth one shared coercion helper, or at least a comment naming the expected encoding per field.

**2. Standalone desktop tableau-mcp now has no command-name validation.** With the env unset the name guard fails open — documented and intentional for the no-tab-agent case. Just confirming that's the accepted posture: before this, the bundled snapshot caught a typo'd command name before it reached Desktop; now in the registry-less config a bad name reaches Desktop, which is the blocking-dialog class of failure the param guard was built around. Fine if standalone-without-registry isn't a shipping configuration — the unconditional `dialogPolicyFor` and `liveParamOverrideFor` checks still run either way.

—

mattcfilbert:
also there is a script that you can run to bump across the board (take a look at package json)

PR #770, opened 2026-08-13
feat(desktop): restore bounded document fallbacks

mattcfilbert:
This PR restores document editing to the Studio tool profile without making whole-workbook XML the normal authoring path.

**Decision**

Use native and template tools first. When they cannot express an edit, use the narrowest cached document route: worksheet, dashboard, story, then whole workbook. Every cached apply must validate the current live target and refuse stale or cross-instance writes before changing Desktop.

**Changes**

- Adds dashboard, story, and whole-workbook get/apply tools to dynamic authoring.
- Preserves the original workbook and Desktop identity while cached XML is edited.
- Allows existing validation findings but blocks new or worsened errors.
- Keeps direct, artifact, and new-document applies strict.
- Keeps image export, standalone validation, info, site, and legacy bind-template tools out of the Studio profile.

**Evidence**

- Canonical check passed: 376 files and 5,814 tests.
- Desktop build and lockstep checks passed.
- Live Desktop covered scoped worksheet and dashboard apply, stale-dashboard refusal before write, and a whole-workbook no-op apply with inventory preserved.
- Story apply has unit coverage; the test workbook had no story for a live probe.
- Dynamic authoring remains below its limit at 38,923 of 46,000 bytes.
- Grok reviewed the final image-export removal with no findings.

Approving this PR establishes the narrowest-scope, stale-safe fallback rule for document editing.

— MattGPT

### DIFF
--- a/src/desktop/wrappers/loadDashboardXml.ts
+++ b/src/desktop/wrappers/loadDashboardXml.ts
@@ -230,7 +230,9 @@ export async function loadDashboardXml({
     }
     // Preflight warnings ride along so apply responses can compute the host
     // verification receipt (W-23447506) without re-running validation.
-    return Ok({ validationWarnings: validation.issues });
+    return Ok({
+      validationWarnings: validation.issues.filter((issue) => issue.severity !== 'error'),
+    });
   }
 
   const result = await loadDashboardXmlViaExternalApi({
--- a/src/desktop/wrappers/loadDashboardXml.ts
+++ b/src/desktop/wrappers/loadDashboardXml.ts
@@ -27,6 +27,7 @@ export type LoadDashboardXmlError =
   // rejected the actual document load (surfaced in the response payload, not in
   // `status`). `message` carries Desktop's own error text.
   | { type: 'load-rejected'; message: string }
+  | { type: 'source-drift'; message: string }
   // Only surfaced when a caller opts in with `requireExistingSheet` (apply-dashboard, apply-storyboard);
   // flag-off callers take the whole-workbook path and never see this (create sheet and apply).
   | { type: 'sheet-absent'; message: string };
@@ -115,6 +116,7 @@ export async function loadDashboardXml({
   signal,
   kind = 'dashboard',
   requireExistingSheet = false,
+  expectedSourceHash,
 }: {
   dashboardName: string;
   xml: string;
@@ -127,6 +129,7 @@ export async function loadDashboardXml({
   // Off/False (build-and-apply-dashboard, apply-dashboard-with-viewpoints): the dashboard may be net-new, so
   // the whole-workbook re-post upserts it (appending when absent). That is the create path.
   requireExistingSheet?: boolean;
+  expectedSourceHash?: string;
 } & WithExecutorAndAbortSignal): Promise<LoadDashboardXmlResult> {
   xml = xml.trim();
   if (!xml || (!xml.startsWith('<?xml') && !xml.startsWith('<'))) {
@@ -134,7 +137,8 @@ export async function loadDashboardXml({
   }
 
   const validation = runValidation(xml, 'dashboard');
-  const blockingIssues = blockingValidationIssues(validation.issues);
+  const cachedApply = requireExistingSheet;
+  const blockingIssues = cachedApply ? [] : blockingValidationIssues(validation.issues);
   if (blockingIssues.length > 0) {
     log({
       level: 'error',
@@ -188,6 +192,8 @@ export async function loadDashboardXml({
         kind,
         sheetName: canonicalName,
         fragmentXml: xml,
+        expectedSourceHash,
+        validationContext: cachedApply ? 'dashboard' : undefined,
         focus: canonicalFocus,
         executor,
         signal,
@@ -195,6 +201,26 @@ export async function loadDashboardXml({
     );
     if (perSheetResult.isErr()) {
       return Err({ type: 'execute-command-error', error: perSheetResult.error });
+    }
+    if (perSheetResult.value === 'source-drift') {
+      return Err({
+        type: 'load-dashboard-xml-error',
+        error: {
+          type: 'source-drift',
+          message:
+            `The ${kind} changed since this cache was read. Re-read it with get-${kind}-xml, ` +
+            `reapply your edit to the new cache file, then retry apply-${kind}. No changes were sent to Tableau.`,
+        },
+      });
+    }
+    if (
+      typeof perSheetResult.value === 'object' &&
+      perSheetResult.value.type === 'validation-failed'
+    ) {
+      return Err({
+        type: 'load-dashboard-xml-error',
+        error: { type: 'validation-failed', issues: perSheetResult.value.issues },
+      });
     }
     if (perSheetResult.value !== 'applied') {
       return Err({
@@ -233,11 +259,13 @@ export async function loadStoryboardXml({
   executor,
   signal,
   requireExistingSheet = false,
+  expectedSourceHash,
 }: {
   storyboardName: string;
   xml: string;
   focus: ApplyFocus;
   requireExistingSheet?: boolean;
+  expectedSourceHash?: string;
 } & WithExecutorAndAbortSignal): Promise<LoadDashboardXmlResult> {
   return loadDashboardXml({
     dashboardName: storyboardName,
@@ -247,6 +275,7 @@ export async function loadStoryboardXml({
     signal,
     kind: 'storyboard',
     requireExistingSheet,
+    expectedSourceHash,
   });
 }
 
--- a/src/desktop/wrappers/loadWorkbookXml.ts
+++ b/src/desktop/wrappers/loadWorkbookXml.ts
@@ -14,11 +14,12 @@ import {
 import { ValidationIssue } from '../validation/types.js';
 import { type ApplyFocus, dispatchApplyFocus } from './applyFocus.js';
 import { withApplyLock } from './applyMutex.js';
+import { sourceSha256 } from './cacheFingerprint.js';
 
 export type LoadWorkbookXmlError =
   | { type: 'invalid-xml' }
   | { type: 'validation-failed'; issues: Array<ValidationIssue> }
-  | { type: 'workbook-drift' }
+  | { type: 'workbook-drift'; message?: string }
   // The workbook document POST reported transport-level completion, but Tableau
   // rejected the actual document load (e.g. "Qualified Name Parse Error").
   // `message` carries Desktop's own error text.
@@ -47,6 +48,8 @@ export async function loadWorkbookXml({
   xml,
   baselineXml,
   expectedWorkbookXml,
+  expectedSourceHash,
+  cachedLiveRelative,
   focus,
   applyOptions,
   executor,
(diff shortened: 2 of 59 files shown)


mattcfilbert:
**Andy-lens review: ship it.**

Nothing blocks merge. The one thing that could have been a real bug — a false "changed since you read it" refusal — doesn't happen: the freshness check re-fetches through the same `get*Document` route the read used, so the hash compare can't drift on a serialization mismatch. That was the load-bearing risk and it holds.

The fallback design is right: native and template tools first, then the narrowest cached route, and every cached apply refuses a stale or cross-instance write before touching Desktop. The differential validation gate correctly lets a pre-existing error ride while blocking a new or worsened one. Test coverage is strong — the new/worsened matrix and the cross-session provenance cases pin the seams that matter. Good catch fingerprinting the accepted live readback in `batchCreateAndCacheSheets`, not the submitted body.

Two optional items, neither blocking:

- `restampSidecarAfterEdit` (cacheFingerprint.ts) is a no-op, and that's the correct behavior — leaving the sidecar untouched preserves the original source hash and producer identity. But the name claims a restamp it doesn't do, and it carries three unused params. Consider a rename (`preserveSidecarProvenance`) and dropping the dead resolver arg, or inline a comment at the four call sites and delete it.
- The apply-* descriptions open with "When the cache has source metadata" — the agent can't observe that, so it just spends tokens. The useful half is "if this changed since you read it, the apply is refused; re-read and reapply."

Approving this establishes the narrowest-scope, stale-safe fallback as the rule for document editing. Ship it.

— MattGPT

tableaukyler (commented):
**KylerGPT review — restore bounded document fallbacks**

This adds scoped dashboard/story/whole-workbook cached-XML fallback tools, a source-hash staleness guard on top of the existing cross-instance sidecar, and a differential-validation rule ("block only new/worsened findings, allow pre-existing ones"). The narrowest-scope-first design and the stale-safe-refusal-before-write are the right shape. Traced the three new mechanisms through their actual callers and confirmed three issues, all independently re-verified against the code:

🔴 **apply-workbook's freshness check false-positives on pure navigation, not just conflicting edits** — `loadWorkbookXml.ts`

🟠 **Pre-existing validation "errors" are reported as generic warnings for dashboard/worksheet/storyboard cached applies, but suppressed for workbook** — inconsistent receipt across four tools built on the same design

🟠 **refine-worksheet.ts (untouched by this PR) silently gains a provably-useless extra live GET on every call** — blast radius from the new `validationContext` wiring

Details inline. None of these need a redesign of the mechanism — they're fixable in place.

src/desktop/wrappers/loadWorkbookXml.ts
```diff
@@ -70,55 +75,72 @@ export async function loadWorkbookXml({
...
-      if (currentWorkbook.value.xml !== expectedWorkbookXml) {
+      if (expectedWorkbookXml !== undefined && currentWorkbook.value.xml !== expectedWorkbookXml) {
         return Err({
           type: 'load-workbook-xml-error',
           error: { type: 'workbook-drift' },
         });
       }
+      if (
+        expectedSourceHash !== undefined &&
+        sourceSha256(currentWorkbook.value.xml) !== expectedSourceHash
```
tableaukyler, line 117: 🔴 apply-workbook's freshness check hashes the *entire* live workbook document (`getWorkbookDocument().xml`), and that document's `<windows>` section carries which sheet is `active`/`maximized` (see `activeWindowName` in `applyFocus.ts`, `inspectWorkbookForActivation` in `activateSheet.ts`). A `goto-sheet` between the `get-workbook-xml` read and this apply — an explicit `activate-sheet` call, or just the focus dispatch that automatically follows *any* successful per-sheet apply (`dispatchApplyFocus`, called unconditionally on success in `applyWorkbookText` and `tryApplyViaPerSheetRoute`) — changes those bytes and flips `sourceSha256`, tripping `workbook-drift` with zero real content conflict.

This is exactly the sequence `instructions.ts`'s own `whole-workbook-fallback` route promotes: scoped edits first, whole-workbook last. Any scoped edit or navigation in between and the whole-workbook cache goes stale for no real reason, forcing a needless re-read/re-edit cycle. Worth hashing a windows-stripped view of the document (mirror the `<windows>` scrub already added to `undeclaredCalcReference.ts` in this same PR) rather than the raw full-document bytes.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -127,14 +129,16 @@ export async function loadDashboardXml({
...
 } & WithExecutorAndAbortSignal): Promise<LoadDashboardXmlResult> {
   xml = xml.trim();
   if (!xml || (!xml.startsWith('<?xml') && !xml.startsWith('<'))) {
     return Err({ type: 'load-dashboard-xml-error', error: { type: 'invalid-xml' } });
   }
 
   const validation = runValidation(xml, 'dashboard');
-  const blockingIssues = blockingValidationIssues(validation.issues);
+  const cachedApply = requireExistingSheet;
+  const blockingIssues = cachedApply ? [] : blockingValidationIssues(validation.issues);
```
tableaukyler, line 141: 🟠 `loadWorkbookXml`'s `cachedLiveRelative` path explicitly strips severity `'error'` issues before returning `validationWarnings` (`loadWorkbookXml.ts:140`) — a pre-existing, intentionally-allowed error doesn't get counted. This path (and the worksheet analog at `loadWorksheetXml.ts:510,527`) returns the raw `validation.issues` computed on the candidate alone at line 139, unfiltered, all the way out to `Ok({ validationWarnings: validation.issues })` below. Downstream, `formatDashboardPromiseCheck`/`formatWorksheetPromiseCheck` just count `validationWarnings.length` for the "N warning(s)" line — so the *same* "existing error, nothing new" scenario reports "0 warning(s)" for apply-workbook and "N warning(s)" for apply-dashboard/apply-storyboard/apply-worksheet. Same feature, inconsistent receipt across the three new tools.

src/desktop/wrappers/loadWorksheetXml.ts
```diff
@@ -459,13 +463,31 @@ export async function loadWorksheetXml({
         kind: 'worksheet',
         sheetName: canonicalName,
         fragmentXml: xml,
+        expectedSourceHash,
+        validationContext: cachedApply ? 'worksheet' : undefined,
```
tableaukyler: 🟠 `validationContext` is derived unconditionally from `cachedApply` (`requireExistingSheet`) here, not from whether the caller actually wants the new differential check. `refine-worksheet.ts` — untouched by this PR — already calls this with `requireExistingSheet: true` and already runs its own preflight (`refineWorksheet.ts:270-276`) that refuses on any blocking issue in the exact same XML before it ever reaches here. So this now forces an extra live `getWorksheetDocument` fetch (`perSheetDocumentApply.ts:81`) on every refine-worksheet call, feeding a check (`introducedBlockingValidationIssues`) that can only ever return non-empty if the candidate has a blocking issue — impossible here, since refine's own gate already ruled that out. Real per-call Desktop round-trip added to a file whose own header calls itself "the refine fast lane... ONE fetch... apply ONCE." Consider gating this on `expectedSourceHash !== undefined` (or a caller-controlled flag) instead of piggybacking on `requireExistingSheet`, which pre-existing callers use for unrelated reasons.

mattcfilbert:
Thanks Kyler — all three fixed in 42a6aabf.

**1. Drift on navigation (🔴).** `sourceSha256` now strips the `active`/`maximized` attributes off `<window>` tags before hashing, so switching sheets no longer trips `workbook-drift`. I kept it to just those two attributes rather than scrubbing the whole `<windows>` subtree — that subtree also holds real structure (worksheet `cards`, dashboard `viewpoints`, `simple-id`, and the separate `<active id>` child), and the guard still needs to catch edits to those. The scrub lives inside `sourceSha256`, so every mint and compare site stays in step. Tests cover both directions: a navigation-only flip hashes equal, a structural change (including the `<active id>` child) still hashes different.

**2. Inconsistent warning receipt (🟠).** The cached dashboard/worksheet/storyboard returns now filter `severity === 'error'` before returning `validationWarnings`, matching apply-workbook. Those remaining errors are provably pre-existing — the per-sheet route blocks any newly introduced error before POST — so this only evens out the count, it doesn't hide a new problem.

**3. refine-worksheet's extra GET (🟠).** Added a `callerPreflightsBlockingIssues` flag to `loadWorksheetXml`; refine-worksheet passes it, since its own preflight already refuses on any blocking issue, so the per-sheet introduced-issue check (and its live GET) is redundant there. The apply-* tools don't pass it, so they keep the check regardless of whether they carry a source hash.

Full suite green (5821), types and lint clean.

🤖 Posted by MattGPT on Matt Filbert's behalf.

src/desktop/wrappers/loadWorksheetXml.ts
mattcfilbert: addressed below

src/desktop/wrappers/loadDashboardXml.ts
mattcfilbert: addressed below

src/desktop/wrappers/loadWorkbookXml.ts
mattcfilbert: addressed below

tableaukyler (commented):
**KylerGPT re-review — 1 new commit since last review (`42a6aabf`)**

Verified all three findings from my prior review against the fix commit, independent of the author's own summary:

✅ **Navigation false-positive on apply-workbook freshness (🔴)** — fixed at the shared root, not just the one call site. `sourceSha256` (`cacheFingerprint.ts`) now strips `active`/`maximized` off `<window>` tags before hashing, so every caller of it (workbook freshness, per-sheet drift, all cache sidecars) gets the fix, not just `loadWorkbookXml.ts:117`. Confirmed the strip is narrow — it leaves the `<active>` *child element* alone (new test at `cacheFingerprint.test.ts` proves it), and confirmed it's a no-op for worksheet/dashboard fragments (they carry no `<window>` tags) so no other caller's semantics shifted.

✅ **Inconsistent pre-existing-error suppression across cached applies (🟠)** — `loadDashboardXml.ts:234` and `loadWorksheetXml.ts:514` now filter `severity !== 'error'` before returning `validationWarnings`, matching the workbook path. Confirmed the filter is scoped to the *cached-apply* branch only — the direct/external-API branch in each file (`loadDashboardXml.ts:250`, `loadWorksheetXml.ts:532`) is untouched, which is correct since there's no "pre-existing" baseline to diff against there. Storyboard rides the same fix since it shares `loadDashboardXml`.

✅ **Needless live GET on every refine-worksheet call (🟠)** — new `callerPreflightsBlockingIssues` flag threaded through `loadWorksheetXml` → `tryApplyViaPerSheetRoute`. `refine-worksheet.ts` now passes it, and since it never passes `expectedSourceHash` either, the extra `getWorksheetDocument` fetch is fully skipped for its call path (confirmed against `perSheetDocumentApply.ts:85`'s `expectedSourceHash !== undefined || validationContext !== undefined` gate). New tests (`loadWorksheetXml.test.ts`) pin both directions: the GET is skipped when preflighted, and still runs for a hash-less caller that *hasn't* preflighted.

Ran the full suite against the head commit in an isolated worktree: 5,821/5,821 tests pass, lint clean, typecheck clean.

No new findings on this commit.

tableaukyler (approved):

PR #771, opened 2026-08-14
fix(desktop): don't retry when Desktop is blocked on a modal dialog

lauren-jacksonSFDC:
## Decision

A Tableau Desktop modal that only a person can dismiss (Sign In, connection/credential prompts, a hung call behind a dialog) is now surfaced **consistently as a non-retryable, action-required condition** across the desktop tool surface: the agent is told to dismiss the dialog and re-check with `list-instances`, never to retry — a retry just re-hangs against the same dialog. This is a rule the codebase now keeps, not a one-off.

## Why

Production cluster "Desktop blocked by modal dialogs" (~446 est. occurrences), where dialog-blocked failures surfaced through many tools as generic `other_error` and, worse, as **retryable**:

- `get-summary-data` classified every `statusCode >= 500` as `retryable` (`DesktopCommandExecutionError` is always 500). So an `awaiting-user` failure (`recoverable: false`) and a call timeout were handed back as "retry once" — telling the agent to retry a dialog a retry can never clear.
- The async **poll-timeout** mapping emitted a bare "did not reach a terminal state" string with none of the dismiss/`list-instances` guidance the call-deadline timeout already carried.

## What changed

1. **Unify the hang guidance.** Extract the existing call-deadline dialog guidance into a shared `BLOCKING_DIALOG_GUIDANCE` constant and reuse it on the async `poll-timeout` / `operation-expired` mappings, so every hang surface reads identically (dismiss the dialog, do not retry, `list-instances`).
2. **Don't retry a dialog.** Add `DesktopCommandExecutionError.blockedByDesktopDialog` (true for an `awaiting-user` command failure or a timeout) and route those in `get-summary-data` to `action-required` / `desktop-blocked` instead of `retryable`.
3. **Don't assert a dialog we can't see.** The `desktop-blocked` next-action reads "Dismiss **any open** Tableau dialog, then call `list-instances`" rather than "Dismiss the Tableau dialog" — a `command-timed-out` hang can't confirm a dialog from the client (only `awaiting-user` carries `blockingWindows`), so a genuine no-dialog timeout no longer tells the user to dismiss one that isn't there. (Matt's review.)

No version bump: rebased onto the current `feature/desktop` (already at 2.63.5, set by #768), so the PR's own duplicate 2.63.5 bump was dropped on rebase.

## Scope

Client-side only. The terminal-`FAILED`-with-a-cryptic-code case (e.g. `Unable to complete action` / `8F2A4D91`) **cannot** be recognized as dialog-related from the client: the Desktop server only probes for blocking windows on a *non-terminal* operation, so a terminal failure never carries `blockingWindows` (verified against the monolith `tabexternalapihost` handler — the probe is gated on `!IsTerminal`, and `AWAITING_USER` is never assigned). Closing that subset needs a server-side change, tracked in GUS **W-23855650**.

## Tests

- `DesktopCommandExecutionError.blockedByDesktopDialog` for awaiting-user / timeout / ordinary failure.
- Poll-timeout now carries the dismiss/`list-instances` guidance.
- `get-summary-data` integration test: a blocking dialog on a non-terminal poll → `action-required` / `desktop-blocked`, not `retryable`.

## Validation

Lint, `tsc --noEmit`, and the default + desktop bundle builds all pass; 101/101 unit tests on the touched files.

Live smoke test against Tableau Desktop (External Client API 0.2.6): with a modal ("Edit Title") blocking the command queue, `get-summary-data` returned `action-required` / `reason: desktop-blocked` (**not** `retryable`) carrying the dismiss/`list-instances` guidance and the surfaced dialog title; after dismissal, `list-instances` + a re-read recovered cleanly. Happy path (real rows) and the `worksheet-not-found` error path were also confirmed.

### DIFF
--- a/src/desktop/callDeadline.ts
+++ b/src/desktop/callDeadline.ts
@@ -37,6 +37,11 @@ export function formatBudget(budgetMs: number): string {
   return budgetMs % 1000 === 0 ? `${budgetMs / 1000}s` : `${budgetMs}ms`;
 }
 
+export const BLOCKING_DIALOG_GUIDANCE =
+  'Desktop is most likely showing a blocking dialog that a person has to dismiss, or the instance is wedged. ' +
+  'Do not retry this call — it will hang against the same dialog. ' +
+  'Tell the user to dismiss any open Tableau dialog, then call list-instances to confirm the session is still reachable and re-target if the pid changed.';
+
 /**
  * The agent-facing text for an expired call. It names the budget, says what is most likely
  * wrong, and forbids a blind retry — retrying a call that raised a blocking Desktop dialog
@@ -59,9 +64,7 @@ export function desktopCallTimeoutMessage({
     `Tableau Desktop did not respond within ${formatBudget(budgetMs)} and the call was aborted${
       scope ? ` (${scope})` : ''
     }.`,
-    'Desktop is most likely showing a blocking dialog that a person has to dismiss, or the instance is wedged.',
-    'Do not retry this call — it will hang against the same dialog.',
-    'Tell the user to dismiss any open Tableau dialog, then call list-instances to confirm the session is still reachable and re-target if the pid changed.',
+    BLOCKING_DIALOG_GUIDANCE,
   ].join(' ');
 }
 
--- a/src/desktop/externalApi/externalApiToolExecutor.test.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.test.ts
@@ -826,7 +826,14 @@ describe('ExternalApiToolExecutor', () => {
       });
 
       expect(result.isErr()).toBe(true);
-      expect(result.unwrapErr().type).toBe('command-timed-out');
+      const error = result.unwrapErr();
+      expect(error.type).toBe('command-timed-out');
+      // A poll-timeout is a hang, most likely a modal Desktop can't clear — surface the same
+      // dismiss-the-dialog / do-not-retry / list-instances guidance the call-deadline path uses.
+      if (error.type === 'command-timed-out') {
+        expect(error.error).toContain('Do not retry');
+        expect(error.error).toContain('list-instances');
+      }
     });
 
     it('maps a CANCELLED terminal operation to a command-failed error', async () => {
--- a/src/desktop/externalApi/externalApiToolExecutor.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.ts
@@ -2,7 +2,11 @@ import { Err, Ok, Result } from 'ts-results-es';
 import { z } from 'zod';
 
 import { log } from '../../logging/logger.js';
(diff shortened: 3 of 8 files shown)


mattcfilbert (approved):
**MattGPT** — automated review on Matt's behalf (andy-review lens).

Reviewed at `05e9037`. Clean, well-scoped fix — approving.

The retry-vs-blocked routing is correct end to end. A non-terminal poll that carries `blockingWindows` becomes `awaiting-user` → `command-failed`/`recoverable:false` → `blockedByDesktopDialog` → `action-required`/`desktop-blocked`, so the agent is told to dismiss the dialog and check `list-instances` instead of retrying into the same wedge. Consolidating the three inline strings into `BLOCKING_DIALOG_GUIDANCE` removes the duplication. Tests cover the cases that matter (awaiting-user, timed-out, and an ordinary failure that must *not* flag), and the terminal-FAILED-with-cryptic-code case that can't be recognized client-side is disclosed and tracked (W-23855650), not hidden. Version bump is present.

One question, your call:

- `src/errors/mcpToolError.ts` — every `command-timed-out` now flags `blockedByDesktopDialog`, so a plain slow-but-not-blocked timeout routes to `action-required` ("dismiss the dialog") instead of the old `retryable`. That's the right call for the wedge case, and a hung call rarely benefits from a retry — but it does mean a genuine timeout with no dialog now tells the user to dismiss a dialog that isn't there. Fine to keep if that's the deliberate default.

Fine as-is:

- The guidance names `list-instances`. Naming another tool is normally risky (it may not be enabled), but here it's an always-registered desktop tool, it's Tableau's own, and `emittedToolReferences` validates the token — so a rename can't silently break it.

lauren-jacksonSFDC:
@mattcfilbert thanks — and good catch on the timeout case.

Yes, flagging **every** `command-timed-out` as `blockedByDesktopDialog` is the deliberate default. A hung Desktop call almost never benefits from a retry (it just re-hangs), and the client genuinely can't tell a dialog-blocked hang from a plain slow timeout at that point: `command-timed-out` carries only a string, and only `awaiting-user` (a non-terminal poll with `blockingWindows`) actually confirms a dialog. So we err toward non-retry rather than risk re-wedging Desktop.

You're right, though, that the wording shouldn't *assert* a dialog when there may not be one. The shared `BLOCKING_DIALOG_GUIDANCE` already hedges ("most likely a blocking dialog … **or the instance is wedged**" → `list-instances`), but the next-action label still said "Dismiss the Tableau dialog." I softened it to **"Dismiss any open Tableau dialog, then call `list-instances`"** so a genuine no-dialog timeout no longer tells the user to dismiss one that isn't there — `3d661156`.

Also rebased onto the latest `feature/desktop`; the PR's own 2.63.5 bump was dropped as a duplicate (already set by #768), so it now ships at 2.63.5 like its siblings.

tableaukyler (approved):

PR #773, opened 2026-08-14
feat(desktop): add export-storyboard-image tool

myu404:
## Summary

Adds `export-storyboard-image`, a desktop MCP tool that renders a storyboard as an image. It mirrors the existing `export-worksheet-image` and `export-dashboard-image` tools exactly: same route shape (`GET /v0/workbook/storyboards/{id}/image`), same optional `filePath`/`mimeType` query params, same response schema, and the same inline-vs-file-cache result handling. The storyboard endpoint has no story-point selector — it always renders the storyboard's currently active story point.

## Changes

- `src/desktop/externalApi/types.ts` — new `storyboardImage` route and `storyboardImageRoute()` helper
- `src/desktop/externalApi/externalApiToolExecutor.ts` — new `exportStoryboardImage()` method
- `src/desktop/externalApi/mockExternalApiServer.ts` / `executor.mock.ts` — mock support for tests
- `src/tools/desktop/api/exportStoryboardImage.ts` — new tool, structured like `exportDashboardImage.ts`
- `src/tools/desktop/toolName.ts` / `tools.ts` — tool registration
- `src/tools/desktop/api/exportSheetImage.test.ts` — storyboard coverage added to the shared test suite

`externalApiContract.test.ts` is intentionally left untouched — its captured fixture predates this endpoint.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run lint`
- [x] `npx vitest run src/tools/desktop/api/exportSheetImage.test.ts src/desktop/externalApi/externalApiContract.test.ts` (116/116 passing)

### DIFF
--- a/src/desktop/externalApi/executor.mock.ts
+++ b/src/desktop/externalApi/executor.mock.ts
@@ -44,6 +44,7 @@ export function makeExecutorMock(
     getWorksheetUnderlyingData: vi.fn(),
     exportWorksheetImage: vi.fn(),
     exportDashboardImage: vi.fn(),
+    exportStoryboardImage: vi.fn(),
     validateWorkbookDocument: vi.fn(),
     applyWorkbookDocument: vi.fn(),
     applyWorksheetDocument: vi.fn(),
--- a/src/desktop/externalApi/externalApiToolExecutor.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.ts
@@ -53,6 +53,7 @@ import {
   SiteWorkbookList,
   siteWorkbookListSchema,
   storyboardDocumentRoute,
+  storyboardImageRoute,
   StoryboardItem,
   storyboardItemSchema,
   StoryboardList,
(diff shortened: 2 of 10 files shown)


mattcfilbert (approved):
Nice, @myu404 — this is a clean copy of the dashboard image tool with only the storyboard swaps, and the test coverage matches the siblings one-for-one (happy path, deprecated-alias/conflict rejection, honest too-new-endpoint 404, and raw relative-path forwarding). Ran the export-image suite locally: 27 passing. Approving.

One thing worth confirming, not a blocker:

- `exportStoryboardImage.ts` — the description says it renders "the active story point." So a storyboard with several story points exports only the one that's currently active, and there's no param to pick a different one. If that's the intended V0 scope, great — just want it to be a decision, not an accident. A caller who expects "the whole story" will get one point and no signal about the others.

Kept the tool out of the lean authoring profile, same as `export-worksheet-image` and `export-dashboard-image` — good, that's the call we made in #770.

— MattGPT (posted on Matt's behalf)

lauren-jacksonSFDC:
Same sitch as 775 - adding new routes needs to be careful about versioning ATM: https://github.com/tableau/tableau-mcp/pull/775#issuecomment-5296749309

myu404:
Thanks @mattcfilbert — it's a deliberate V0 decision, not an accident. Active-point-only matches how `export-worksheet-image` / `export-dashboard-image` behave over the External Client API today; picking a specific story point would need a new API affordance, so it's a follow-up rather than something this PR should fake.

To your "no signal about the others" point, I've made the scope explicit in the description so the model can't assume it reached the whole story:

> Render a storyboard's active story point as an image. Only the active point is rendered; other story points are not included and cannot be selected.

PR #774, opened 2026-08-14
@W-23854238 - Add download workbook tool (clean branch)

takashibinns:
### Summary

- add [Download Workbook](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooks_and_views.htm#download_workbook) SDK support for downloading TWB and TWBX files
- gate the tool behind the authoring-tools feature flag

### Testing

- npm test
- npm run lint
- npm run build


features.json
```diff
@@ -1,5 +1,7 @@
 {
   "mcp-apps": false,
   "view-file-mode": false,
-  "view-data-file-mode": false
+  "view-data-file-mode": false,
+  "workbook-file-mode": true,
```
jarhun88, line 5: what is this for?

features.json
```diff
@@ -1,5 +1,7 @@
...
+  "authoring-tools": true
```
jarhun88: default should be false

features.json
takashibinns: if workbook-file-mode is set to true, the tool will just save the workbook to a local tmp directory (tool returns the filepath).  
if set to false AND the S3 bucket env var is set, it will try uploading the workbook to the S3 bucket and return a presigned url.

i think this should be true by default, so it will just work when running tab-mcp locally

jarhun88: ok gotcha, maybe a better naming would be something more generic to blob storage, i think well have several tools that could gate on blob storage download + upload

jarhun88 (approved):

features.json
takashibinns: i was trying to mirror the logic of the viewdata and viewimage toolresults, which use the ```view-file-mode``` and ```view-file-data-mode``` flags.  since they each had their own flags, i figured i should follow the pattern. but maybe they should all be using the same flag? we could call it ```stage-files-in-s3``` or something like that?

features.json
takashibinns: done

mattcfilbert (approved):
Nice work, @takashibinns — this is a clean new tool. The S3-or-temp-path result builder with the graceful fallback is the right shape, the feature-flag gating lines up (the tool registers only when `authoring-tools` is on, and the e2e test now keys on that same flag), and the test coverage is real: both output modes, the S3-failure fallback, the disabled path, access denial, and API errors. Docs page, unit + e2e tests, and the version bump are all here. I traced the two things that could have bitten — path traversal from the server-supplied filename, and any listed-but-broken flag combination — and both hold. `sanitizeFileName` kills path separators and the `${randomUUID()}-` prefix means the basename can never be `..`, so nothing escapes the temp dir.

A few things to consider, none blocking:

**Please consider — trim the desktop-open guidance out of the descriptions.** The tool description and the new server-level `description` in `src/server.ts` both tell the agent how to "derive the full Tableau Desktop application path" and compare version numbers. Two concerns: the MCP client is an LLM, so every sentence in a tool description spends its context budget on every tool list — and "how to find the Desktop app on disk" is guidance about what happens *after* the download, not what the tool does or how to call it. The server-level copy is the sharper one: `Server` is the base class for the web server too, so this ships in the `initialize` response to every client, including Cloud/Server clients that will never open a local Desktop app. Suggest dropping the desktop-path sentences from the tool description, and either removing the server-level `description` or moving it to the desktop server only.

**Question — do we need both `workbook-file-mode` and `authoring-tools`?** Registration is gated on `authoring-tools` (default false), but S3 mode is gated on `workbook-file-mode` (default true). So `workbook-file-mode: true` does nothing until `authoring-tools` is also flipped — the flag whose name implies it's the gate isn't the gate. Worth collapsing to one, or gating registration on `workbook-file-mode` too, so the names mean what they say. What do you think?

nits (optional):
- `src/sdks/tableau/methods/workbooksMethods.ts:85` — `downloadWorkbook`'s JSDoc reads `@param workbookId - The ID of the workbook to delete.` (copy-paste from `deleteWorkbook`) — should say download.
- `src/server/oauth/scopes.ts:52-53` and `src/restApiInstance.ts:40-41` — `tableau:workbooks:download` is listed just before `tableau:workbooks:delete`; alphabetized it'd be delete-then-download.

Fine as-is, tradeoffs noted:
- The extension inference `mimeType === 'application/xml' ? 'twb' : 'twbx'` breaks if Tableau ever returns `application/xml; charset=utf-8`. Low stakes today because the user-facing filename comes from `Content-Disposition`; this only affects the S3 object-key suffix and the no-filename fallback. If you touch it, compare on the parsed media type and share one helper instead of the three copies.
- Temp files under `tmpdir()/tableau-mcp-workbooks` aren't cleaned up. Fine for the local single-user Desktop target this tool is for; worth a line in the docs that non-local deployments should configure S3, since these files may carry extract data and land at default umask.

---
*Posted by MattGPT on Matt Filbert's behalf.*

features.json
takashibinns: update: we decided to address this in a future PR

PR #775, opened 2026-08-14
feat(desktop): add workbook-export-as tool over the workbook:exportAs route

myu404:
## What

Adds `workbook-export-as`, a Desktop MCP tool that exports the currently open workbook to a file and leaves the open document unchanged. It drives the External Client API `workbook:exportAs` route through the standard async operation pattern (POST then poll).

Formats and their required extensions:
- `pdf` → `.pdf`
- `powerpoint` → `.pptx`
- `packaged-workbook` → `.twbx`
- `prior-version` → `.twb`/`.twbx` — a down-saved copy for an older release; also requires `targetVersion` (e.g. `2024.1`)

The write is headless: it goes straight to `filePath` with no Save dialog and returns no bytes. The one pre-flight guard rejects `prior-version` with no `targetVersion` (with a precise message); every other check — path shape, extension↔format match, unknown version — is left to the host.

The `workbook:exportAs` route ships in External Client API 0.2.7, so the tool declares a minimum API version of `0.2.7`: it is hidden from a connected Desktop that predates the route and appears once the route is available.

## Changes
- New tool `src/tools/desktop/api/workbookExportAs.ts` (+ unit tests), registered in the dynamic-authoring profile.
- API layer: `exportWorkbookAs` executor method, `ExportAsWorkbookRequest`/`ExportAsFormat` types, and the `workbook:exportAs` route plus its mock-server branch.
- Declares `minApiVersion: '0.2.7'`, with tests pinning the floor and asserting the tool is hidden below 0.2.7.
- Re-pinned the desktop tool-surface byte budgets to account for the new tool.

## Testing
- `npm run lint`, `tsc --noEmit`, `npm run build:desktop` — all clean.
- Full unit suite passes, including the new tool's tests and the version-floor tests.


lauren-jacksonSFDC:
Standup just reminded me:
1. We have to be "backwards compatible" at the moment due to the external builds, since this needs a new monolith build we need to make sure we're only making the new endpoints available with the correct version of the API
2. Lately I've been checking in the new openapi spec with API bumps
3. We've got to bump the version of the MCP in here as well

mattcfilbert:
Clean, well-scoped tool. It's one thin door over `workbook:exportAs`, the executor, type, and mock-server layers follow the existing pattern, and the tests cover the branches that matter — happy path, prior-version forwarding, the pre-flight guard, and a server-side rejection surfaced back. Nice call leaving all validation except the prior-version/targetVersion pair to the host and saying so in the comment, and thanks for re-pinning the byte budgets so the tool-search deferral stays honest.

**One thing to change before merge:**

The missing-`targetVersion` guard returns success, not an error. A prior-version export with no `targetVersion` returns `new Ok({ message: '...requires targetVersion...' })`, and the test asserts `result.isError` is falsy. Everywhere else we reject bad caller input with an error — `sortWorksheet` returns `new ArgsValidationError(...).toErr()` when a field isn't on the shelves. For the model, `isError: true` is what routes the call into its retry path; a success envelope with a corrective string can read as "done, nothing to do." I'd return an `ArgsValidationError` (or the repo equivalent) here so the signal matches the rest of the surface. If the success envelope is deliberate, a one-line why in the PR would settle it.

**Version bump:** I don't see `package.json` in the diff. A new tool usually gets `npm run version:patch` — flagging in case it was missed, or tell me if the desktop variant's release convention differs.

**One question, your call:** `targetVersion` is forwarded on any format (`...(targetVersion !== undefined ? { targetVersion } : {})`), so a stray `targetVersion` on a pdf export goes to the server to ignore. The comment acknowledges this. Fine to leave — just checking you'd rather trust the server than catch that caller mistake client-side.

Two things that look off but are correct, so no action:
- Skipping save-workbook's post-write inventory re-read is right here. save-workbook re-reads because a dismissed Save As dialog reports SUCCEEDED; exportAs requires `filePath` (`z.string().min(1)`, not optional), so there's no dialog to dismiss and a `completed` status is authoritative. The comment explains it.
- Routing a file-writing tool through `runExternalApiReadTool` reads odd, but that wrapper is the shared session-resolve/route-missing/error-map path that open-file, save-workbook, add-worksheet, and undo/redo all use for writes. The "read" name is legacy, not this PR's problem.

nit (optional): only pdf and prior-version have happy-path tests. powerpoint and packaged-workbook go through the identical body-forwarding path so it's not a behavior gap, but one parametrized case over all four formats would document the extension matrix the description promises.

Out of scope for me: the PR notes the contract test deliberately keeps `workbook:exportAs` out of the captured-spec route list until a fixture recapture. Worth confirming that recapture is tracked somewhere so the route doesn't stay untested against the real spec.

---
*Posted by MattGPT on behalf of Matt Filbert.*

src/tools/desktop/api/workbookExportAs.ts
```diff
@@ -0,0 +1,98 @@
...
+    name: 'workbook-export-as',
+    title,
+    description:
+      'Export the open workbook to a file, leaving the open document unchanged. ' +
+      'format→extension: pdf→.pdf, powerpoint→.pptx, packaged-workbook→.twbx, ' +
+      'prior-version→.twb/.twbx (a down-saved copy for an older release; also pass ' +
+      'targetVersion). Writes headlessly to filePath; no dialog, no returned bytes.',
+    annotations: {
+      readOnlyHint: false,
+      destructiveHint: false, // writes a new artifact; the open document is left unchanged
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

`destructiveHint: false` looks wrong here. This tool always writes to a **required** caller-supplied absolute `filePath` (`z.string().min(1)`, no inline-only mode), and per its own line-40 comment "re-running overwrites the same file" — so it can silently clobber a pre-existing, unrelated file at that path.

**Trigger:** An agent calls `workbook-export-as` with `filePath` pointing at an existing, unrelated file → the export writes headlessly with no dialog → that file is overwritten. Because `destructiveHint: false`, an MCP host that gates destructive ops behind a confirmation prompt skips the prompt.

**Why this matters:** The two closest structural siblings — `export-dashboard-image` and `export-worksheet-image` — set `destructiveHint: true` for exactly this overwrite case ("a caller-supplied filePath can overwrite an existing file at that path"), and their `filePath` is only *optional*, so this tool is at least as destructive. The line-39 comment justifies `false` via the *in-memory* document, but `destructiveHint` measures the *filesystem* effect. (`save-workbook`, the model for this tool, does use `false`, but its `filePath` is optional with save-in-place semantics — a mixed precedent worth reconciling across these four tools.)

myu404: Good catch — fixed in 0563df1a. Flipped `destructiveHint` to `true` and rewrote the rationale comment to the axis it actually measures:

```ts
destructiveHint: true, // the required filePath can overwrite an existing file at that path
```

The old comment reasoned about the in-memory document (which is left unchanged) — the wrong axis, as you note. `destructiveHint` is about the filesystem write. And since `filePath` is required here (`z.string().min(1)`), this tool is at least as destructive as `export-dashboard-image`/`export-worksheet-image` — whose `filePath` is only optional and which already set `true` — so it now matches them.

I left `save-workbook` on `false`: its `filePath` is optional with save-in-place semantics, so it's a genuinely different case. Reconciling the `destructiveHint` convention across all four tools is its own change — happy to file a follow-up if you'd like it tracked.

lauren-jacksonSFDC:

lauren-jacksonSFDC (approved):

PR #778, opened 2026-08-14
@W-23865341 - consolidate the FFs for staging files in S3

takashibinns:
### Summary

there are 3 feature flags that essentially determine the same behavior:

- ```view-file-mode```
  - when true, upload view image files to S3 for staging and return presigned url
  - when false, return a base64 encoded data uri
- ```view-data-file-mode```
  - when true, upload view data csv files to S3 for staging and return presigned url
  - when false, return raw csv text
- ```workbook-file-mode```
  - when true, store the workbook in a local tmp drive and return the file path
  - when false, upload the workbook file to S3 and return a presigned URL

Instead of keeping identical logic, we should be keying off the MCP_S3_BUCKET env var. If it's set, then we should attempt to stage these files in S3. If not set, we should return the data inline (csv, dataUri, file path).

### Testing

- npm test
- npm run lint
- npm run build

### DIFF
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -698,7 +698,7 @@ instead of inlining it:
 - `get-view-image` / `get-custom-view-image` — rendered image (otherwise inline base64)
 - `get-view-data` / `get-custom-view-data` — CSV data (otherwise inline text)
 - `download-workbook` — workbook file (otherwise a local temp-file path)
-- `request-workbook-upload` / `validate-upload-and-publish-workbook`- stage workbook in S3 before publishing,(otherwise a local file path)
+- `request-workbook-upload` / `validate-upload-and-publish-workbook`— stage workbook in S3 before publishing (otherwise a local file path)
 
 The client fetches the files directly from S3, so the payload never streams back through the MCP
 server on read.
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -1,6 +1,6 @@
 ---
-sidebar_position: 1
----
+
+## sidebar_position: 1
 
 # Environment Variables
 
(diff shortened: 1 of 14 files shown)


docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -626,19 +626,22 @@ This allows embedding Tableau visualizations from custom Tableau Server domains
...
-`get-custom-view-image` tools upload the rendered image to this bucket and return a short-lived
-presigned URL (as a `resource_link` content block) instead of inlining the image as base64. The
-client fetches the image bytes directly from S3, so the image never streams back through the MCP
+Enables offloading large tool payloads to Amazon S3. When set, the following tools upload the
+payload to this bucket and return a short-lived presigned URL (as a `resource_link` content block)
+instead of inlining it:
+
+- `get-view-image` / `get-custom-view-image` — rendered image (otherwise inline base64)
+- `get-view-data` / `get-custom-view-data` — CSV data (otherwise inline text)
+- `download-workbook` — workbook file (otherwise a local temp-file path)
```
jarhun88, line 700: nit: can you add a line for the publish workbook tool i wrote as well? it also uses a local temp file path

jarhun88 (approved):

jarhun88:
Also a minor cleanup nit in the same file: the `mockIsFeatureEnabled` hoist and `vi.mock('../../../features/init.js', ...)` scaffolding are now dead code since nothing under test calls the feature gate anymore.

docs/docs/configuration/mcp-config/env-vars.md
takashibinns: done

src/tools/web/workbooks/downloadWorkbook.test.ts
jarhun88: Also a minor cleanup nit in the same file: the mockIsFeatureEnabled hoist and vi.mock('../../../features/init.js', ...) scaffolding are now dead code since nothing under test calls the feature gate anymore.

takashibinns: done

PR #779, opened 2026-08-14
feat(style): apply one verified workbook style pack

mattcfilbert:
## Decision

Add one guarded `apply-workbook-style` tool for document-driven workbook styling. This sets the rule for this path: Studio passes a validated style pack, Tableau MCP changes only known presentation fields, and one guarded workbook update must preserve the workbook's analysis and protected style state.

Future style rules should extend this transformer. They should not teach the agent to hand-edit workbook XML.

## Scope

This first slice updates existing worksheet title fonts and colors, body fonts, text and background colors, and existing categorical, sequential, and diverging palettes.

It reports unsupported formats, dashboard spacing and alignment, advisory rules, and global datasource styles instead of inventing XML. It does not add image export or visual grading.

The tool is in the Studio dynamic-authoring profile. The profile is 42,648 bytes across 51 tools, below the 46,000-byte ceiling.

## Safety

- Selects worksheets and dashboards by stable inventory IDs, including hidden sheets used by dashboards.
- Treats repeated dashboard zones across layout variants as one membership set, while still rejecting a genuinely missing or extra sheet.
- Rejects stale or cross-instance writes before dispatch.
- Uses one workbook update, then checks analysis, all supported presentation state, and idempotence on readback.
- Returns `unknown` and says not to retry after an uncertain dispatch.

## Evidence

- `scripts/agent-check`: 380 test files and 5,949 tests passed; lint, typecheck, Desktop build, and lockstep also passed.
- Adversarial review found no remaining P1/P2 issue.
- Live Book2 probe verified the no-op path with zero writes.
- Live Superstore testing found and fixed a real multi-layout dashboard membership bug. The next apply dispatched once, then the External Client API reported Tableau's transient “Processing Request” window as human-blocking. The tool returned `unknown`, `retrySafe: false`, and did not retry. A read-only check confirmed the operation had partially or differently settled, so that response was correct.

This is stacked on #770 and will shrink after #770 merges. The attachment path is split across [agent-chat-ui !31](https://gitlab.tableausoftware.com/browser-clients/agent-chat-ui/-/merge_requests/31) and [tab-agent-south !147](https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/147).

Approving this PR accepts the typed, guarded styling contract and its first supported presentation fields. A clean live success receipt still depends on the External Client API reserving `blockingWindows` for dialogs that need human action.

— MattGPT, on behalf of Matt Filbert


mattcfilbert:
Superseded by #790 — closing. (Posted by MattGPT on Matt's behalf.)

PR #780, opened 2026-08-17
feat(desktop): make the knowledge root configurable and multi-root

tableaukyler:
## Summary
- Adds `TABLEAU_KNOWLEDGE_DIR`, read directly in `assets.ts` (mirrors the
  `TEMPLATES_DIR` module-level pattern, not `Config`) since the multi-root
  logic lives right alongside `listKnowledgeSlugs`/`readKnowledgeBySlug`.
- `listKnowledgeSlugs()`/`readKnowledgeBySlug()` resolve against the external
  root, in precedence order, even under SEA:
  `.vendored/protected` > top-level user content > `.vendored/overridable`.
  Slugs are normalized (tier prefix stripped) so all three map to the same
  logical `expertise://tableau/<slug>`.
- Falls back to the existing bundled `resources/desktop/knowledge` tree
  (SEA-embedded or disk) when the env var is unset — no change in that mode.
- `getKnowledgeDir()` reports the configured root for the empty-corpus guard
  message.

```mermaid
flowchart LR
  ENV[TABLEAU_KNOWLEDGE_DIR] --> P[.vendored/protected]
  ENV --> U[top-level user content]
  ENV --> O[.vendored/overridable]
  P -->|wins| SLUG[normalized slug]
  U -->|shadows overridable| SLUG
  O -->|fallback| SLUG
```

Left unchanged, verified unaffected: `vendoredAssets.test.ts` and
`corpusIntegrity.test.ts` ("serves each topic slug from exactly one
directory") — both run with the env var unset, so they still only exercise
the single bundled tree; the layered precedence only activates when
`TABLEAU_KNOWLEDGE_DIR` is set.

Deferred (long-term, per discussion): retiring `resources/desktop/knowledge`
from `DESKTOP_ASSET_DIRS` once tab-agent-south always supplies the dir.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `yarn lint`
- [x] `yarn test` — 376 files / 5826 tests passed
- [x] New `knowledgeExternalRoot.test.ts` covering precedence, slug
      normalization, `.vendored`-leak guard

### DIFF
--- a/src/desktop/assets.ts
+++ b/src/desktop/assets.ts
@@ -206,17 +206,42 @@ export function readResourceAsset(relPath: string): string | null {
   }
 }
 
-// Knowledge module slugs (forward-slash, no .md) under resources/desktop/knowledge.
-// SEA reads the manifest; disk walks the tree.
-export function listKnowledgeSlugs(): string[] {
-  if (runningAsSea()) {
-    const prefix = 'resources/desktop/knowledge/';
-    return listSeaAssetKeys('resources/desktop/knowledge')
-      .filter((key) => key.endsWith('.md'))
-      .map((key) => key.slice(prefix.length).replace(/\.md$/, ''))
-      .sort();
-  }
-  const root = join(getResourcesRoot(), 'knowledge');
+// TABLEAU_KNOWLEDGE_DIR (set by tab-agent-south once it materializes the corpus
+// to <agent-data-dir>/knowledge) points at a tree with the same tiering skills/
+// commands already use: `.vendored/protected` (rebuilt every startup from the
+// vendored source, authoritative), `.vendored/overridable` (operator-editable
+// vendored default), and everything else at the top level (user-authored,
+// excluded from the vendored tiers). When unset, knowledge falls back to the
+// tree this repo ships itself (SEA-embedded or resources/desktop/knowledge on
+// disk) — the pre-external-corpus behavior, kept as a transitional default.
+const KNOWLEDGE_DIR_ENV_NAME = 'TABLEAU_KNOWLEDGE_DIR';
+const VENDORED_DIR_NAME = '.vendored';
+const PROTECTED_DIR_NAME = 'protected';
+const OVERRIDABLE_DIR_NAME = 'overridable';
+
+function externalKnowledgeRoot(): string | undefined {
+  const raw = process.env[KNOWLEDGE_DIR_ENV_NAME]?.trim();
+  return raw ? raw : undefined;
+}
+
+// Precedence order when a slug is claimed by more than one tier: protected is
+// authoritative and always wins; a user file shadows the vendored overridable
+// default; overridable is the fallback when neither above provides the slug.
+function externalKnowledgeRootsByPrecedence(baseDir: string): string[] {
+  return [
+    join(baseDir, VENDORED_DIR_NAME, PROTECTED_DIR_NAME),
+    baseDir,
+    join(baseDir, VENDORED_DIR_NAME, OVERRIDABLE_DIR_NAME),
+  ];
+}
+
+/**
+ * The `.md` slugs (forward-slash, no extension) under `root`, walked
+ * recursively. `skipTopLevelDirNames` excludes named directories one level
+ * below `root` only — used to keep the user-content root from also walking
+ * back into `.vendored`.
+ */
+function walkMarkdownSlugs(root: string, skipTopLevelDirNames: readonly string[] = []): string[] {
   const slugs: string[] = [];
   const walk = (dir: string, prefixParts: string[]): void => {
     let entries;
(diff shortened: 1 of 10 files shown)


src/desktop/assets.ts
```diff
@@ -206,17 +206,42 @@ export function readResourceAsset(relPath: string): string | null {
...
-export function listKnowledgeSlugs(): string[] {
-  if (runningAsSea()) {
-    const prefix = 'resources/desktop/knowledge/';
-    return listSeaAssetKeys('resources/desktop/knowledge')
-      .filter((key) => key.endsWith('.md'))
-      .map((key) => key.slice(prefix.length).replace(/\.md$/, ''))
-      .sort();
-  }
-  const root = join(getResourcesRoot(), 'knowledge');
+// TABLEAU_KNOWLEDGE_DIR (set by tab-agent-south once it materializes the corpus
```
WillMartin: why an env var for this? Usually been trying to be explicit with arguments for everything except secrets (which will go in the env)

src/desktop/assets.ts
```diff
@@ -206,17 +206,42 @@ export function readResourceAsset(relPath: string): string | null {
...
+// to <agent-data-dir>/knowledge) points at a tree with the same tiering skills/
+// commands already use: `.vendored/protected` (rebuilt every startup from the
+// vendored source, authoritative), `.vendored/overridable` (operator-editable
+// vendored default), and everything else at the top level (user-authored,
+// excluded from the vendored tiers). When unset, knowledge falls back to the
+// tree this repo ships itself (SEA-embedded or resources/desktop/knowledge on
+// disk) — the pre-external-corpus behavior, kept as a transitional default.
```
WillMartin: we should probably have a gus item in a later epic tracking removing the internally bundled versions

src/desktop/knowledge/index.ts
```diff
@@ -11,7 +10,7 @@ export interface KnowledgeResource {
 }
 
 export function getKnowledgeDir(): string {
```
WillMartin: This feels slightly odd. "index" normally would be the "export" surface but we're exporting something defined in a module below this

src/desktop/assets.ts
tableaukyler: fair question, but everything is currently using ENV variables. I'd be introducing a new concept in this PR to pass args... which I agree we should totally do. Will file a follow-up item.

mattcfilbert:
Re-reviewed with the port-fidelity question front of mind: does any knowledge module served today stop being served after this change?

Short answer, with the env var unset: no. I checked the shipped tree at head — 105 markdown files, and the only dot-prefixed entry is `tactics/governance/.gitkeep`, which was never served. The new hidden-name skip drops nothing today, and the SEA path still lists all 105. The default mode is safe.

The risk is the new external-root mode. Two things worth fixing before merge:

**1. A partial external corpus fails silently.** When `TABLEAU_KNOWLEDGE_DIR` is set, the bundled tree is ignored — the external roots are the whole corpus, with no union against the bundle. The only guard is the empty-corpus error, which fires only at zero slugs. If the external roots ever ship a subset (say 3 of the 105 modules), the server comes up healthy and serves 3; the agent quietly loses the other 102, with no error and no log. The completeness of the port then rests entirely on the vendoring step in tab-agent-south, which this repo can't see. Worth a floor check here — a required-slug manifest, or a minimum count diffed against the bundled tree — so an incomplete vendor fails loud instead of degrading quietly.

**2. Version bump.** No `version:patch` in this PR, and head (2.63.9) is behind `feature/desktop` (2.63.11). Needs a rebase and a bump to at least 2.63.12 — CI has no increment guard, so a stale version merges clean.

A few smaller notes, none blocking:

- List and read can disagree on odd filenames: `walkMarkdownSlugs` lists a slug like `guides/v1..v2`, but `isSafeKnowledgeSlug` rejects it on read, so it 404s. The `..`/backslash/leading-slash check now lives in three spots — worth one shared gate both paths call.
- The hidden-name filter is on the disk walk but not the SEA branch. Harmless now (only `.gitkeep`), but they would diverge if a hidden `.md` is ever bundled. One line mirrors it.
- Symlink containment (a `leak.md` pointing outside the root would be listed and read) is not a boundary on Desktop, where the user owns the filesystem. It only matters if these roots ever run server-side or multi-tenant — a cheap realpath-under-root check if that day comes.

Nice change overall — the multi-root precedence and the `.vendored` leak guard are well tested. The one thing I would protect is the corpus floor: silent partial loss is the failure that wouldn't surface until an agent starts giving worse answers.

—posted by MattGPT on Matt Filbert's behalf

src/desktop/assets.ts
tableaukyler: added GUS item

src/desktop/assets.ts
tableaukyler: added GUS item

mattcfilbert (approved):
lgtm

src/desktop/knowledge/index.ts
tableaukyler: fixed. good call

PR #782, opened 2026-08-18
fix(desktop): accept null index and storyPointCount from External Client API

lauren-jacksonSFDC:
**Decision:** our External Client API read schemas now accept an explicit `null` for the sheet-item fields the API declares nullable — `index` (worksheet/dashboard/storyboard) and `storyPointCount` (storyboard). This is a rule we keep going forward: where the API types a field as `["integer","null"]`, our Zod uses `.nullish()`, not `.optional()`.

**Why:** the API contract for `index` changed. It used to be a plain `integer` that hidden sheets *omitted*; it is now required-and-nullable, and hidden sheets send `index: null`. `.optional()` accepts a missing key but rejects an explicit `null`, so `list-worksheets` (and the dashboard/storyboard reads) threw a ZodError on any workbook containing a hidden sheet. `storyPointCount` is the same class — the spec marks it nullable and we were rejecting null.

**Tests:** added contract-test cases asserting the three item schemas accept `index: null`, and that `StoryboardItem` accepts `storyPointCount: null`.

**Follow-up (not in this PR):** the captured contract fixture is still v0.2.6 — it documents `index` as an omitted plain integer, not nullable — and the contract harness checks property presence and requiredness but *not* nullability. That gap is why this drift shipped silently. Refreshing the fixture and adding nullability parity to the harness would catch the next one automatically.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.63.6",
+  "version": "2.63.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.63.6",
+      "version": "2.63.7",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.63.6",
+  "version": "2.63.7",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/externalApi/externalApiContract.test.ts
+++ b/src/desktop/externalApi/externalApiContract.test.ts
@@ -237,6 +237,27 @@ describe('external client API contract (captured openapi fixture)', () => {
         true,
       );
     });
+
+    it.each([
+      ['WorksheetItem', worksheetItemSchema],
+      ['DashboardItem', dashboardItemSchema],
+      ['StoryboardItem', storyboardItemSchema],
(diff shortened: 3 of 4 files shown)


mattcfilbert (approved):
**MattGPT** — review on Matt's behalf.

Approving. Small, correct, well-scoped fix. `.optional()` rejects an explicit `null`, so `list-worksheets` threw a ZodError the moment a workbook had a hidden sheet (the API now sends `index: null` for those). Switching to `.nullish()` accepts both a missing key and an explicit `null`, which is what the contract now requires. I verified the semantics directly: `.nullish()` accepts missing / `null` / a valid int and still rejects a string or a float; the old `.optional()` returned false on explicit `null`. `.nullish()` is already the house idiom in this file. Version bump is right. Thanks Lauren.

No downstream risk from the wider type. `WorkbookReadGateway` reads only `.name` off these items; nothing in the repo sorts, compares, or does arithmetic on `.index` or `.storyPointCount`, so `null` behaves the same as absent everywhere it flows.

**One follow-up worth pointing at the harness, not the fixture.** The deferred work should add type/nullability assertions to the contract harness, not just refresh the captured fixture. The parity test (`externalApiContract.test.ts:179-189`) compares declared-key and required-key sets but never inspects `type` — so a refreshed fixture documenting `index` as `["integer","null"]` would pass identically to today's `"integer"` and still enforce nothing, leaving the same silent-drift gap open for the next nullability change. The `.success` cases this PR adds are the correct stopgap; the durable fix is asserting `type` in the harness.

**Heads-up, not a blocker on this PR:** the `build (>=22.7.5 <23)` job is red, but it's unrelated to this diff. All 23 failures are one Playwright strict-mode violation at `tests/oauth/tableau-authz/flows/consentFlow.ts:20` — `button[type="submit"]` now matches three buttons (Switch site / Switch username / Allow) after an upstream Tableau OAuth consent-page HTML change. This PR touches no OAuth code, and `build (24.x)` passed the same steps. Worth a separate fix to the selector (`#allow-button`), independent of this change.

PR #783, opened 2026-08-18
feat(desktop): gate tools by minimum External Client API version

lauren-jacksonSFDC:
Adds a per-tool `minApiVersion` floor for desktop MCP tools. At registration the server reads the connected Desktop's External Client API version from its discovery file and drops any tool whose floor exceeds it.

**Fail-open:** when the connected version is unknown (no discoverable instance yet), every tool registers and the existing per-call `endpointNotInThisBuild` guard stays the backstop. This is a rule the codebase now keeps — any tool can set `minApiVersion` and be gated — not a one-off.

**Interim:** the planned `GET /v0/capabilities` endpoint is the eventual source of truth for whether an endpoint is served; this discovery-file version read is the stopgap until it lands.

### Floors set
| Tools | Floor |
| --- | --- |
| `pause-auto-updates`, `resume-auto-updates` | 0.2.5 |
| `open-file`, `save-workbook`, `add-worksheet`, `add-dashboard`, `add-storyboard` | 0.2.6 |
| `export-storyboard-image` | 0.2.7 |

### Verified
- Unit tests: version compare, `resolveConnectedApiVersion`, `filterToolsByApiVersion`, plus a floor-pinning test. Full desktop suite green (3567 tests).
- Live against Tableau Desktop @ apiVersion 0.2.6: the seven ≤0.2.6 tools register and pause→resume round-trips; under `TOOL_PROFILE=full` the gate drops exactly `export-storyboard-image` (needs 0.2.7) and re-exposes it at 0.2.7.

### Note
`supportsOperationResult` now reuses the new `apiVersionAtLeast` helper (behavior-identical, ≥0.1.1).

**Reviewer checkpoint:** the floor *values* — the mechanism is version-agnostic; the specific numbers are the thing to sanity-check.

### DIFF
--- a/src/desktop/externalApi/apiVersion.ts
+++ b/src/desktop/externalApi/apiVersion.ts
@@ -1,0 +1,19 @@
+/**
+ * Numeric SemVer comparison ("0.2.6" >= "0.2.5"). Missing/unparseable parts read as 0,
+ * so `undefined` and `""` compare as 0.0.0 — below every real floor.
+ */
+export function apiVersionAtLeast(current: string | undefined, minimum: string): boolean {
+  const [cMajor, cMinor, cPatch] = parseApiVersion(current);
+  const [mMajor, mMinor, mPatch] = parseApiVersion(minimum);
+  if (cMajor !== mMajor) return cMajor > mMajor;
+  if (cMinor !== mMinor) return cMinor > mMinor;
+  return cPatch >= mPatch;
+}
+
+function parseApiVersion(version: string | undefined): [number, number, number] {
+  const [major = 0, minor = 0, patch = 0] = (version ?? '')
+    .split('.')
+    .map((part) => Number.parseInt(part, 10))
+    .map((part) => (Number.isFinite(part) ? part : 0));
+  return [major, minor, patch];
+}
--- a/src/desktop/externalApi/apiVersion.ts
+++ b/src/desktop/externalApi/apiVersion.ts
@@ -1,3 +1,5 @@
+export type ApiVersionFloor = `${number}.${number}.${number}`;
+
 /**
  * Numeric SemVer comparison ("0.2.6" >= "0.2.5"). Missing/unparseable parts read as 0,
  * so `undefined` and `""` compare as 0.0.0 — below every real floor.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.63.7",
+  "version": "2.63.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.63.7",
+      "version": "2.63.8",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.63.7",
+  "version": "2.63.8",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/externalApi/apiVersion.test.ts
+++ b/src/desktop/externalApi/apiVersion.test.ts
@@ -1,0 +1,33 @@
+import { apiVersionAtLeast } from './apiVersion.js';
+
+describe('apiVersionAtLeast', () => {
+  it.each([
+    ['0.2.6', '0.2.6', true],
+    ['0.2.6', '0.2.5', true],
+    ['0.2.5', '0.2.6', false],
+    ['0.3.0', '0.2.9', true],
+    ['1.0.0', '0.9.9', true],
+    ['0.9.9', '1.0.0', false],
+    ['0.2.10', '0.2.9', true],
+    ['0.1.1', '0.1.1', true],
+    ['0.1.0', '0.1.1', false],
+  ])('%s >= %s → %s', (current, minimum, expected) => {
+    expect(apiVersionAtLeast(current, minimum)).toBe(expected);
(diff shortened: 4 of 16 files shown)


mattcfilbert (commented):
**MattGPT** — review on Matt's behalf (andy-review lens).

Clean, well-shaped change with genuinely thorough tests. The `minApiVersion` field is a version-agnostic rule rather than a one-off, the fail-open design is deliberate and documented, and the tests cover every branch — pinned/unpinned/no-match/mixed-fleet/fail-open plus a floor-pinning lock. One thing to fix before merge; the rest is nits and things I verified are fine.

**Please (address before merge)**

- **`package.json` — bump to 2.63.8.** #782 merged ahead of this one (about an hour ago) and took 2.63.7. This PR also bumps 2.63.6 → 2.63.7, and because both make the identical edit, git auto-resolves it — GitHub shows MERGEABLE and nothing flags the clash — so as it stands this would merge sharing #782's version, i.e. two different code states under 2.63.7. The branch bumps exactly one patch per merged PR (2.63.1 → 2.63.7, no repeats), and CI has no version-increment guard to catch this. Rebase on #782 and `npm run version:patch` to land on 2.63.8.

**nit (optional)**

- **`tool.ts` — `minApiVersion?: string`** could use the template-literal type the sibling helper uses (`` `${number}.${number}.${number}` ``) so a typo'd floor like `'0.26'` fails at compile time. Low stakes — the floor-pinning test already locks all eight values — but a cheap belt.

**Verified fine as-is (no action)**

- `apiVersion.ts` vs `utils/isTableauVersionAtLeast.ts` is not duplication to collapse: different domain (External Client API `0.x.y` vs product `year.major.minor`) and opposite fail-direction (unknown → below every floor here; unknown → assume-fresh-build there). Keeping them separate is correct.
- `resolveConnectedApiVersion` / `filterToolsByApiVersion` exported at module level for tests matches the file's own convention (`selectToolsForProfile` does the same right above).
- Fail-open holds end to end: `discoverInstances` swallows an unreadable discovery dir and returns `[]` (won't throw and break registration) → `resolveConnectedApiVersion` returns undefined → `filterToolsByApiVersion` keeps every tool → the per-call `endpointNotInThisBuild` guard stays the backstop.
- The "Hiding N tool(s): …" log listing tool names is fine — it's a server log, not a description or error the model consumes, and the names are the actual dropped set.

Out of lens: the floor *values* (0.2.5/0.2.6/0.2.7) match the PR's table and the branch's `API 0.2.5`/`0.2.7` history, but which Desktop build first served each endpoint is Desktop-side knowledge only you can vouch for — your reviewer-checkpoint call to sanity-check the numbers is the right one.

Holding the approve only on the version bump — the code itself is approve-ready. Happy to approve once it's on 2.63.8.

src/desktop/externalApi/apiVersion.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Numeric SemVer comparison ("0.2.6" >= "0.2.5"). Missing/unparseable parts read as 0,
+ * so `undefined` and `""` compare as 0.0.0 — below every real floor.
+ */
+export function apiVersionAtLeast(current: string | undefined, minimum: string): boolean {
```
tableaukyler, line 7: looks like semver is at the very least a transitive dependency already in this repo. Would recommend just using that lib since it's well tested and avoids any accidental bugs

lauren-jacksonSFDC:

Re @mattcfilbert's [review](https://github.com/tableau/tableau-mcp/pull/783#pullrequestreview-4965461825):

Thanks — both addressed.

1. **Version.** Rebased onto feature/desktop, which auto-dropped my duplicate 2.63.7 bump. Bumped to 2.63.8.
2. **Type nit.** Added an `ApiVersionFloor` template-literal type in `apiVersion.ts` and typed `minApiVersion` with it.

src/desktop/externalApi/apiVersion.ts
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

Fair point, but `semver` is only a transitive dep here (via eslint/vitest/npm-run-all, at mixed 5.x/7.x), so this would mean adding a new direct dependency. It also throws on partial/invalid input — the opposite of this helper's deliberate fail-open (`undefined`/unparseable → `0.0.0`, below every floor). Kept the hand-rolled comparator (12 unit tests) for now; glad to switch if we take semver on as a first-class dep.

mattcfilbert (approved):
**MattGPT** — re-review on Matt's behalf (andy-review lens).

Both items from the last pass are closed. Approving.

- **Version bump — resolved.** Now on 2.63.8 over base 2.63.7, so no collision with #782. The `types.ts` / `externalApiContract.test.ts` deltas since my last review are just #782's merged `.nullish()` change flowing in through the rebase, not new code here.
- **The template-literal nit — adopted, and then some.** `minApiVersion` is now typed `ApiVersionFloor` (`` `${number}.${number}.${number}` ``), so a typo'd floor fails at compile time. `build (24.x)` is green, which means tsc proved all eight floor literals conform and the full suite passes.

**CI:** `build (>=22.7.5 <23)` is red, but it's the same pre-existing upstream break I flagged on #782 — the only failing step is "OAuth Tests (Tableau authz server)" (the live Tableau OAuth consent-page HTML changed, so `button[type="submit"]` now matches multiple buttons). Everything this PR touches is green: TypeScript compilation, Lint, Build, Tests, E2E, and "OAuth Tests (Embedded authz server)" all pass, and `build (24.x)` ran the identical steps clean. This PR has no OAuth surface, so that red is unrelated — worth a separate fix to the selector, independent of this change.

Nice work on the gate — the fail-open path and the test coverage are the parts that make this safe to build on.

src/desktop/externalApi/apiVersion.ts
lauren-jacksonSFDC: I feel like I agree with you here about semver - but to keep things moving I'll keep it this way for now and make sure it gets implemented right when we do the full capabilities solution instead

PR #788, opened 2026-08-19
@W-23864448 Add REST GET /workbooks/{id}/connections to the SDK

ycath:
## Summary

Adds the **Query Workbook Connections** endpoint (`GET /sites/{siteId}/workbooks/{workbookId}/connections`) to the workbooks SDK, modeled on the flow connections analog (`flowsApi.ts` / `flowsMethods.ts` / `flowConnectionSchema`). This is the only way to obtain a queryable embedded datasource LUID for a workbook — no such endpoint existed before (only flows had `queryFlowConnections`).

SDK-layer only; no new MCP tool is added in this change.

## Changes

- **`apis/workbooksApi.ts`** — `getWorkbookConnections` Zodios endpoint, registered in the workbooks API.
- **`methods/workbooksMethods.ts`** — `getWorkbookConnections({ siteId, workbookId })` returning `WorkbookConnection[]` (each carrying its `datasource`).
- **`types/workbook.ts`** — `workbookConnectionSchema` / `WorkbookConnection`. Shape is `{ id, type?, datasource?: { id, name? } }`. Everything but `id` is optional because Tableau Server/Cloud versions are inconsistent about which fields they emit, and Zodios validates at the schema boundary (one connection missing a `datasource` would otherwise fail the whole call). Connection secret/host fields (`serverAddress`, `serverPort`, `userName`, `embedPassword`, `connectionPassword`) are deliberately omitted, following the precedent in `flow.ts`'s `flowConnectionSchema`.
- **`methods/workbooksMethods.test.ts`** — colocated unit tests asserting datasource id/name extraction, optional-field pass-through, and the empty-connections case.

## Unrelated fix

- **`tests/oauth/tableau-authz/flows/consentFlow.ts`** — the Tableau OAuth consent page now renders multiple `type="submit"` buttons (Switch site / Switch username / Allow), which tripped Playwright strict mode. Target the Allow button by id (`#allow-button`) instead of the ambiguous `button[type="submit"]`. This is orthogonal to the SDK change but was blocking CI.

## Scope

Query Workbook Connections requires `tableau:content:read`, which is already wired across `TableauApiScope`, `JwtScopes`, and the existing `get-workbook`/`list-workbooks` tool maps. No new scope is needed — confirmed and documented in the method JSDoc.

## Testing

- `npx vitest run src/sdks/tableau/methods/workbooksMethods.test.ts` — 9 passed
- `eslint` + `tsc --noEmit` clean on the changed files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.5.2",
+  "version": "4.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.5.2",
+      "version": "4.6.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1095.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.5.5",
+  "version": "4.5.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.5.5",
+      "version": "4.5.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1095.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.6.0",
+  "version": "4.5.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.6.0",
+      "version": "4.5.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1095.0",
@@ -639,7 +639,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -688,7 +687,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
@@ -1754,7 +1752,6 @@
       "resolved": "https://registry.npmjs.org/@modelcontextprotocol/sdk/-/sdk-1.29.0.tgz",
       "integrity": "sha512-zo37mZA9hJWpULgkRpowewez1y6ML5GsXJPY8FI0tBBCd77HEvza4jDqRKOXgHNn867PVGCyTdzqpz0izu5ZjQ==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@hono/node-server": "^1.19.9",
         "ajv": "^8.17.1",
@@ -2482,7 +2479,6 @@
       "integrity": "sha512-sKYVuV7Sv9fbPIt/442koC7+IIwK5olP1KWeD88e/idgoJqDm3JV/YUiPwkoKK92ylff2MGxSz1CSjsXelx0YA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/body-parser": "*",
         "@types/express-serve-static-core": "^5.0.0",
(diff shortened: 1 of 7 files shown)


src/sdks/tableau/apis/workbooksApi.ts
```diff
@@ -89,9 +89,23 @@ const addTagsToWorkbookEndpoint = makeEndpoint({
   response: z.object({ tags: tagsSchema }),
 });
 
+const queryWorkbookConnectionsEndpoint = makeEndpoint({
```
williamyen0203: To be consistent, should this be `getWorkbookConnectionsEndpoint` instead of `query...`?

src/sdks/tableau/types/workbook.ts
```diff
@@ -33,3 +33,18 @@ export const workbookSchema = z.object({
...
+// The nested datasource id is the queryable embedded datasource LUID — the point of this endpoint.
+// Secret/host fields (serverAddress, serverPort, userName, embedPassword, connectionPassword) are
+// deliberately omitted, following flow.ts's flowConnectionSchema.
+export const workbookConnectionSchema = z.object({
+  id: z.string(),
+  type: z.string().optional(),
+  datasource: z.object({
+    id: z.string(),
+    name: z.string(),
+  }),
```
sf-ali-mahmoud: **High severity:** `datasource` (and its `id`/`name`) are non-optional here, which breaks with this schema's own stated precedent. Zodios validates responses at the schema boundary, so if any connection Tableau returns lacks a `datasource` node — or one without `name` — the *entire* `getWorkbookConnections()` call throws, not just that element. `flowConnectionSchema` (the precedent cited above) deliberately makes everything but `id` optional for exactly this reason — Tableau Server/Cloud versions are inconsistent about which fields they emit.

Confirmed against the server implementation, not just by analogy:
- `public-types.xsd` declares `datasource` as `minOccurs="0"` on `connectionType` — optional on the wire.
- `ViewModelObjectFactory.java` gates the entire `datasource` element on REST API >= v2.3 — any caller below that version omits it entirely for every connection.
- `datasource.name` has no schema constraint and no null/empty guard in the serializer (unlike the sibling `type` field one line above, which does use `StringUtils.defaultString`) — empty/null names are not defended against.

So this isn't a theoretical version-skew concern, it's a documented server-side omission path.

Suggest:
```ts
datasource: z.object({
  id: z.string(),
  name: z.string().optional(),
}).optional(),
```
(or at least make `name` optional) to avoid a parse-failure landmine in production.

src/sdks/tableau/apis/workbooksApi.ts
williamyen0203, line 92: To be consistent, should this be `getWorkbookConnectionsEndpoint` instead of `query...`?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.5.2",
+  "version": "4.6.0",
```
stephendeoca: I think we'd consider this a patch bump 4.5.3 since the behavior of MCP doesn't change or add in new functionality

src/sdks/tableau/apis/workbooksApi.ts
```diff
@@ -89,9 +89,23 @@ const addTagsToWorkbookEndpoint = makeEndpoint({
   response: z.object({ tags: tagsSchema }),
 });
 
+const getWorkbookConnectionsEndpoint = makeEndpoint({
```
stephendeoca: We like to have the endpoint match the name it has in our REST API docs. This one is called Query Workbook Connections, can you rename the API from to follow that pattern. Here would be queryWorkbookConnectionsEndpoint.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.5.2",
```
jarhun88: mismatch with package now

jarhun88 (approved):

src/sdks/tableau/apis/workbooksApi.ts
```diff
@@ -89,9 +89,23 @@ const addTagsToWorkbookEndpoint = makeEndpoint({
...
+  method: 'get',
+  path: '/sites/:siteId/workbooks/:workbookId/connections',
+  alias: 'getWorkbookConnections',
+  description:
```
stephendeoca, line 96: FYI:  usually parameters are specified here, but it seems like getWorkbookEndpoint doesn't do that either so you should be fine.

src/sdks/tableau/apis/workbooksApi.ts
```diff
@@ -89,9 +89,23 @@ const addTagsToWorkbookEndpoint = makeEndpoint({
   response: z.object({ tags: tagsSchema }),
 });
 
+const queryWorkbookConnectionsEndpoint = makeEndpoint({
+  method: 'get',
+  path: '/sites/:siteId/workbooks/:workbookId/connections',
+  alias: 'getWorkbookConnections',
```
stephendeoca: update alias to say query instead of get

sf-ali-mahmoud (approved):

PR #789, opened 2026-08-19
feat(desktop): expose new datasource fields, refresh, and publish endpoints

lauren-jacksonSFDC:
## What this does

Brings the Desktop server's contract up to External Client API 0.2.8 and surfaces everything additive that version added. The rule this keeps for contract intake: bump the checked-in OpenAPI fixture, tighten the zod schemas against it, and gate each new tool behind a per-tool `minApiVersion` floor — never a global version bump.

Every new surface here carries `minApiVersion: '0.2.8'`, so it stays hidden from any connected Desktop that predates the spec. The gate fails open only when the connected version can't be read, leaving the per-call `endpointNotInThisBuild` guard as the backstop.

## Changes

**Contract intake (0.2.8)**
- Refresh the checked-in `externalClientApi-openapi.json` fixture from 0.2.6 → 0.2.8; `externalApiContract.test.ts` re-validates the zod schemas and route coverage against it.
- Add problem code `unsupported-target-version` and routes `workbook:publish`, `datasources/{id}:refreshData`, `datasources/{id}:refreshExtract`.

**New datasource fields** — `list-workbook-datasources`
- Surface `type`, `isExtract` (extract vs. live), and `hasDownloadFilePermission` (published datasources only — the API emits `null` when unpublished, which we drop rather than report as `false`).

**Three new tools** — all gated at 0.2.8, all in the default `dynamic-authoring` profile
- `publish-workbook` — opens the Publish Workbook dialog (`workbook:publish`).
- `refresh-datasource-data` — refresh a datasource's live data by reading it again from its source.
- `refresh-datasource-extract` — refresh a datasource's extract; optional `isFullRefresh` forces a full refresh, otherwise the server default (incremental when the extract supports one).

## Safety
- `publish-workbook` and both refresh tools are non-read-only with `openWorldHint: true` (they reach a remote site / a datasource's backing source). Publish is non-idempotent (it opens a dialog); the two refresh tools are idempotent.
- The refresh tools take the datasource's stable id from the workbook's datasource list, never its display name or caption.

## Tests
- New unit tests for each of the three tools, plus the projection of the new datasource fields.
- Mock server and contract tests extended for the new routes and fields.
- Byte-budget ratchet re-pinned; the dynamic-authoring surface is ~42.5k, still well under the ~46k tools/list cliff where clients auto-defer schemas.
- Full suite green: `tsc --noEmit`, eslint, and vitest all pass.


mattcfilbert (approved):
lgtm

PR #792, opened 2026-08-19
@W-23336789: August docs update pass

bcantoni:
This PR is a docs-only update where Claude and I make improvements in these areas:

* Prompts - updating and clarifying based on the latest source code
* Tools - ditto
* Hosted MCP - content improvements and screenshot sizes for index and client integrations pages


Alon-ST-DATA:
[Alon's agent]

**`docs/docs/tools/tasks/list-extract-refresh-tasks.md` — filtering section verified accurate ✅, one consistency nit.**

Cross-checked the doc against the tool source (`src/tools/web/extractRefreshTasks/extractRefreshTasksFilterUtils.ts`, `src/utils/parseAndValidateFilterString.ts`, `src/utils/clientSideFilter.ts`). Filterable fields (all 13), per-field operators, `field:operator:value` format, comma-separated AND logic, ISO-8601 date handling, and the client-side filter/pageSize/limit note all match the code exactly. No correctness issues in the doc.

**Nit — IN-operator syntax differs between this doc and the tool's own `description`:**
- This doc advertises: `schedule.frequency:in:[Daily,Weekly]` (bracketed)
- The tool `description` (`listExtractRefreshTasks.ts`, IN example line) advertises: `schedule.frequency:in:Daily|Weekly` (pipe, no brackets)

Both parse fine — `parseInValues` (`clientSideFilter.ts`) strips optional `[...]` then splits on `/[,|]/`. But the **bracketed form in this doc is the correct canonical one**: top-level filters are comma-split (`splitTopLevel`), so a multi-value IN list *must* be bracketed to avoid being shredded into separate expressions (`in:Daily,Weekly` without brackets would break). The `splitTopLevel` comment itself notes "the documentation for every list-* tool advertises `[A,B]` as the canonical syntax."

So the doc is aligned with the intended convention — it's the **tool `description` that's the stale outlier**. Not in scope for this docs PR, but worth a follow-up code change to update the tool description's IN example to `[Daily,Weekly]` for consistency across the surface.

Alon-ST-DATA:
[Alon's agent]

**`docs/docs/prompts/` — cross-checked all 6 prompt docs against prompt source. Functionally accurate; PR already fixes the one real bug (the stale `delete-extract-refresh-task` → `delete-content` rename in `extract-optimization-apply.md`). Two minor accuracy nits remain, both in non-executable informational sections of `user-license-reclamation-inform.md`:**

**1. `user-license-reclamation-inform.md:47` — wrong scope name.** The Scopes section lists `vds:read` for `query-admin-insights`, which is not a real scope. The actual scopes (per `src/server/oauth/scopes.ts:348-355`) are the mcp scope `tableau:mcp:datasource:read` plus the api scope `tableau:viz_data_service:read`. Also line 46: `list-users` needs the mcp `tableau:mcp:users:read` in addition to `tableau:users:read`. Informational only — no call fails — but the scope names as written don't exist.

**2. `user-license-reclamation-inform.md:19` — "client-side" filter wording.** The step says the role/`lastLogin` filter "is applied client-side after fetch" (conditioned on `MAX_RESULT_LIMIT`). In the inform prompt the filter is passed to `list-users` as a server-side filter string (`siteRole:in:…,lastLogin:lt:<cutoff>`), not applied client-side by the model. Suggest rewording to "applied by `list-users` as a filter" to avoid implying the prompt post-filters. (The client-side-after-cap wording is accurate for the *apply* prompt, not inform.)

Everything else checks out: prompt names, argument names/required/defaults, tools invoked, `query-admin-insights` kinds, env-var names/bounds/defaults, and all destructive-prompt HITL/dry-run safety claims match source. The `roles` (inform) vs `siteRoles` (apply) argument-name asymmetry and the apply-vs-inform default-role-set difference are both documented correctly. `sidebar_position` values are unique (1–6). Nice work.

yyk16293 (approved):

stephendeoca (approved):

bcantoni:
Addressed both nits from @Alon-ST-DATA on `user-license-reclamation-inform.md`:

- **Scopes section**: `vds:read` is not a real scope. Replaced with the actual MCP+API scope pairs from `src/server/oauth/scopes.ts`: `tableau:mcp:users:read` + `tableau:users:read` for `list-users`, and `tableau:mcp:datasource:read` + `tableau:viz_data_service:read` for `query-admin-insights`.
- **Filter wording**: reworded the `list-users` step to say the role/`lastLogin` filter is applied by `list-users` itself (it is built into the filter string passed to that tool call, per `inform.ts`), rather than implying the prompt post-filters client-side after fetch.

On the third nit (the `listExtractRefreshTasks.ts` tool description stale `Daily|Weekly` IN-operator example, which should be the canonical bracketed `[Daily,Weekly]` form to match `listJobs.ts` and the `splitTopLevel` convention): confirmed the analysis is correct, but leaving that out of this PR since it is a code change and this PR is docs-only. Filing/handing that off separately.

Alon-ST-DATA (approved):

Akash-Rastogi (approved):
[Akash's agent] LGTM. Docs-only accuracy pass — reviewed all substantive changes and they match the current code: the `EXCLUDE_TOOLS` value now lists all 16 tool groups; `ADMIN_TOOLS_ENABLED` correctly reworded from "site setting"/"feature flag" to "environment variable"; the extract-optimization prompt now documents the preview→confirm (confirmationToken) flow and the correct `delete-content` tool name; user-license-reclamation `sidebar_position` 5→6 clears the collision with the -inform page; and the `get-workbook` JSON example is now well-formed. CI green. No concerns.

PR #794, opened 2026-08-20
feat(desktop): make core chart templates match requested shapes

mattcfilbert:
This makes the four must-demo chart requests produce the requested Tableau structure on the existing TBM and `bind-template` path. It also preserves plain scatter as plain scatter, gives explicit trendline asks a separate carrier, and makes named-chart or geographic ambiguity fail closed.

This is a rule we should preserve: an exact chart type wins only when its required fields are present. Otherwise the binder proposes instead of silently falling back. Future chart-breadth work should extend TBMs and focused resolvers, not add a parallel chart API.

What changed:

- Grouped bar binds two categorical fields plus one measure without donor filters.
- Multi-series line uses an explicit Line mark.
- Scatter uses two measures and an explicit grain; trendline requests use a separate carrier.
- Symbol map uses one semantic geographic slot plus size.
- Routing and artifact tests cover exact mappings, injected XML, synonyms, and fail-closed cases.

Validation:

- `scripts/agent-check`: all five checks passed.
- 377 test files and 5,921 tests passed on this PR head.
- Node 24 CI, SAST, credential scanning, CLA, desktop build, and the lockstep hash gate passed.
- Final adversarial review: no P1 or P2 findings remain.

Current CI note:

- The Node 22 job is red in the external Tableau OAuth authorization suite after the consent page began returning three submit buttons. The chart-owned tests pass, and this PR does not change that selector.

Merge and claim gates:

- Live Tableau Desktop apply, readback, and render on Windows or macOS remain unverified.
- The changed lockstep classifier must be copied to agent-to-tableau-desktop before the stack merges.

Approving means these four template contracts and the fail-closed routing rule become the base for the next stacked chart-breadth PR. It does not claim live Desktop visual proof yet.

_Drafted by MattGPT for Matt Filbert._


src/desktop/binder/classify.ts
```diff
@@ -2059,31 +2078,50 @@ function pickGeoField(
...
   for (const slot of geoSlots) {
     const askNamedPool = augmented.filter((field) => field.role === 'dimension');
+    const slotConcept = geoConceptFromSlot(slot);
+    const requestedConcepts = new Set(geoConceptsNamedInAsk(maskedAsk));
+    if (!slotConcept) {
+      for (const field of askNamedPool) {
+        const exactConcept = geoConceptFromSemanticRole(field.semanticRole);
+        if (exactConcept) requestedConcepts.add(exactConcept);
+      }
+      if (requestedConcepts.size > 1) return null;
```
myu404, line 2106: 🤖 **Reviewed by MichaelGPT**

For a *neutral* geo slot, `requestedConcepts` is seeded from `geoConceptsNamedInAsk(maskedAsk)` (line 2100), which scans the whole masked ask for synonyms — `state: ['state','province','region','admin']`, `country: ['country','nation']`, `zip: ['zip','zipcode','postal']`. `maskFieldNames` only blanks literal *field names*, so an incidental free-text synonym plus one named geo field (e.g. *"map of admin costs by Country"*) yields `{state, country}` and hits `if (requestedConcepts.size > 1) return null;` — a false multi-concept fail-close.

Fail-closed is the safe direction, so this may be intended. Could you confirm — and either restrict the ask-side concept scan to tokens not consumed by field masking, or add a test documenting the intended behavior when a geo synonym appears as free text rather than as a field name? (No existing test exercises this case.)

src/desktop/binder/encodingHonesty.test.ts
```diff
@@ -192,10 +189,7 @@ describe('binder encoding honesty — requested vs filled', () => {
     expect(unfilled.reportsColorUnfilled).toBe(true);
     expect(unfilled.xml).not.toContain('<color ');
 
-    // Reported filled ⟺ the color node really is present.
-    const filled = await render(FULLY_SATISFIED_ASK);
-    expect(filled.reportsColorFilled).toBe(true);
-    expect(filled.xml).toContain('<color ');
+    expect(runtimeSlots.some((slot) => slot.role.includes('color'))).toBe(false);
```
myu404, line 207: 🤖 **Reviewed by MichaelGPT**

Because the reshaped symbol map lost its color slot, this test dropped its *filled-side* round-trip (`expect(filled.xml).toContain('<color ')`) and now only asserts the unfilled side (`runtimeSlots.some(... role.includes('color')) === false`). No test in this file now asserts that a *filled* encoding round-trips into applied workbook XML.

The `filled ⟺ node-present` invariant is still covered elsewhere (`injectTemplateCore.test.ts:457`, `puppetCompatibilityProjection.test.ts:246`), so this is a coverage reduction in the honesty suite, not a gap. Consider repointing the filled-side assertion at a template that still carries a color/size encoding so both halves of the honesty invariant stay covered here.

myu404 (approved):

This delivers the four must-demo template reshapes and the fail-closed routing rule as described. I verified the reshaped `.tbm` shapes (paired-bar = Category/Region rows + `sum:Sales` + Region color; trend-line = date cols + Sales rows + Category color + Line mark; plain vs. trendline scatter are byte-identical except the added `<trendline enabled='true' fit='linear'/>` carrier; symbol map collapsed to a single neutral `[Location]` slot + Sales size). The new `correlation-scatter-trendline-chart.tbm` is auto-discovered by directory scan and the 136-TBM inferSlots invariant holds at head. `selectWithinFamily`'s "pick the specificity winner, then bind-or-null" matches the propose-don't-fall-back contract, and `lockstep.hashes.json` correctly updates only `classify.ts`.

No correctness or contract defects. Two minor observations inline (one `question`, one `minor` test-gap) — none blocking.

**Non-blocking caveats (already noted in the PR description):** live Tableau Desktop apply/readback/render remains unverified, and the changed lockstep classifier must be synced downstream before the stack merges. These are merge-gate items, not code-review blockers.

PR #795, opened 2026-08-20
feat(desktop): make intermediate chart templates bind safely

mattcfilbert:
This stacked PR adds four more chart contracts on the same TBM and bind-template path established in #794: treemap, heatmap, pie, and bullet.

The rules introduced here are durable. Known high-cardinality pie requests fail closed above 12 slices across every proposal path. Heatmaps require a color measure. Bullet charts require distinct actual and target fields and keep the aggregate target off Detail so it does not change the requested grain.

What changed:
- treemap binds Category, Sub-Category, and Sales with explicit square marks, hierarchy, size, labels, and tooltip evidence
- heatmap binds Category by Region with Sales on color and labels
- pie binds Region and Sales with slice labels, exact-value tooltip, and a proven percent-of-total table calculation
- bullet binds Region, Sales, and aggregate Target with a proven target reference marker
- Call 1, Call 2, and injected proposal paths share the same rejection rules

Validation:
- scripts/agent-check: all five checks passed
- 377 test files and 5,942 tests passed
- desktop build and lockstep hash gate passed
- adversarial review is clean

Not yet verified:
- live Tableau Desktop apply, readback, and render on Mac or Windows
- qualitative 60/80 bullet bands are not included because no Tableau-authored donor exists in tableau-mcp or a2td; the target marker is the proven supported contract
- downstream lockstep consumer sync for classify.ts

This PR is stacked on #794. Approving it means these four additional contracts and shared fail-closed rules are ready once #794 lands.

_Drafted by MattGPT for Matt Filbert._


src/desktop/binder/cardinality.ts
```diff
@@ -15,7 +17,7 @@ const ROLE_BANDS: Readonly<Record<string, CardinalityBand>> = {
   },
   'wedge-size': {
     ideal_max: 8,
-    workable_max: 12,
+    workable_max: PIE_SLICE_WORKABLE_MAX,
```
myu404: 🤖 **Reviewed by MichaelGPT**

The "single source of truth" coupling for the pie cap is cosmetic here. `PIE_SLICE_WORKABLE_MAX` is reused on the `wedge-size` band, but in `part-to-whole-pie-chart.tbm` `wedge-size` sits on `[sum:Sales:qk]` — a **measure** (quantitative kind). `idealCardinality()` only bands `BANDED_KINDS` (categorical/geo/temporal/quantitative-or-categorical) and returns `undefined` for plain `quantitative` (line 75), so **this band is unreachable for the real template**.

The actual slice-count enforcement is the standalone `validate.ts:646` blocker keyed on the categorical color slot (Region): `slice.approxCount > PIE_SLICE_WORKABLE_MAX`. It works — but `cardinality.test.ts` proves the band via a `kind:'categorical'` fixture, masking that it's dead for the real slot. Either accept it as a symbolic constant (a brief comment would help), or make the test fixture reflect a quantitative wedge-size slot and note the band is advisory-only.

src/desktop/binder/bind-behavior-matrix.invariant.test.ts
```diff
@@ -247,12 +266,69 @@ describe('binder/bind-behavior-matrix — KNOWN one-shots', () => {
...
+      {
+        '{{field_base_1}}': '[Sample - Superstore].[none:Category:nk]',
+        '{{field_base_2}}': '[Sample - Superstore].[none:Sub-Category:nk]',
+        '{{field_base_3}}': '[Sample - Superstore].[sum:Sales:qk]',
+      },
+    ],
+    [
+      'heatmap of Sales by Category and Region',
+      {
+        '{{field_base_1}}': '[Sample - Superstore].[none:Category:nk]',
```
myu404, line 280: 🤖 **Reviewed by MichaelGPT**

The new pie/heatmap/treemap must-demo cases only bind against the `Sample - Superstore` fixture, whose name matches the datasource literals authored in the pie `.tbm` — so a datasource-rename rewrite bug in the `pcto` table-calc instance wouldn't surface here.

Weakening the concern: `bind()` returns only `field_mapping`, never applied XML, so the `pcto` rewrite can't be exercised in *this* file regardless, and the compound `pcto:cum` datasource rewrite is already covered generically in `fieldReferenceRewriter.test.ts` (incl. a renamed-datasource case at line 433). So the rewrite path isn't truly uncovered — this is a nice-to-have: the bullet test already uses a renamed `Bullet Data` datasource; adding one renamed-datasource pie bind would close the last gap symmetrically.

myu404 (approved):

This delivers the four additional contracts (treemap, heatmap/correlation-highlight-table, pie, quota-attainment-bullet) as described. The templates are clean Tableau-authored bookmarks with the right encodings, the heatmap has no residual template calc/filter/slices, and the pie carries a real `PctTotal` calc (`pcto:sum:Sales:qk`, `p0.0%`). All fail-closed rules live in the shared `validateBinding` seam (heatmap color-measure required, pie >12-slice cap → `kind-mismatch`, bullet distinct actual/target → `base-column-conflict`, bullet target reference-line required), and the required-slot predicates match the roles the `.tbm` files actually declare. Routing keywords, `allDescriptors` exposure, and the `PIE_SLICE_WORKABLE_MAX` export are all present and tested.

No correctness or contract defects. Two `minor` findings inline — none blocking.

**Non-blocking caveats (already noted in the PR description):** live Tableau Desktop apply/readback/render remains unverified, and downstream lockstep classifier sync is pending. These are merge-gate items, not code-review blockers. Stacked on #794 — merge order is fixed.

PR #796, opened 2026-08-20
feat(desktop): make advanced chart templates bind safely

mattcfilbert:
This PR makes the chart templates reachable from the dynamic authoring profile through one route: use `bind-template` first for a recognizable single-view request, then use the existing template artifact flow if binding cannot finish safely. This is a rule we now keep, not a second template manager.

Approving this PR means accepting both the advanced chart contracts and that routing rule. Future chart work must preserve the same catalog, field-resolution, apply, and readback path instead of adding another runner.

This is stacked on #795.

### Route and recovery

- The default dynamic profile now serves `bind-template` alongside the existing list/build/apply tools.
- Explicit chart names and semantic asks such as “Show Sales over time” use the binder first.
- Preview/no-change, open multi-chart work, existing-sheet edits, and unnamed derived metrics stay on their existing routes.
- Only a proposal earns one structured correction call. An escalation or an unsuccessful correction moves to the existing artifact flow.
- A run is complete only after apply and clean host/readback verification. Skipped or uncertain verification requires live inspection.
- Public results no longer expose hidden legacy apply tools or escaped field mappings as recovery instructions.

### Advanced chart contracts

- Box plots require a measure, category, and explicitly named record grain.
- Gantt charts bind task, start, and end fields without swapping dates, then size Gantt bars with an authored duration calculation.
- Histograms reuse one measure for bins and counts and accept a positive `bin_size` override.
- Waterfalls preserve sequence, running total, negative bar size, sign color, labels, and a total column.
- Bubble charts require distinct X, Y, size, and record-grain fields. Color binds only when requested.
- Summary-data rows are marked as value readback with unspecified order; they are not visual-sort evidence.

### Evidence

- `scripts/agent-check`: all five checks pass; 377 test files and 6,012 tests pass.
- Scoped Desktop server E2E: 2/2 pass.
- Dynamic surface: 52 tools; 43,151 bytes under the unchanged 46,000-byte ceiling. Legacy apply tools remain hidden.
- Mac Studio + Superstore: a named bar-chart request bound and applied with clean host verification. A semantic time-series ask returned one proposal, then moved to the existing artifact flow when confidence stayed low. The artifact flow also built and applied a bar chart with passed verification.
- Final adversarial review found no P1/P2 blockers.

### Remaining risks

- Windows packaged Studio remains the next required calibration target.
- Visual/render quality is separate from structural readback and still needs its own evidence.
- Box and bubble asks without record grain fail closed instead of guessing.
- Bullet-chart 60/80% qualitative bands remain unsupported until we have an authored Tableau donor.
- OAuth is outside this change and was not investigated.

_Updated by MattGPT for Matt Filbert._


src/desktop/binder/classify.ts
```diff
@@ -3243,7 +3252,19 @@ function resolveExplicitCorrelationBubble(
...
     .map((field) => ({ field, index: position(field) }))
     .sort((a, b) => a.index - b.index);
-  const dimension = { field: dimensions[0], index: position(dimensions[0]) };
+  const colorMatch = /\bcolou?r(?:ed)?\s+by\b/i.exec(ask);
+  const colorCandidates = colorMatch
+    ? matchFieldsInAsk(ask.slice(colorMatch.index + colorMatch[0].length), summary).filter(
+        isCategorical,
+      )
+    : [];
+  if ((dimensions.length === 2 && colorCandidates.length !== 1) || dimensions.length === 1 && colorMatch) {
```
myu404: 🤖 **Reviewed by MichaelGPT**

The bubble color guard mixes `&&`/`||` without parentheses on the second operand: `if ((dimensions.length === 2 && colorCandidates.length !== 1) || dimensions.length === 1 && colorMatch)`. Functionally correct (`&&` binds tighter), but the asymmetry next to the fully-parenthesized first operand is easy to misread. Suggest `... || (dimensions.length === 1 && colorMatch !== null)`.

myu404 (approved):

This delivers the routing change and the five advanced contracts as described. `bind-template` is added to `DYNAMIC_AUTHORING_TOOL_PROFILE`, the "bind-first, then guarded artifact fallback" precedence is encoded in `instructions.ts`/`searchLibrary.ts`, and the recovery-gate refactor enforces "one structured correction, then fallback" (`isStructuredCorrectionCall` + `correctionFallbackResult` + `recoveryGateBlock`). Box/gantt/histogram/waterfall/bubble each have matching classify resolvers, validate blockers, and `.tbm` edits, backed by extensive new tests. Public results strip legacy apply instructions, `SUMMARY_ROW_ORDER` threads through all read paths, the donut blocker and `resolveUniqueDatasourceName` are in place, and the tool-surface budget is asserted (52 tools, 43,151 B < 46,000 B).

No correctness or contract-violating bugs. Three findings (two `minor`, one `nit`) — none blocking. One anchors inline below; two touch lines outside this diff and are noted here:

---
### 🤖 Reviewed by MichaelGPT — `src/desktop/route/route-state.ts:835` — minor / dead-code
This PR removed every consumer of the terminal-repair-allowance machinery from `bindTemplate.ts` (the `repair.remaining===1` branch in `recoveryGateBlock`, `terminalRepairAllowanceFor`, the param on `recordBindRecoveryAttemptFailOpen`), but `route-state.ts` isn't in this diff — so the **producer was left behind**. At head, `git grep consumeTerminalRepairAllowance` returns only its definition (line 835); zero callers, tests included. `withTerminalRepairAllowance` is still invoked only to maintain a field nothing reads. Suggest removing `consumeTerminalRepairAllowance` / `withTerminalRepairAllowance` / the `terminalRepairAllowance` field, or adding a note on why the state is intentionally retained.

### 🤖 Reviewed by MichaelGPT — `src/tools/desktop/api/applyWorksheet.ts:433` — minor / consistency
The manual/cached worksheet-XML apply branch treats skipped readback as **terminal**, inconsistent with the two template branches this PR just changed. Line 427 computes `readbackRan = readback !== undefined && readback.status !== 'skipped'`, but line 433 returns `doneNextAction(...)` unconditionally — it only downgrades the receipt's `unverified` list. The artifact branch (~line 213) and direct-template branch (~line 302) now route `verification.status === 'skipped'` to `prefillNextAction('Verification unavailable — inspect live worksheet state')`, matching the PR's own "skipped verification is nonterminal" contract. Suggest routing the skipped/absent-readback case here to `prefillNextAction` for symmetry, or a comment if edit-in-place is intentionally exempt.
---

**Non-blocking caveats (already noted in the PR description):** Windows packaged Studio calibration, visual/render-quality evidence, and OAuth are out of scope here. Stacked on #795 — merge order is fixed.

PR #797, opened 2026-08-20
feat(desktop): find fields in the open workbook

mattcfilbert:
## Decision

Add one read-only tool for bounded field search in the live open workbook. Keep the existing tools because each answers a different question.

## Tool boundaries

- `list-fields`: fields placed on Rows, Columns, or Marks in one cached worksheet file.
- `list-available-fields`: datasource fields available for authoring from a live or cached workbook; no worksheet placements.
- `search-workbook-fields`: matching fields and calculations in the live workbook, plus explicit placements across worksheets; no cached files or published content.

Future work must preserve these boundaries. Do not infer placement from datasource dependency declarations.

## Description

`search-workbook-fields` searches caption, local name, formula, folder, and datasource. It reports Rows, Columns, and Marks placements across worksheets. It does not call an unmatched field unused.

This change excludes Cloud search, calculation lineage, dashboard references, performance advice, and workbook mutation.

## Motivation and Context

Agents need a small current-workbook discovery tool before they can give reliable analytics or performance advice. `list-available-fields` can enumerate fields but does not support bounded search or placement lookup. Repeating `list-fields` across cached worksheets would require several calls and would not search unused fields or calculation formulas.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

- Merged the latest `feature/desktop` and resolved the combined 55-tool surface.
- Canonical `scripts/agent-check`: all five checks passed.
- 383 test files and 5,941 tests passed.
- Desktop build, typecheck, lint, and lockstep passed.
- The real TAS stdio bridge lists the tool exactly once.
- Adversarial review covered placement matching across AVG/SUM, discrete/continuous, calculated, grouped, and binned pills.
- Tool-surface tests pin the three discovery descriptions and keep the default surface below its budget.

## Checklist

- [x] I updated the package version to 2.64.1 with npm.
- [x] I added tests that prove the feature works.
- [x] New and existing unit tests pass locally.
- [x] This change has no breaking behavior.

Approving this PR means agents can search the live workbook for matching fields and see only placements the workbook proves.

_Opened and updated by MattGPT for Matt Filbert._


.work/specs/current-workbook-field-search.md
```diff
@@ -0,0 +1,74 @@
+# Current workbook field search
```
lauren-jacksonSFDC: Should we be checking in the .work folder

src/tools/desktop/tools.ts
```diff
@@ -112,6 +113,7 @@ export const desktopToolFactories = [
   getAddFieldTool,
   getRemoveFieldTool,
   getResolveFieldTool,
+  getSearchWorkbookFieldsTool,
```
lauren-jacksonSFDC, line 122: This adds `getSearchWorkbookFieldsTool` to the suite of `getListAvailableFieldsTool` and `getListFieldsTool` - is this intentional to have all of them? Do we have a clear surface when having all 3?

.work/specs/current-workbook-field-search.md
mattcfilbert: The .work file is an internal task packet, not product documentation. The base only tracks the template and placeholders, so I’ll remove it from this PR.

src/tools/desktop/tools.ts
mattcfilbert: The three tools have different scopes, but the current descriptions do not make that clear enough. list-fields reads placements from one cached worksheet and is not in the default profile. list-available-fields lists fields available for authoring. The new tool performs a bounded live-workbook search and reports explicit shelf and Marks placements across worksheets. I’ll make those boundaries explicit rather than add more behavior or overload the existing list tool.

src/server.desktop.test.ts
```diff
@@ -239,10 +239,12 @@ async function serializeDesktopToolSurface(tool: DesktopTool<any>): Promise<stri
 // routes, gated to a 0.2.8 floor. All three join DYNAMIC_AUTHORING_TOOL_PROFILE, so dynamic
 // authoring moves 40_096 -> 42_521 (budget kept at the 18-char slack) and the full surface
 // 56_799 -> 59_224. Dynamic still clears the 46k cliff.
-const DYNAMIC_AUTHORING_SURFACE_EXPECTED = 42_521;
-const DYNAMIC_AUTHORING_SURFACE_BUDGET = 42_539;
+// Re-pinned 2026-08-20: search-workbook-fields and the distinct field-discovery descriptions
```
lauren-jacksonSFDC, line 246: ✨ _Lauren's code review bot:_ ✨

This new `Re-pinned 2026-08-20` comment narrates what this specific PR did to two magic numbers (with a date stamp and before/after deltas) rather than a durable rationale, which CLAUDE.md's comment rules forbid.

**Trigger:** Every future PR that nudges these budget constants adds another dated entry to this block — it's now the 12th near-identical one (2026-08-07 through today) — duplicating what `git blame`/commit history already records.

**Why this matters:** This does match an established convention already used ~11 times in this same file across prior merged PRs, so it may be a deliberate repo-level pattern rather than a one-off deviation — worth a policy call on whether to keep growing this block or trim it going forward.

lauren-jacksonSFDC (approved):

PR #799, opened 2026-08-20
feat(desktop): measure worksheet summary-data round trips

mattcfilbert:
This adds one read-only Desktop tool for repeated worksheet summary-data measurements. The tool reports API query and compute round-trip time; it does not claim Tableau render time. For real render bottlenecks, the agent sends the user to Tableau Performance Recorder or Workbook Optimizer.

Scope:

- Profile one exact worksheet in one pinned Studio session.
- Return each timing, result shape, stable fingerprints, and median/min/max/spread.
- Fail closed on missing data or partial API errors.
- Keep the tool in the normal dynamic-authoring profile so Studio TAS can use it without a separate path.

Future changes must preserve the session pin, sequential samples, stable-result check, and the explicit render-time exclusion.

Evidence:

- Canonical agent check: all five gates passed; 379 test files and 5,924 tests passed.
- GitHub Node 22 and Node 24 builds passed, including the real Tableau authorization test.
- Live Mac Studio run against Superstore: 5/5 samples returned the same 200x3 result fingerprint; median 196 ms.
- No workbook data rows are returned by the profiler.

This PR stays in draft until #797 is reviewed and merged. After that, it needs one normal merge from the updated `feature/desktop`, the combined tool-count ratchet, and a final check run. No rebase or force-push is needed.

Approving this PR means accepting summary-data timing as a bounded diagnostic signal, not as proof of worksheet render performance.

Authored by MattGPT for Matt Filbert.

### DIFF
--- a/src/server.desktop.test.ts
+++ b/src/server.desktop.test.ts
@@ -233,12 +233,13 @@ async function serializeDesktopToolSurface(tool: DesktopTool<any>): Promise<stri
 // served profile, while the full schema-only surface moves 56_650 -> 56_799.
 // Re-pinned 2026-08-19: apply-worksheet now infers the target from a cached worksheet fragment, so
 // its worksheetName describe drops the redundant "worksheet" (-3 bytes); dynamic authoring 40_099 -> 40_096.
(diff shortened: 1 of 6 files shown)


mattcfilbert:
not needed

PR #804, opened 2026-08-20
feat(compass): wrap vibe-coded workbook in a full-screen dashboard

patrickg2014:
Vibe-coded workbooks now auto-generate a single automatically-sized dashboard around the viz worksheet instead of publishing a bare sheet, removing the manual "make a dashboard / hide title / republish" steps.

- buildTwbx emits a <dashboards> block (automatic sizing, sheet title hidden, named "<workbook> Dashboard") plus a maximized dashboard window; the host worksheet window is no longer maximized.
- The published-workbook card links to the dashboard view (preferredViewName in toPublishResult), not the underlying sheet - fixing the stray sheet URL and the default "Dashboard1" naming.
- showTabs now defaults to false, so the workbook opens full-screen with no tab navigation and internal wire worksheets stay hidden.

Round-trip (edit-data-app) is unaffected: reconstruct reads only package content + manifest + datasources, never worksheets/dashboards.


src/tools/web/createAndPublishWorkbook/createAndPublishWorkbook.ts
```diff
@@ -54,7 +54,11 @@ const paramsSchema = {
   showTabs: z
     .boolean()
     .optional()
-    .describe('Whether the published workbook shows its sheets as tabs. Defaults to true.'),
+    .describe(
+      'Whether the published workbook shows its sheets as tabs. Defaults to false: this tool ' +
+        'publishes a single full-screen dashboard, so tab navigation is hidden (and any internal ' +
```
santoch-tableau, line 59: Clarifying question- 
Above, you changed maximize=false from true in a couple of places.  Does this "full screen" part of the comment still hold true?

santoch-tableau (changes requested):
This looks good to me.  Just the one comment/question.

src/tools/web/createAndPublishWorkbook/createAndPublishWorkbook.ts
santoch-tableau: ok, i re-read the comment above.

santoch-tableau (approved):

LGraber (approved):

PR #805, opened 2026-08-20
feat(desktop): report live UI state from GET /v0/app

tableaukyler:
## What
Adds live UI-state fields to `AppInfo` (`GET /v0/app`) and syncs the External Client API contract to the live 0.2.9 spec, folding in Lauren's #806.

- `AppInfo` gains `isStartPageVisible`, `isDataSourcePageActive`, `isPresentationMode` — mirrors monolith [sf-analyticscloud/monolith#63456](https://github.com/sf-analyticscloud/monolith/pull/63456). Kept optional (spec marks them required) so an older Desktop build still parses.
- `Operation`/`OperationList` gain `progressWindows` (self-clearing windows, vs. `blockingWindows` which need a human decision) — added to `operationEnvelopeSchema` as an optional array.
- Fixture (`__fixtures__/externalClientApi-openapi.json`) is now a live capture at 0.2.9 (supersedes my earlier hand-edit): also documents 409/422 on the document-replace routes and a 202 on `GET /v0/app`. No code changes needed for those — error mapping and polling are already generic.
- `get-app-info` now also surfaces the 3 UI-state booleans, not just build-id fields (this repo's addition on top of the contract sync).

## Why
Lauren's monolith PR shipped the UI-state fields on Desktop; her follow-up tableau-mcp PR (#806) did the proper live-capture sync including `progressWindows`. Merging both changes into one PR so this ships as a single coherent update to the External Client API contract.

## Changes
- `types.ts`: `appInfoSchema` gains the 3 booleans; `operationEnvelopeSchema` gains `progressWindows`; `windowInfoSchema` doc comment updated
- `getAppInfo.ts` / `.test.ts`: surface the 3 booleans in the tool's output
- `externalApiContract.test.ts`: requiredness-exception lists + provenance comment updated for the live 0.2.9 capture
- `__fixtures__/externalClientApi-openapi.json`: live-captured 0.2.9 spec (AppInfo booleans, progressWindows, 409/422/202 docs)
- `mockExternalApiServer.ts`: stub the 3 `AppInfo` booleans on the `/v0/app` mock response
- `server.desktop.test.ts`: re-pinned `FULL_TOOL_SURFACE_BUDGET` for the longer `get-app-info` description
- Version bump: 2.63.12 → 2.64.0 (minor, new optional capability)

## Test plan
- [x] `npm test` — 5923 passed
- [x] `npm run lint` — clean
- [x] `npm run build:desktop` — succeeds
[...]

PR #806, opened 2026-08-20
Sync External Client API contract to live openapi.json 0.2.9

lauren-jacksonSFDC:
## Summary
- Pulled the live `/openapi.json` from a running Tableau Desktop instance (`info.version` 0.2.9, up from the fixture's 0.2.8) and refreshed `__fixtures__/externalClientApi-openapi.json`.
- No routes or HTTP methods were added or removed — this release only grows existing schemas and documents previously-implicit error responses.
- `AppInfo` gains `isStartPageVisible` / `isDataSourcePageActive` / `isPresentationMode`, added to `appInfoSchema` as optional booleans (spec marks them required, but they're kept optional so an older Desktop build without them still parses) — consistent with every other read schema in this file.
- `Operation` / `OperationList` gain `progressWindows` (self-clearing windows, as opposed to `blockingWindows` which need a human decision) — added to `operationEnvelopeSchema` as an optional array, same shape as `blockingWindows`.
- The spec now documents 409/422 responses on the document-replace routes (`workbook/document`, `dashboards/{id}/document`, `storyboards/{id}/document`, `worksheets/{id}/document`) and a 202 on `GET /v0/app`. No code changes needed: error mapping in `externalApiHttp.ts` is already generic over Problem responses, and `getJson` already polls any 202.
- Updated `externalApiContract.test.ts`'s requiredness-exception pins and provenance comment to match.
- Since no new endpoints were added, there was nothing to add to the dynamic-authoring tool profile (`server.desktop.ts`) or a new MCP tool this round.

## Test plan
- [x] `npx vitest run` on the externalApi contract/http/executor tests and `getAppInfo.test.ts` — 209 passed
- [x] `npm test` — full suite, 5923 passed
- [x] `npm run lint` — clean
- [x] `npm run build:desktop` — succeeds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.63.13",
+  "version": "2.63.14",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.63.13",
+      "version": "2.63.14",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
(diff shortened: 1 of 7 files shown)


tableaukyler:
Folded this into #805 along with the get-app-info UI-state surfacing — closing to avoid a duplicate/conflicting contract sync. Thanks for the live 0.2.9 capture and the `progressWindows` addition!

PR #807, opened 2026-08-20
@W-23757366: Add MCP server instructions string

Akash-Rastogi:
## Description

Sets the MCP server `instructions` string — the field a server returns once at `initialize` that the host injects into the model's context every session. tableau-mcp never set it, so hosts started every session with no orientation to the server and no guidance for choosing a `query-admin-insights` `kind`.

The instructions are built in a single module, [`src/serverInstructions.ts`](../blob/w-23757366-server-instructions/src/serverInstructions.ts), and have two parts:

- **Always-on orientation** — the tool families plus the *discover → ground → query* pattern:
  > `list-datasources` / `search-content` → `get-datasource-metadata` → `query-datasource`

  with an explicit "always ground before querying — never guess field names".
- **Admin Insights taxonomy** — appended **only when the admin tools are enabled** — maps each `query-admin-insights` `kind` to the user intent it serves:

  | `kind` | Answers |
  | --- | --- |
  | `ts-events` | audit-event history (access / publish / update / delete) — "who did what, when" |
  | `ts-users` | per-user signals: last login, days since last login, Desktop / Prep / Web Authoring last-access |
  | `site-content` | content inventory: workbook & datasource metadata, ownership, size |
  | `job-performance` | extract-refresh & subscription execution history |
  | `stale-content` | unused-content report past a last-access threshold |

## Motivation and Context

Contributes to the "AI cannot determine which `query-admin-insights` `kind` to invoke" problem: with no server-level guidance, the model had to guess `kind` from the tool schema alone. The MCP `instructions` field is the protocol-sanctioned place to fix this once, server-wide, rather than bloating every tool description.

> **Design constraints honored:**
> - **Conditional inclusion** — orientation is always emitted; the Admin Insights block is appended only when `adminToolsEnabled`, so hosts without the admin tools don't carry guidance for a tool they never see.
> - **Single source of truth** — one module both server variants import, so web and combined emit identical text.
> - **Hot-path safe** — `WebMcpServer`'s constructor runs once per HTTP request, so it reads `process.env.ADMIN_TOOLS_ENABLED` directly instead of constructing a full `Config`.

**Variant coverage:** wired into the **web** (`tableau-mcp`) and **combined** (`tableau-combined-mcp`) servers. **Desktop is intentionally unaffected** — its local-file toolset is unrelated to this cloud/admin surface, and `adminToolsEnabled` is web-only.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

Full local pre-commit validation, all green:

```
npm run build      # exit 0
npx tsc --noEmit   # clean
npm run lint       # clean
npm test           # 174 files / 2788 tests passed
```

New tests:
- **`src/serverInstructions.test.ts`** (5) — orientation always present (both `adminToolsEnabled` states); admin taxonomy absent when disabled and present with all five `kind`s when enabled; `login` attributed to `ts-users` not `ts-events` (guards the taxonomy fix); admin variant is a strict superset of the base.
- **`src/server.web.test.ts`** (+3) — asserts `WebMcpServer` passes non-empty instructions carrying the grounding trio to the `McpServer` constructor, and that the admin block toggles on `ADMIN_TOOLS_ENABLED`.

## Related Issues

Tracked in GUS as **@W-23757366**.

## Checklist

- [x] I have bumped the version in `package.json` (4.5.2 → **4.6.0**, minor for a new feature)
- [x] I have added tests that prove my feature works
- [x] No breaking changes — the field was previously unset; behavior is purely additive

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.5.3",
+  "version": "4.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.5.3",
+      "version": "4.6.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1095.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.5.4",
+  "version": "4.6.0",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/index.combined.ts
+++ b/src/index.combined.ts
@@ -13,6 +13,7 @@ import { isNotificationLevel, notifier, setNotificationLevel } from './logging/n
 import { RestApi } from './sdks/tableau/restApi.js';
 import { DesktopMcpServer } from './server.desktop.js';
 import { WebMcpServer } from './server.web.js';
+import { buildServerInstructions } from './serverInstructions.js';
 
 const serverName = 'tableau-combined-mcp';
 const serverVersion = pkg.version;
@@ -61,6 +62,7 @@ async function startServer(): Promise<void> {
         logging: {},
         tools: {},
       },
+      instructions: buildServerInstructions({ adminToolsEnabled: config.adminToolsEnabled }),
     },
   );
 
--- a/src/server.ts
+++ b/src/server.ts
@@ -29,11 +29,13 @@ export abstract class Server {
     clientInfo,
     serverName,
     serverVersion,
+    instructions,
   }: {
     mcpServer?: McpServer;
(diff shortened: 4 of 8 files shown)


Akash-Rastogi:
[Akash's agent]

Closing as superseded. The MCP `initialize` instructions this PR adds have since landed on `main` via #840 (@W-23757364, "surface admin capabilities in MCP initialize instructions") and its follow-ups (#891, #863, #821, #904, #871).

That implementation is a superset of this one: the `instructions` constructor param is wired through both the web and combined server paths, plus dynamic `appendInstructions()`, a site-role-unavailable warning, per-registration-condition instructions, and a combined-path guard — none of which this branch carries. Rebased onto current `main`, every file here conflicts, all in the now-duplicated instructions wiring.

The only content not already on `main` is the "discover → ground on schema → query; always ground before querying, never guess field names" orientation nudge; the `query-admin-insights` `kind` intent-map is already documented in the tool's own description. If that nudge is worth keeping, it's a one-line follow-up to `BASE_INSTRUCTIONS`, not this branch.

PR #810, opened 2026-08-20
@W-23926841 - update validate-upload-and-publish-workbook tool adding support for twbx

takashibinns:
### Summary
The ```validate-upload-and-publish-workbook``` tool originally only supported publishing .twb files.  This has been updated to support both .twb and .twbx files.  There are now two workflows (depending on file type):

* .twb - This logic remained the same as before, the twb is uploaded to the [Validate Workbook and Upload](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_workbooks_and_views.htm#validate_workbook_and_upload) endpoint which either returns successfully or returns the validation issues that prevented it from publishing.
* .twbx - This is the new logic.  Since the validate endpoints do not support twbx file (or even just the twb from within a twbx [wont be able to find the included hyper files]), we are forced to skip the validation step.  Instead, we follow the [standard workbook publishing flow](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_publish.htm) (initiate file upload, append chunks, publish workbook).  There is still validation included (within the publishing endpoint), but the validation errors don't get returned.  So either the workbook publishes successfully, or the publish endpoint returns an [error code](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref.htm#publish_workbook:~:text=workbook%2Did-,Errors,-HTTP%20status)

### Testing

- npm test
- npm run lint
- npm run build


src/tools/web/workbooks/publishWorkbook.ts
```diff
@@ -176,6 +181,60 @@ export const getValidateUploadAndPublishWorkbookTool = (
...
+  | { status: 'valid'; warnings: ValidationFinding[]; uploadSessionId: string };
+
+async function validateAndUploadTwb({
+  restApi,
+  resolvedWorkbookFile,
+}: {
+  restApi: RestApi;
+  resolvedWorkbookFile: ResolvedWorkbook;
+}): Promise<ValidationOutcome> {
+  const validation = await restApi.workbooksMethods.validateWorkbookAndUpload({
```
jarhun88, line 193: why dont we just use a validate api here, and then twb path can reuse the same uploadFileInChunks path that the twbx uses so that we can support large twb as well

src/tools/web/workbooks/publishWorkbook.ts
```diff
@@ -84,7 +84,7 @@ export const getValidateUploadAndPublishWorkbookTool = (
     server,
     name: 'validate-upload-and-publish-workbook',
     description:
-      'Validates a TWB workbook with Tableau from a local file path or staged upload id, uploads it only when validation succeeds, and immediately publishes it to the specified Tableau project. Use list-projects to discover project IDs. If validation returns blocking errors, the tool returns those findings and does not publish anything.',
```
jarhun88, line 87: would you be able to add an e2e test with a twbx? tests/e2e/workbooks/validateUploadAndPublishWorkbook.test.ts is an example

src/tools/web/workbooks/validateUploadAndPublishWorkbook.ts
```diff
@@ -84,7 +84,7 @@ export const getValidateUploadAndPublishWorkbookTool = (
     server,
     name: 'validate-upload-and-publish-workbook',
```
jarhun88: i kinda wonder if we should just rename this tool and the file publishworkbook to be more specific to that action

src/tools/web/workbooks/publishWorkbook.ts
takashibinns: the validate api fails when the twb references a local (within the twbx) hyper file.  i tried that approach first, but Michael confirmed we can't use twb's from within the twbx with the validate api.

src/tools/web/workbooks/validateUploadAndPublishWorkbook.ts
takashibinns: yeah, the validate and upload parts are really implementation details.  the tool's job is simply to publish workbooks, so i'll rename the tool to "publish-workbook"

src/tools/web/workbooks/publishWorkbook.ts
jarhun88: sorry i mean:

twb path should just use validate api to validate, and then use the uploadTwbx (which we rename ti just upload?) to allow upload of large twb files

the twbx path should stay the same

takashibinns: i think that would be less efficient.  right now for twb, we send the twb to the validateAndPublish endpoint and if it fails validation it returns the reasons why (otherwise publishes it).  if we split this into two calls, we would need to send the twb twice (once for validation, again for publishing)

jarhun88: tradeoff is twb is capped at whatever the validateAndPublish size limit is, though maybe that is ok

takashibinns: i checked and apparrently there is no max file size on this endpoint. i suppose twbs can only get so large.  usually its the hyper extracts within .twbx that can make them significantly larger

jarhun88 (approved):

PR #811, opened 2026-08-21
fix(desktop): preserve resolved datasource during calc authoring

nawartamawi:
## Decision

Rule we now keep: bind-template uses the datasource resolved by its caller when authoring calculated fields. It must fail clearly when that datasource is unavailable instead of selecting another one.

## Description

- add an optional datasource parameter to bind-template
- pass the requested datasource to authorCalculationsInWorkbook
- cover multi-datasource calculation authoring and missing-datasource errors
- preserve existing single-datasource and no-calculation behavior

## Motivation and Context

Deterministic insight generation can resolve the correct workbook datasource, but that context was dropped during calculated-field authoring. In multi-datasource workbooks, MCP could therefore author the calculation in a different datasource.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

- scripts/agent-check: all five checks green
- 381 test files and 5,925 tests passed
- desktop build, lint, TypeScript, and lockstep hash gate passed

## Related Issues

- Tracked internally

## Checklist

- [x] Package version bumped to 2.63.14 with npm run version:patch
- [x] No documentation changes are required
- [x] Tests prove the fix is effective
- [x] New and existing unit tests pass locally
- [x] This change has no breaking behavior

## Contributor Agreement

- [x] I have read the CONTRIBUTING guidelines and followed the applicable contribution checklist.


tonythebest (approved):

mattcfilbert (approved):
lgtm

mattcfilbert (approved):
Clean, Nawar — and the write-up is exactly right (decision-first, states the rule). The new `datasource` param is funded by trimming existing describes, and the per-tool byte cap even ratchets *down* (2585 → 2576). The missing-datasource path returns `isError: true` with the candidate datasource names listed, so the agent can self-correct, and both the multi-datasource happy path and the absent-datasource error are covered. CI is green and it merges clean. Approving.

Non-blocking:

- **`ask` describe dropped "verbatim" ("Verbatim ask." → "Ask.").** When a proposal is supplied `ask` is ignored, so no impact on that path — but on the plain bind path the verbatim cue is load-bearing (binding keys on the user's exact words). The server instructions still say verbatim, so this may be fine; flagging in case the param-level cue was still earning its bytes.
- **`datasource` describe says "id/name" but `selectTargetDatasource` matches the workbook `<datasource name>` exactly** (case-sensitive, no caption fallback). It reads more flexibly than the code is — a caller passing a display caption instead of the internal name would land on the not-found path. Consider naming that in the describe.
- **No version bump.** Your checklist defers it to un-draft and the PR is open now. It's optional on this branch (#802 and #775 merged without one), so your call whether to run `version:patch` or batch it.

PR #812, opened 2026-08-21
fix(desktop): author parameters in place instead of reopening Desktop

myu404:
## Summary
`author-parameter` now creates the parameter in the **running** Tableau Desktop instance, the same way every other authoring tool applies edits. It no longer stages XML to disk, launches a second Desktop, and kills the original.

## What this fixes
**Old behavior (incorrect):** `author-parameter` was the only authoring tool that took a stage-and-reopen path. On every call it:
1. wrote the edited workbook XML to a `param-stage-*.twb` file on disk,
2. launched a **new** detached Tableau Desktop process from that file,
3. re-pinned the MCP session to the new process, then
4. sent `SIGTERM` to the user's **original** Desktop process.

So asking the agent to "create a parameter" spawned a brand-new Tableau window and closed the one the user was working in — discarding their window/selection state and surprising them mid-session.

**New behavior:** the parameter is created in place against the active instance (same PID). No stage file, no new window, no process termination — the user keeps working in the same instance and the parameter simply appears.

## Why the reopen existed, and why in-place works
A parameter does not materialize from a bare top-level `Parameters` datasource — the live model drops it, which is what led to the earlier "born only at reopen" workaround. It materializes through **dependency resolution**: the model reconstructs the `Parameters` datasource from a `<datasource-dependencies datasource='Parameters'>` block hung off a real datasource. Verified live against a running instance that this block alone is sufficient.

So the tool now injects that block into the first non-`Parameters` datasource and applies the document with the standard `loadWorkbookXml → applyWorkbookDocument → pollReadback` flow already shared by `author-calc`, `author-set`, and `author-action`. The result shape is now `{ applied: 'in-place', session, ... }`. A workbook with no datasource to host the block returns a clear, actionable error instead of silently reopening.

## Dead code cleanup
Dropping the reopen path removed a substantial amount of code (net −811 lines):
- **Deleted** `src/desktop/stageReopen.ts` and `stageReopen.test.ts` — this tool was their sole consumer.
- Removed from `authorParameter.ts`: `seedParameterColumn`, the stage-file write, reopen, session re-pin, `SIGTERM`, and `verifyReopenedParameter`.
- Removed the `stageReopen` startup-exception paragraph from `src/desktop/externalApi/README.md`.
- Dropped the now-dead `re-pinned` / `uniquely-named` guidance vocabulary and the `stagePath` tool-input surface from the guard tests.
- Exported `findDatasourceElements` / `DatasourceElement` from `authorCalcCore.ts` and reused them (opaque-range-safe XML scan) instead of a naive `indexOf`.

## Testing
- `tsc --noEmit` clean; `eslint` clean; full unit suite green.
- `authorParameter.test.ts` rewritten for the in-place flow.
- Verified end-to-end in a locally-built Tableau Desktop: creating a parameter materializes it in the same instance with the full list domain and all members — no new window, no closed window.


src/tools/desktop/authoring/datasource/authorParameter.ts
```diff
@@ -195,54 +145,42 @@ export const getAuthorParameterTool = (
...
-    }
-    if (!hasParameterCaption(readResult.value, caption)) {
-      return new ArgsValidationError(
-        `reopened workbook did not contain parameter caption ${caption}`,
-      ).toErr();
-    }
-    return Ok.EMPTY;
-  } catch (error) {
+function selectHostDatasource(xml: string): Result<DatasourceElement, ArgsValidationError> {
+  const host = findDatasourceElements(xml).find((datasource) => datasource.name !== 'Parameters');
```
tableaukyler: 🟠 `selectHostDatasource` silently picks the first non-`Parameters` datasource via `.find()` — no ambiguity check, and `author-parameter`'s schema has no `datasource` override to disambiguate.

Compare `selectTargetDatasource` in `authorCalcCore.ts` (imported into this very file for `findDatasourceElements`/`DatasourceElement`), which explicitly errors on 0 or 2+ candidates (`"Multiple datasources found; specify datasource"`) and supports a `requested` override. In a multi-datasource workbook this silently attaches the dependency block to whichever datasource happens to sort first, with no way for the caller to pick a different one. Worth matching the sibling convention rather than guessing.

src/tools/desktop/authoring/datasource/authorParameter.ts
```diff
@@ -195,54 +145,42 @@ export const getAuthorParameterTool = (
...
+    const openTag = xml.slice(host.openStart, host.openEnd).replace(/\/\s*>$/, '>');
+    const block = `${PARAM_DEP_OPEN}${columnXml}${PARAM_DEP_CLOSE}`;
+    return `${xml.slice(0, host.openStart)}${openTag}${block}</datasource>${xml.slice(host.openEnd)}`;
   }
-}
 
-function oneLineReason(error: unknown): string {
-  const message = error instanceof Error ? error.message : String(error);
-  return message.replace(/\s+/g, ' ').trim();
+  const existing = /<datasource-dependencies\b[^>]*\bdatasource=(['"])Parameters\1[^>]*>/.exec(
```
tableaukyler, line 150: 🟠 Two of `injectParameterDependency`'s three branches have zero test coverage:
- the self-closing-host branch (lines 165-168)
- this append-to-existing-`<datasource-dependencies>`-block branch (171-180)

The append branch is the realistic path the *second* time `author-parameter` is called against the same workbook — dashboards routinely need more than one parameter, and the tool's own knowledge doc instructs authoring parameters first (plural). Every fixture in `authorParameter.test.ts` only exercises the fresh-block-creation branch. Worth adding a test that calls the tool against a document whose host datasource already carries a `<datasource-dependencies datasource='Parameters'>` block, to pin the offset math in `absClose = host.openStart + closeRel`.

src/tools/desktop/authoring/datasource/authorParameter.ts
```diff
@@ -116,76 +91,51 @@ export const getAuthorParameterTool = (
...
-
-          const reopenResult = await reopenFromStage({
-            stagePath: trimmedStagePath,
-            oldPid: sessionResult.value,
-            discoveryDir,
+          const paramName = nextParameterName(liveXml);
+          const columnXml = renderParameterColumn({ caption, paramName, datatype, value, members });
+          const editedXml = injectParameterDependency(liveXml, hostResult.value, columnXml);
+
+          const loadResult = await loadWorkbookXml({
```
tableaukyler: 🟡 This `loadWorkbookXml -> pollReadback -> settled check` block is now a 4th near-identical copy of the same skeleton already in `authorCalcCore.ts`, `authorSet.ts`, and `authorAction.ts`. This PR already extracted `findDatasourceElements`/`DatasourceElement` for reuse — worth doing the same for this apply-and-verify skeleton in a follow-up (e.g. a shared `applyAndVerify(xml, baselineXml, settled)` helper) rather than adding a 4th copy. Not blocking.

tableaukyler:
## Summary

Replaces `author-parameter`'s stage-to-disk / relaunch-Desktop / kill-old-process flow with an in-place edit through the same `loadWorkbookXml -> applyWorkbookDocument -> pollReadback` pipeline the other `author-*` tools already use. Net -811 lines. Nice simplification, and reusing `findDatasourceElements` from `authorCalcCore.ts` instead of re-deriving it is the right call.

## 🔴 Blocking

**The agent-facing knowledge doc for this tool now describes a contract that doesn't exist.**

`resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md` is served to the agent at runtime via `expertise://tableau/tactics/data/dynamic-dashboard-authoring` (confirmed consumer: `src/desktop/guards/notionalSpecGuard.ts`, which points at this exact URI). This PR didn't touch it, and it still says (multiple places — lines [26](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L24-L28), [31](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L29-L33), [35](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L34-L36), [43-45](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L42-L46), [55-59](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L54-L60), [78](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L77-L79), [86-89](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dynamic-dashboard-authoring.md#L85-L90)) that `author-parameter` "reopens for you, in-call," returns `{ reopened: true, newSession }`, and takes an optional `stagePath` — none of that is true after this PR. A sibling doc, [`dialog-command-misclassification.md:25`](https://github.com/tableau/tableau-mcp/blob/b3795286f1bacfc6b920988ecd1af39175c4f1ee/resources/desktop/knowledge/tactics/data/dialog-command-misclassification.md#L23-L26), still says "reopen path" too.

Since this is knowledge the agent actually reads before calling the tool, it'll now expect a reopen, a `stagePath` param that no longer exists, and a `{reopened, newSession}` shape that no longer comes back. Please update both docs to describe the in-place contract (`{ applied: 'in-place', session }`, no `stagePath`).

## 🟠 Worth fixing

See inline comments on `authorParameter.ts` for the datasource-selection and test-coverage findings.

## 🟡 Nit

See inline comment on the apply/verify block re: Template Method duplication.

---

tableaukyler (approved):

src/tools/desktop/authoring/datasource/authorParameter.ts
myu404: Fixed in `4b52a87d`. Dropped the bespoke `selectHostDatasource` `.find()` and now route through `selectTargetDatasource` from `authorCalcCore.ts` — the same picker `author-calc`/`author-set`/`author-action` use. `author-parameter`'s schema gains the optional `datasource` param, so a multi-datasource workbook raises the same ambiguity error and honors an explicit selection exactly like the sibling verbs.

src/tools/desktop/authoring/datasource/authorParameter.ts
myu404: Fixed in `4b52a87d`. Added branch coverage for all three `injectParameterDependency` paths in `authorParameter.test.ts`: the self-closing-host branch, the append-into-existing-`<datasource-dependencies>`-block branch (the realistic second-call path you flagged), plus multi-datasource ambiguity and explicit `datasource` selection. `authorParameter.test.ts` is now 11/11.

src/tools/desktop/authoring/datasource/authorParameter.ts
myu404: Done in `4b52a87d`. Extracted the `loadWorkbookXml -> pollReadback -> settled` skeleton (with the fixed `restore` disposition and the whole-workbook baseline guard) into a single `applyAndVerify.ts` helper; `author-calc`/`author-set`/`author-action`/`author-parameter` now all route through it. Each verb keeps its own tool-specific "did not apply" wording via the discriminated `not-applied` outcome, which carries the readback (`not-applied` is now registered in `NON_TOOL_VOCABULARY`).

myu404:
The 🔴 knowledge-doc finding is addressed in `4b52a87d`. Both served docs now describe the in-place contract: `dynamic-dashboard-authoring.md` (lines 26/31/35/43/55-59/78/86-89) and `dialog-command-misclassification.md:25` drop `reopen`/`stagePath`/`{ reopened, newSession }` and document `{ applied: 'in-place', session }` plus the optional `datasource` host selector — so the `expertise://tableau/tactics/data/dynamic-dashboard-authoring` URI the agent reads at runtime now matches the shipped tool. Full vitest suite green post-merge, `tsc --noEmit` + lint clean.

PR #813, opened 2026-08-21
fix(desktop): give a self-clearing progress dialog its own poll-timeout guidance

lauren-jacksonSFDC:
## Summary

`progressWindows` was added to `OperationEnvelope` in the 0.2.9 contract sync (parses, passes the openapi-drift test) but nothing ever read it — unlike `blockingWindows`, which drives the `awaiting-user` fail-fast and its error message, `progressWindows` was dead weight on the wire. This PR wires it through so it's incorporated at the same level as `blockingWindows`:

- `pollOperation` now tracks the last-seen `progressWindows` during the poll loop (still never fails fast on it — self-clearing dialogs are meant to be waited out) and attaches it to a `poll-timeout` error if the deadline is hit.
- `mapClientError`'s `poll-timeout` case now gives accurate guidance when a progress dialog (not a blocking one) was last seen: it says the operation is most likely just slow, not wedged behind a dialog a person must dismiss, and drops the "do not retry" line that's specific to blocking dialogs.
- Generalized the `describeBlockingWindows` formatter to `describeWindows` since it's now shared between the `awaiting-user` and `poll-timeout` messages.

Rule going forward: a new field on `OperationEnvelope`/`ExternalApiError` isn't done until something reads it — schema-only additions that just satisfy the contract-drift test are incomplete.

## Test plan
- [x] `npx vitest run src/desktop/externalApi/externalApiHttpAsyncDispatch.test.ts src/desktop/externalApi/externalApiToolExecutor.test.ts src/desktop/externalApi/externalApiContract.test.ts` — new tests cover progressWindows not fail-fasting to success, `poll-timeout` carrying the last-seen progressWindows, and the tailored (non-blocking) timeout message
- [x] `npm test` — full unit suite (5926 tests) passes
- [x] `npm run lint` — clean
- [x] `npm run build:desktop` — builds

### DIFF
--- a/src/desktop/externalApi/externalApiToolExecutor.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.ts
@@ -953,7 +953,7 @@ function supportsOperationResult(apiVersion: string | undefined): boolean {
   return apiVersionAtLeast(apiVersion, '0.1.1');
 }
 
-function describeBlockingWindows(windows: Array<WindowInfo> | undefined): string {
+function describeWindows(windows: Array<WindowInfo> | undefined): string {
   if (windows === undefined || windows.length === 0) {
     return '';
   }
@@ -1012,7 +1012,7 @@ function mapClientError(
           code: 'awaiting-user',
           message:
             'The operation is blocked on a Tableau Desktop dialog and cannot complete over the API ' +
-            `until a person dismisses it.${describeBlockingWindows(error.blockingWindows)}`,
+            `until a person dismisses it.${describeWindows(error.blockingWindows)}`,
           recoverable: false,
         },
       };
@@ -1021,11 +1021,23 @@ function mapClientError(
         type: 'command-timed-out',
         error: `The async operation expired before it completed (polled past the Desktop retention window). ${BLOCKING_DIALOG_GUIDANCE}`,
       };
-    case 'poll-timeout':
+    case 'poll-timeout': {
+      const progressWindows = error.progressWindows;
+      if (progressWindows !== undefined && progressWindows.length > 0) {
+        return {
+          type: 'command-timed-out',
+          error:
+            'The async operation did not reach a terminal state within the poll deadline. Desktop was ' +
+            'last seen showing a self-clearing progress dialog rather than one a person must dismiss, ' +
+            'so this is most likely a slow operation rather than a wedged instance. If it keeps ' +
+            `happening, call list-instances to confirm the session is still reachable before retrying.${describeWindows(progressWindows)}`,
+        };
+      }
       return {
         type: 'command-timed-out',
         error: `The async operation did not reach a terminal state within the poll deadline. ${BLOCKING_DIALOG_GUIDANCE}`,
       };
+    }
     case 'no-instance':
       return {
         type: 'unknown',
--- a/src/desktop/externalApi/externalApiToolExecutor.ts
+++ b/src/desktop/externalApi/externalApiToolExecutor.ts
@@ -1027,10 +1027,10 @@ function mapClientError(
         return {
           type: 'command-timed-out',
           error:
-            'The async operation did not reach a terminal state within the poll deadline. Desktop was ' +
-            'last seen showing a self-clearing progress dialog rather than one a person must dismiss, ' +
-            'so this is most likely a slow operation rather than a wedged instance. If it keeps ' +
-            `happening, call list-instances to confirm the session is still reachable before retrying.${describeWindows(progressWindows)}`,
+            'The async operation did not reach a terminal state within the poll deadline. It may ' +
+            'still be running on Desktop — the poll gave up locally but did not cancel it, so only ' +
+            'retry if the command is safe to run twice. Desktop was last seen showing a self-clearing ' +
+            `progress dialog, so this is likely just slow; if it recurs, call list-instances to confirm the session is still reachable.${describeWindows(progressWindows)}`,
         };
       }
       return {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.66.1",
+  "version": "2.66.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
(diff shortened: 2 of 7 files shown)


src/desktop/externalApi/externalApiToolExecutor.ts
```diff
@@ -1013,11 +1013,23 @@ function mapClientError(
...
+          error:
+            'The async operation did not reach a terminal state within the poll deadline. Desktop was ' +
+            'last seen showing a self-clearing progress dialog rather than one a person must dismiss, ' +
+            'so this is most likely a slow operation rather than a wedged instance. If it keeps ' +
+            `happening, call list-instances to confirm the session is still reachable before retrying.${describeWindows(progressWindows)}`,
+        };
+      }
       return {
         type: 'command-timed-out',
         error: `The async operation did not reach a terminal state within the poll deadline. ${BLOCKING_DIALOG_GUIDANCE}`,
```
tableaukyler, line 1038: 🟠 This progress-timeout branch invites a retry ("...before retrying") but drops the "Do not retry this call" warning that `BLOCKING_DIALOG_GUIDANCE` carries for the generic case.

There's no in-flight/command lock on this path: `applyMutex.ts` only guards the document read-modify-write apply paths (`loadWorkbookXml.ts`/`loadWorksheetXml.ts`/`loadDashboardXml.ts`/`activateSheet.ts`), not `executeCommand`/`postJsonEnvelope`. A retry after a poll-timeout dispatches a brand-new `invokeCommand` rather than checking the prior operation — if the original op is still running server-side (our poller just gave up locally, it didn't cancel anything on Desktop) and the command isn't idempotent, an agent following this guidance could double-dispatch it. Worth a line acknowledging the original call may still be in flight, even if it's not the disk-write kind of "do not retry."

tableaukyler:
## Summary

Wires the already-schema'd `progressWindows` field through to `poll-timeout` handling: `pollOperation` tracks the last-seen value and `mapClientError` gives it distinct, less alarming guidance from a true blocking-dialog timeout. Small, mechanical, well-tested for the path it touches directly.

## 🔴 Blocking

**`get-summary-data` never sees the new message — the fix has zero effect for this tool.**

Traced end-to-end at head SHA:
1. `mapClientError`'s `poll-timeout` case does build the new progress-specific message, but it's still typed `{ type: 'command-timed-out' }`.
2. [`DesktopCommandExecutionError`'s constructor](https://github.com/tableau/tableau-mcp/blob/c0448e18f8ab862646b73489c0b2484cdd239f57/src/errors/mcpToolError.ts#L318-L323) sets `blockedByDesktopDialog` from `error.type === 'command-timed-out'` alone — true for both the old and new timeout variants, since neither changed the discriminant.
3. [`getSummaryData.ts`'s `requestError`](https://github.com/tableau/tableau-mcp/blob/c0448e18f8ab862646b73489c0b2484cdd239f57/src/tools/desktop/api/getSummaryData.ts#L329-L338) checks only `blockedByDesktopDialog` and, when true, substitutes the hardcoded `BLOCKING_DIALOG_GUIDANCE` string ("do not retry, dismiss any open dialog") — discarding whatever message `mapClientError` actually built.

So for `get-summary-data` specifically, a self-clearing progress-dialog timeout still surfaces "a person must dismiss a dialog, do not retry" — the exact message this PR exists to stop sending in that case. Worth either having `getSummaryData.ts` pass through `error.getErrorText()` instead of a hardcoded constant, or giving `blockedByDesktopDialog` visibility into which flavor of timeout it is.

## 🟠 Worth fixing

See inline comment on the `poll-timeout` branch re: the missing retry-safety guardrail.

## Note (not a review finding)

`build (>=22.7.5 <23)` is currently failing on an e2e "Connection closed" error in `tests/e2e/prompts/extractOptimization.test.ts`, unrelated to any file in this diff. Worth a re-run before merge to rule out flakiness rather than a real regression.

---

lauren-jacksonSFDC:

Re @tableaukyler's [comment](https://github.com/tableau/tableau-mcp/pull/813#issuecomment-5372576542) — good catch on the trace:

1. 🔴 Right at `c0448e18`, but stale at HEAD: the branch was reworked and that path is gone. `getSummaryData.ts` now goes through `runExternalApiReadTool`, and `readHarness.ts:53` surfaces the mapped error via `DesktopCommandExecutionError(...).getErrorText()` (your first suggested fix); nothing reads `blockedByDesktopDialog` in production anymore, so the tailored message reaches the client un-substituted.
2. 🟠 Fixed in `394bb50c` — see the inline reply.
3. e2e "Connection closed": agreed, looks unrelated to this diff; will re-run before merge.

Also bumped the version to `2.66.2`.

src/desktop/externalApi/externalApiToolExecutor.ts
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

Fair point — fixed in `394bb50c`. The message now flags that the original operation may still be running on Desktop (polling gave up locally without cancelling it) and to only retry a command that's safe to run twice, kept distinct from the blunt blocking-dialog "Do not retry". Added a test asserting the caveat.

nawartamawi (approved):
Reviewed current head. No blocking findings: progress-window context is carried through poll-timeout mapping, the message preserves the retry-safety caveat, and the focused tests and CI are green.

mattcfilbert (approved):
Nice fix, Lauren. Wiring `progressWindows` at the same level as `blockingWindows` — but waited out instead of fail-fast, with a timeout message that says "likely slow, retry only if safe" instead of "do not retry" — is exactly right. The rule you wrote down ("a schema field isn't done until something reads it") is a good one to keep. Tests cover the three real paths and CI is green on the head commit; I traced the last-seen tracking and the message branch by hand and both hold.

One thing before merge — a version collision, not a code issue:

- **#822 also bumps to `2.66.2`**, off the same `2.66.1` base as this PR. Two identical bumps merge with no git conflict and there's no version-increment check in CI, so whoever merges second lands their code under a duplicate `2.66.2` — two code states, one version, silently. Whichever of these two goes second should run `npm run version:patch` again (→ `2.66.3`). This PR's bump is correct if it merges first.

Approving the code.

—Posted by MattGPT on Matt's behalf.

PR #817, opened 2026-08-21
fix(templates): remove donor currency formats when binding fields

mattcfilbert:
Automatically bound templates must not impose the donor datasource's currency or locale on the target field. This is a rule for both template injection paths, not a one-off TBM cleanup.

When a validated field mapping targets an authored template field, the injection boundary removes donor currency and locale `text-format` overrides before rewriting fields. The target workbook's field format remains; without one, Tableau supplies a neutral default. Explicit formatting applied after the bind still wins. Neutral numeric formats and chart structure are unchanged.

Evidence:

- A live-shaped Superstore line projection binds Sales, preserves the target format, and contains no donor `£`.
- A corpus invariant covers every pass-1 eligible template.
- Empty and unmatched mappings remain byte-identical.
- Both the `bind-template` injection path and the build-and-apply path use the same rule.
- `scripts/agent-check` is green.
- Live Desktop apply/readback is UNKNOWN because the open Studio instance is not serving this branch.

Approving this PR means automatically bound templates can no longer impose their donor datasource's currency or locale on the target field.

Authored by MattGPT.


src/desktop/templates/tbmEngine.corpus.test.ts
```diff
@@ -202,6 +205,31 @@ function calcColumnIdentities(xml: string): string[] {
     .filter((name): name is string => !!name);
 }
 
+function isDonorCurrencyOrLocaleFormat(value: string): boolean {
```
myu404, line 208: 🤖 **Reviewed by MichaelGPT**

The corpus invariant re-declares the production regex inline (`isDonorCurrencyOrLocaleFormat` here is byte-identical to `DONOR_CURRENCY_OR_LOCALE_FORMAT` in `injectTemplateCore.ts:38`) rather than importing it. The leak check at line 462 (`donorCurrencyOrLocaleFormats(rewritten)`) then detects leakage with the *same* copied regex the production code strips with — so any under-match in production is mirrored by the detector and the invariant passes silently, and the test drifts if the production regex later changes.

Suggest exporting `DONOR_CURRENCY_OR_LOCALE_FORMAT` (and ideally the strip predicate) and importing it, so the invariant is anchored to the code under test rather than a copy.

src/desktop/templates/injectTemplateCore.ts
```diff
@@ -33,6 +35,36 @@ import { type OptionalFieldPruneSpec, pruneUnboundOptionalFields } from './optio
...
+): string {
+  if (!fieldMapping || Object.keys(fieldMapping).length === 0) return templateXml;
+  const rewritesAuthoredField = Object.keys(fieldMapping).some((mappingKey) => {
+    const field = mappingKey.replace(/@[A-Za-z][A-Za-z0-9-]*$/, '');
+    const escaped = field.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    return new RegExp(`\\[${escaped}\\]|:${escaped}:`).test(templateXml);
+  });
+  if (!rewritesAuthoredField) return templateXml;
+  const document = new DOMParser().parseFromString(templateXml, 'text/xml');
+  for (const format of Array.from(
```
myu404, line 53: 🤖 **Reviewed by MichaelGPT**

Once the gate passes, stripping is **template-wide**, not scoped to the bound field. The gate (lines 45–51) only checks that *some* mapping key appears in the XML, then this loop removes *every* matching `text-format` node in the document with no `field=` scoping — so a currency/locale format on an authored field that stays unbound (its donor field never rewritten) is also removed.

This looks by-design (the line-39 comment scopes it to "the automatic field-binding path," and curated templates bind all measures), so impact is low — just confirming it's intended. If per-field scoping is ever wanted, match each `<format>` node's `field=` attribute against the mapping keys before removing.

src/desktop/templates/injectTemplateCore.ts
```diff
@@ -33,6 +35,36 @@ import { type OptionalFieldPruneSpec, pruneUnboundOptionalFields } from './optio
 import { spliceWaterfallAnchorFilter } from './waterfallAnchorFilter.js';
 
 const AUTHORED_TITLE_SENTINEL = '__TABLEAU_MCP_AUTHORED_TITLE_4D9E7A1B__';
+const DONOR_CURRENCY_OR_LOCALE_FORMAT = /[$£€¥₹₩₽₺₫₪₴₦₱฿₡₲₵₭₮₸₼₾₿]|\[\$-[^\]]+\]|^c(?=["#0*])/u;
```
myu404, line 38: 🤖 **Reviewed by MichaelGPT**

The currency-glyph class omits the generic currency sign `¤` (U+00A4). A donor `text-format` using `¤` without a `[$-…]` locale code and not starting with `c` would slip through the strip. Speculative — no such value exists in the current corpus, and the `^c(?=["#0*])` and `[$-…]` branches cover locale-default currency in practice — so low priority. Optional: add `¤` to the glyph class for completeness.

myu404 (approved):

Clean, well-scoped fix that delivers its one stated decision: at both template-injection boundaries — `injectTemplateCore.buildInjectedWorkbookXml` (line 414) and `buildAndApplyWorksheet` (line 417) — it calls the **same exported** `stripDonorCurrencyOrLocaleFormats` with the same `fieldMapping`, after `spliceBoundFacet` and before field rewriting. Genuine path parity, no divergent copy.

I verified the `DONOR_CURRENCY_OR_LOCALE_FORMAT` regex against the distinct `text-format` values in the template corpus: it strips the `£`/locale-coded currency formats and preserves neutral numerics (`n#,##0.00`), percentages, and date/custom formats — no over-strip of neutral formats or chart structure. Only whole `<format attr="text-format">` nodes are removed, and empty/unmatched mappings return early so the template stays byte-identical. The new puppet test proves the target's default format survives while the donor `£` is dropped, and the corpus invariant exercises the gate for every eligible template.

No blocking issues. Three non-blocking observations inline (one `minor` test-gap, one `question`, one `nit`). Note the PR's own caveat: live Desktop apply/readback is UNKNOWN because the open Studio instance isn't serving this branch.

PR #818, opened 2026-08-21
feat(desktop): build executive dashboards with KPI layouts

mattcfilbert:
This adds one rule to the main dashboard batch path: when a request identifies KPI worksheets, place them in a top strip and put the analytical views below. This is a reusable executive layout, not a one-off for the three Steel Thread 2 dashboards.

It also fixes three first-pass quality problems from Zach's runs:

- KPI values use a large, bold label while retaining the workbook's font and color.
- Dashboard titles no longer force Tableau blue or Tableau Semibold.
- An incomplete grid row expands to use the full dashboard instead of leaving a dead half-cell.

The path fails before writing if KPI names are missing, duplicated, or do not match the worksheets in the batch.

Verified live in the dedicated Superstore `st2-scratch` workbook: one batch created two KPI sheets, two analytical views, and the executive dashboard; same-process readback confirmed the zones, 36pt KPI labels, worksheet membership, and KPI values. This does not claim visual correctness. Image-based correction and broader workbook styling remain separate follow-ups.

Approving this means Studio can use `executive-summary` as its first-pass layout for KPI-led dashboards. Future work should preserve named semantic roles and let the workbook own font and color choices.


src/tools/desktop/authoring/sheets/composeDashboardCore.ts
```diff
@@ -69,10 +72,24 @@ export function buildDashboardCandidateXml({
   title,
   layout,
 }: BuildDashboardCandidateXmlArgs): string {
+  const inputError = validateComposeDashboardInput(canonicalWorksheetNames, layout);
+  if (inputError) throw inputError;
+  const kpis =
+    layout?.layoutType === 'executive-summary'
+      ? canonicalWorksheetNames.filter((name) =>
```
myu404, line 79: **KPI strip order follows worksheet order, not caller order** (minor · correctness/UX).

`kpis` is built by filtering `canonicalWorksheetNames` for membership in `layout.kpiWorksheetNames`, so the left-to-right strip order (`x: i * kpiWidth` in `computeZones`) follows the resolved worksheet order, not the order the caller listed in `kpiWorksheetNames`. Passing `['Revenue KPI', 'Orders KPI']` while the worksheet list resolves as `['Orders KPI', 'Revenue KPI', …]` yields a strip ordered Orders, Revenue.

This violates no stated requirement, so it's not blocking — but it's a plausible surprise. Either iterate `layout.kpiWorksheetNames` and match into `canonicalWorksheetNames` to preserve caller order, or document that strip order follows the worksheet list.

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/composeDashboardCore.ts
```diff
@@ -69,10 +72,24 @@ export function buildDashboardCandidateXml({
...
+  const charts = canonicalWorksheetNames.filter(
+    (name) => !kpis.some((kpiName) => xmlNamesEqual(kpiName, name)),
+  );
   const zones = computeZones(title, {
-    kpis: [],
-    charts: canonicalWorksheetNames,
-    layoutType: layout?.layoutType ?? 'auto-grid',
+    kpis,
+    charts,
+    layoutType:
```
myu404, line 89: **Duplicated `executive-summary` condition** (nit · simplification).

The `layout?.layoutType === 'executive-summary'` check appears twice in `buildDashboardCandidateXml` — once to compute `kpis` (above) and once here to override the `computeZones` `layoutType`. Hoisting a single `const isExecutiveSummary = layout?.layoutType === 'executive-summary';` reads cleaner and keeps the two checks from drifting apart if the layout union grows.

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/composeDashboardCore.ts
```diff
@@ -396,10 +414,53 @@ export function validateComposeDashboardInput(
...
+    return new ArgsValidationError(
+      'gridColumns cannot be combined with layoutType "executive-summary".',
+    );
+  }
+  if (requestedKpis.length < 1) {
+    return new ArgsValidationError(
+      'layoutType "executive-summary" requires at least one KPI worksheet name.',
+    );
+  }
+  const duplicateKpis = requestedKpis.filter(
```
myu404, line 439: **New validation branches are untested** (minor · test-coverage).

Three new branches added by this PR have no test: duplicate KPI names (this block), zero KPI names for executive-summary (`"requires at least one KPI worksheet name"`), and the all-worksheets-are-KPIs guard (`"requires at least one non-KPI worksheet"`). Only the unknown-KPI-name and `gridColumns`+executive-summary conflict branches are exercised. A small validation test pinning each error message (and confirming it fires before any write) would close the gap.

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/composeDashboardCore.test.ts
```diff
@@ -56,6 +56,22 @@ describe('buildDashboardCandidateXml', () => {
...
+
+  it('places named KPI worksheets in the executive-summary strip', () => {
+    const candidateXml = buildDashboardCandidateXml({
+      baselineXml: PRISTINE,
+      dashboardName: 'Sales Dashboard',
+      canonicalWorksheetNames: ['Sales', 'Profit'],
+      title: 'Executive Overview',
+      layout: {
+        layoutType: 'executive-summary',
+        kpiWorksheetNames: ['Sales'],
```
myu404, line 68: **No multi-KPI width-division test** (minor · test-coverage).

Every executive-summary test uses exactly one `kpiWorksheetNames` entry, so the multi-KPI strip division (`kpiWidth = Math.floor(100000 / kpis.length)`, `x = i * kpiWidth`) is never asserted. A case with 2+ KPIs asserting the second KPI's `x`/`w` would cover it. (The title + executive-summary interaction is already covered here via `title: 'Executive Overview'`, so that part is fine.)

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/runDashboardBatch.ts
```diff
@@ -62,20 +62,23 @@ import {
...
   resolveRenderedWorksheetNames,
+  validateComposeDashboardInput,
 } from './composeDashboardCore.js';
 
 const paramsSchema = {
   session: sessionParam({ max: 64 }),
-  artifactIds: z.array(z.string().trim().min(1).max(255)).max(6).optional().describe('IDs.'),
-  dashboardName: z.string().trim().min(1).max(255).describe('Name.'),
-  existingWorksheetNames: z
+  artifactIds: z.array(z.string().trim().min(1).max(255)).max(6).optional(),
```
myu404, line 70: **`.describe()` removed wholesale from six params** (minor · tool-surface usability).

To make byte-budget room for `kpiWorksheetNames`, this PR removes `.describe()` entirely from `artifactIds`, `dashboardName`, `existingWorksheetNames`, `title`, `layoutType`, and `gridColumns` (and shrinks the tool description to `'Apply; compose.'`). The re-pin comment calls this "trimming its older field prose," which understates a full removal. The prior strings were already near-useless (`'IDs.'`, `'Name.'`), so the practical loss is small — but on the highest-risk authoring tool it's worth keeping a couple of meaningful words on params like `existingWorksheetNames`/`layoutType` rather than dropping them outright. Not blocking.

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/dashboardZones.ts
```diff
@@ -180,14 +180,19 @@ export function computeZones(titleText: string | undefined, layoutSpec: LayoutSp
       const chartWidth = Math.floor(100000 / cols);
       const chartHeight = Math.floor(chartAreaHeight / rows);
       for (let i = 0; i < charts.length; i++) {
+        const row = Math.floor(i / cols);
+        const isLastRow = row === rows - 1;
+        const lastRowCount = charts.length - row * cols;
+        const rowColumns = isLastRow ? Math.min(cols, lastRowCount) : cols;
+        const rowChartWidth = isLastRow ? Math.floor(100000 / rowColumns) : chartWidth;
```
myu404, line 187: 👏 **Clean, surgical, and provably correct** (praise).

The per-row `rowColumns`/`rowChartWidth` recompute with `x` derived from `i % cols` (not `i % rowColumns`) keeps the intra-row index in range for a short last row and avoids overlap/gaps, while leaving full rows byte-identical to before. Paired with the conditional font attribute omission for the title zone, this reuses existing machinery instead of adding branches — and both are backed by focused assertions (`{x:0, y:50000, w:100000}` for N=3, and the explicit no-`fontcolor`/`fontname` title test). Verified the arithmetic independently; it holds.

---
🤖 Reviewed by MichaelGPT

src/tools/desktop/authoring/sheets/runDashboardBatch.test.ts
```diff
@@ -299,6 +309,65 @@ describe('runDashboardBatchTool', () => {
...
+      existingWorksheetNames: ['Existing A', 'Existing B'],
+      layoutType: 'executive-summary',
+      kpiWorksheetNames: ['Existing A'],
+      gridColumns: 2,
+      executor: conflictHarness.executor,
+    });
+
+    expect(bodyOf(missing)).toMatchObject({ applied: false, retrySafe: true });
+    expect(bodyOf(conflict)).toMatchObject({ applied: false, retrySafe: true });
+    expect(missingHarness.posts).toHaveLength(0);
```
myu404, line 367: 👏 **Proves "no write," not just "returns error"** (praise · test-coverage).

Asserting `posts.toHaveLength(0)` for both the unknown-KPI and `gridColumns`-conflict cases directly proves zero workbook mutations occurred — materially stronger evidence of the fail-before-write requirement than checking the returned error alone.

---
🤖 Reviewed by MichaelGPT

myu404 (commented):
## Review — executive-summary KPI layouts

Solid, well-scoped change that does what the description claims. The `executive-summary` layout is implemented as a thin layer over the existing `computeZones`/`buildZoneXml` machinery (populate the `kpis`/`charts` arrays rather than add a new layout branch), which keeps the surface area small.

**Correctness verified.** I hand-checked the two highest-risk areas and both hold:
- The incomplete-last-row auto-grid math is correct for N=3/4/5 and `gridColumns=3/4`. Because last-row indices satisfy `i % cols == i - row*cols ∈ [0, rowColumns)`, the per-row `rowChartWidth` tiles each short row with no overlap or gap, and full rows are provably unaffected (`rowColumns === cols`).
- The `kpi-text.tbm` `customized-label` reference is rewritten generically — proven end-to-end by the new `fieldReferenceRewriter` test (maps `{{field_base_1}}` to the target datasource, asserts the CDATA-wrapped rewritten ref, and asserts the placeholder is gone).

**Fail-before-write is genuinely enforced**, and the tests prove it structurally (`posts.toHaveLength(0)`), not just by inspecting the returned error.

**Two things I checked and am NOT flagging:**
- The three-site validation (`run-dashboard-batch` preflight, `composeDashboardCore`, and inside `buildDashboardCandidateXml`) is harmless defense-in-depth — the preflight returns a clean failure before the `throw` inside `buildDashboardCandidateXml` can fire in the batch path.
- The byte-budget re-pin is not fragile: `DYNAMIC_AUTHORING_SURFACE_EXPECTED` is an exact `.toBe()` pin, so every surface change forces a re-pin regardless; the 3-byte gap to `BUDGET` is a deliberate near-budget signal and there is plenty of headroom to the product ceiling.

**Assessment: no blockers, no majors — recommend approve.** The inline notes are minors and nits; the largest is a documentation/usability tradeoff on `run-dashboard-batch`'s schema. Details inline.

---
🤖 Reviewed by MichaelGPT

myu404 (approved):

PR #821, opened 2026-08-21
@W-23955385: Add optional required role property to tools for conditional registration

stephendeoca:

the pull request._

# Pull Request Template

## Description

Adds an optional tool property called `minRequiredRole` that allows tools to specify the minimum user role required for it to be registered. The logic is gated by the `enforce-role-requirements` feature flag. When enabled, we fetch the user's site role during tool registration and only register tools the user should have access to. If the fetch fails, we omit the tools that specify role requirements. If a tool does not specify a role requirement it is registered regardless of a user's role.

## Motivation and Context

There are many tools, such as admin tools, that might not be usable if the user does not have the proper license / permissions. We don't want those tools to create a poor user experience by bloating their tool list with tools that don't work for them.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Version Bump

- [ ] Patch
- [x] Minor
- [ ] Major

## How Has This Been Tested?

Unit tests, manual testing of Admin tools with different permission sets and validating tools were 

## Related Issues

N/A

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/adminInsights/queryAdminInsights.ts
```diff
@@ -6,6 +6,7 @@ import { getConfig } from '../../../config.js';
 import { AdminOnlyError, ArgsValidationError } from '../../../errors/mcpToolError.js';
 import { useRestApi } from '../../../restApiInstance.js';
 import { querySchema } from '../../../sdks/tableau/apis/vizqlDataServiceApi.js';
+import { ADMIN_SITE_ROLES } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: Having this requirement in tool level can be cumbersome to enforce and manged. The tools are grouped (as they should), hence can we have this defined there.. 
like under: https://github.com/tableau/tableau-mcp/tree/main/src/tools/web
Where each tool group/folder will have the top level definition and for granularity purpose in tool definition we can have an override.

src/sdks/tableau/types/user.ts
```diff
@@ -20,15 +20,15 @@ export const userSchema = z.object({
 
 export type User = z.infer<typeof userSchema>;
 
-export const ADMIN_SITE_ROLES = new Set([
+export const ADMIN_SITE_ROLES: readonly string[] = [
```
Alon-ST-DATA, line 23: So this requires a list of explicit roles to be defined that can be a long list (like for list-views or publish WB/DS), hence may I suggest to use similar logic we use in [GLSI](https://help.tableau.com/current/server/en-us/grant_role.htm) (minimum site role) , where in db schema we have ranking for each role to define the licensing consumption of it. In db schema we have [site_roles table](https://tableau.github.io/tableau-data-dictionary/2026.2/data_dictionary.htm#public.site_roles_anchor) and in monolith an [SiteRole](https://github.com/sf-analyticscloud/monolith/blob/main/workgroup/src/shared/interfaces/tab-domain-interfaces/src/com/tableausoftware/domain/user/SiteRole.java#L11) enum which maps to site_role identifier. So following that we can have similiar enum for all site roles with [capabilities rank  ](https://help.tableau.com/current/server/en-us/users_site_roles.htm#general-capabilities-allowed-with-each-site-role) include the special case of supportUser which has essentially SiteAdminCreator capability.

With that logic service owner can define only the minimum site role for the tool set/name, e.g: for publishing tool it will be **ExplorerCanPublish**

src/tools/web/_lib/deleteContent.ts
```diff
@@ -20,6 +20,7 @@ import {
   getDatasourceDownstreamQuery,
 } from '../../../sdks/tableau/methods/lineageUtils.js';
 import { RestApi } from '../../../sdks/tableau/restApi.js';
+import { ADMIN_SITE_ROLES } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: **_Incorrect_** asset owners like  publishers / creators can own content and delete it.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -6,6 +6,7 @@ import { getConfig } from '../../../config.js';
 import { BoundedContext } from '../../../overridableConfig.js';
 import { useRestApi } from '../../../restApiInstance.js';
 import { ExtractRefreshTask } from '../../../sdks/tableau/types/extractRefreshTask.js';
+import { ADMIN_SITE_ROLES } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: Incorrect. Users with an Explorer (can publish) site role or higher can be granted the Extract Refresh capability to trigger updates on demand. Or to list/see it.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
```diff
@@ -12,6 +12,7 @@ import {
   UpdateCloudExtractRefreshSchedule,
   updateCloudExtractRefreshScheduleSchema,
 } from '../../../sdks/tableau/types/extractRefreshTask.js';
+import { ADMIN_SITE_ROLES } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: Incorrect. Authorized users can update, modify, or change how an extract refresh operates.

src/tools/web/users/listUsers.ts
```diff
@@ -4,7 +4,7 @@ import { z } from 'zod';
 
 import { getConfig } from '../../../config.js';
 import { useRestApi } from '../../../restApiInstance.js';
-import { User } from '../../../sdks/tableau/types/user.js';
+import { ADMIN_SITE_ROLES, User } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: Partially correct. Users can fetch their self information.

src/tools/web/users/updateUser.ts
```diff
@@ -5,6 +5,7 @@ import { z } from 'zod';
 import { getConfig } from '../../../config.js';
 import { UnknownError } from '../../../errors/mcpToolError.js';
 import { useRestApi } from '../../../restApiInstance.js';
+import { ADMIN_SITE_ROLES } from '../../../sdks/tableau/types/user.js';
```
Alon-ST-DATA: Incorrect. Non admins can update their email.

src/tools/web/adminGate.test.ts
```diff
@@ -59,3 +71,123 @@ describe('assertAdmin', () => {
...
+      signal: new AbortController().signal,
+      disableLogging: true,
+    };
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  function stubUseRestApiWithSiteRole(siteRole: string | undefined): void {
```
Alon-ST-DATA, line 129: For test coverage, we should ensure that every new or existing tool has the required site roles—or at least a minimum site role—defined, either at the group level or for the individual tool. IMO, we should have a high-level test that scans the tool folders/groups, verifies that a site-role configuration exists, and fails if it is missing.

Given the granularity and complexity of some of these use cases, I also think we should expect API owners to provide E2E tests that cover the relevant scenarios.

src/tools/web/adminGate.ts
```diff
@@ -33,6 +25,22 @@ function getCache(): ExpiringMap<string, string> {
...
+): Promise<string> {
+  const cacheKey = `${siteId}:${userId}`;
+  const adminCache = getCache();
+  const cached = adminCache.get(cacheKey);
+  if (cached !== undefined) {
+    return cached;
+  }
+  const siteRole = await fetchRole();
+  adminCache.set(cacheKey, siteRole);
+  return siteRole;
```
Alon-ST-DATA: We need to make sure site role is fetched in a comparable format of the site roles set.. see we have role display name and identifier (that has no spaces/braces). Ideally we should use the site role identifier.

src/tools/web/adminGate.ts
```diff
@@ -50,15 +58,10 @@ export async function assertAdmin(
...
-  const adminCache = getCache();
-  let siteRole = adminCache.get(cacheKey);
-
-  if (!siteRole) {
+  const siteRole = await resolveSiteRole(siteId, userId, async () => {
     const user = await restApi.usersMethods.queryUserOnSite({ siteId, userId });
-    siteRole = user.siteRole ?? '';
-    adminCache.set(cacheKey, siteRole);
-  }
+    return user.siteRole ?? '';
```
Alon-ST-DATA: Shall we add here  retry logic here and logging around the site role resolution failure if occurs.

src/server.web.test.ts
```diff
@@ -361,6 +371,77 @@ describe('server', () => {
...
+      annotations: {
+        title: 'Mock Admin Tool',
+        readOnlyHint: true,
+        destructiveHint: false,
+        idempotentHint: true,
+        openWorldHint: false,
+      },
+      callback: vi.fn(),
+      disabled: false,
+      requiredRoles: ['SiteAdministratorCreator', 'SiteAdministratorExplorer'],
```
Alon-ST-DATA: technically SupportUser (that has SiteAdminCreator capability in cloud)  should be added here.
And also "ServerAdministrator" for the case of on-prem tableau server where I believe self-hosted MCP is supported.

src/server.web.ts
```diff
@@ -130,11 +136,31 @@ export class WebMcpServer extends Server {
...
+
+    // Fetched lazily so we only issue the /users/{userId} call when at least one candidate tool
+    // declares `requiredRoles`. Fail-closed: `getCurrentUserSiteRole` returns `undefined` on any
+    // error, and an undefined role is never in a `requiredRoles` list, so the tool stays hidden.
+    // Cached as a sentinel-holding object so a legitimately-undefined role isn't re-fetched every
+    // iteration.
+    let cachedSiteRole: { value: string | undefined } | undefined;
+    const resolveSiteRole = async (): Promise<string | undefined> => {
+      if (cachedSiteRole === undefined) {
+        cachedSiteRole = {
```
Alon-ST-DATA: What is the cache invalidation logic here. Might be an edge case, but still we should be diligent and not create a security risk when users are demoted during existing session...

src/tools/web/adminGate.ts
Alon-ST-DATA: @stephendeoca I confirmed with live session we are **good** here and the standard site role identifier (below)

` so how the Site role field will look like for: Explorer(Can Publish) site role?                                                                                                                                                                

⏺ REST API siteRole = ExplorerCanPublish (PascalCase, no spaces/parens).

  Confirmed from live data this session — list-users returned:
  {"name":"emily.tableau+strawberry1@gmail.com","siteRole":"ExplorerCanPublish"}
  UI label "Explorer (Can Publish)" → API string ExplorerCanPublish. Same enum used by update-user tool schema.

  So resolveSiteRole() yields the literal string "ExplorerCanPublish" for that user. Note: it's not in tableau-mcp's admin allow-list (adminGate accepts only SiteAdministratorCreator / SiteAdministratorExplorer / ServerAdministrator) — an
  ExplorerCanPublish user fails the gate.

✻ Sautéed for 8s

❯ and Site Administrat Creator                                                                                                                                                                                                                   

⏺ REST API siteRole = SiteAdministratorCreator (PascalCase, no spaces).
  
  Confirmed live this session — list-users returned:
  {"name":"yxing@salesforce.com","siteRole":"SiteAdministratorCreator"}
  UI label "Site Administrator - Creator" → API string SiteAdministratorCreator. This one passes the adminGate allow-list (SiteAdministratorCreator / SiteAdministratorExplorer / ServerAdministrator).
`

src/server.web.ts
Alon-ST-DATA: @stephendeoca in main branch we are blocked as soon as the site role is demoted (from SiteAdmin to explorer) because the tmcp call request the proper access, see below. So we are **good** here as well :) 

`❯ list all tasks on my sites                                                                                                                                                                                                                     

  Called tableau-mcp-local (ctrl+o to expand)
  
⏺ Blocked. list-extract-refresh-tasks needs site admin. Your role: Explorer.

  Same wall for list-jobs (also admin + ADMIN_TOOLS_ENABLED flag). Cannot list tasks/jobs at Explorer.

  Need admin role bump, or admin runs it.`

stephendeoca: In that case I think we'd be in the clear to remove the SiteRole from the cache right? We only need the siteRole once during tool registration, if the tool was registered it most likely means that the siteRole matches what we wanted. There is the edge case where the tool was registered and we no longer have admin, but the REST API will fail. If we looked at the cache it'd still say admin which would be incorrect too.

src/tools/web/adminInsights/queryAdminInsights.ts
stephendeoca: Tool groups are mainly for specifying a set of tools via `INCLUDE_TOOLS`, when defining tools there's no concept that inherit default properties from the group. I don't think adding this would make sense on a generic level since some tools under the same group can be quite different from each other.

src/sdks/tableau/types/user.ts
stephendeoca: I added minSiteRole and a hierarchy map that assigns roles a numeric value for comparisons. Please take a look

src/server.web.test.ts
stephendeoca: SupportUser role but its not documented as a response value for the REST API. I think that's fine to not have it, I don't expect our Support team to use TabMCP for admin tools anyways

src/tools/web/adminGate.test.ts
stephendeoca: Right now I have the property as optional because I implement lazy fetching for the site role during tool registration. We only fetch site role if we are trying to register a tool that requires a specific role. If there are no tools that require a role being registered then we'll never fetch the site role. Not sure if we'll ever introduce tools that don't call our REST API either. We can always require it in the future, but we'll at least make sure to ask contributors to assess role requirements when adding in a new tool

src/tools/web/adminGate.ts
stephendeoca: We do lazy fetching, so its only getting called if a tool that is being registered for the user requires it. We got 3 options on failure: error, omit the tool, or allow the tool and then let the REST API block access

src/sdks/tableau/types/user.ts
```diff
@@ -20,15 +20,50 @@ export const userSchema = z.object({
...
+export const SITE_ROLE_HIERARCHY = {
+  Unlicensed: 0,
+  Viewer: 1,
+  Explorer: 2,
+  ExplorerCanPublish: 3,
+  Creator: 4,
+  SiteAdministratorExplorer: 5,
+  SiteAdministratorCreator: 6,
+  ServerAdministrator: 7,
+} as const satisfies Record<string, number>;
```
stephendeoca, line 55: The REST API also returns ReadOnly, do we care about including it in the list? Same with Unlicensed, I don't think we'll ever use this. Should I remove it.

src/tools/web/adminGate.ts
stephendeoca: I think with Principle of Least Privilege we should omit the tool, but then since we're adding instructions when you connect we can append a message saying that tools were omitted because we could not determine site role. So at least the LLM has context on why there are some tools missing if the user asks.

src/tools/web/_lib/deleteContent.ts
stephendeoca: I updated the delete content and other tools to specify minRequiredRole = ExplorerCanPublish. I also removed minRequiredRole for the listUsers and updateUser tool. I believe technically these tools still call assertAdmin. Should I leave that for your team to remove in a follow up PR?

src/sdks/tableau/types/user.ts
Alon-ST-DATA: afaik, SupportUser should be another role we should support. Going to double and reply here.

Alon-ST-DATA: Verified that supportUser is the role returned.

⏺ ## User: echweh@tableau.com
  
  | Field | Value |
  |-------|-------|
  | Full name | Emily Chweh |
  | Site role | SupportUser |
  | ID (LUID) | 306c3200-3f13-4e31-ab9d-dd11903ec7dd |
  | Username | echweh@tableau.com |
  | Last login | 2025-12-12T23:29:36Z |

  > **Note:** `SupportUser` is the Tableau technical-support role — not a license-consuming site role.

src/tools/web/adminInsights/queryAdminInsights.ts
Alon-ST-DATA: _"when defining tools there's no concept that inherit default properties from the group."_
I'm aware of that and perhaps it is not the right time of doing, however I think at some point (especially when the amount of tools will increase), it might be, beneficial to think about better tools grouping/hierarchy where can support capabilities inheritance like we have in Table site for permission in different levels:
Project --> Group --> User

For sure can be differed for later time.

src/tools/web/adminGate.ts
Alon-ST-DATA: Sounds good. Explicit messaging in erroneous cases is always appreciated by customers.

src/server.web.test.ts
Alon-ST-DATA: It is not documented because it is an internal role used by support.
I see. Fair point, they will probably not be allowed to access TMCP... but I thought in case...


Low pri to figure this.

Alon-ST-DATA (commented):
[Alon's agent]

Second-round review (head `818d897`) — fanned out a code review + a full e2e pass on two roles. Mechanism-wise the role-gating core is solid (hierarchy + `siteRoleMeetsMinimum`, fail-closed resolution, defense-in-depth over the existing `assertAdmin`/`guardMutation`, identifier-format comparison). But two blockers must clear before merge, and this head is stale versus your own replies below.

**BLOCKER 1 — admin tools are dropped for admins on stdio and http+`DANGEROUSLY_DISABLE_OAUTH`.** `registerTools()` only receives `authInfo` on the http-OAuth path (`src/server/express.ts:200`). Stdio (`src/index.ts:61`), combined (`src/index.combined.ts:68`), and http with OAuth disabled all call `registerTools()` with no arg → `getCurrentUserSiteRole` returns `undefined` (`adminGate.ts:92`) → every `minRequiredRole` tool is dropped for *everyone*, including real admins, even with `ADMIN_TOOLS_ENABLED=true`. Proven live: the unchanged e2e `tests/e2e/server.test.ts` now goes RED (default + combined variants) — `query-admin-insights` and the 6 other admin tools vanish from `tools/list`. This is a genuine regression, not just a test artifact. The stdio/no-OAuth paths need a role source (or the gate must not fail-closed to hidden when role is simply unknowable on that transport).

**BLOCKER 2 — version collision.** `package.json` is `4.6.1`, but `origin/main` already shipped `4.6.1` (via #840). This branch is 6 commits behind main and collides exactly. Rebase onto main and bump to `4.6.2`.

**Stale head vs your replies.** Several fixes you describe in-thread are NOT in `818d897`: delete-content/refresh/update still carry `MIN_ADMIN_SITE_ROLE` (your `ExplorerCanPublish` retag isn't committed), and `listUsers`/`updateUser` still tag `MIN_ADMIN_SITE_ROLE` + still call `assertAdmin`/`guardMutation` (not removed). Please push so review tracks real code. Inline notes below.

Full disposition of my 12 first-round comments: #2 (minSiteRole hierarchy) and #9 (identifier-format compare) resolved; #12 (demotion) behaviorally safe (live REST blocks regardless of cache); #1 (group-level) deferred by mutual agreement; #11 partial (ServerAdministrator mapped, SupportUser deferred low-pri); #3–#8, #10 still open at this head.

src/server.web.ts
```diff
@@ -130,11 +137,32 @@ export class WebMcpServer extends Server {
...
+    };
+
     const toolsToRegister: typeof allTools = [];
     for (const tool of allTools) {
       if (await Provider.from(tool.disabled)) continue;
       if (includeTools.length > 0 && !includeTools.includes(tool.name)) continue;
       if (excludeTools.length > 0 && excludeTools.includes(tool.name)) continue;
+      if (
+        tool.minRequiredRole &&
+        !siteRoleMeetsMinimum(await resolveSiteRole(), tool.minRequiredRole)
```
Alon-ST-DATA: [Alon's agent]

**BLOCKER (regression).** This drop fires whenever `resolveSiteRole()` returns `undefined` — which is *always* on transports that don't thread `authInfo` into `registerTools()`. Only `src/server/express.ts:200` (http-OAuth) passes it. Stdio (`src/index.ts:61`), combined (`src/index.combined.ts:68`), and http+`DANGEROUSLY_DISABLE_OAUTH` all call `registerTools()` with no arg, so `getCurrentUserSiteRole` early-returns `undefined` (`adminGate.ts:92`) and every `minRequiredRole` tool is stripped for real admins too.

Reproduced live: `ADMIN_TOOLS_ENABLED=true npx vitest run --config ./vitest.config.e2e.ts tests/e2e/server.test.ts` → 2 FAIL (default + combined `should list tools`); `query-admin-insights` + the 6 admin tools are dropped. That e2e file is unchanged by this PR, so it's a true regression, not a stale assertion.

Fix direction: resolve the caller's role on the stdio/PAT path too (the server has a PAT identity it can look up), or don't fail-closed-to-hidden when role is structurally unavailable on the transport. As-is, the feature only works on the http-OAuth path and silently breaks admins everywhere else.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.6.0",
+  "version": "4.6.1",
```
Alon-ST-DATA: [Alon's agent]

**BLOCKER.** `4.6.1` already shipped on `main` via #840. This branch is 6 commits behind main and collides on the exact version. Rebase onto main and bump to `4.7.0` (`npm run version:minor`).

src/tools/web/users/listUsers.ts
```diff
@@ -25,6 +25,7 @@ export const getListUsersTool = (server: WebMcpServer): WebTool<typeof paramsSch
     server,
     name: 'list-users',
     disabled: !config.adminToolsEnabled,
+    minRequiredRole: MIN_ADMIN_SITE_ROLE,
```
Alon-ST-DATA: [Alon's agent]

Still `MIN_ADMIN_SITE_ROLE` at this head, and `assertAdmin` is still called (`listUsers.ts:103`). Your reply said minRole was removed here — it isn't in `818d897`. Also the self-service model still stands: a user can read their own info, so strict admin-only over-restricts. If you keep it admin-only for now that's defensible, but the tag and the thread disagree — please reconcile.

src/tools/web/_lib/deleteContent.ts
```diff
@@ -141,6 +142,7 @@ export const getDeleteContentTool = async (
     server,
     name: 'delete-content',
     disabled: !config.adminToolsEnabled,
+    minRequiredRole: MIN_ADMIN_SITE_ROLE,
```
Alon-ST-DATA: [Alon's agent]

Still `MIN_ADMIN_SITE_ROLE` here (your `ExplorerCanPublish` retag from the thread isn't committed to `818d897`). One caution for when you do lower it: drop `minRequiredRole` to `ExplorerCanPublish` **and** loosen the call-time gate (`guardMutation`/`assertAdmin`) together — otherwise the tool registers for an Explorer-can-publish user but then the REST/admin gate rejects the call, which is exactly the registered-but-blocked UX this PR is trying to remove. Registration gate and runtime gate must move in lockstep.

src/tools/web/_lib/deleteContent.ts
Alon-ST-DATA: sure

src/server.web.test.ts
Alon-ST-DATA: Followed up with the admin tools PM and he said:
_"[they should] as they are full admins on customer sites when so enabled. but this is a situation where it's even more important to prevent unexpected behavior on the write / destructive side, to a greater degree even than customer admins."_

src/sdks/tableau/types/user.ts
stephendeoca: I added SupportUser, but MIN_ADMIN_SITE_ROLE is still SiteAdministratorExplorer. Lmk if we should change this

Alon-ST-DATA: Yes please change min role to SupportUser since it i a valid role for admin_tools.

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
```diff
@@ -29,6 +29,7 @@ export const getListExtractRefreshTasksTool = (
     server,
     name: 'list-extract-refresh-tasks',
     disabled: !config.adminToolsEnabled,
+    minRequiredRole: 'ExplorerCanPublish',
```
Alon-ST-DATA, line 32: [Alon's agent] **MAJOR — registration gate contradicts the runtime gate.** `minRequiredRole: 'ExplorerCanPublish'` here, but the callback still calls `assertAdmin` (admin-only, ~line 104). Net: an ExplorerCanPublish caller SEES this tool in `tools/list`, invokes it, and gets *"This tool requires administrative permissions."* The two gates disagree. The backing Cloud REST endpoint is admin-scoped anyway → set `minRequiredRole: MIN_ADMIN_SITE_ROLE`, or drop/loosen the `assertAdmin` (transitional per the planned assertion-removal, but it ships as a user-visible inconsistency now).

src/tools/web/_lib/deleteContent.ts
```diff
@@ -141,6 +141,7 @@ export const getDeleteContentTool = async (
     server,
     name: 'delete-content',
     disabled: !config.adminToolsEnabled,
+    minRequiredRole: 'ExplorerCanPublish',
```
Alon-ST-DATA, line 144: [Alon's agent] **MAJOR — preview/confirm role asymmetry → dead-end HITL flow.** `delete-content` (preview) is gated `ExplorerCanPublish` here with no `assertAdmin`, but `confirm-delete-content` is gated `SupportUser`. An ExplorerCanPublish user can run the preview and receive a `confirmationToken`, but the matching confirm tool is never registered for them → they start a destructive flow they can never complete. Align both halves of the pair. (Same shape on the update-extract pair — see that comment.)

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
```diff
@@ -118,6 +118,7 @@ export const getUpdateCloudExtractRefreshTaskTool = async (
     server,
     name: 'update-cloud-extract-refresh-task',
     disabled: !config.adminToolsEnabled,
+    minRequiredRole: 'ExplorerCanPublish',
```
Alon-ST-DATA, line 121: [Alon's agent] **MAJOR — preview/confirm role asymmetry (pair #2).** `update-cloud-extract-refresh-task` (preview) gated `ExplorerCanPublish` here, no `assertAdmin`, but `confirm-update-cloud-extract-refresh-task` is gated `SupportUser`. Same dead-end as delete-content: an ExplorerCanPublish user gets a token from the preview but the confirm tool is never registered for them. Align the two halves.

src/sdks/tableau/types/user.ts
```diff
@@ -20,15 +20,52 @@ export const userSchema = z.object({
 
 export type User = z.infer<typeof userSchema>;
 
-export const ADMIN_SITE_ROLES = new Set([
+export const ADMIN_SITE_ROLES: readonly string[] = [
+  'SupportUser',
```
Alon-ST-DATA, line 24: [Alon's agent] **MAJOR — SupportUser added to `ADMIN_SITE_ROLES` unconditionally (not behind the flag).** This widens `assertAdmin` to accept SupportUser for *every* assertAdmin-gated tool the moment this merges — independent of `enforce-role-requirements`. So a runtime authz change ships flag-or-no-flag. `SITE_ROLE_HIERARCHY` also ranks SupportUser=5 (above Creator=4) and `MIN_ADMIN_SITE_ROLE='SupportUser'`, so SupportUser clears both registration gating and runtime `assertAdmin`, including destructive `confirm-delete-content`. The role ordering is total & internally consistent for the 9 listed roles — please just confirm SupportUser-as-admin outside the feature flag is the intended product decision.

src/server.web.ts
```diff
@@ -172,14 +193,64 @@ export class WebMcpServer extends Server {
...
+    // Names of role-gated tools hidden specifically because the role fetch FAILED
+    // rather than because the caller's role was genuinely too low.
+    const toolsOmittedForRoleFetchFailure: string[] = [];
+
     const toolsToRegister: typeof allTools = [];
     for (const tool of allTools) {
       if (await Provider.from(tool.disabled)) continue;
       if (includeTools.length > 0 && !includeTools.includes(tool.name)) continue;
       if (excludeTools.length > 0 && excludeTools.includes(tool.name)) continue;
+      if (enforceRoleRequirements && tool.minRequiredRole) {
```
Alon-ST-DATA, line 248: [Alon's agent] **MAJOR / DESIGN — fail-closed on a TRANSIENT role-fetch failure drops tools for genuine admins (the reproduced flap).** When `enforce-role-requirements` is on and any candidate tool declares `minRequiredRole`, registration does a synchronous `GET /sessions/current` (2 retries, exp backoff ~100/200ms + jitter ≈ 300–700ms total). Any blip spanning all 3 attempts → role resolves `undefined` → `siteRoleMeetsMinimum` false → **all** role-gated tools omitted for that whole connection until reconnect.

**Reproduced live:** 6 *concurrent* connects on the same admin PAT → **5/6 sessions dropped to 23 tools** (the 5 admin tools omitted) for a genuine SiteAdministratorCreator; 1/6 kept 28. Cause: registration-time PAT self-sign-in hits PAT-concurrency 401s under simultaneous init; all retries (concurrent) fail. Each dropped session correctly logged the omission and carried `SITE_ROLE_UNAVAILABLE_WARNING`, so it's honestly surfaced — but the practical effect is an admin transiently losing admin tools on reconnect races / multiple clients. (Sequential reconnects: 10/10 deterministic, no warning.)

Since the real enforcement is runtime `assertAdmin` (a non-admin is still blocked at call time), options: (a) **fail-open** on fetch-failure — register the tools and let `assertAdmin`/the API reject — eliminates the flap with zero security loss for the admin-gated tools; (b) de-dupe/cache the registration role fetch across concurrent session inits; (c) at minimum widen retries/backoff. Note: the OAuth/embedded path passes a pre-issued bearer token to `/sessions/current` (no fresh sign-in) so it likely avoids this PAT-signin wall — untested here. Minor: typo two lines down — "ommited" → "omitted".

Alon-ST-DATA (commented):
[Alon's agent] **Round-3 review + live e2e** (head e2170f86, v4.8.0) — three independent passes: code/build, standards diff-review, and live 3-PAT registration testing on a real site.

**Round-1 BLOCKERs — verified FIXED:** the no-authInfo drop (role now resolved via `GET /sessions/current` in `getCurrentUserSiteRole`, needs no `userId` claim / `users:read` scope, so it works even when `registerTools` has no `authInfo` — stdio and http+DANGEROUSLY_DISABLE_OAUTH included) and the 4.6.1 version collision (now 4.8.0).

**Core feature — VERIFIED CORRECT & deterministic** under normal (sequential) connects. Live matrix on strawberrysugar10ay, `enforce-role-requirements=true`, `ADMIN_TOOLS_ENABLED=true`, http/PAT:

| Tool (minRole) | Explorer(2) | SupportUser(5) | SiteAdminCreator(7) |
|---|---|---|---|
| query-admin-insights (5) | ✗ | ✓ | ✓ |
| list-jobs (5) | ✗ | ✓ | ✓ |
| delete-content (3) | ✗ | ✓ | ✓ |
| list-extract-refresh-tasks (3) | ✗ | ✓ | ✓ |
| update-cloud-extract-refresh-task (3) | ✗ | ✓ | ✓ |
| list-workbooks (ungated) | ✓ | ✓ | ✓ |

Tool counts 23 / 28 / 28 (delta = exactly the 5 gated tools). Every cell matches the `minRequiredRole` model; ground-truth roles confirmed independently via raw REST `/sessions/current`.

**Tests:** `npm run build` clean; `npm test` **2867/2867 pass**; role-specific unit `server.web.test.ts` (24) + `adminGate.test.ts` (14) + `user.test.ts` (14) = **52/52**; `tests/e2e/server.test.ts` **9/9** (run WITHOUT process-level `ADMIN_TOOLS_ENABLED` — see harness note below). No e2e coverage exists for the role-gating paths themselves — the matrix above was produced manually; an e2e that boots flag-on and asserts an admin sees the admin tools would close the loop and would have caught the flap.

**Additional finding — registration-visibility gap (not anchorable to a diff line):** `list-users` and `update-user` carry **no `minRequiredRole`**, so with `ADMIN_TOOLS_ENABLED=true` they REGISTER for an Explorer (rank 2) and appear in a non-admin's `tools/list` — unlike the other 5 admin tools. Runtime `assertAdmin` still blocks execution, but they're advertised to non-admins. Intentional, or should these two also declare `minRequiredRole`?

**Harness note (not a PR bug):** `tests/e2e/server.test.ts` keys its expected tool set off the runner's `process.env.ADMIN_TOOLS_ENABLED` while default/combined variants spawn via `getDefaultEnv()` (which omits the flag) — running the suite WITH `ADMIN_TOOLS_ENABLED=true` exported spuriously fails 2/9. Run it with `env -u ADMIN_TOOLS_ENABLED`.

Inline comments below carry the 5 finding-level items. Framing: the security boundary is the runtime `assertAdmin` + Tableau API perms (unchanged); registration-time role gating is a discoverability/UX layer — several items are availability/UX, not security. Round-1 blockers are resolved; safe to merge with the flag default-OFF, but items 1–3 and a decision on 4 should land before the flag is ever enabled.

src/sdks/tableau/types/user.ts
stephendeoca: SupportUser as admin outside the feature flag is the intended product design

src/tools/web/extractRefreshTasks/listExtractRefreshTasks.ts
Alon-ST-DATA: [Alon's agent] Filed follow-up: **W-24079182** (assigned to Alon Siman Tov, Epic [TMCP] 266). Tracks reconciling the registration `minRequiredRole` gate with the runtime `assertAdmin` — align this tool's registration gate to `MIN_ADMIN_SITE_ROLE` (or loosen `assertAdmin`) so a single capability declaration drives both and the two can't drift.

src/tools/web/_lib/deleteContent.ts
Alon-ST-DATA: [Alon's agent] Filed follow-up: **W-24079182** (assigned to Alon Siman Tov, Epic [TMCP] 266). Covers this preview/confirm asymmetry — the WI requires both halves of every destructive pair (delete-content / confirm-delete-content, update-extract pair) to be gated consistently so mutating flows stay admin-guarded end-to-end and no role can start a flow it can't finish.

src/tools/web/extractRefreshTasks/updateCloudExtractRefreshTask.ts
Alon-ST-DATA: [Alon's agent] Rolled into the same follow-up: **W-24079182**. This is the update-extract half of the preview/confirm-asymmetry class already in scope there (WI item #3 names `update-cloud-extract-refresh-task` and its confirm half). No separate WI — same fix (align both halves of every destructive preview/confirm pair under one assert-admin capability).

src/server.web.ts
Alon-ST-DATA: Ok this is known limitation for PAT / self-hosted. I did check with oAuth for roles and verified the proper tools are registred for each role:

V/E - 25 tools
E(cp)/C - 28 tools
SupU/SAE/SAC - 30 tools

Alon-ST-DATA: [Alon's agent] **Registration-load test results (live, `enforce-role-requirements` ON).** Captured the actual registered tool set observed by a client session against a running server, to attach data to this thread.

**Observed set — 25 tools registered:**
- 14 read/discovery: `list-datasources`, `list-workbooks`, `list-views`, `list-projects`, `list-custom-views`, `get-workbook`, `get-view`, `get-view-data`, `get-view-image`, `get-custom-view-data`, `get-custom-view-image`, `get-datasource-metadata`, `search-content`, `query-datasource`
- 7 Pulse
- 2 admin: `list-users`, `update-user`
- 2 OAuth session: `reset-consent`, `revoke-access-token`

**Omitted (5):** `query-admin-insights`, `list-jobs`, `list-extract-refresh-tasks`, `delete-content`, `update-cloud-extract-refresh-task`.

**The set is internally inconsistent — this is the registration↔runtime drift, not a clean role verdict:**
- `update-user` (admin-only, **mutating**) is registered, but read-only admin tools `query-admin-insights` / `list-jobs` / `list-extract-refresh-tasks` are dropped. The least-dangerous tools are hidden while a mutating admin tool is exposed — backwards from what a role gate should produce.
- `delete-content` (destructive) dropped but `update-user` (mutating) kept — two mutating/destructive admin tools, opposite verdicts in the same session.

So the omission set doesn't line up with each tool's real runtime `assertAdmin` requirement: it reflects the per-tool `minRequiredRole` values as authored, which disagree with the callback gates. That's the same drift called out on `listExtractRefreshTasks.ts` (r3914895511) and the preview/confirm asymmetry on `deleteContent.ts` (r3914895516) / `updateCloudExtractRefreshTask.ts` (r3914895525).

Combined with the concurrent-flap in this thread (transient `/sessions/current` failure → *all* gated tools dropped), the net is: the registered set is a function of (authored `minRequiredRole` per tool) × (a fail-closed role fetch that can flap), neither of which matches the runtime `assertAdmin` truth. Reconciling those is tracked in follow-up **W-24079182**.

Alon-ST-DATA: [Alon's agent] Filed as its own follow-up: **W-24079830** — *Registration role-fetch resilience: eliminate fail-closed tool-drop for genuine admins on transient `/sessions/current` failure.*

Tracked separately from W-24079182 (which covers the static `minRequiredRole` ↔ `assertAdmin` drift + preview/confirm asymmetry) because this is a distinct root cause: registration-time role-fetch reliability/policy, not a static mismatch. Reconciling the drift wouldn't fix this flap — a perfectly-aligned gate still drops tools when the fetch flaps.

Scope on the WI: (1) decide fail-open vs fail-closed on fetch failure (recommend fail-open — runtime `assertAdmin` still blocks non-admins, so fail-closed at registration buys no security on admin-gated tools, only harms genuine admins); (2) de-dupe/cache the registration role fetch across concurrent session inits (kills the PAT-concurrency 401 wall); (3) widen retry/backoff as a floor; (4) verify the OAuth/embedded path (pre-issued bearer, no fresh sign-in) is immune. Includes the live 5/6-concurrent-drop repro.

src/tools/web/adminGate.ts
```diff
@@ -47,25 +21,105 @@ export async function assertAdmin(
...
+ * would otherwise be empty and produce a malformed `/users/` request.
+ *
+ * Retries a failing fetch (up to {@link MAX_SITE_ROLE_FETCH_ATTEMPTS} total attempts) via the shared
+ * {@link retry} helper. A deterministic failure is *not* retried: an `unauthorized` session, or a
+ * 4xx thrown by the sign-in step on the pat/uat/direct-trust path (e.g. 401/403), can't be changed
+ * by retrying, so we fail fast rather than delaying registration for every unauthorized caller.
+ *
+ * Fail-closed: once retries are exhausted (or a non-retryable error is thrown), returns `undefined`
+ * (as does any missing-auth case, since the sign-in path throws) so a failure never grants access
+ * via a falsy role predicate.
```
yyk16293, line 78: Let's simplify these comments with description about retry strategy

src/tools/web/adminGate.ts
```diff
@@ -47,25 +21,105 @@ export async function assertAdmin(
...
+): Promise<string | undefined> {
+  try {
+    return await retry(
+      async () =>
+        await useRestApi({
+          ...restApiArgs,
+          // `/sessions/current` requires no scopes; this scope is retained only so the Connected App
+          // sign-in used by the pat/uat/direct-trust paths keeps a valid, non-empty scope set. It is
+          // ignored on the OAuth path, where the pre-issued bearer token is passed through as-is.
+          jwtScopes: ['tableau:users:read'],
```
yyk16293, line 91: Verify on scope dependency for **_/sessions/current_** API call

src/tools/web/tool.ts
```diff
@@ -63,8 +64,14 @@ export type WebToolParams<Args extends ZodRawShape | undefined = undefined> = To
   TableauWebRequestHandlerExtra,
   TableauWebToolCallback<Args>,
   Args
-> &
-  (
+> & {
+  /**
+   * Lowest site role allowed to see this tool at registration time. Omitted means everyone. When
```
yyk16293, line 69: Let's add in BOLD letters about default behavior when this field is omitted.

src/tools/web/adminGate.test.ts
```diff
@@ -51,11 +66,289 @@ describe('assertAdmin', () => {
...
+    };
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  // getCurrentUserSiteRole retries with real backoff delays. Fake timers keep the retry-path tests
+  // fast: schedule the call, drain all pending timers/microtasks, then await the settled result.
+  async function resolveWithFakeTimers<T>(op: () => Promise<T>): Promise<T> {
```
jarhun88, line 118: wait why do we need this in a test? if were trying to test getCurrentUserSiteRole then i would suggest returning mock data instead of actually calling, lmk if i have the wrong understanding

src/tools/web/adminGate.ts
```diff
@@ -47,25 +21,105 @@ export async function assertAdmin(
...
+ * Retries a failing fetch (up to {@link MAX_SITE_ROLE_FETCH_ATTEMPTS} total attempts) via the shared
+ * {@link retry} helper. A deterministic failure is *not* retried: an `unauthorized` session, or a
+ * 4xx thrown by the sign-in step on the pat/uat/direct-trust path (e.g. 401/403), can't be changed
+ * by retrying, so we fail fast rather than delaying registration for every unauthorized caller.
+ *
+ * Fail-closed: once retries are exhausted (or a non-retryable error is thrown), returns `undefined`
+ * (as does any missing-auth case, since the sign-in path throws) so a failure never grants access
+ * via a falsy role predicate.
+ */
+export async function getCurrentUserSiteRole(
```
jarhun88, line 80: i think this function should be in its own folder. something like userRole. Were going to be using for other tools beyond admin right?

src/tools/web/_lib/confirmDeleteContent.ts
```diff
@@ -46,6 +47,7 @@ export const getConfirmDeleteContentTool = (server: WebMcpServer): WebTool<typeo
       async () =>
         !config.adminToolsEnabled || !(await getFeatureGate().isFeatureEnabled('mcp-apps')),
     ),
+    minRequiredRole: MIN_ADMIN_SITE_ROLE,
```
jarhun88, line 50: **Role-gate mismatch breaks the HITL confirm flow for non-admin publishers.** `delete-content` (deleteContent.ts) is gated at `minRequiredRole: 'ExplorerCanPublish'` (rank 3), but this confirm tool is gated at `MIN_ADMIN_SITE_ROLE` ('SupportUser', rank 5). With `enforce-role-requirements` on, a caller ranked ExplorerCanPublish or Creator (ranks 3–4) will have `delete-content` registered (can preview a deletion) but NOT `confirm-delete-content` registered — so they can never actually confirm/execute the deletion they just previewed. The preview and confirm halves of the same HITL flow should share the same `minRequiredRole` (presumably `ExplorerCanPublish`, matching the preview tool's actual authorization requirement per the REST API), not silently jump to the admin threshold.

— Jaehun Bot

src/tools/web/extractRefreshTasks/confirmUpdateCloudExtractRefreshTask.ts
```diff
@@ -52,6 +53,7 @@ export const getConfirmUpdateCloudExtractRefreshTaskTool = (
       async () =>
         !config.adminToolsEnabled || !(await getFeatureGate().isFeatureEnabled('mcp-apps')),
     ),
+    minRequiredRole: MIN_ADMIN_SITE_ROLE,
```
jarhun88, line 56: **Same role-gate mismatch as confirmDeleteContent.ts.** `update-cloud-extract-refresh-task` is gated at `minRequiredRole: 'ExplorerCanPublish'` (rank 3), but this confirm tool requires `MIN_ADMIN_SITE_ROLE` ('SupportUser', rank 5). With `enforce-role-requirements` enabled, an ExplorerCanPublish/Creator caller sees `update-cloud-extract-refresh-task` in their tool list but never sees its confirm tool, so they can preview a schedule change but can't complete it. This confirm tool's `minRequiredRole` should match its preview tool's (`ExplorerCanPublish`), not the admin threshold.

— Jaehun Bot

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.6.0",
+  "version": "4.8.0",
```
jarhun88: **Version collision (recurring blocker — please recheck before merge).** This branch bumps to `4.8.0`, but `main` has already shipped `v4.8.1` (tag exists, and `gh pr view --json mergeable` currently reports `CONFLICTING`/`DIRTY`). This is the same class of issue flagged twice already in review (first at 4.6.1, then again after a rebase) — the branch keeps falling behind main's version. Please rebase onto current `main` and bump to `4.8.2` (or whatever is next after rebase) via `npm run version:minor`/`patch` before merge.

— Jaehun Bot

jarhun88:
**Summary review (logic-focused, lighter path for an already-open external PR).**

CI is green (build ×2, tests, SAST, credential scan, CLA — `gh pr checks 821`). This is my independent pass on top of the extensive prior review thread (Alon's agent, 3 rounds incl. live e2e). Two rounds of prior blockers (no-authInfo tool drop, version collision at 4.6.1) are confirmed fixed at this head via `getCurrentUserSiteRole`'s `/sessions/current` resolution — good fix, it correctly avoids depending on `userId`/`users:read` scope so it works uniformly across stdio, http+OAuth, and http+DANGEROUSLY_DISABLE_OAUTH.

**New findings this pass (posted inline):**
1. **HITL preview/confirm role mismatch** (confirmDeleteContent.ts, confirmUpdateCloudExtractRefreshTask.ts) — the confirm tools require `MIN_ADMIN_SITE_ROLE` while their paired preview tools only require `ExplorerCanPublish`. An ExplorerCanPublish/Creator caller can preview but never see the confirm tool to execute. This looks like a genuine oversight, not an intentional restriction — the confirm tool should mirror its preview tool's role floor.
2. **Version collision has recurred** — branch bumps to `4.8.0`, but `main` already shipped `v4.8.1` (tag exists) and GitHub reports `mergeable: CONFLICTING`/`mergeStateStatus: DIRTY`. Needs another rebase + version bump before merge — same class of issue flagged twice before.

**Still-open items from the existing thread I'd flag as pre-merge-worthy but non-blocking** (not re-litigating, just surfacing since they weren't visibly resolved at this head):
- `list-users`/`update-user` still register with no `minRequiredRole` even with `ADMIN_TOOLS_ENABLED=true` — discoverability/UX gap only (runtime `assertAdmin` still blocks execution), author's call on whether that's intentional.
- No e2e test boots the server with `enforce-role-requirements=true` and asserts the registered tool set differs by role — the round-3 reviewer's manual matrix is solid evidence but isn't codified as a regression test.

**Strengths:** fail-closed design throughout (`siteRoleMeetsMinimum`, `getCurrentUserSiteRole`, registration loop) is consistent and well-tested; removing the old `ExpiringMap` admin cache closes the demotion-during-session risk raised earlier in the thread; the lazy single-fetch-per-registration-call pattern (`Object.hasOwn` guard) avoids redundant REST calls; retry-vs-fail-fast split (401/403 vs transient) in `getCurrentUserSiteRole` is correct and tested.

— Jaehun Bot

src/server.web.ts
```diff
@@ -202,14 +223,64 @@ export class WebMcpServer extends Server {
     const allTools = await Promise.all(
       webToolFactories.map((toolFactory) => toolFactory(this, tableauServerInfo.productVersion)),
     );
+
+    // The registration-time role check is gated behind `enforce-role-requirements`. When it's off
+    // (the default), tools register regardless of the caller's site role and no /users call is made.
+    const enforceRoleRequirements = await getFeatureGate().isFeatureEnabled(
+      'enforce-role-requirements',
```
jarhun88, line 232: are you planning on using a different FF for non role related conditional registration? eg pulse

stephendeoca: yes that's in my other MR https://github.com/tableau/tableau-mcp/pull/871

src/server.web.ts
```diff
@@ -202,14 +223,64 @@ export class WebMcpServer extends Server {
...
       if (await Provider.from(tool.disabled)) continue;
       if (includeTools.length > 0 && !includeTools.includes(tool.name)) continue;
       if (excludeTools.length > 0 && excludeTools.includes(tool.name)) continue;
+      if (enforceRoleRequirements && tool.minRequiredRole) {
+        // Site role is fetched lazily when at least one candidate tool declares a `minRequiredRole`.
+        // Fail-closed: tools with role requirements are ommited during registration if a user's role
+        // is unable to be fetched. Using `Object.hasOwn` to check if site role context has been populated, as
+        // site role may be undefined as a result of failure to fetch. Retries are used when fetching site role,
+        // so there is no need to keep calling `getCurrentUserSiteRole` if it has already failed before.
+        if (!Object.hasOwn(registrationContext, 'siteRole')) {
```
jarhun88, line 254: im abit confused, when does this get populated?

src/server.web.ts
```diff
@@ -202,14 +223,64 @@ export class WebMcpServer extends Server {
...
+
+    // The registration-time role check is gated behind `enforce-role-requirements`. When it's off
+    // (the default), tools register regardless of the caller's site role and no /users call is made.
+    const enforceRoleRequirements = await getFeatureGate().isFeatureEnabled(
+      'enforce-role-requirements',
+    );
+
+    // Stores context that is used for determining if registration conditions have been met.
+    // Registration context is uninitialized, but then gets populated with each condition checked.
+    const registrationContext: RegistrationContext = {};
```
jarhun88, line 237: can you explain to me what exactly is the purpose of this? is it so that we only call the getCurrentUserSiteRole once? 

if this is meant to be a caching impl, how is this more efficient solution then just calling the getCurrentUserSiteRole api once outside of the tools for loop?

stephendeoca: This is a context object that's going to be used for conditional registration, it doesn't seem to do much right now but in my follow up MR it will have more properties like isPulseEnabled and hasPulsePremium.

src/server.web.ts
stephendeoca: This gets populated the first time a tool that requires a min role is included in the list of tools to register. At first there will be no site role, but then we fetch it and store the context so that other tools that specify min role requirements don't have to refetch site role.

src/tools/web/extractRefreshTasks/confirmUpdateCloudExtractRefreshTask.ts
stephendeoca: This is to be fixed in a follow up by Alon. Right now we're focusing on the role based registration logic.

src/tools/web/adminGate.test.ts
stephendeoca: We're mocking REST API to test when getCurrentUserSiteRole fails, but there's logic to retry REST API calls after a given time has passed but this mocks out the timer so that the tests aren't actually waiting. So we're testing the retry logic

src/tools/web/adminGate.ts
stephendeoca: I'm going to leave this here for now, there should be follow up work that might involve caching for admin tools. Then in the follow up we can organize

jarhun88: hmm ok just note that both flow tools and authoring tools are going to consume this soon, which as consequence may require additional work to cleanup later

stephendeoca: Their tool logic doesn't call this code, if they do specify a min required role the only surface area of that code would be the role value in the tool definition.

src/server.web.ts
jarhun88: what is the benefit of doing this? based on impl, I see that we will call getCurrentUserSiteRole once. But this is the same as just calling it outside the tool for loop.

src/tools/web/adminGate.ts
jarhun88: ah yes sounds good

src/server.web.ts
stephendeoca: Since tool registration is a critical path for MCP connectivity, I wanted to make sure we only make calls if its actually necessary. If we move the call to outside the tool loop we'll always fetch role during registration, but its possible after the tool exclusion logic that we have no tools with role requirements and we made an unnecessary fetch.

This practice of checking conditions only if tools require makes more sense for other conditions such as if pulse is enabled. Why ping the pulse endpoint if no pulse tools are being registered? But fetching site role would be a condition that's pretty safe to move outside the loop. Lmk if you think we should do that.

src/tools/web/adminGate.ts
stephendeoca: Turns out empty scope in JWT token is allowed, so I removed the unnecessary scope

src/server.web.ts
jarhun88: the optimization feels abit premature, I wouldnt expect a single api to delay connection time significantly. We could just wrap the api function with a cache if we want to reduce fetches for future connections which would reduce calls and have cleaner impl

src/server.web.ts
```diff
@@ -246,17 +251,10 @@ export class WebMcpServer extends Server {
...
-        // so there is no need to keep calling `getCurrentUserSiteRole` if it has already failed before.
-        if (!Object.hasOwn(registrationContext, 'siteRole')) {
-          registrationContext.siteRole = await getCurrentUserSiteRole(restApiArgs);
-        }
         const siteRole = registrationContext.siteRole;
         if (!siteRoleMeetsMinimum(siteRole, tool.minRequiredRole)) {
-          // An `undefined` role means the fetch failed
+          // Tools with role requirements are ommited during registration if a user's role
+          // is unable to be fetched. An `undefined` role means the fetch failed.
           if (siteRole === undefined) {
```
jarhun88, line 259: might want a FF check here, cuz siteRole is going to be undefined when FF is off right?

src/server.web.ts
```diff
@@ -246,17 +251,10 @@ export class WebMcpServer extends Server {
...
-        // is unable to be fetched. Using `Object.hasOwn` to check if site role context has been populated, as
-        // site role may be undefined as a result of failure to fetch. Retries are used when fetching site role,
-        // so there is no need to keep calling `getCurrentUserSiteRole` if it has already failed before.
-        if (!Object.hasOwn(registrationContext, 'siteRole')) {
-          registrationContext.siteRole = await getCurrentUserSiteRole(restApiArgs);
-        }
         const siteRole = registrationContext.siteRole;
         if (!siteRoleMeetsMinimum(siteRole, tool.minRequiredRole)) {
-          // An `undefined` role means the fetch failed
+          // Tools with role requirements are ommited during registration if a user's role
```
jarhun88: or FF is off

jarhun88 (approved):

src/server.web.ts
stephendeoca: There's already a FF check before getting into this branch of code. It's the parent if statement

Alon-ST-DATA (approved):
LGTM.
We need to figure out the UAT connection issue, but the can be followed later.

PR #822, opened 2026-08-22
feat(desktop): format existing dashboards without rebuilding them

mattcfilbert:
Existing dashboards can now be formatted without rebuilding them. The rule is: edit the named constituent worksheets, verify both formatting and worksheet semantics after write, and leave dashboard structure alone. Approving this PR accepts that boundary.

What changes

- Adds `format-worksheets` for labels and explicit number, currency, and percentage formats across 1–12 named worksheets.
- Requires the caller to provide currency symbols; the tool never guesses locale.
- Re-reads each worksheet under the apply lock and rejects readback that changed shelves or other worksheet semantics.
- Routes dashboard formatting through `list-dashboards` and the dashboard's constituent worksheets.

What stays out

- No fonts, colors, themes, layout changes, or visual autocorrection. Workbook themes and style guides own the overall look.
- No silent dashboard rebuilds. `run-dashboard-batch` remains a new-dashboard path.

Proof

- Direct MCP smoke on a disposable packaged Superstore workbook formatted the `Performance` worksheet twice with verified readback.
- A live-shaped regression proves formatting cannot report success if a shelf disappears.
- Canonical gate: 390 test files, 6,211 tests, lint, typecheck, Desktop build, and lockstep checks.

Stack: base change for #823 and #824.

_Updated by MattGPT._


src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+                  (finding) => finding.severity === 'error',
+                ),
+              signal: extra.signal,
+            });
+            if (!readback.ok) {
+              return new DesktopCommandExecutionError(readback.error).toErr();
+            }
+            if (!readback.settled) {
+              return new XmlModificationError(
+                `Desktop accepted formatting for "${worksheet.name}", but the requested formatting did not survive readback or Tableau dropped worksheet semantics.`,
```
myu404: 🤖 **Reviewed by MichaelGPT**

Only the drift branch discloses partial writes — it appends a *"N earlier formatted sheets may already be updated"* clause. This not-settled branch (*"did not survive readback or Tableau dropped worksheet semantics"*) and both command-error branches carry no such caveat. Sheets are applied one at a time, so a failure on sheet N leaves sheets 1..N-1 applied and verified without telling the caller.

Impact is limited — the tool is idempotent (the `alreadyFormatted` short-circuit makes a full-batch retry safe) — so this is honesty/consistency, not a correctness bug. Suggest adding the same "N earlier formatted sheets may already be updated" clause to the readback-failure and command-error messages.

src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+            if (applied.kind === 'drift') {
+              return new XmlModificationError(
+                `Worksheet "${worksheet.name}" changed while formatting was being applied. ${formatted.length} earlier formatted sheets may already be updated.`,
+              ).toErr();
+            }
+
+            const readback = await pollReadback({
+              read: async () => await executor.getWorksheetDocument(worksheet.id, extra.signal),
+              settled: (value) =>
+                hasRequestedFormatting(value.xml, xml) &&
```
myu404, line 193: 🤖 **Reviewed by MichaelGPT**

Readback "settled" compares via `hasRequestedFormatting(value.xml, xml)`, which keys on the **raw, undecoded** `value` attribute. Currency is emitted as the literal `c&quot;<symbol>&quot;#,##0…`. If a live Desktop re-serializes that attribute with single-quote delimiters and literal double-quotes (`c"$"#,##0`), the readback string is no longer byte-equal to the intended `c&quot;$&quot;#,##0`, `hasRequestedFormatting` returns false, and **every currency apply falsely reports "did not survive readback."**

The mock harness only perturbs `/>`→` />` spacing and re-stores the exact intended string, so entity/quote normalization is never exercised. This path is fail-closed (safe), but it could make the currency case unusable against a live Desktop. Suggest normalizing the `value` attribute (decode entities / canonicalize quoting) on both sides before comparison, or adding a test that round-trips the format string through a real XML parse/serialize.

src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+  DesktopCommandExecutionError,
+  XmlModificationError,
+} from '../../../../errors/mcpToolError.js';
+import { DesktopMcpServer } from '../../../../server.desktop.js';
+import { resolveShelfField } from '../../api/resolveShelfField.js';
+import { sessionParam } from '../../params.js';
+import { DesktopTool } from '../../tool.js';
+
+const displayUnitsSchema = z.enum(['none', 'thousands', 'millions', 'billions']);
+const currencySymbolSchema = z.enum(['$', '€', '£', '¥']);
```
myu404, line 24: 🤖 **Reviewed by MichaelGPT**

`currencySymbol` is a closed enum of exactly four glyphs (`$ € £ ¥`); any other currency (₹, A$, R$, CHF…) is unrepresentable and rejected by Zod. The "never guess locale" property is good and holds. Just confirming this four-currency cap is the intended scope and not a gap — if broader support is expected, accept a validated + escaped short symbol string instead of a four-value enum; otherwise a note in the tool description would set expectations.

src/tools/desktop/authoring/sheets/runDashboardBatch.ts
```diff
@@ -66,7 +65,7 @@ import {
 } from './composeDashboardCore.js';
 
 const paramsSchema = {
-  session: sessionParam({ max: 64 }),
+  session: z.string().max(64).optional(),
```
myu404: 🤖 **Reviewed by MichaelGPT**

The `session` param switched from the shared `sessionParam({ max: 64 })` to a raw `z.string().max(64).optional()`, dropping the shared `.describe('Session ID; optional if pinned or unique.')` that every other desktop tool — including this PR's own `format-worksheets` — carries. No validation/trimming is lost (only the LLM-facing description), and it reads as byte-funding for the new `replaceExisting` field under the per-entry cap. Still an avoidable divergence from the shared factory — suggest keeping `sessionParam({ max: 64 })` and funding `replaceExisting`'s bytes elsewhere, or noting the divergence in a comment.

myu404 (approved):

`format-worksheets` delivers what the description claims: it edits existing worksheet cached XML for mark labels plus explicit number/currency/percentage formats, applies under the apply lock with a `sourceSha256` drift check, and verifies **fail-closed** via `pollReadback` — requiring *both* `hasRequestedFormatting` *and* zero error-severity findings from `verifyWorksheetReadback` (so shelf/encoding/mark/filter/sort semantics can't silently change).

Currency-symbol enforcement (`superRefine`, never guessing locale), the per-worksheet 1–12 schemas, duplicate-name rejection, the idempotent `alreadyFormatted` short-circuit, and `run-dashboard-batch`'s hoisted fail-closed existing-dashboard check (moved *before* any workbook apply, gated on the new `replaceExisting`) are all implemented as described and well-tested — including a shelf-drop and a pane-move regression. The `existingDashboardName` move from `candidateXml` to `workbookXml` is semantically equivalent (a worksheet upsert adds no dashboards) and doesn't break the replace path.

No blocker- or major-severity defects. Four non-blocking observations inline (one `minor`, two `question`, one `nit`).

src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+      destructiveHint: false,
+      idempotentHint: true,
+      openWorldHint: false,
+    },
+    callback: async ({ session, worksheets }, extra): Promise<CallToolResult> => {
+      return await tool.logAndExecute<{ formatted: FormattedWorksheet[] }>({
+        extra,
+        args: { session, worksheets },
+        callback: async () => {
+          const worksheetNames = worksheets.map(({ name }) => name);
```
nawartamawi: [P2] Deduplicate after resolving worksheet IDs

The duplicate guard compares raw selectors, but each selector accepts either a worksheet name or ID. Passing the same worksheet once by name and once by ID passes preflight; the first edit applies, then the second re-read sees the first change and fails the source-hash check, leaving a partial operation. Resolve every selector first and reject duplicate resolved IDs before applying anything.

src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+    const paneXml = xml.slice(pane.start, pane.end);
+    const nextPane = upsertStyleFormat(paneXml, 'mark', {
+      attr: 'mark-labels-show',
+      value: request.showLabels ? 'true' : 'false',
+    });
+    xml = xml.slice(0, pane.start) + nextPane + xml.slice(pane.end);
+  }
+
+  for (const format of request.numberFormats ?? []) {
+    const resolved = resolveShelfField(xml, format.field);
```
nawartamawi, line 241: [P2] Reject ambiguous friendly field names

Friendly field resolution returns the first canonical pill whose local name matches. A worksheet containing SUM(Sales) and AVG(Sales), or identically named fields from multiple datasources, therefore accepts field Sales, formats only the first match, and reports verified success. Reject ambiguous names and return the canonical candidates, or require an exact canonical reference.

nawartamawi (changes requested):
Requesting changes for two correctness issues: duplicate name/ID selectors can cause a partial write, and ambiguous friendly field names can format the wrong canonical pill while reporting verified success. Details are inline.

src/tools/desktop/authoring/style/formatWorksheets.ts
lauren-jacksonSFDC: Seems like we would have this list somewhere else as well - should it be shared? Same with the display units?

src/tools/desktop/authoring/style/formatWorksheets.ts
lauren-jacksonSFDC: We also should not be using worksheet names here unless actually needed - the ID is the safe alternative and is how we talk to the API as well as finding specific sheets in the XML (via the simple-identifier tag)

src/tools/desktop/authoring/style/formatWorksheets.ts
```diff
@@ -0,0 +1,405 @@
...
+  });
+}
+
+function numberFormatElements(xml: string, column: string): Array<'cell' | 'label'> {
+  const markValue = [...xml.matchAll(/<encodings\b[^>]*>[\s\S]*?<\/encodings>/g)].some(
+    ([encodings]) =>
+      new RegExp(`\\bcolumn=(['"])${escapeRegExp(escapeAttribute(column))}\\1`).test(encodings),
+  );
+  const shelf = [...xml.matchAll(/<(rows|cols)\b[^>]*>([\s\S]*?)<\/\1>/g)].some((match) =>
+    match[2].includes(column),
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

`numberFormatElements` compares the column two different ways. The encodings (mark-value) check re-escapes the column before matching — it builds its regex from `escapeAttribute(column)` — but the shelf check on line 378 matches the raw value: `match[2].includes(column)`.

`resolveShelfField` returns the entity-**decoded** column (the XML parser runs with `processEntities` enabled), so for a measure like `R&D Spend` (or a datasource named `Sales & Ops`), `column` holds a literal `&` while the raw `<cols>`/`<rows>` text carries `&amp;`. The shelf `.includes()` therefore fails while the encodings check still matches.

Consequence: for a shelf-only field with such a name, `numberFormatElements` returns `[]` and `formatWorksheetDocument` wrongly reports *"Field … is not used as a mark value or on a worksheet shelf"*; if the field is also a mark value, the `label` element is silently dropped so the axis/header label never gets formatted. Escaping the column in the shelf check too — e.g. `match[2].includes(escapeAttribute(column))` — would keep the two checks consistent.

src/server.desktop.ts
```diff
@@ -185,7 +185,7 @@ export const DYNAMIC_AUTHORING_TOOL_PROFILE: ReadonlySet<DesktopToolName> =
     'author-set',
     'author-parameter',
     'author-action',
-    'format-labels',
+    'format-worksheets',
```
lauren-jacksonSFDC, line 190: ✨ _Lauren's code review bot:_ ✨

This changes the profile Set member from `format-labels` to `format-worksheets`, but the `DYNAMIC_AUTHORING_TOOL_PROFILE` docstring above (lines 102-103) still lists the author-* verbs as "… author-action (parameter-change wiring), format-labels (mark labels)". `format-labels` is no longer in the Set, so the comment now names a tool the profile doesn't ship — update the docstring to say `format-worksheets`.

src/server.desktop.test.ts
```diff
@@ -186,126 +186,25 @@ describe('DESKTOP_INSTRUCTIONS (generated from DESKTOP_ROUTE_TABLE)', () => {
...
-    // The default served surface includes instructions. Full-profile tool schemas are
-    // pinned separately so intentional route prose does not fund schema growth.
-    // Re-pinned 2026-08-10: explicit single-view writes use apply-worksheet.templatePlan;
-    // preview/no-change requests retain the read-only artifact path.
-    expect(DESKTOP_INSTRUCTIONS).toHaveLength(4_265);
-    expect(dynamicAuthoringTotal).toBe(DYNAMIC_AUTHORING_SURFACE_EXPECTED);
-    expect(dynamicAuthoringTotal).toBeLessThanOrEqual(DYNAMIC_AUTHORING_SURFACE_BUDGET);
-    expect(dynamicAuthoringTotal).toBeLessThanOrEqual(DYNAMIC_AUTHORING_PRODUCT_CEILING);
-    expect(fullToolSurfaceTotal).toBeLessThanOrEqual(FULL_TOOL_SURFACE_BUDGET);
+    // Tool search, not an aggregate tools/list payload, owns discovery. Keep only the
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

This comment ("Tool search, not an aggregate tools/list payload, owns discovery. Keep only the instruction pin and per-tool budget…") restates the rationale already given at lines 194-195 ("Aggregate tools/list size is no longer a product contract: TAS discovers tools through search. Keep the selected profile and its instruction text pinned, plus per-tool caps below.") ~11 lines up, and this `describe` block holds only the one test. Consider dropping this second copy — one statement of the rationale is enough. (Minor.)

lauren-jacksonSFDC (approved):

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: Fixed in eabdfdd7. Every failure after a worksheet mutation now reports the exact number of earlier worksheets that may already be updated. Regressions cover apply, readback, unsettled readback, and drift paths.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: Fixed in eabdfdd7. Formatting keys now compare decoded XML attribute values, so `&quot;` and literal quotes are equivalent. The regression uses a live-shaped currency reserialization.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: This first contract stays intentionally narrow: $, €, £, and ¥ only. The caller must select the symbol; the tool never infers locale. Broader symbols should be added only with a reviewed public contract.

_Responded by MattGPT for Matt Filbert._

mattcfilbert: I checked the production tree and found no second definition of this contract. Keeping these schemas local avoids a premature shared abstraction; we should extract them when another tool actually shares the contract.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/runDashboardBatch.ts
mattcfilbert: Fixed in eabdfdd7. `run-dashboard-batch` now uses the shared `sessionParam({ max: 64 })` contract and retains the KPI ordering guidance. The tool remains at the existing 1,020-byte cap.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: Fixed in 113668c0. The tool resolves every selector to a worksheet ID and rejects duplicate IDs before any document read or apply. A name-plus-ID regression covers the partial-write case.

_Responded by MattGPT for Matt Filbert._

mattcfilbert: Agreed. The public selector still accepts a worksheet name or ID, but every write uses the resolved worksheet ID. The name is retained only for the response.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: Fixed in 17de4db7. A friendly field name that matches more than one canonical shelf pill now fails before apply and lists the candidates; an exact canonical reference remains allowed.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/style/formatWorksheets.ts
mattcfilbert: Fixed in 1a32dade. Shelf text is decoded before field comparison. Regressions cover entity-encoded names and alternate encodings.

_Responded by MattGPT for Matt Filbert._

src/server.desktop.ts
mattcfilbert: Fixed in 113668c0. The profile description now names `format-worksheets`.

_Responded by MattGPT for Matt Filbert._

src/server.desktop.test.ts
mattcfilbert: Fixed in 113668c0. The duplicate rationale was removed.

_Responded by MattGPT for Matt Filbert._

PR #823, opened 2026-08-22
feat(desktop): compose fixed-size dashboard stories

mattcfilbert:
Blank stories can now be composed from ordered, same-size dashboards. The rule is: produce a fixed-size story, preserve populated stories unless replacement is explicit, and verify the saved point order and targets. Approving this PR accepts that boundary.

What changes

- Adds `compose-story` for 1–12 distinct dashboard points.
- Builds caption navigation and arrows, then verifies size, order, captions, and targets after write.
- Forces `sizing-mode='fixed'`; automatic or range sizing cannot pass as a verified result.
- Keeps exact replay idempotent and rejects source drift.

What stays out

- No dashboard creation, inferred narrative, styling, or visual correction.
- No silent replacement of existing story points.

Proof

- Direct MCP smoke on a disposable packaged Superstore workbook composed a two-point 1400×1000 story, then replayed it successfully without changing the result.
- A regression covers the former false-success case for an automatic-size story.
- Canonical gate: 391 test files, 6,226 tests, lint, typecheck, Desktop build, and lockstep checks.

Stack: based on #822; #824 is based on this PR.

_Updated by MattGPT._


src/tools/desktop/authoring/sheets/composeStory.ts
```diff
@@ -0,0 +1,325 @@
...
+    })
+    .join('');
+  const nextStoryPoints = `<story-points>${renderedPoints}</story-points>`;
+
+  return {
+    ok: true,
+    xml: storyboardXml
+      .replace(size[0], nextSize)
+      .replace(flipboard[0], nextFlipboard)
+      .replace(storyPoints[0], nextStoryPoints),
```
myu404: 🤖 **Reviewed by MichaelGPT**

`composeStoryDocument` writes the new XML with `String.prototype.replace(searchString, replacementString)` (`.replace(storyPoints[0], nextStoryPoints)`). When the **replacement** string contains JavaScript's `$` patterns (`$&`, `$'`, `$` + backtick, `$$`), they are interpreted, not inserted literally. Dashboard names/captions are user-controlled and flow into `nextStoryPoints` via `escapeXml`, which escapes `& < > " '` but **not `$`** (nor backtick). Because each value is immediately followed by its closing `'` delimiter, any value ending in `$` produces the `$'` pattern.

Verified against the actual code with node:
- `Revenue $` → `captured-sheet='Revenue ` followed by the text *after* the match spliced in (`$'`).
- `A$&B` → the entire matched story-points block re-inserted (`$&`).
- **`Cost $$5` → collapses to `Cost $5` in *both* the composed XML and the readback**, so `storyMatches` passes and the tool returns `verified: true` while the story point references a non-existent sheet — silently defeating the verify-after-write contract.

Fix is small: use a function replacer for all three transforms — `.replace(size[0], () => nextSize)`, `.replace(flipboard[0], () => nextFlipboard)`, `.replace(storyPoints[0], () => nextStoryPoints)` — so text is inserted literally; or escape `$` in attribute values. Add a test with a `$`-bearing name (including the `$$` case) to lock it.

src/tools/desktop/authoring/sheets/composeStory.ts
```diff
@@ -0,0 +1,325 @@
...
+    ['nav-type', 'caption'],
+    ['show-nav-arrows', 'true'],
+  ] as const) {
+    nextFlipboard = upsertAttribute(nextFlipboard, name, value);
+  }
+
+  const renderedPoints = request.points
+    .map((point, index) => {
+      const caption = point.caption ?? point.dashboard;
+      return `<story-point captured-sheet='${escapeXml(point.dashboard)}' caption='${escapeXml(caption)}' id='${index + 1}' />`;
```
myu404, line 311: 🤖 **Reviewed by MichaelGPT**

When `caption` is omitted it defaults to the resolved dashboard name (`point.caption ?? point.dashboard`), which can be up to 255 chars — bypassing the `caption` schema's `.max(80)` that's enforced for explicit captions. Suggest truncating/validating the defaulted caption to the same 80-char bound, or documenting that the max applies only to caller-supplied captions.

myu404 (changes requested):
## 🤖 **Reviewed by MichaelGPT** — Recommendation: **Request changes**

On the whole this delivers: `compose-story` is registered across all five points with matching byte-pin/count updates, and the core contracts hold — sizing is forced `fixed` and the readback **truly** fails closed on automatic sizing (`readFixedSize` returns `undefined` for `sizing-mode !== 'fixed'`, so a fixed intent can't match, confirmed by the `persistAutomaticSizingMode` regression test). Distinct-dashboard, same-size, populated-story refusal, explicit `replaceExisting`, exact-replay idempotency, and the sha256 drift-under-lock guard are all implemented and tested.

I'm requesting changes for **one `major` correctness bug** (details inline): a `$` in a user-supplied dashboard name or caption is interpreted as a JavaScript replacement pattern in `String.prototype.replace`, corrupting the written XML — and in the `$$` case it can pass readback verification while writing a wrong `captured-sheet`, silently defeating the "verified implies correct" contract. The fix is a one-liner (function replacer). One `nit` also noted inline. This is the middle of the stack, so the fix should land before the dependent navigation PR merges.

src/tools/desktop/authoring/sheets/composeStory.ts
```diff
@@ -0,0 +1,325 @@
...
+    })
+    .join('');
+  const nextStoryPoints = `<story-points>${renderedPoints}</story-points>`;
```
nawartamawi, line 314: [P2] Reconcile point-scoped story zones during replacement

Replacing only the story-points container preserves sibling zones keyed by story-point-id and flipboard-zone-id. Real Tableau stories use these zones for point-specific descriptions and floating objects. Because this code regenerates point IDs from 1, old annotations can attach to unrelated new points or become dangling when the new story is shorter. Remove or reconcile those zones during explicit replacement, or reject stories that contain them.

src/tools/desktop/authoring/sheets/composeStory.ts
```diff
@@ -0,0 +1,325 @@
...
+            runValidation(composed.xml, 'dashboard').issues,
+          );
+          if (introduced.length > 0) {
+            return new ArgsValidationError(
+              `The composed story failed preflight: ${introduced.map(({ message }) => message).join(' ')}`,
+            ).toErr();
+          }
+
+          if (!alreadyMatches) {
+            const sourceHash = sourceSha256(source.value.xml);
```
nawartamawi, line 164: [P2] Revalidate dashboard dependencies under the apply lock

The selected dashboard documents determine captured-sheet names and the common story size, but they are read only during preflight. Inside the apply lock only the storyboard source is re-read. If a selected dashboard is resized, renamed, or deleted in that interval, the tool can apply and verify a candidate built from stale dimensions or targets. Fingerprint and re-read every selected dashboard before applying.

src/tools/desktop/authoring/sheets/composeStory.ts
```diff
@@ -0,0 +1,325 @@
...
+    ['maxwidth', request.width],
+    ['minheight', request.height],
+    ['minwidth', request.width],
+  ] as const) {
+    nextSize = upsertAttribute(nextSize, name, String(value));
+  }
+
+  let nextFlipboard = flipboard[0];
+  for (const [name, value] of [
+    ['active-id', '1'],
```
nawartamawi: [P3] Keep active point out of content idempotence

The request does not select an active story point, but composition always forces active-id 1 and storySignature includes that UI state. Merely viewing point 2 makes an otherwise identical populated story fail replay unless replaceExisting is authorized. Preserve a still-valid active ID or exclude it from request-equivalence checks.

nawartamawi (changes requested):
Requesting changes for two additional correctness issues: replacement leaves point-scoped zones attached to regenerated IDs, and dashboard size/name dependencies are not revalidated under the guarded apply. I also left one non-blocking active-point idempotence follow-up inline.

src/tools/desktop/authoring/sheets/composeStory.ts
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

**XML corruption: `String.replace` expands `$` patterns in the replacement string.**

`composeStoryDocument` builds the output with `storyboardXml.replace(storyPoints[0], nextStoryPoints)` (and the two `.replace()` lines above it). Even though the search argument is a plain string, JavaScript still interprets the special replacement patterns (`$&`, `` $` ``, `$'`, `$$`, `$n`) inside the *replacement* string. `escapeXml` escapes `& < > " '` but never `$`, so those sequences reach `.replace()` intact.

Because `nextStoryPoints` embeds the user-supplied `caption` and the resolved dashboard `captured-sheet` name, a value like `Sales $& Marketing` makes `$&` expand to the entire matched `<story-points>…</story-points>` node — splicing it inside the `caption='…'` attribute and producing malformed XML that is then handed to `applyStoryboardDocument`. It also fires on tamer input: an apostrophe escapes to `&apos;`, so a caption `A$'B` becomes `A$&apos;B`, which contains `$&`.

**Fix** — use the function form of the replacement, whose return value is used verbatim (no `$` expansion). That's already the idiom in this file's own `upsertAttribute` (line 309) and in the sibling `formatWorksheets.ts`:

```ts
    xml: storyboardXml
      .replace(size[0], () => nextSize)
      .replace(flipboard[0], () => nextFlipboard)
      .replace(storyPoints[0], () => nextStoryPoints),
```

lauren-jacksonSFDC (approved):

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in f92c4a63. Function replacers preserve dollar signs literally; regressions cover both `$$` and `$&` captions.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in 751caa68. An implicit caption now obeys the same 80-character limit as an explicit caption and fails before apply with a request for a shorter caption.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in 14903c46. A populated story with point-scoped mappings is rejected before apply unless the request is an exact replay; the regression proves no mutation occurs.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in 14903c46 and df2e5f32. The tool relists, rereads, and hashes every selected dashboard under the apply lock. Regressions cover rename, deletion, and layout drift.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in 751caa68. Composition preserves the current active ID when it still exists in the new point set and falls back to 1 only when it is missing or invalid. An active-point-2 exact replay now skips apply.

_Responded by MattGPT for Matt Filbert._

src/tools/desktop/authoring/sheets/composeStory.ts
mattcfilbert: Fixed in f92c4a63. Function replacers preserve dollar signs literally; regressions cover both `$$` and `$&` captions.

_Responded by MattGPT for Matt Filbert._

PR #824, opened 2026-08-22
feat(desktop): add safe dashboard navigation

mattcfilbert:
Dashboard navigation is now allowed only when every selected dashboard has a safe, unoccupied top title band and unambiguous existing navigation. The rule is fail closed: preserve all user zones and refuse uncertain geometry or collisions. Approving this PR accepts that boundary.

What changes

- Adds `set-dashboard-navigation` for 2–4 dashboards using native root-level buttons and real dashboard window IDs.
- Traverses nested layout zones when checking overlap and existing navigation.
- Allocates IDs from exact `zone id` attributes and parses `<zone>` separately from `<zone-style>`.
- Verifies button type, root parentage, geometry, label, and target after write.
- Keeps exact replay idempotent while still checking for newly introduced overlaps.

What stays out

- No font, color, theme, or broader layout changes.
- No replacement of conflicting user-authored navigation.

Proof

- Direct MCP smoke on a disposable packaged Superstore workbook created two dashboards, applied navigation, verified both dashboards, and replayed the same request successfully.
- That smoke exposed the `<zone-style>` parser bug; the live-shaped regression and exact-tag fix are included here.
- The same smoke composed and replayed the downstream fixed-size story, proving the full stack works together.
- Canonical gate: 392 test files, 6,245 tests, lint, typecheck, Desktop build, and lockstep checks.

Stack: based on #823, which is based on #822.

_Updated by MattGPT._


src/tools/desktop/authoring/sheets/setDashboardNavigation.ts
```diff
@@ -0,0 +1,659 @@
...
+
+          const completed: Array<{ dashboard: string; links: string[]; verified: true }> = [];
+          for (const dashboard of prepared) {
+            if (!navigationMatches(dashboard.sourceXml, dashboard.intendedXml)) {
+              const sourceHash = sourceSha256(dashboard.sourceXml);
+              const applied = await withApplyLock(async () => {
+                const latest = await executor.getDashboardDocument(dashboard.id, extra.signal);
+                if (latest.isErr()) return { kind: 'error' as const, error: latest.error };
+                if (sourceSha256(latest.value.xml) !== sourceHash) {
+                  return { kind: 'drift' as const };
```
myu404, line 155: 🤖 **Reviewed by MichaelGPT**

Three fail-closed branches have no tests: the `sourceSha256` drift return (this line), the `applyDashboardDocument` Err branch (~line 166), and the preflight-rejection gate on `introducedBlockingValidationIssues` (~lines 129–135). The tool-level tests cover the happy path, UUID-resolution failure, and three readback rejections — but the `applyDashboardDocument` mock always returns `Ok` and `getDashboardDocument` returns the unchanged documents map, so drift never triggers. Suggest adding tests for a `sourceSha256` mismatch, an apply Err, and a preflight-validation regression to lock the safety gates.

src/tools/desktop/authoring/sheets/setDashboardNavigation.ts
```diff
@@ -0,0 +1,659 @@
...
+                signal: extra.signal,
+              });
+              if (!readback.ok) return new DesktopCommandExecutionError(readback.error).toErr();
+              if (!readback.settled) {
+                return new XmlModificationError(
+                  `Desktop accepted navigation for "${dashboard.name}", but the title band or buttons did not survive readback.`,
+                ).toErr();
+              }
+            }
+            completed.push({ dashboard: dashboard.name, links: dashboard.links, verified: true });
```
myu404, line 187: 🤖 **Reviewed by MichaelGPT**

The idempotent-skip path reports `verified: true` without an apply or readback: the `if (!navigationMatches(dashboard.sourceXml, dashboard.intendedXml))` block guards apply/`pollReadback`, but the `completed.push({ …, verified: true })` sits *outside* it. It's defensible — `sourceXml` was live-read at prepare time so the document already matched — but it conflates "verified by readback" with "already matched at prepare." Suggest distinguishing the skip case or documenting the semantics.

myu404 (approved):

`set-dashboard-navigation` delivers its claims. It's consistently registered (`toolName.ts`, `tools.ts`, `DYNAMIC_AUTHORING_TOOL_PROFILE`, `instructions.ts`) with matching byte-pin and vocabulary updates. The fail-closed gates — schema, window-UUID resolution, full-width top title-band detection, overlap, unknown geometry, and conflicting existing navigation — are correct and mostly well-tested.

The `<zone>` vs `<zone-style>` parser fix is real and regression-tested (boundary lookahead so `zone` never consumes `zone-style`). The verify-after-write signature covers title geometry plus button type/geometry/action/caption/parentage/order, so drift is rejected and replay is idempotent.

No blocker- or major-severity defects. Two non-blocking observations inline (one `minor` test-gap, one `nit`).

src/tools/desktop/authoring/sheets/setDashboardNavigation.ts
```diff
@@ -0,0 +1,659 @@
...
+    rootChildren: children,
+    layout,
+    layoutInner,
+    titleBand,
+    titleHeight: title.height,
+    titleWidth: title.width,
+    expectedButtons: validButtons,
+  });
+  if (!safety.ok) return undefined;
+  return JSON.stringify({
```
nawartamawi, line 541: [P1] Verify preservation of every pre-existing dashboard zone

navigationSignature serializes only title geometry and navigation buttons. If Desktop readback drops or mutates a pre-existing chart below the title band, removing that clear zone does not affect the safety result or either signature, so navigationMatches can settle and the tool returns verified true despite content loss. Compare preserved non-navigation subtrees while allowing only the intended title-width change and newly added button zones.

nawartamawi (changes requested):
Requesting changes because readback verification does not fingerprint preserved non-navigation zones. A dropped or mutated chart below the title band can still produce verified true, which breaks the safety and preservation contract. Details are inline.

lauren-jacksonSFDC:
I'm not 100% sure this should be a separate tool from activate-sheet (the normal sheet navigation/wrapper around gotosheet API call). Should we not just be hardening activate-sheet instead if it has unsafe behaviour in it?

PR #826, opened 2026-08-24
feat(author-set): support condition (rule-based) sets

myu404:
## Summary

Adds a `condition` mode to `author-set` so callers can create condition (rule-based) sets from a boolean expression (e.g. `SUM([Sales]) >= 60000`), and fixes the `malformed-set-groupfilter` pre-flight validator so it no longer rejects the valid nested filter form. Together these let condition-set requests succeed end to end — previously they failed both at the tool (no condition mode existed) and at validation (a false positive on the correct XML).

## Description

- `author-set` accepts `mode: 'condition'` with a `conditionExpression` (e.g. `SUM([Sales]) >= 60000`) and emits the nested `groupfilter` shape a rule-based set requires: a `function='filter'` group wrapping a `level-members` child. Callers pass the raw expression; the tool handles XML escaping. Cross-mode argument validation is tightened so blank/whitespace optional params are treated as absent.
- The `malformed-set-groupfilter` validation rule is now quote-aware and comment-tolerant. A literal `>` inside a quoted `expression` attribute (e.g. `expression='SUM([Sales])>=60000'`) no longer terminates attribute scanning, and XML comments between a filter group and its child are skipped before the child-presence check.

## Motivation and Context

Condition/rule-based set requests could not be fulfilled before this change: the tool had no condition mode, and the pre-flight validator misread the correct nested form as the malformed flat form and rejected it. The flat (childless) filter form is genuinely unparseable and the set is dropped on reload, so the rule is still needed — it just no longer produces false positives on the valid form.

## Type of Change

- [x] Bug fix
- [x] New feature

## How Has This Been Tested?

- Added unit tests for the new `condition` mode (raw operator input → correctly escaped XML) and for the tightened cross-mode argument guards.
- Added detector tests for the expression-first / unescaped-`>` case and the comment-between-filter-and-child case.
- Full existing unit-test suite and lint/format pass locally.


mattcfilbert (approved):
Nice work, Michael. The `condition` mode emits Desktop's own nested `filter-group` shape, and the validator rewrite fixes the false positive on the valid form without letting the flat (childless) form through — I traced the quote-aware scan by hand and it holds. Tests are well-targeted and CI is green on the head commit.

One thing before merge:

- **Add the version bump.** Run `npm run version:patch` — it's on the PR checklist and every recent desktop PR bumps it; this one doesn't touch `package.json` yet.

One note: a few open PRs also change the `server.desktop.test.ts` byte pins. If one of them merges first, re-derive the `+52` and re-pin against the new base before merging.

Approving on the code — just add the version bump first.

—Posted by MattGPT on Matt's behalf.

PR #828, opened 2026-08-24
fix(desktop): reject continuous fields in worksheet sort

### LINKED ISSUE #827: sort-worksheet reports success for continuous fields Desktop rejects
Desktop can return HTTP 200 for a worksheet sort while logging InvalidField when fieldName resolves to a quantitative or continuous shelf field. tableau-mcp then reports success, and the worksheet can be left with a red pill and an incomplete render.

Expected behavior: sort-worksheet should reject continuous shelf fields before dispatch. Measure-based ranking should use refine-worksheet with operation sort_by_field. Discrete member sorts should continue to use the native sort endpoint.

This was reproduced with SUM(Sales) on a bar chart. The template had already authored the correct descending computed sort, so the follow-up native sort was also unnecessary.

— MattGPT

mattcfilbert:
## Decision

Rule we now keep: sort-worksheet sends native sort requests only for nominal or ordinal shelf fields. Measure-based ranking uses refine-worksheet with operation sort_by_field.

This is compatibility protection for the worksheet sort endpoint. It does not close the broader External Client API issue where Desktop can log a semantic failure while the API reports SUCCEEDED. Native sort receipts therefore say Desktop accepted the request and that the result was not independently verified.

Future changes must keep the type check and native POST under the workbook mutation lock, route measure ranking to the readback-verified computed-sort path, and avoid claiming a native sort landed without evidence.

## Description

- Resolve the exact on-shelf column instance and its declared type.
- Reject quantitative, ambiguous, and unverified fields before any POST.
- Preserve native member sorting and clearSort for discrete fields.
- Resolve calculated fields by their Tableau caption.
- Keep the read, type check, and native sort in the same workbook mutation lock.
- Point measure-ranking requests to refine-worksheet.
- Report native sort acceptance without claiming verified success.
- Merge the #822/#823 stack while preserving its shared resolver contract and tool-search checks.

## Motivation and Context

A live Sales-by-State run sent SUM(Sales) to the native sort endpoint. Desktop logged InvalidField and ran no query, but returned HTTP 200 with a SUCCEEDED operation, so tableau-mcp reported success and left the sheet with a red pill and an incomplete render.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- TDD regressions cover continuous rejection with zero POSTs, discrete sort and clearSort, captioned calculations, ambiguous shelf fields, ordinal collision suffixes, unknown-type failure, mutation locking, and truthful terminal receipts.
- scripts/agent-check: all five checks green; 6,281 tests passed.
- A fresh adversarial stack review found the missing mutation lock; the fix and regression are included.
- Live post-fix Desktop smoke has not been rerun.

## Related Issues

Closes #827.

## Checklist

- [x] Bumped package version to 2.66.6 with npm.
- [x] Updated the public tool copy; no separate docs change is needed.
- [x] Added tests that prove the fix.
- [x] New and existing tests pass locally.
- [x] No breaking changes.
- [x] Read and followed the contributing guidelines.

Approving this PR blocks invalid native sort inputs and keeps native sort receipts within what MCP can prove. The ECA false-success behavior still needs an upstream fix.

— MattGPT


src/tools/desktop/api/resolveShelfField.ts
```diff
@@ -13,27 +14,60 @@ import { listFields } from '../../../desktop/metadata/index.js';
...
   const wanted = stripBrackets(trimmed).toLowerCase();
   for (const column of onShelf) {
-    const parsed = parseCanonicalColumnRef(column);
-    const localName = parsed?.localFieldName;
-    if (localName !== undefined && localName.toLowerCase() === wanted) {
-      return { ok: true, column };
+    const declaration = metadata.get(column);
+    const localNames = declaration?.localNames ?? [parseCanonicalColumnRef(column)?.localFieldName];
+    if (localNames.some((localName) => localName?.toLowerCase() === wanted)) {
+      return { ok: true, column, type: declaration?.type };
```
nawartamawi: [P2] Reject ambiguous friendly-name matches

`fieldName` may be a friendly base or caption name, but this loop returns the first shelf pill with that name. A worksheet can legitimately contain multiple date levels for the same field—for example, Month and Year `Order Date` pills on different shelves—so sorting `Order Date` can silently target whichever entry appears first and still report success. Please collect all friendly-name matches and return an ambiguity error listing their canonical tokens unless the input exactly matches one canonical token, with regression coverage for two shelf pills sharing the same friendly name.

nawartamawi (changes requested):
Requesting changes for one correctness issue: friendly-name resolution can select the wrong shelf pill when multiple pills share a caption or base name. Exact canonical references remain unambiguous; friendly lookup should reject multiple candidates and surface the canonical options. CI is green, and the rest of the guard behavior looks solid.

src/tools/desktop/api/resolveShelfField.ts
mattcfilbert: MattGPT: Fixed in 01157ac5. Friendly names now resolve only when exactly one shelf token matches; ambiguous names return every matching canonical token and exit before the sort POST. Exact canonical tokens still work. Full local agent-check passed: 6,201 tests plus lint, typecheck, desktop build, and lockstep checks. Ready for rereview.

mattcfilbert: fixed

lauren-jacksonSFDC (approved):

PR #830, opened 2026-08-24
fix(desktop): build dense scatter plots with detail and color

mattcfilbert:
## Decision

Rule we now keep: when a scatter or bubble request asks for many marks plus categorical color, use a bounded semantic Detail finer than Color and verify that both encodings persist. This is a general authoring rule, not an image-only exception.

This PR does not add an image tool or transport. It skips the repair for one-mark-per-category views and continuous color. If no useful lower grain exists, the agent reports that limit or asks the user; it never invents a field.

Future changes must resolve an exact `column_ref` before adding Detail, preserve the Detail/Color readback, and avoid claiming rendered mark density was verified.

## What changed

- Read the current worksheet before repairing a dense scatter or bubble chart.
- List real workbook fields and choose a bounded discrete Detail finer than Color.
- Add the exact resolved field, apply the worksheet, and verify the two encodings structurally.
- Preserve the merged tool-search contract and bump the package to `2.66.7`.

## Evidence

- `scripts/agent-check`: lint, typecheck, 6,282 tests, Desktop build, and lockstep hashes passed.
- Focused instruction and server-surface tests: 126 passed.
- Live image-reference smoke previously rendered a 9,340-mark Sales vs. Profit scatter with Sub-Category on Detail and Category on Color.
- Final integrated code review found no P1 or P2.

Approving this PR makes dense scatter and bubble requests repair missing grain and categorical color with real workbook fields, while keeping rendered density explicitly unverified.

— MattGPT


src/desktop/instructions.ts
```diff
@@ -66,6 +66,11 @@ export const DESKTOP_ROUTE_TABLE: readonly DesktopInstructionEntry[] = [
     ],
     requiredEvidence: ['each applied worksheet returns a success receipt'],
   },
+  {
+    kind: 'prose',
+    id: 'dense-scatter-bubble',
+    text: 'Many/diverse scatter/bubble + categorical color overrides bind-template “Done—no further tool calls needed”: get-worksheet-xml(mode:inline) → add-field → apply-worksheet. Detail: bounded, semantic, discrete, finer than Color; absent/equal/no-finer Detail is missing—replace/supplement. Claim neither until apply receipt proves both. Skip one-mark-per-category, continuous color, no semantic lower grain.',
```
nawartamawi: [P2] Resolve the Detail field before `add-field`

`add-field.columnRef` accepts only an exact datasource-qualified reference obtained through field resolution. `get-worksheet-xml` normally includes dependencies only for fields already used on the sheet, so the missing finer-grain field this rule wants to add frequently is unavailable from that read. Although `list-available-fields` appears in the generic artifact fallback, this repair runs after a terminal bind and explicitly advertises a different three-step sequence, leaving the agent to invent a reference or fail first. Add `resolve-field` for a named field, or `list-available-fields`/`search-workbook-fields` when selecting a lower grain, before `add-field`, and pin that sequence in the test.

nawartamawi (changes requested):
Requesting changes for one repair-path blocker. The dense scatter/bubble rule is sound, but the advertised sequence does not obtain the canonical column reference needed for a missing lower-grain Detail field. Please add field resolution before `add-field`.

Non-blocking: when no suitable lower grain exists, explicitly report the limitation or ask the user; also keep the claim scoped to structural encoding persistence because the apply receipt does not verify rendered mark density.

src/desktop/instructions.ts
mattcfilbert: MattGPT: Addressed in b5ae3bc5. The route now discovers a lower-grain field before add-field, uses its exact column_ref, reports or asks instead of inventing when none exists, and limits readback claims to Detail/Color persistence; rendered density remains unverified. Local agent-check passed: 6,187 tests plus lint, typecheck, desktop build, and lockstep checks. Ready for rereview.

lauren-jacksonSFDC (approved):

PR #833, opened 2026-08-25
fix(desktop): fail a worksheet apply when Tableau drops part of the document

lauren-jacksonSFDC:
## Decision

A worksheet apply now **fails** whenever the External Client API reports the apply SUCCEEDED but attaches a `document-warning` — Desktop's signal that it accepted the document while dropping part of what we submitted (a filter, a shelf field, an encoding). Previously that warning was discarded on every apply path, so the tool returned success while the rendered chart silently did not match the intent. This is a rule the worksheet apply paths now keep, not a one-off.

This is the bug behind the report: `apply-worksheet` returning success while Tableau dropped the Country/Region filter and calc-heavy rows/cols.

## Why hard-failing is safe

Investigation of every `document-warning` source reachable from the document-apply path in the monolith found they are all the same class: *"the client asked for X, Desktop dropped/ignored/fell back."* There are no benign notices on this path, and a clean apply emits an **empty** warnings list. So failing on any non-empty `document-warning` cannot over-trigger on a healthy apply.

The per-node Tableau code is collapsed to the single generic `document-warning` wire code before it reaches us, so the warning's **message text** is the only field-level detail that survives — we surface it verbatim for the agent to act on.

## What changed

- **Consume the drop signal.** Thread `documentWarnings` through `applyWorkbookText` and the per-sheet apply, and hard-fail with a new `document-warning` error on the three worksheet apply routes (per-sheet, whole-workbook, artifact). This is complementary to the existing post-apply readback verifier: it is an authoritative in-band signal that does not depend on a re-read succeeding.
- **Keep the real error code on a rejection.** On the client-rejected 4xx (RFC-9457 Problem) path, the specific Tableau code rides in the `tableauErrorCode` extension member while `code` is the generic `operation-failed`. We now surface `tableauErrorCode` when present instead of the generic code, so a rejection reports the code Tableau actually raised.

## Deliberate scope

Only the three **worksheet** apply routes fail on a `document-warning` here. `apply-workbook`, `apply-dashboard`, and `bind-template` go through `loadWorkbookXml`/`loadDashboardXml`, whose callers (dashboard compose, batch) run a retry classifier that would mistake this new error for a retry-safe one and loop. Extending the hard-fail to those paths is a documented follow-up gated on that classifier work.

## Testing

- New unit tests assert the hard-fail on all three worksheet routes (per-sheet, whole-workbook, artifact) and that the dropped-node message is surfaced.
- New unit test asserts `tableauErrorCode` is preferred over the generic `code` on a Problem rejection; contract test updated for the new schema extension member.
- `npx tsc --noEmit`, `npm run lint`, and `npm run build:desktop` all clean. Full unit suite green except a pre-existing flaky timeout test (`unreachableInstance.test.ts`, `timeoutMs: 60` on real timers) unrelated to these paths.

## Verified live

Smoke-tested against Tableau Desktop (External Client API 0.2.9) with Sample - Superstore:

- **No-regression control:** a clean worksheet apply still returns success — the empty warnings list does not trip the new hard-fail.
- **Reproduced the bug:** applying a worksheet with a quantitative (range) filter on a real string dimension (`Category`) made Desktop return SUCCEEDED while silently dropping the filter (*"Error parsing filter for field 'Category', ignoring filter"*). Before this change the tool reported success; with the fix it now hard-fails with the dropped-filter message, and a readback confirmed the filter was in fact dropped.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.66.2",
+  "version": "2.66.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "2.66.2",
+      "version": "2.66.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@modelcontextprotocol/ext-apps": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.66.2",
+  "version": "2.66.3",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/desktop/externalApi/externalApiContract.test.ts
+++ b/src/desktop/externalApi/externalApiContract.test.ts
@@ -287,7 +287,7 @@ describe('external client API contract (captured openapi fixture)', () => {
   describe('Problem ↔ problemResponseSchema', () => {
     const problem = specSchema('Problem');
 
-    it('declares every spec property (extras `type`/`detail` are RFC-9457 members additionalProperties admits)', () => {
+    it('declares every spec property (extras `type`/`detail` are standard RFC-9457 members, `tableauErrorCode` Tableau’s extension member)', () => {
       const missing = Object.keys(problem.properties ?? {}).filter(
         (key) => !declaredKeys(problemResponseSchema).includes(key),
       );
@@ -295,7 +295,7 @@ describe('external client API contract (captured openapi fixture)', () => {
       const extras = declaredKeys(problemResponseSchema).filter(
         (key) => !(key in (problem.properties ?? {})),
       );
-      expect(extras.sort()).toEqual(['detail', 'type']);
+      expect(extras.sort()).toEqual(['detail', 'tableauErrorCode', 'type']);
     });
 
     it('PROBLEM_CODES equals the spec x-extensible-enum exactly', () => {
--- a/src/desktop/externalApi/externalApiHttp.ts
+++ b/src/desktop/externalApi/externalApiHttp.ts
@@ -505,6 +505,7 @@ async function mapErrorResponse(res: Response): Promise<ExternalApiError> {
       code: problem.data.code,
       title: problem.data.title,
       detail: problem.data.detail ?? (text || undefined),
+      tableauErrorCode: problem.data.tableauErrorCode,
     };
   }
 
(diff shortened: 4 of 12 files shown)


mattcfilbert (commented):
**MattGPT review** — posted by Matt's assistant (MattGPT), andy-review lens.

Nice fix, Lauren. The important call here is right: a worksheet apply that Tableau only partly honored used to return success with a warning, so the agent trusted a chart that didn't match the intent. This turns that into a hard error whose message names what got dropped and tells the agent to fix the XML and re-apply — so the repair path actually runs instead of trusting a lossy result. It rides the existing `load-worksheet-xml-error` path cleanly (all three worksheet tools and the artifact path surface it as an error), and preferring the real Tableau error code over the generic "operation failed" is a good touch. CI is green.

Two things before merge:

**Please — bump the version and coordinate the number.** The PR is still at `2.66.2`, which is the current `feature/desktop` version, so `npm run version:patch` hasn't run yet. It needs to. One catch: #828 and #822 are both also sitting at `2.66.2` right now, so all three of you land on `2.66.3` the moment you bump. Two PRs making the same `2.66.2 → 2.66.3` edit merge with no conflict and CI has no version guard, so whoever merges second silently ships a second `2.66.3` with different code. Whoever lands first takes `2.66.3`; the rest bump again.

**Question — are all of these warnings really "dropped part of the document"?** `documentWarningOutcome` hard-fails on any `ExecuteCommandWarning` from the apply, with no filtering by code. That's the right default if every warning on this path means Tableau dropped something. But if the External Client API can return a benign or informational warning on an apply it fully honored, this now fails a correct chart and tells the agent to "fix" it — a false failure that could send it into a repair loop. Do we know the warnings on this path are all loss/drop, or is it worth gating on the specific codes that mean a drop? Your call — I can't see the API's warning taxonomy from here.

Scope looks deliberate and fine: only the worksheet routes hard-fail; the workbook and dashboard apply paths still ignore the warnings (they compile against the new return type, they just don't act on it). Fair line for this PR — worth a follow-up if those paths can drop content too.

lauren-jacksonSFDC:

Re @mattcfilbert's [review](https://github.com/tableau/tableau-mcp/pull/833#pullrequestreview-5014458816):

Thanks Matt.

1. **Version** — bumped to `2.66.3` (pushed). #828 and #822 are both still at `2.66.2`, so it's a merge-race: first to land takes `2.66.3`, the rest re-bump.

2. **Benign warnings** — verified in the monolith that every warning on the apply path (`WorkbookApplyDocument`/`SheetApplyDocument`) is a drop: the only source is `GetWorkbookImportWarnings()`, wrapped as the single `document-warning` code, and a clean apply returns an empty list (so a correct chart never false-fails). Keeping fail-on-any deliberately — the spec leaves `code` free-form so an allow-list can't be authoritative, and missing a drop under a new code would silently reintroduce this bug, whereas a benign warning would be a loud, fixable false-fail.

3. **Scope** — agreed; only the worksheet routes hard-fail, and the workbook/dashboard paths compile against the new return type but don't act on it yet. Worth a follow-up if those paths can drop content too.

PR #838, opened 2026-08-25
fix(oauth): pin TLS SNI to original hostname when fetching CIMD client metadata

### LINKED ISSUE #837: CIMD client metadata fetch fails (403) for client_id hosts behind SNI-routing CDNs (e.g. Cloudflare)
## Summary

Authorizing with a CIMD `client_id` whose host is fronted by a CDN that routes by TLS SNI (e.g. Cloudflare) fails with:

```json
{"error":"invalid_request","error_description":"Unable to fetch client metadata"}
```

This reproduces for `client_id=https://claude.ai/oauth/claude-code-client-metadata`, used by Claude Code / Claude Desktop, when authorizing against a self-hosted tableau-mcp server.

## Root cause

In `getClientFromMetadataDoc()` (`src/server/oauth/authorize.ts`):

1. The hostname of the client metadata URL is DNS-resolved to a raw IP address (DNS-rebinding SSRF mitigation).
2. The URL's hostname is rewritten to that IP.
3. `isSSRFSafeURL` validates the IP-based URL.
4. `axios.get()` requests the IP-based URL, overriding only the `Host` header to the original hostname.

Step 4 connects to the raw IP over HTTPS. Node/axios derive the TLS **SNI** from the request URL's host — which is now the IP, not the original hostname. CDNs like Cloudflare route and select which certificate to present based on SNI, not the `Host` header. Since the SNI is an unrecognized IP, Cloudflare's edge rejects the request (403), which the server logs as e.g.:

```
Failed to fetch client metadata from https://claude.ai/oauth/claude-code-client-metadata
AxiosError: Request failed with status code 403
url: 'https://<resolved-ip>/oauth/claude-code-client-metadata'
```

...and surfaces to the OAuth caller as the generic `Unable to fetch client metadata`, indistinguishable from a real network/DNS failure.

This affects **any** CIMD client whose `client_id` domain sits behind an SNI-routing CDN — not specific to claude.ai — so it's likely to recur for other MCP clients adopting CIMD (a newer part of the [MCP authorization spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents)).

Related: #193 (proposed disabling CIMD entirely as a workaround; closed without a fix for CIMD-enabled deployments).

## Fix

Opened as # — pins the outbound TLS SNI (via `https.Agent({ servername })`) to the original hostname while still connecting to the pre-resolved IP, so the existing SSRF protection is unaffected but CDN-fronted client_id hosts resolve correctly.

## Environment

- tableau-mcp: self-hosted, OAuth/CIMD enabled
- Reported via an enterprise deployment behind a corporate proxy; confirmed via manual `axios.get()` against the raw resolved IP that the 403 is independent of proxy config — it reproduces identically from outside that network.

jenwitteng:
## Summary

Fixes #837.

`getClientFromMetadataDoc()` resolves the CIMD `client_id` metadata URL's hostname to an IP address (DNS-rebinding SSRF mitigation) and rewrites the request URL to that IP before fetching it, overriding only the `Host` header to the original hostname.

For HTTPS targets, Node/axios derive TLS **SNI** from the request URL's host, which is now the raw IP. CDNs that route/serve certificates by SNI (e.g. Cloudflare) reject requests with an unrecognized IP-based SNI, regardless of the `Host` header — surfacing as `invalid_request: Unable to fetch client metadata` to the OAuth caller. This reproduces for `client_id=https://claude.ai/oauth/claude-code-client-metadata` (Claude Code / Claude Desktop) against any self-hosted tableau-mcp server, and will affect any CIMD client whose `client_id` host sits behind a similar CDN.

## Fix

Pin the outbound TLS SNI to the original hostname via `https.Agent({ servername: originalHostname })`, passed as `httpsAgent` on the existing `axios.get()` call. The connection still only ever talks to the pre-resolved, SSRF-checked IP — `isSSRFSafeURL` and the IP-resolution step are unchanged — so DNS-rebinding protection is preserved. This also fixes TLS certificate hostname validation, which previously validated against the wrong identity (the IP) rather than the original hostname.

## Before / after

- Before: `axios.get(<ip-url>, { headers: { Host: originalHostname } })` → SNI = IP → 403 from SNI-routing CDNs.
- After: same call + `httpsAgent: new https.Agent({ servername: originalHostname })` → SNI = original hostname → CDN routes correctly; SSRF pinning to the resolved IP unchanged.

## Testing

- `tsc --noEmit`: clean.
- `eslint` / `prettier --check`: clean.
- Full unit suite (`npm test`, Node 22): **177 files / 2823 tests pass**, no regressions.
- Added `src/server/oauth/authorize.test.ts`, a new regression test for `getClientFromMetadataDoc` (exported for testability). It mocks DNS resolution, the SSRF check, and axios, then asserts:
  - the outbound request URL is still the resolved IP (SSRF pinning to the vetted IP is unaffected), and
  - the `httpsAgent`'s TLS `servername` is the original hostname, not the IP.
- Verified the test is meaningful: reverting just the `authorize.ts` fix makes this test fail; restoring the fix makes it pass again.
- Manually reproduced the underlying failure by resolving `claude.ai` and issuing a raw `axios.get()` against the resolved IP with only a `Host` header override (403); confirmed the fix resolves it locally with the same `servername`-pinned `https.Agent`.

## Additional info

Root cause and reproduction details are in #837.

### DIFF
--- a/src/server/oauth/authorize.test.ts
+++ b/src/server/oauth/authorize.test.ts
@@ -1,0 +1,63 @@
+import https from 'https';
+import { isSSRFSafeURL } from 'ssrfcheck';
+import { describe, expect, it, vi } from 'vitest';
+
+import { axios } from '../../utils/axios.js';
+import { getClientFromMetadataDoc } from './authorize.js';
+import { getDnsResolver } from './dnsResolver.js';
+
+vi.mock('ssrfcheck', () => ({
+  isSSRFSafeURL: vi.fn(),
+}));
+
+vi.mock('./dnsResolver.js', () => ({
+  getDnsResolver: vi.fn(),
+}));
+
+vi.mock('../../utils/axios.js', async () => {
+  const actual =
+    await vi.importActual<typeof import('../../utils/axios.js')>('../../utils/axios.js');
+  return {
+    ...actual,
+    axios: { create: vi.fn() },
+  };
+});
+
+describe('getClientFromMetadataDoc', () => {
+  it('pins TLS SNI to the original hostname, not the resolved IP', async () => {
+    const resolvedIp = '93.184.216.34';
+    vi.mocked(getDnsResolver).mockReturnValue({
+      resolve4: vi.fn().mockResolvedValue([resolvedIp]),
+      resolve6: vi.fn(),
+    } as unknown as ReturnType<typeof getDnsResolver>);
+    vi.mocked(isSSRFSafeURL).mockReturnValue(true);
+
+    const get = vi.fn().mockResolvedValue({
+      headers: { 'content-type': 'application/json' },
+      data: {
+        client_id: 'https://claude.ai/oauth/claude-code-client-metadata',
+        redirect_uris: ['http://localhost:12345/callback'],
+      },
+    });
+    vi.mocked(axios.create).mockReturnValue({ get } as unknown as ReturnType<typeof axios.create>);
+
+    const result = await getClientFromMetadataDoc(
+      new URL('https://claude.ai/oauth/claude-code-client-metadata'),
+    );
+
+    expect(result.isOk()).toBe(true);
+    expect(get).toHaveBeenCalledTimes(1);
+
+    const [requestedUrl, requestConfig] = get.mock.calls[0];
+
+    // The request is still pinned to the resolved IP (SSRF protection intact).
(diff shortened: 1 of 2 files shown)


jenwitteng:
recheck

PR #840, opened 2026-08-25
@W-23757364 - surface admin capabilities in MCP initialize instructions

Alon-ST-DATA:
## Summary

Fixes **W-23757364** (P2): authenticated site admins asking a **generic** admin-health question — *"what should I pay attention to as an admin on this site?"* — did not get the model to surface or invoke admin-specific capabilities (`query-admin-insights`, the admin prompts). It fell back to general read tools and skipped the admin surface entirely.

**Root cause:** the MCP server sent **no server-level `instructions` string** in the `initialize` handshake, and the admin tool/prompt descriptions were worded around named workflows. So a broad admin prompt had no signal that admin capabilities were relevant. The known workaround — self-declaring *"I am a Tableau Cloud site admin"* — worked only because it keyword-matched the admin tool descriptions. Pure discoverability/signaling gap: the tools themselves work once invoked, and `assertAdmin` gating is unchanged.

## Fix

- **`src/server.web.ts`** — new exported `buildWebInstructions()`: a BASE block always, plus an ADMIN block (names the admin prompts + `query-admin-insights` kinds, and instructs proactive surfacing *even when the user asks broadly*) appended only when `ADMIN_TOOLS_ENABLED=true`. Single source of truth.
- **`src/server.ts`** — base `Server` accepts an optional `instructions` param. Added a **structural guard** (`invariant`): if an `McpServer` is supplied *and* instructions are expected, it asserts the supplied server actually carries them — so the provided-mcpServer path can never silently drop instructions again (fails fast at startup).
- **`src/index.combined.ts`** — the `combined` bundle builds its own `McpServer`; it now constructs it with `buildWebInstructions()`. **Previously the fix no-oped for this shipped variant** (the SDK stores `instructions` only at construction — no post-hoc setter — so a supplied server never received them).
- **`src/tools/web/adminInsights/queryAdminInsights.ts`** + **`src/prompts/{staleContent,userLicenseReclamation,jobOptimization}/inform.ts`** — one umbrella discoverability clause each, as belt-and-suspenders for clients that surface tool/prompt descriptions but not server instructions.

## Not a security/scope change

Discoverability/signaling only. Instructions are advertised on the build-time `ADMIN_TOOLS_ENABLED` flag (pre-auth), but this leaks nothing: when the flag is on the admin tools are already registered and listed to any caller, and **execution is still gated by `assertAdmin` at call time**. Advertising the guidance text exposes nothing beyond the existing `tools/list` / `prompts/list`.

## Testing

| Tier | Result |
|---|---|
| `npm run build` (all variants incl. combined) | clean |
| `npx tsc --noEmit` | clean |
| `npm run lint` | clean |
| Unit (`npx vitest run`) | **2811 passed / 0 failed** |
| e2e (`vitest.config.e2e.ts`, live over stdio) | **9 passed / 0 failed** |

- **Unit** (`src/server.web.test.ts`): flag on → admin guidance present; unset → base only.
- **e2e** (`tests/e2e/server.test.ts`): asserts the **real** `initialize` handshake `instructions` for the `default` variant (admin on/off) **and** the `combined` variant. The combined assertion is a real regression guard — with the `index.combined.ts` fix removed it fails (startup guard fires); restored, it passes.
- **Manual live verification** (read-only, populated admin site): confirmed the composed instructions are on the wire, and every advertised admin capability returns real data — `query-admin-insights` (stale-content: 14 stale items / ~23.8 GB; job-performance; ts-users), `list-users` (74 users), `list-jobs`. No mutations run.

### Out of scope / follow-ups
- **Model tool-selection** (does the LLM, given a generic prompt, now *choose* admin tools) is LLM-behavioral → belongs to the **eval tier**, filed as a separate follow-up (env-gated, lives in the eval clone). Manual e2e proves the signal is present + the surface is live, not the model's choice.
- **Version bump:** this PR sets `4.6.1`. Heads-up — `dev/cyang/w-23864452` also claims `4.6.1` unmerged; whichever lands second must re-bump to `4.6.2`.


Akash-Rastogi (approved):
[Akash's agent] LGTM. Reviewed the full diff:

- `buildWebInstructions()` is a clean single source of truth, and the code correctly handles the SDK reading `instructions` only at `McpServer` construction time.
- The combined-bundle regression is genuinely fixed (`index.combined.ts` builds its `McpServer` with the instructions) and guarded by the fail-loud `invariant` in `server.ts`, so the provided-server path can't silently drop them again — the combined e2e test proves it (fails against pre-fix code).
- Security reasoning holds: advertising the admin guidance leaks nothing (those tools are already in `tools/list` when `ADMIN_TOOLS_ENABLED` is on, and execution stays gated by `assertAdmin`).
- Comprehensive unit + e2e coverage; CI green.

No concerns.

jarhun88 (approved):
Reviewed diff and CI (build all variants, tsc, lint, 2811 unit / 9 e2e tests — all green). Correctness, security posture (assertAdmin still gates execution; no new leak beyond existing tools/list exposure), and test coverage (including a real regression test for the combined-variant bug) all check out. Left one non-blocking robustness note inline on src/server.ts.

— Jaehun Bot

src/server.web.ts
```diff
@@ -36,9 +36,51 @@ export const serverName = 'tableau-mcp';
...
+ * getConfig() mocks. Direct env read keeps composition side-effect-free.
+ *
+ * Both the default/web path (WebMcpServer constructs its own McpServer) and the combined path
+ * (index.combined.ts supplies a pre-built McpServer) MUST feed instructions through this function.
+ * The SDK reads `instructions` only from the McpServer constructor options (it is never settable
+ * post-construction), so a supplied McpServer that was not built with these instructions would
+ * silently drop them — see the guard in server.ts.
+ */
+export function buildWebInstructions(): string {
+  const adminToolsEnabled = process.env.ADMIN_TOOLS_ENABLED === 'true';
```
stephendeoca, line 71: We'll be adding in condition registration. So its possible admin tools are enabled, but a non admin connects and this message is still there for them. This might have sense to start with the base instructions and then add the admin instructions later if there are admin tools being registered.

Alon-ST-DATA: Make sense.
A good follow up item for the conditional tool registration work. Happy to file a tracking work item. Let me know.

PR #842, opened 2026-08-25
feat(desktop): author URL actions on worksheets and dashboards

myu404:
## What

Adds a `url` mode to the `author-action` tool so it can create **URL (hyperlink) actions**, alongside the existing parameter and set actions.

The action is written as a legacy `<action>` element carrying a `<link expression=...>` payload — the shape Tableau classifies as a URL action. (A hand-authored `<command>`-based hyperlink is instead classified as an unknown action type and does nothing.) Supported options:

- **Worksheet source**, **dashboard source** (with optional `<exclude-sheet>` opt-outs), or a worksheet scoped within a dashboard
- Open target: default, browser, or a specific dashboard zone
- Field substitution via `<[Field Name]>` in the URL, with optional URL-encoding

## What it resolves

- **No first-class URL action before this.** There was no supported way to author a URL action; the only workaround produced an action Tableau did not recognize as a URL action, so it silently did nothing. This emits the correct `<link>` shape and verifies it on readback.
- **Source miscategorization.** Worksheet, dashboard, and story names share one namespace. Passing a dashboard name as `sourceWorksheet` (or a worksheet name as `sourceDashboard`) is now rejected up front with guidance to use the correct field. This prevents emitting an invalid `<source worksheet='<dashboard>'>` that Tableau accepts on apply but errors on when the action is later opened for editing.

## Tests

- New unit tests in `authorAction.test.ts` cover the emitted `<link>` shape (worksheet / dashboard / combined sources, browser and specific-zone targets, URL-encoding), duplicate and caption-collision detection, raw-vs-escaped URL handling, and both source-miscategorization rejections.
- Full suite green; the tool-surface byte-budget and emitted-reference guards are updated to match.


src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -492,6 +670,129 @@ function renderSetAction({
...
+    const linkTag = block.match(/<link\b[^>]*>/)?.[0];
+    if (linkTag === undefined) {
+      return false;
+    }
+    const linkExpression = getAttr(linkTag, 'expression');
+    if (linkExpression === undefined) {
+      return false;
+    }
+    // A double-escaped expression (e.g. &amp;lt;[City]&amp;gt;) means a field reference
+    // degraded into a literal string. It still round-trips through unescapeXml back to the
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

The rationale here is inaccurate. `unescapeXml` only unescapes one level, so a double-escaped readback like `expression='...&amp;lt;[City]&amp;gt;'` becomes `...&lt;[City]&gt;` — not the caller's raw `...<[City]>`. So the equality check on line 764 (`unescapeXml(linkExpression) === expression`) already returns `false` for that input *without* this guard; it can't "falsely report success." And because line 169 rejects any pre-escaped input, the only reachable `expression` is raw, which can never round-trip back through a double-escaped stored value.

**Why this matters:** the comment tells a maintainer tracing a URL-action readback failure that the plain equality would falsely succeed, which is impossible for the reachable (raw) inputs. Either restate the guard's real purpose (an explicit belt-and-suspenders rejection of the double-escape signature) or drop it as redundant with line 764.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -492,6 +670,129 @@ function renderSetAction({
   );
 }
 
+// A URL action is the legacy <action> tag whose payload is a <link> (NOT a
+// <command>): monolith GetActionType classifies it URL(3) iff IsLink() is true and
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

A few of the new comments lean on monolith internals a tableau-mcp reader can't resolve and that rot silently — the unlinked, renumber-prone kind the comment standard steers away from. Worth stating the contract directly and moving the monolith symbols/rationale to the PR description:

- `:674` (here) — `GetActionType` / `IsLink()` / `URL(3)`
- `:737` — `Unknown(0)` / `URL(3)`, plus "the agent used to hand-author" (change narration)
- `:259` — `GetWorksheet -> AsWorksheet()` and hex code `0x5CCCC2BD`
- `authorAction.test.ts:653` — "the exact type-0 failure mode"
- `authorAction.test.ts:953` — "the 0x5CCCC2BD reproduction"

The durable contract each of these is really documenting — a URL action is an `<action>` with a `<link>` child (never `<command>`), no `tsl:` prefix — reads fine without the enum ordinals and hex codes. The zoneId comment at `:198` is a good model: it states Tableau's parsing behavior directly, with no internal symbol.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: Good catch — dropped the guard as redundant in 7c7bb440. Pre-escaped URLs are already rejected at validation, so the only reachable expression is raw, and the single-level `unescapeXml` in the final equality check already fails a double-escaped readback. Rewrote the comment to describe that mechanism directly.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: Done in 7c7bb440 — rewrote the `renderUrlAction`, `hasUrlActionWithLink`, and source-classification comments (plus the two test comments) to state the observable XML contract and Tableau's parsing behavior directly, modeled on the zoneId comment, with no internal symbols or status codes.

mattcfilbert (approved):
Approving — this is careful work.

The URL action is written as a `<link>` (not a `<command>`), the escape/unescape round-trip is correct, and the error messages tell the agent exactly how to fix a bad input. The byte budget and the emitted-reference guard are both updated to match, and CI is green. The pre-escaped-URL rejection and the worksheet-vs-dashboard miscategorization guard are good defensive calls. Thanks Michael.

One open question, non-blocking:

Readback (`hasUrlActionWithLink`) confirms the caption-matched `<action>` kept a `<link>` with the right expression and no `<command>` child — that covers the core failure this PR fixes. It does not check that the `<source>` scoping (worksheet / dashboard / `<exclude-sheet>`), the `<url-action-type>` / `<url-action-target>`, or `url-escape` survived apply. For the default target there are no children to verify, so the common path is fully covered; for browser / specific-zone / encoded URLs, if Tableau drops or rewrites those children the tool still returns `Ok` with the requested target echoed back. Intended — trust the pre-apply validation for those — or worth folding source and target into the survival check?

---
🤖 Posted by MattGPT (automation) on Matt's behalf.

myu404:
@mattcfilbert thanks for the review and the question. Intended — and I think leaving readback as-is is the consistent call.

Each mode verifies one defining payload and trusts pre-apply validation for the rest: `set` checks the target group, `parameter` the target parameter, and `url` the `<link>` expression (via unescape-and-compare, which is what the double-escaped test exercises). They don't re-verify their own modifiers or source scoping, so `url` not re-checking source/type/target matches its siblings rather than being a gap.

I also looked at making it a post-apply presence check for the type/target children and backed off: those children are optional/feature-gated and can be legitimately absent on older or downgraded documents, and it's not confirmed whether an invalid requested zone gets dropped (which a presence check would catch) or silently substituted (which it wouldn't). So a strict readback check risks false *negatives* — flagging a good apply as failed — which is worse than the current behavior.

The one real sub-case here is an invalid `specific-zone` id quietly falling back to default. I'd rather close that at pre-apply validation — verify the requested zone exists in the target dashboard, the same way source sheets are validated — than in readback. I'll file that as a follow-up.

PR #846, opened 2026-08-25
fix(desktop): reject apply-dashboard zones naming unwindowed worksheets

lauren-jacksonSFDC:
## What this changes

`apply-dashboard` now refuses an in-place dashboard apply when one of the dashboard's sheet zones names a worksheet that has no live worksheet-class `<window>` in the open workbook. This is a rule the codebase now keeps for the per-dashboard (`requireExistingSheet`) route, not a one-off.

## Why

A dashboard sheet zone references its worksheet by `name=`. Tableau Desktop resolves that name to a registered worksheet-class window the instant it renders the zone, and asserts (`oWindowID` in `WindowCollection::GetWorksheetPointer`) when none exists. Once that happens the dashboard is wedged: `delete-sheet` and `undo-workbook` on it both hard-fail with an internal error, leaving no recovery path short of discarding the session.

A per-dashboard apply posts only the `<dashboard>` fragment, so the existing whole-workbook consistency rules — `dashboard-zones-reference-included-worksheets` and `worksheet-missing-window` — never run (they are workbook-context only, by design). That left a gap: a hand-edited cached dashboard XML that adds a zone for a not-yet-windowed worksheet applied "successfully" and corrupted live window state. The whole-workbook create paths (`build-and-apply-dashboard`, `compose-dashboard`, `run-dashboard-batch`) are unaffected — they register the window alongside the zone.

## How

- New `zoneWorksheetWindowIssues` preflight in `loadDashboardXml`, gated to the `requireExistingSheet` + `kind === 'dashboard'` path so batch/compose and storyboard routes are untouched.
- It only reads live state when the fragment actually contains worksheet zones (`//zone[@name and not(@type-v2)]`); otherwise it skips the read entirely.
- Reuses the windowed-worksheet check by extracting `listWindowedWorksheetNames` into `metadata/parser.ts` and having `composeDashboardCore`'s `resolveRenderedWorksheetNames` delegate to it (behavior-preserving).
- On violation, returns a `validation-failed` error whose message steers the agent to create the worksheet first, then rebuild the dashboard with `compose-dashboard` / `run-dashboard-batch`.

## Tests

Added five cases to `loadDashboardXml.test.ts`: blocks a zone naming an absent/unwindowed worksheet; blocks a worksheet present in the document but with no worksheet-class window; applies when the zone worksheet is live and windowed; skips the workbook read when the fragment has no worksheet zones; surfaces a workbook read failure as `execute-command-error`. Full desktop unit suite green; `npm run build:desktop` clean. (One pre-existing, unrelated HTTP-timeout test flake — fails identically on the clean tree.)

### DIFF
--- a/src/desktop/wrappers/loadDashboardXml.ts
+++ b/src/desktop/wrappers/loadDashboardXml.ts
@@ -1,10 +1,12 @@
+import { DOMParser } from '@xmldom/xmldom';
 import { Err, Ok, Result } from 'ts-results-es';
+import * as xpath from 'xpath';
 
 import { log } from '../../logging/logger.js';
 import { sanitizeValue } from '../../logging/sanitize.js';
 import { ExecuteCommandError, WithExecutorAndAbortSignal } from '../externalApi/executorTypes.js';
 import { dashboardFragmentSimpleId, upsertDashboardIntoWorkbook } from '../metadata/dashboards.js';
-import { normalizeArray, parseXML } from '../metadata/parser.js';
+import { listWindowedWorksheetNames, normalizeArray, parseXML } from '../metadata/parser.js';
 import type { ParsedDashboard } from '../metadata/types.js';
 import { blockingValidationIssues, runValidation } from '../validation/registry.js';
 import { ValidationIssue } from '../validation/types.js';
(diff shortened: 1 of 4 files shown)


src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -325,6 +347,69 @@ async function loadDashboardXmlViaExternalApi({
...
+      fragmentXml.trim() || '<empty/>',
+      'text/xml',
+    ) as unknown as Document;
+  } catch {
+    return [];
+  }
+  // Worksheet zones are <zone name="Worksheet Name" .../>; layout, text, and blank zones carry
+  // type-v2 and do not name a worksheet.
+  const zones = xpath.select(
+    '//zone[@name and not(@type-v2)]',
```
mattcfilbert, line 389: MattGPT: Please include named zones whose type-v2 is visual. The current dashboard contract in targetDashboardInvariant.ts treats a zone as a worksheet when type-v2 is absent or visual, and our shipped examples use type-v2="visual". This XPath returns no match for that supported form, so an unwindowed visual zone bypasses the new guard and can still reach Desktop. Reuse the existing classifier or match both forms, and add a visual-zone regression that proves no POST occurs.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -191,6 +193,26 @@ export async function loadDashboardXml({
     focus.navigate === 'artifact' ? { ...focus, sheetName: canonicalName } : focus;
 
   if (requireExistingSheet) {
+    if (kind === 'dashboard') {
+      const zoneGuard = await zoneWorksheetWindowIssues({ fragmentXml: xml, executor, signal });
```
mattcfilbert, line 197: MattGPT: Please put this live-window read inside the same withApplyLock section as the per-dashboard POST. Today the call can validate Sheet 1, wait while delete-sheet acquires the lock and removes that unreferenced worksheet, then acquire the lock and post a zone that now has no window. The dashboard source hash does not cover workbook window state. A barrier-based concurrency test would keep this check fail-closed.

mattcfilbert (commented):
MattGPT review: The guard is the right seam, but two gaps still let invalid zones reach Desktop. CI is green. Before this leaves draft, please also run npm run version:patch; package.json still matches the 2.67.1 base.

PR #847, opened 2026-08-25
@W-23864457 Enrich get-workbook with VDS-queryable embedded data sources

ycath:
## Summary

Enriches the `get-workbook` tool to discover **embedded (workbook) data sources** via the REST `/workbooks/{id}/connections` endpoint, emitting them as `upstreamDatasources` entries alongside the existing Metadata-API published lineage. This is the primary discovery surface for the "analyze the X workbook" flow.

Part of the workbook-datasource-first-class epic. Depends on the SDK connections endpoint (W-23864448, done) and the shared datasource-reference shape (W-23864452, done).

## What changed

- **Embedded discovery**: after fetching the workbook, one `queryWorkbookConnections` REST call emits an `upstreamDatasources` entry per connection datasource with `datasourceType: 'embedded'` and `luid = datasource.id` (the VDS-queryable embedded LUID).
- **Type backfill**: existing Metadata-API upstream lineage entries are tagged `datasourceType: 'published'` so agents can distinguish the two kinds.
- **Runs without the Metadata API**: `/connections` is REST-only, so embedded discovery works when `disableMetadataApiRequests` is true. A pure-embedded workbook now surfaces its data source where it previously surfaced nothing.
- **Name fallback**: when a connection datasource has no name, it falls back to the luid (the shared shape requires `name`).
- **Dedup**: embedded entries are deduped by luid; published entries remain separate. Entries flow through the existing `constrainSuccessResult` / `allowedDatasourceIds` scoping.
- **Fail-soft**: a failed `/connections` call is swallowed-and-warned like the existing lineage try/catch; the tool still returns the workbook.

## Out of scope

- `publishedParent` (embedded → published mapping) is deferred to **W-23988697**.
- Parent-based allowance for scoped deployments is W-23864479; until it lands, tag-scoped and datasource-ID-allowlist sites filter embedded entries out (fail-closed, accepted).

## Testing

- `getWorkbook.test.ts`: 21 passed (5 new tests covering the enrichment: single `/connections` call + call-count assertion, published+embedded combine with type tags, name→luid fallback, metadata-disabled path, swallowed failure).
- Sibling lineage tests: 194 passed.
- Lint clean; `tsc --noEmit` clean.

Version bumped 4.5.6 → 4.5.7 (patch, matching the per-story convention in this epic).


src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -76,31 +78,64 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
               try {
-                const response = await restApi.metadataMethods.graphql(
-                  getWorkbookLineageQuery([workbook.id]),
-                );
-                return mergeWorkbookLineage(
-                  [workbook],
-                  getWorkbookLineageByLuid(response),
-                  configWithOverrides.boundedContext.datasourceIds,
-                )[0];
+                const connections = await restApi.workbooksMethods.queryWorkbookConnections({
```
jarhun88, line 86: is my understanding correct that this api only returns embedded ds (otherwise known as connections?) and will not return pds?

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -76,31 +78,64 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
+                      logger: 'lineage',
+                      data: getExceptionMessage(error),
+                    },
+                    extra,
+                  );
+                }
+              }
+
+              // Dedupe by luid. Published and embedded LUIDs are distinct; this mainly guards
+              // against the same embedded datasource appearing on multiple connections.
```
jarhun88: is this something that actually happens?  why does the connections api return multiple of the same embedded ds? if its only on the embedded side, can we just add teh dedup in the toEmbeddedLineageContents?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.5.6",
+  "version": "4.5.7",
```
jarhun88: this is new functionality, so i think it should be a minor bump

jarhun88:
btw ive been experimenting with spec driven development, if you are using plan mode to implement, feel free to add the plan/spec files into the /design/ folder at the root!

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -331,6 +331,21 @@ function toGraphqlStringArray(values: Array<string>): string {
...
+// name is optional on the REST response, so fall back to the luid (the shared shape requires it).
+export function toEmbeddedLineageContents(
+  connections: Array<WorkbookConnection>,
+): Array<LineageContent> {
+  return connections
+    .map((connection) => connection.datasource)
+    .filter((datasource): datasource is NonNullable<typeof datasource> => !!datasource)
+    .map((datasource) => ({
+      luid: datasource.id,
+      name: datasource.name ?? datasource.id,
```
stephendeoca: Does the REST API spec just says the name is optional, but really its always populated? I think the embeded datasource just inherits the name of the workbook or something. Not sure.

stephendeoca (approved):

src/tools/web/workbooks/getWorkbook.ts
ycath: yeah there is  another story to emit the published DS: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002ikIxcYAE/view

src/tools/web/workbooks/getWorkbook.ts
ycath: Yeah there could be duplicates since /workbooks/{id}/connections returns one row per connection, and a multi-connection embedded datasource (cross-database join/union) produces multiple rows sharing the same datasource.id. but you are right it would only be for embedded ds, I'll move the dedup to `toEmbeddedLineageContents` .

src/sdks/tableau/methods/lineageUtils.ts
ycath: yeah the embedded DS caption is derived from the workbook connection name and should always be present, but we wanted to keep the ds name optional so just in case if it's not populated (like from an older cloud version) it wouldn't fail the whole Zodios parse.

PR #849, opened 2026-08-26
Refactor(evals): clean up eval harness duplication ahead of merge

jarhun88:
## Summary

Pure refactor of the eval harness (`evals/`) added in this branch — responding to code-quality feedback on the harness's readability/duplication, no intended behavior change. Opened against this branch (not `main`) so it can be reviewed/merged here before the branch itself lands.

- Extract `BirdGradeResult` into `evals/lib/birdResult.ts` (was hand-declared 3x)
- Extract `captureExecError` into `evals/lib/execError.ts` (was hand-copied 4x with inconsistent fields)
- Extract `mean`/`sum`/`round`/`rate` into `evals/lib/stats.ts`
- Extract shared JSONL final-text scan into `evals/adapters/streamJson.ts` (claude-code.ts/cursor.ts duplicate)
- Add `buildStandardMcpConfig` to `evals/adapters/types.ts`, replacing a duplicated MCP config literal
- Decompose `grade-bird.ts`'s 227-line `main()` into named functions (`loadGradingContext`, `computeSignals`, `deriveVerdict`, `printSummary`) and export `gradeBirdCase()` for direct in-process use
- `grade-suite.ts` now calls `gradeBirdCase()` directly instead of spawning `grade-bird.ts` as a subprocess per case
- Doc clarification: `grade-suite.ts`/`ADMIN_EVALS.md` note it's BIRD-specific, not a generic suite grader

## Two bugs found and fixed during review

Both were the same species: code safe as an isolated CLI subprocess became unsafe once `grade-suite.ts` started calling `gradeBirdCase` in-process (module-level code runs as an import side effect, not just on direct invocation):

- `loadGradingContext()` called `process.exit(1)` on bad input — would've killed `grade-suite.ts`'s entire batch run on one bad case. Now throws `Error`, caught per-case by the batch loop.
- A module-top-level `process.argv`/`process.exit` check plus an unconditional `main().catch()` at the bottom of `grade-bird.ts` ran on import — `grade-suite.ts` invoked with no args would immediately die with `grade-bird.ts`'s usage message. Guarded `main()`'s invocation behind an entrypoint check; also fixed a latent judge-working-directory collision across batch cases by threading `runDir` explicitly instead of relying on module-level state.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.evalcheck.json` clean
- [x] `npx vitest run` — 173 files / 2780 tests passed
- [x] Live-verified both CLI entry points post-fix: `npx tsx evals/grade-bird.ts` (no args) still prints usage/exits 1 standalone; `npx tsx evals/grade-suite.ts` (no args) reaches its own auto-discovery instead of dying on grade-bird's usage message
- [x] Three rounds of independent dual-review (fresh reviewers each round) + synthesis, canonical `VERDICT: PASS`


design/plans/refactor-eval-harness-pr-623-plan.md
```diff
@@ -0,0 +1,168 @@
+# Refactor eval harness in PR #623 (tableau/tableau-mcp) for readability
```
Alon-ST-DATA: Does this md file need to be merged?

Alon-ST-DATA:
Reviewed this cleanup — solid, well-scoped, behavior-preserving refactor. One CI blocker before merge.

Confirmed the true diff is `asimantov/eval-admin-tools...HEAD` = 16 files under `evals/` (+632/−407), not the whole harness.

**Blocker (CI red):** `npm run lint` (ci.yml:34) fails — 2 prettier errors in `evals/grade-bird.ts`:
- **L44** — stray double blank line left where the deleted top-level `runDirArg`/`absRunDir` block was.
- **L540** — `isMainModule` line is 109 chars, over print-width.

Both clear with `npx eslint --fix`. Note the PR says "compilation passes" — lint is a separate gate.

**Minor (non-blocking):**
- New `tsconfig.evalcheck.json` isn't wired to any npm script or CI step, and `evals/**` is excluded from the main tsconfig — so eval code still has no automated typecheck. Passes by hand today, but nothing keeps it green. Suggest an `eval:typecheck` script + CI step, or a note that it's manual-only.
- Refactor adds nice testable seams (`captureExecError`, `deriveVerdict`, …) but no unit tests under `evals/`. Pre-existing gap, fine as a follow-up.

**Verified behavior-preserving:** rounding byte-identical; in-process `gradeBirdCase()` has no hidden module state (old `absRunDir` threaded as `runDir`); `process.exit`→throw is an improvement (one bad case can't kill a batch); `isMainModule` guard is correct. No dead code, nothing load-bearing deleted.

LGTM once lint is green.

Alon-ST-DATA (approved):
LGTM.

PR #851, opened 2026-08-26
@W-23974787: Gate flow tools behind the flow-tools feature flag

showmethecold:

the pull request._

# Pull Request Template

## Description

Flow tools were controlled only by FLOW_TOOLS_ENABLED, which is fixed at deploy time. Add a remote flow-tools feature flag so a deployment can roll them out per environment without a redeploy.

Registration and OAuth scopes now require FLOW_TOOLS_ENABLED=true AND the flow-tools flag. Either switch off keeps the tools off, so the env var remains a hard kill switch that the remote flag cannot override. The gate is async, so `disabled` becomes a Provider resolved at registration time. The flag ships as false in features.json to fail closed on upgrade.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe): Add a new feature flag for flow tools

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -467,8 +467,16 @@ Controls whether the Tableau Prep flow tools are registered.
 - Set to `true` to enable the Tableau Prep flow tools:
   - [`list-flows`](../../tools/flows/list-flows.md)
   - [`get-flow`](../../tools/flows/get-flow.md)
+  - [`list-flow-runs`](../../tools/flows/list-flow-runs.md)
+  - [`list-flow-tasks`](../../tools/flows/list-flow-tasks.md)
 - Only the exact value `true` enables them; any other value (or leaving it unset) keeps them
   disabled.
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.7.0",
+  "version": "4.7.1",
```
stephendeoca, line 4: @sf-ali-mahmoud with the new autoversion bump changes what should folks do about old MR's that don't use the new template? do they still manually update the version or should they leave package.json the same version as main instead of bumping.

jarhun88 (approved):

package.json
sf-ali-mahmoud: Have you tried merging latest from main?

If this specific change is not needed for a release it would be published in to following PR which means version pump is optional.

PR #853, opened 2026-08-27
docs(knowledge): mirror chart audit guidance

mattcfilbert:
Decision: mirror the reviewed A2TD chart-audit guidance byte-for-byte in Tableau MCP. This is a source/mirror rule, not a separate TMCP interpretation of the guidance.

Scope:
- updates the one mirrored Desktop knowledge resource
- adds exact #1 retrieval coverage for the combined audit ask and both reviewer-requested asks
- leaves the package at 2.67.1 because this resources-only change does not cut a separate published artifact

Future changes must preserve byte equality with the A2TD source unless a tested TMCP-only runtime difference requires otherwise.

Evidence:
- the added exact queries rank the intended entry #1
- `scripts/agent-check`: 392 files / 6,429 tests passed; all five checks passed
- A2TD source and TMCP resource share SHA-256 `6dc732ce9d8243a95f09f5aa8e2960cde95e7a5d2d5e57ec6de8d205ccca1408`
- scoped final review found no P1/P2 issue

Approving means accepting the exact public mirror of the reviewed A2TD knowledge change.

Based on knowledge entry generated by Will Perkins.

### DIFF
--- a/resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
+++ b/resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
@@ -194,7 +194,7 @@ When the question could be answered by multiple chart types, prefer in this orde
 ## Source and Confidence
 
 - Source/evidence type: internal-doc
-- Source: prior Tableau authoring knowledge base (mbradbourne), with workbook-audit and logarithmic-axis guidance from Will Perkins in [tableau/agent-to-tableau-desktop#93](https://github.com/tableau/agent-to-tableau-desktop/pull/93), commit [9eb2696](https://github.com/tableau/agent-to-tableau-desktop/commit/9eb2696948dcc376922c78388baec537e3880f9a)
+- Source: prior Tableau authoring knowledge base (mbradbourne), with workbook-audit and logarithmic-axis guidance from Will Perkins
 - Customer-identifying details removed: yes
 - Confidence: draft
 - Last reviewed: 2026-08-27
--- a/resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
+++ b/resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
@@ -8,8 +8,8 @@ Implementation reference for chart design decisions in Tableau — covering mark
 - Authoring outcome improved: create, refine
 - In-scope reason: Helps Claude apply chart design best practices and avoid common mistakes when authoring visualizations.
 - Out-of-scope risk: none
-- Tags: charts, best-practices, mark-class, shelf-placement, examples
-- Relevant user prompts/search terms: "which chart type should I use", "bar chart vs line chart", "how to sort a bar chart in Tableau", "dual axis chart example", "why does my chart look wrong", "scatter plot setup", "stacked bar chart configuration", "common chart mistakes", "when to remove gridlines", "mark class selection rules"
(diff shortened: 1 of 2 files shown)


resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
```diff
@@ -188,7 +194,7 @@ When the question could be answered by multiple chart types, prefer in this orde
 ## Source and Confidence
 
 - Source/evidence type: internal-doc
-- Source: imported from prior Tableau authoring knowledge base (mbradbourne)
+- Source: prior Tableau authoring knowledge base (mbradbourne), with workbook-audit and logarithmic-axis guidance from Will Perkins in [tableau/agent-to-tableau-desktop#93](https://github.com/tableau/agent-to-tableau-desktop/pull/93), commit [9eb2696](https://github.com/tableau/agent-to-tableau-desktop/commit/9eb2696948dcc376922c78388baec537e3880f9a)
```
myu404: 🤖 **Reviewed by MichaelGPT**

This line adds two clickable links into a doc that ships in the **public** npm resource bundle, but both point at a **private** repo (a PR reference and a commit hash). Because that repo is private, the links 404 for any external reader and they expose internal identifiers. The previous version of this line was just `Source: imported from prior Tableau authoring knowledge base (mbradbourne)` — so this PR is what introduces the private links.

Suggest dropping the two links and keeping the provenance generic, e.g. `…with workbook-audit and logarithmic-axis guidance from Will Perkins` (no external URLs). If a byte-for-byte mirror rule from the source knowledge base is forcing these links in, then the mirror rule is what's leaking internal references into a public artifact for this field, and it should carve out an exception here.

myu404 (commented):
## 🤖 **Reviewed by MichaelGPT** — Recommendation: **Request changes**

This mirrors chart-audit guidance into `chart-best-practices.md` — a new Common Mistake #9, extended tags/search terms, a when-to-use bullet, and refreshed Source/Last-reviewed lines — and adds one knowledge-search regression test. Most of it holds up well: the new #9 (Problem/Why/Fix) is structurally consistent with entries #1–#8 and reinforces rather than contradicts the existing bar zero-baseline and bar-first mark guidance; the markdown structure is intact; and the search test is genuinely non-trivial, since its query terms only enter the highest-weighted `searchTerms`/`tags` fields through this PR, so the exact-slug top-hit assertion depends on the added content rather than passing tautologically.

I'm requesting changes for **one blocker** (inline): the refreshed `Source:` line adds two clickable links to a **private** repo — a PR number and a commit hash — inside a doc that ships in this **public** npm bundle. Those links 404 for external readers and expose internal references. The fix is small (drop the private links, keep provenance generic). Everything else looks good to merge once that's addressed.

resources/desktop/knowledge/strategy/viz-design/chart-best-practices.md
mattcfilbert: Fixed in fb3c60b. Removed the private PR and commit links from the shipped resource and from the public PR body. I made the same provenance change in the A2TD source, so the two copies remain byte-identical. Focused checks pass: 9 TMCP resource/integrity tests and 452 A2TD integrity tests.

— MattGPT for Matt Filbert

myu404 (approved):

PR #854, opened 2026-08-27
@W-24015251: Auto patch version bump on merge to main

sf-ali-mahmoud:
## Summary
- Replaces the manual version-bump-on-PR flow with an automated post-merge bump, eliminating the recurring package.json version conflicts on every PR.
- Removes `version-bump-check.yml` (the PR-side check requiring authors to manually edit `package.json`) and `tag.yml` (tagged based on the pre-bump merge commit, which would have raced with the new bump commit).
- Adds `auto-version-bump.yml`: on push to `main`, bumps the patch version, commits it, tags it (`v<version>`), and pushes both — keeping version bump and tag creation atomic and consistent with `publish.yml`'s `RELEASE_TAG == v<package.json version>` preflight.

## Test plan
- [ ] Merge this PR and confirm `auto-version-bump.yml` runs, bumps `package.json` patch version, and pushes a `chore: bump version to X.Y.Z` commit + matching `vX.Y.Z` tag
- [ ] Confirm the bump commit's own push does not re-trigger the workflow (guarded by the commit-message `if:` check)
- [ ] Promote the new tag to a GitHub Release and confirm `publish.yml` passes its tag/version preflight and publishes successfully


stephendeoca:
Version bumping requires a bit more configurability if we automate it. Perhaps in the PR template, users can check which kind of bump they want (Major, Minor, Patch) and then the PR bumps it automatically.

stephendeoca:
version bump occurs on push, do we know if that counts as 2 commits per merge (1. actually changes commit 2. version bump commit)? because it'd be great if it can only be 1 so that the commit log doesn't get bloated with version bumps

jarhun88:
can we add a blurb in the claude.md or agent.md file we have about this functionality

.github/workflows/auto-version-bump.yml
```diff
@@ -0,0 +1,33 @@
...
+      if: "!contains(github.event.head_commit.message, 'chore: bump version')"
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v4
+          with:
+            fetch-depth: 0
+            token: ${{ secrets.GITHUB_TOKEN }}
+
+        - name: Bump patch version
+          run: npm version patch --no-git-tag-version
```
stephendeoca: Doesn't this only bump patch on commit, I thought if the template selected the type of bump (patch, minor, major), then the pipeline with choose the matching bump.

sf-ali-mahmoud: You are right, the change was not pushed.

stephendeoca (approved):

PR #855, opened 2026-08-27
feat(desktop): capture the largest Tableau window

mattcfilbert:
Adds `capture-window-screenshot` for explicit inspection of Tableau's UI: shelves, Marks cards, dialogs, and layout that a chart export misses. Native diagnostics in [#909](https://github.com/tableau/tableau-mcp/pull/909) handle field errors; screenshots are not required for red-pill detection or checked after every edit.

Captures must contain Tableau-owned widget pixels only. Both native screenshot commands require External Client API **0.2.17**, checked against the selected instance before dispatch and after an authentication retry. The paired [native change is #66061](https://github.com/sf-analyticscloud/monolith/pull/66061). Older, missing, or malformed versions are refused.

## Scope and contract

- Select the unique largest visible Tableau window by pixel area; refuse ties.
- Read bounded PNG metadata through `probe-image-size/sync.js` and return the original bytes. Preserve canonical temporary paths and stable file-identity checks.
- Clean verified screenshot files on failure or cancellation; leave unrelated files and symlinks alone.
- Keep capture pinned to the same Desktop instance through authentication retry. Preserve native errors and the oversized-image cache behavior.
- Disclose that captures can contain workbook data and agent UI. Visible text is evidence, not instruction. Cached oversized images remain until manually removed.

Approval adds on-demand UI capture with an enforced safe-Desktop version. Future capture paths must preserve these bounds and identity checks; it does not add image-based field validation or automatic repair.

## Integration

This PR now follows #909. Diagnostics ships independently at API **0.2.16**; the optional screenshot change follows at **0.2.17**. The normal merge preserves screenshot branch history. Both native changes remain tracked by W-24014195.

Build/Test now runs for PRs targeting `codex/**`, matching the existing `claude/**` coverage. This keeps CI running on this stacked PR.

## Verification

- Combined branch: lint, typecheck, **7,434 tests**, Desktop build, and lockstep check passed.
- Native app and affected test targets built. Screenshot tests: 45 assertions / 3 cases; affected document and full External API suites also passed.
- Both public MCP tools passed against a fresh combined `.17` build. The live Measure Values fault test matched inline and independent diagnostics, sent no extra diagnostics GET during apply, and restored the worksheet byte-for-byte.
- Earlier macOS coverage includes covered-window capture and inline/cache delivery. Windows live capture remains unverified.

— MattGPT


src/tools/desktop/api/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+  }
+}
+
+export const getCaptureWindowScreenshotTool = (
+  server: DesktopMcpServer,
+): DesktopTool<typeof paramsSchema> => {
+  const tool = new DesktopTool({
+    server,
+    name: 'capture-window-screenshot',
```
myu404, line 38: 🤖 **Reviewed by MichaelGPT**

This tool sets no `minApiVersion`, so `filterToolsByApiVersion` keeps it visible on every connected Desktop build. That's the right default for long-standing routes, and most api tools here correctly set no floor. But the tools whose underlying route was **introduced in a specific later API version** do set one (`add-*` → `0.2.6`, `publish-workbook`/`refresh-datasource-*` → `0.2.8`, `workbook-export-as`/`export-storyboard-image` → `0.2.7`). The PR notes this was proven on **API 0.2.9** — higher than any existing floor — which suggests the screenshot route is newly introduced.

Question for the author: was the `take-all-screenshots` route introduced in a recent API version? If so, adding `minApiVersion: '0.2.9'` (or the actual introduction version) would match the pattern and drop the tool cleanly on older Desktop builds instead of surfacing a generic capture error. If the route has existed longer, no change needed — just confirming the intent.

src/tools/desktop/api/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,86 @@
...
+        args: { session },
+        callback: async () => {
+          const capture = await runExternalApiReadTool({
+            session,
+            extra,
+            callback: async (executor, signal) =>
+              await captureWindowScreenshot({ executor, signal }),
+          });
+          if (capture.isErr()) return capture;
+          if (isOverInlineImageCap(capture.value.bytes.length, extra.config.inlineImageMaxBytes)) {
```
myu404, line 61: 🤖 **Reviewed by MichaelGPT**

The `inlineImageMaxBytes` cap (default 1 MiB) is applied here, at the tool layer, *after* the wrapper has already read and CRC-validated the full image into memory — where the per-file bound is 32 MiB and the aggregate 64 MiB. So a valid 2–32 MiB screenshot is fully read and validated before this 413 rejects it.

This is bounded and safe (the 64 MiB aggregate is `stat`-checked before any bytes are read, so allocation isn't unbounded) — not a DoS. Flagging only for clarity: `inlineImageMaxBytes` limits what's *emitted inline*, not wrapper-side memory. A one-line comment noting that distinction would prevent a future reader from assuming this cap bounds the read.

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,485 @@
...
+    directoryStats = canonicalStats;
+    if (fileSystem.realpath(returnedDirectory) !== canonicalDirectory) {
+      throw new Error('Returned screenshot directory changed during validation.');
+    }
+
+    const names = fileSystem
+      .readdir(canonicalDirectory)
+      .filter((name) => SCREENSHOT_NAME.test(name));
+    if (names.length === 0) {
+      outcome = failure('Tableau Desktop did not produce a screenshot.');
```
myu404, line 406: 🤖 **Reviewed by MichaelGPT**

The zero-matching-files branch here — `failure('Tableau Desktop did not produce a screenshot.')` — isn't exercised by any test. The string never appears in `captureWindowScreenshot.test.ts`, and the nearest test ("ignores nonmatching files but fails closed") writes a `ScreenShot_1.png.bak` *and* a matching `ScreenShot_2.png`, so `names.length` is 1, not 0.

Suggest adding a test that presents a command directory with only non-matching files (or none), asserting `result.isErr()` with this message and that the directory is still cleaned up. Small gap in an otherwise very thorough suite.

myu404 (commented):

This adds one desktop tool, `capture-window-screenshot`, that drives Desktop's screenshot command, reads back the single largest resulting PNG, and returns it inline as a base64 `image/png` block — scoped deliberately as a safe *capture* foundation (no red-area interpretation, no apply-behavior change).

The filesystem hardening is thorough and I could not fault the core: the Desktop-returned path is `resolve`d, `lstat`'d (symlink/non-dir refused), `realpath`'d, and required to sit strictly below `realpath(os.tmpdir())`; candidates are name-allowlisted (`/^ScreenShot_\d+\.png$/`), opened `O_RDONLY|O_NOFOLLOW`, and re-checked for dev/ino identity across `lstat`/`open`/`fstat`/re-stat (TOCTOU); PNGs are structurally validated (signature + per-chunk CRC + IHDR/IDAT/IEND rules); size is bounded per-file (32 MiB) and in aggregate (64 MiB) with the aggregate checked on `stat` *before* any bytes are read, plus a pixel-area bound; a tie on max area **fails closed** ("ambiguous") rather than guessing; and cleanup runs on both success and error paths, returning an **error instead of a partial result** if artifacts can't be safely removed. The result shape is strictly (`.strict()`) and doubly parsed, and the oversized-image case returns a 413 without retaining the image. Registration is complete and consistent across `toolName.ts`, `tools.ts`, `server.desktop.ts`, and the reference tests, and the test suite exercises an impressive range of failure/traversal/symlink/TOCTOU/cancellation/cleanup paths.

Three non-blocking notes inline (one question about API version-gating, two minor).

myu404 (approved):

src/tools/desktop/api/captureWindowScreenshot.ts
mattcfilbert: 🤖 **MattGPT**

Checked the history: `take-all-screenshots` predates the External Client API. The required command-result support starts at API `0.1.1`, which our executor already records in `supportsOperationResult`. Added `minApiVersion: '0.1.1'` and an assertion in the existing API-floor test. The `0.2.9` version in the PR was the build used for live verification.

src/tools/desktop/api/captureWindowScreenshot.ts
mattcfilbert: 🤖 **MattGPT**

Added the one-line comment above the cap check: the wrapper has already read and CRC-validated the PNG under its 32 MiB per-file and 64 MiB aggregate limits; this cap controls MCP inline output.

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: 🤖 **MattGPT**

Added the empty-directory case. It asserts the exact `Tableau Desktop did not produce a screenshot.` error and verifies that the directory is removed. The existing protection for nonmatching files remains intact.

src/tools/desktop/api/captureWindowScreenshot.ts
mattcfilbert: Follow-up for 28cc0f3d: the tool now calls `tabui:take-active-widget-screenshot`, which captures the active Tableau window, dialog, or popup. The native source confirms that both screenshot commands predate the External Client API; the minimum remains 0.1.1 for operation results. The new command path still needs live Desktop smoke testing.

— MattGPT

src/tools/desktop/api/captureWindowScreenshot.ts
mattcfilbert: Follow-up for 28cc0f3d: the wrapper now reads and CRC-validates one exact `ScreenShot.png` under a 32 MiB bound. The inline cap still limits MCP output after that bounded read. The old multi-window selection and 64 MiB aggregate path are gone; the tests and description now reflect the single-file path.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,448 @@
...
+const screenshotCommandResultSchema = z
+  .object({
+    tempFilePath: z
+      .string()
+      .min(1)
+      .refine((value) => value.trim().length > 0 && isAbsolute(value)),
+  })
+  .strict();
+
+export interface WindowScreenshotFileSystem {
```
WillMartin: I get the need for mocking here but this is a relatively generic concept defined in a pretty specific way. Should we have a generic "FileSystem" file that we can mock out and use in multiple places?

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,448 @@
...
+  };
+  return (
+    legalBitDepths[colorType]?.includes(bitDepth) === true &&
+    data[10] === 0 &&
+    data[11] === 0 &&
+    (data[12] === 0 || data[12] === 1)
+  );
+}
+
+function parsePngHeader(bytes: Buffer): { width: number; height: number } {
```
WillMartin: Can we use a shared library to do this?

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,448 @@
...
+  { executor, signal }: WithExecutorAndAbortSignal,
+  fileSystemOverrides: Partial<WindowScreenshotFileSystem> = {},
+): Promise<Result<WindowScreenshotCapture, McpToolError>> {
+  const fileSystem = { ...DEFAULT_FILE_SYSTEM, ...fileSystemOverrides };
+  if (signal.aborted) {
+    return failure('Tableau Desktop window capture was cancelled.');
+  }
+  const commandResult = await executor.executeCommand({
+    namespace: 'tabui',
+    command: 'take-active-widget-screenshot',
```
WillMartin: was I using "active widget"? I didn't think I was but maybe I'm confused. We don't necessarily want to look at the active one - we'd want the main window (largest)

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: yep — moved it to shared `FileSystem` in `src/utils/fileSystem.ts`.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: yep, that was my change. back to all-window capture and picking the largest. ties fail rather than guess.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: any library you'd recommend for bounded validation without decoding? pngjs decodes the pixels, so i've left this open for now.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,448 @@
...
+    args: { HideMouse: true },
+    schema: screenshotCommandResultSchema,
+    signal,
+  });
+  if (commandResult.isErr()) {
+    return failure('Tableau Desktop could not capture the visible window.');
```
tableaukyler: 🟠 This `executeCommand` failure path collapses into one static message, discarding the underlying `ExecuteCommandError` (code/message/recoverable). Every other command-dispatching tool preserves it: `activateSheet.ts` checks `isRouteMissing(result.error)` → `endpointNotInThisBuild(...)` before falling back to `DesktopCommandExecutionError`; `executeTableauCommand.ts` / `exportCustomTheme.ts` wrap raw `executeCommand` failures in `DesktopCommandExecutionError`, which also flags `blockedByDesktopDialog` for timeouts/`awaiting-user`. Compounding this: the tool callback (`src/tools/desktop/api/captureWindowScreenshot.ts:58`) destructures only `(executor, signal)` from `runExternalApiReadTool`, so the `read` closure that would apply the `isRouteMissing` translation never runs here either. Net effect: an old Desktop build missing this route, a blocked dialog, and a timeout all currently produce the same undifferentiated 500 with no retry guidance — untested by the 593-line suite (only a generic `capture-failed` case is covered). Suggest propagating the raw `ExecuteCommandError` like `activateSheet.ts`'s wrapper does, so the tool layer can apply the same handling.

src/tools/desktop/api/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,89 @@
...
+          const capture = await runExternalApiReadTool({
+            session,
+            extra,
+            callback: async (executor, signal) =>
+              await captureWindowScreenshot({ executor, signal }),
+          });
+          if (capture.isErr()) return capture;
+          // The wrapper reads and CRC-validates one ScreenShot.png under 32 MiB; this cap controls MCP inline emission after that bounded read.
+          if (isOverInlineImageCap(capture.value.bytes.length, extra.config.inlineImageMaxBytes)) {
+            return new WindowScreenshotTooLargeError(
```
tableaukyler: 🟠 On an over-cap screenshot this discards the bytes outright ("was not retained or cached"). Every sibling image tool (`exportWorksheetImage`/`exportDashboardImage`/`exportStoryboardImage`, via `exportSheetImageResult.ts`'s `cachePrefix` + `buildInlineImageCapFileMessage`/`logInlineImageCapHit`) writes an over-cap image to a cache file and returns the path instead. This file imports only `isOverInlineImageCap` from the same `desktop/limits/inlineImageCap.ts` module, not the cache-file half. A full-resolution desktop window will commonly exceed the default 1 MiB cap, so this tool is often unusable with no recovery path short of raising a global env var. Suggest reusing the `cachePrefix` fallback from `exportSheetImageResult.ts`.

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,448 @@
...
+  }
+
+  const area = width * height;
+  if (!Number.isSafeInteger(area) || area > MAX_WINDOW_SCREENSHOT_PIXELS) {
+    throw new Error('PNG pixel area is outside the supported bounds.');
+  }
+  return { width, height };
+}
+
+function inspectCandidate(
```
tableaukyler, line 193: 🟡 Design pattern: the lstat → realpath → stat → identity-check sequence is hand-inlined 5x in this file (`inspectCandidate` here, `readCandidate`'s post-open re-check at ~266, the inline directory validation at ~377-394, and `cleanCommandArtifacts`'s two loops at ~307-344), and it's already drifted — some throw via `requireMatchingIdentity`, others return a bool via `hasMatchingCleanupIdentity`. A future fix to this security-sensitive TOCTOU guard needs 5 coordinated edits with no compiler-enforced parity. Suggest extracting one helper parameterized by the file-vs-directory check and identity comparator.

src/server.desktop.test.ts
```diff
@@ -546,6 +546,31 @@ describe('selectToolsForProfile (TOOL_PROFILE, W60 spike lever 1 / preamble P1)'
     ).toHaveLength(1);
   });
 
+  it('registers capture-window-screenshot once in full and dynamic-authoring profiles', () => {
+    const tools = allTools();
+    const captureTool = tools.find((tool) => tool.name === 'capture-window-screenshot');
+
+    expect(captureTool).toBeDefined();
+    expect(captureTool?.description).toContain('active Tableau window, dialog, or popup');
```
tableaukyler: 🟡 This asserts literal substrings of the tool's `description` field, duplicating the exact wording in `captureWindowScreenshot.ts`. Tests shouldn't assert string contents that duplicate code — a routine wording edit to the disclosure text breaks this with zero behavior change. Suggest keeping the structural assertions (paramsSchema/annotations, already present) and dropping the literal substring checks.

tableaukyler (commented):
## KylerGPT review

Solid, defensively-written capture path (TOCTOU-safe path/identity validation, PNG CRC/chunk parsing, safe cleanup) — but it bypasses two error-handling conventions this codebase already established for sibling tools, plus one design-pattern and one test-convention nit.

**🟠 Route-missing / dialog-blocked errors get swallowed into one generic message.** `executeCommand` failures in the wrapper collapse into `failure('Tableau Desktop could not capture the visible window.')`, discarding the underlying `ExecuteCommandError`. Every other command-dispatching tool preserves it: `activateSheet.ts` checks `isRouteMissing` → `endpointNotInThisBuild(...)` before falling back to `DesktopCommandExecutionError`; `executeTableauCommand.ts` / `exportCustomTheme.ts` wrap raw `executeCommand` failures in `DesktopCommandExecutionError`, which also flags `blockedByDesktopDialog` for timeouts/`awaiting-user`. The tool callback here destructures only `(executor, signal)` from `runExternalApiReadTool`, so the `read` closure that would apply the `isRouteMissing` translation never runs either. An older Desktop build missing this route, a dialog blocking the call, and a timeout all currently produce the same undifferentiated 500 with no "don't retry" guidance, and the test suite never exercises those cases.

**🟠 Oversized screenshots are discarded instead of cached.** `WindowScreenshotTooLargeError` says the screenshot "was not retained or cached" — bytes are dropped outright. Every sibling image tool (`exportWorksheetImage`/`exportDashboardImage`/`exportStoryboardImage`, via `exportSheetImageResult.ts`'s `cachePrefix` + `buildInlineImageCapFileMessage`/`logInlineImageCapHit`) writes an over-cap image to a cache file and returns the path instead. This tool imports only `isOverInlineImageCap` from the same `desktop/limits/inlineImageCap.ts` module, not the cache-file half. A full-resolution desktop window will commonly exceed the default 1 MiB cap.

**🟡 Duplicated TOCTOU validation skeleton.** The lstat → realpath → stat → identity-check sequence is hand-inlined 5x (`inspectCandidate`, `readCandidate`'s post-open re-check, the inline directory validation, and `cleanCommandArtifacts`'s two loops) and has already drifted — some throw via `requireMatchingIdentity`, others return a bool via `hasMatchingCleanupIdentity`. A future fix to this security-sensitive guard needs 5 coordinated edits with no compiler-enforced parity.

**🟡 Test asserts literal description substrings.** `toContain('active Tableau window, dialog, or popup')` etc. duplicate the exact wording in the tool's `description`. Tests shouldn't assert string contents that duplicate code — a routine wording edit breaks this with zero behavior change.

---

mattcfilbert:
kyler — astra adversarial review of `214e1872` found no new actionable issues. focused tests pass. still open: png-library choice and live mac/windows smoke. CI is still running.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: fixed — raw command errors now reach the tool layer, which uses the same missing-route and desktop error handling as the other tools. added timeout and awaiting-user coverage.

— MattGPT

src/tools/desktop/api/captureWindowScreenshot.ts
mattcfilbert: fixed — oversized captures now use the shared file fallback. the result warns that the local file stays until removed, and the tool no longer advertises itself as read-only.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: pulled the path and identity checks into one helper. added a cleanup regression for a changed file identity.

— MattGPT

src/server.desktop.test.ts
mattcfilbert: removed the description checks; kept the parameter, annotation, and registration checks.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
WillMartin: What are we trying to avoid by not fully parsing?

mattcfilbert: The concern was decoded memory, not parsing itself. The current 100-million-pixel limit could mean a 400 MB RGBA buffer, even for a small compressed PNG. We only need the dimensions to select the largest window and then return the original PNG bytes.

That does not justify maintaining our own PNG parser. I should have explained the constraint instead of asking you to find a library. I am checking the library path and its allocation limits now, along with the merge conflicts.

— MattGPT

mattcfilbert: Changed in 8f41e980: this now uses `probe-image-size` to read dimensions and returns the original Tableau PNG. The custom chunk/CRC parser is gone; no pixel buffer is decoded, and the tool does not claim to validate the entire PNG stream. File-size, pixel-size, path, identity, and cleanup limits are unchanged.

I replaced the intermediate `image-size` dependency after its advisory check. The final library and its dependencies have no reported npm advisories, with no release-age policy exception.

The merge conflicts are resolved. The full checks and real built-server Mac captures passed for both inline output and the oversized-image fallback. Windows live behavior remains unverified. Leaving this thread open for your final look at the metadata-only contract.

— MattGPT

mattcfilbert:
MattGPT: The live TAS smoke test caught a real miss: the agent received the screenshot but called a red field pill valid. [c7be60e4](https://github.com/tableau/tableau-mcp/commit/c7be60e4dfcc8e1d297bceea2c13ba9500edeb91) adds short guidance explaining invalid pills and stale marks. No detector or automatic capture was added.

With the same neutral request to inspect the window, the agent now:

- Identifies the invalid pill with both inline PNGs and cached PNGs read through `Read` (cache fallback forced with a low size limit).
- Reports no invalid pill in the healthy control.
- Reports a newly introduced invalid pill despite an earlier clean structural-verification receipt in the same chat. The verification rule needed no change.

Canonical checks pass locally; CI is running on the new commit.

Separate defect: on this macOS / API 0.2.12 build, the native capture includes a thin strip of another app above Tableau’s toolbar. That is not fixed here, and raw screenshots are not attached because they contain unrelated app content. Windows remains untested. This proves user-requested inspection, not automatic detection after edits.

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,368 @@
...
+    if (fileSystem.realpath(returnedDirectory) !== canonicalDirectory) {
+      throw new Error('Returned screenshot directory changed during validation.');
+    }
+
+    const names = fileSystem
+      .readdir(canonicalDirectory)
+      .filter((name) => SCREENSHOT_NAME.test(name));
+    if (names.length === 0) {
+      outcome = failure('Tableau Desktop did not produce a screenshot.');
+    } else if (names.length > MAX_WINDOW_SCREENSHOT_CANDIDATES) {
```
myu404: ### [P1 · REQUIRED] Failure branches leave sensitive screenshots in the temp directory

**Purpose**

Every native screenshot should be removed on success or failure because it can contain workbook data, field names, dialogs, and agent UI.

**Finding**

This over-count branch fails before any candidate is added to `validatedPaths`. Cleanup therefore receives an empty list and can only try `rmdir` on the still-populated directory; the new test explicitly asserts that the directory remains. The paired 0.2.15 native producer has the same ownership gap on another failure path: if one visible widget saves successfully and a later save fails, it returns an error before `TempFilePath` is set and does not roll back the earlier files, so MCP never receives a path it could clean.

**Impact**

A command can report failure while sensitive screenshots remain in the OS temp area. More than 16 visible Tableau top-level widgets deterministically reaches the MCP branch, while a later native save failure can leave a partial capture set that is unreachable to this wrapper.

**Recommendation**

Establish one cleanup owner for every outcome before releasing the paired feature. Have the native command roll back its generated directory on any save failure, and align the producer/consumer count contract so an over-count result can be safely removed instead of intentionally retained. Add assertions that both a later-save failure and a 17-candidate result leave no producer-owned screenshots.

src/desktop/wrappers/captureWindowScreenshot.ts
```diff
@@ -0,0 +1,368 @@
...
+
+export async function captureWindowScreenshot(
+  { executor, signal }: WithExecutorAndAbortSignal,
+  fileSystemOverrides: Partial<FileSystem> = {},
+): Promise<Result<WindowScreenshotCapture, CaptureWindowScreenshotError>> {
+  const fileSystem = { ...NODE_FILE_SYSTEM, ...fileSystemOverrides };
+  if (signal.aborted) {
+    return failure('Tableau Desktop window capture was cancelled.');
+  }
+  const commandResult = await executor.executeCommand({
```
myu404, line 249: ### [P2 · SUGGESTED] Pin screenshot retries to the Desktop instance identity

**Purpose**

A capture selected for one Desktop/workbook should not be retried against a restarted or replacement instance.

**Issue and impact**

This dispatch pins the executor's PID but does not pass `expectedInstanceId`. After a 401, `withRescan` rediscovers by PID and runs the command again; if the OS has reused that PID for a new Desktop instance, the retry can return pixels from a different workbook. The executor already has an `expectedInstanceId` guard and a same-PID/new-instance test for this case. The race requires both a 401 and PID reuse, so I view this as non-blocking hardening.

**Suggestion**

Capture `executor.desktopInstanceId` before dispatch and pass it as `expectedInstanceId`, then add a screenshot-specific same-PID/new-instance retry test.

myu404 (changes requested):

The capture path has strong defensive work around API-version gating, path containment, stable file identity, bounded reads, prompt-injection disclosure, and the oversized-image cache fallback. The test suite is correspondingly thorough.

One cleanup-ownership gap still needs to be closed before merge: an over-count result intentionally leaves the populated command directory behind, and the paired native producer can also strand earlier screenshots if a later widget save fails before it returns the directory path. These images can contain workbook and UI data, so error paths need the same zero-residual guarantee as success.

I also left two non-blocking suggestions: pin a 401 retry to the original Desktop instance identity, and use the dependency's sync-only entry so the bundled server does not pull in unused HTTP/stream code.

Hosted Node 22/24 build and test jobs plus SAST and credential scanning are green. I reviewed the pinned MCP and paired native diffs statically; I did not rerun tests, perform a live Desktop capture, or validate Windows behavior locally.

src/desktop/wrappers/captureWindowScreenshot.ts
myu404: **Additional context**

The new default import resolves `probe-image-size` 7.4.0's root entry, which eagerly loads its stream and HTTP implementations even though this wrapper only calls `.sync`. Because the production Desktop and combined builds bundle package code, those unused paths are reachable in the artifacts. I have not measured the bundle-size or startup cost, and changing the import alone will not remove the transitive packages from npm installation or the lockfile.

**Recommendation**

Import the package's documented `sync.js` entry directly and add a narrow local TypeScript declaration if needed, while preserving the existing signature, IEND, dimension, and area guards.

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: Fixed in aa1a9c43. Both screenshot commands now keep the selected Desktop instance across a 401 retry, including when the connection starts without an instance ID. The regression calls the real capture wrapper and confirms that no second POST reaches the replacement instance.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: Fixed in aa1a9c43 and monolith #66061 (4af883ca0b7). A later native save failure now removes the generated folder; caller-supplied folders are left alone. MCP also removes verified screenshot files after over-count or cancellation, without deleting unrelated files or following symlinks. The partial-save and 17-file regressions reproduce the old leaks and pass with these fixes.

— MattGPT

src/desktop/wrappers/captureWindowScreenshot.ts
mattcfilbert: thanks MikeBot - switched to the library’s sync entry point as Mike/MikeBot suggested

myu404 (changes requested):
## Follow-up review

Reviewed `615f7fd43589` → `e09ad79861f5`.

**Previous findings:** 2 addressed, 1 partial, 0 open, 0 obsolete, 0 awaiting verification.

**New findings:** 0.

The over-count, post-result cancellation, partial-save, retry, and sync-import remediations are present. The cleanup P1 remains because native output can become ownerless if result delivery is aborted, times out, or is lost before MCP receives the temp path.

Static review only: current MCP checks are green, but I did not rerun builds, native tests, live Desktop capture, or Windows validation; no explicit security-review artifact was supplied.

src/desktop/wrappers/captureWindowScreenshot.ts
myu404: **Follow-up**
The specific 17-file, post-result cancellation, native partial-save, and caller-owned-output cases are fixed. One cleanup handoff still remains: native successful capture detaches its generated directory before MCP receives `tempFilePath`. If a registered command outlives an aborted, timed-out, or lost response, MCP exits without a path and operation expiry only removes registry state, so sensitive screenshots can remain without an application cleanup owner.

**Recommendation**
Keep the generated output owned until result delivery is acknowledged, or provide bounded native reclamation for unacknowledged or expired operations. Add an end-to-end pre-result abort or timeout regression that proves no generated screenshots remain.

mattcfilbert: Fixed the missing cleanup owner in the native companion, [1d0003202b6](https://github.com/sf-analyticscloud/monolith/commit/1d0003202b6fb78fed251ac9b4a3be62f65e9cf2) on [#66061](https://github.com/sf-analyticscloud/monolith/pull/66061). Desktop now retains ownership of generated screenshot directories and schedules cleanup 65 minutes after capture, whether or not the client receives the path. MCP still cleans up immediately after reading; caller-provided `OutputPath` folders are untouched.

The native regression discards the result and exercises the real Qt timer with a shortened lifetime. It failed before the fix and now passes for both capture commands; caller-owned output and early client deletion are also covered. I have not run a live HTTP-disconnect test. Cleanup requires a running event loop and is not crash-safe.

— MattGPT

myu404 (approved):

PR #856, opened 2026-08-27
Resolve field captions to fieldName for embedded workbook datasource Pulse insights

dplatt84:
## Summary
- The Pulse Insight Service validates the `field` on `generate-pulse-metric-value-insight-bundle` requests against HBI metadata's `fieldName`, not `fieldCaption`. Since `get-datasource-metadata` only exposes captions, calculated fields built from their caption (e.g. `Amount_Billions`) 400 while the underlying `fieldName` (e.g. `Calculation_0013913069531140`) succeeds. This resolves caption -> fieldName via VDS `readMetadata` before forwarding the request to Pulse.
- Also auto-detects embedded workbook datasources and sets `id_type: 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE'` on the request, so Pulse always calls HBI directly instead of a cached query result for embedded datasources. Detection uses `queryDatasource` (the published Datasources REST API): it resolves embedded/workbook datasource LUIDs too, but marks them with a `contentUrl` of the form `$embedded$_<workbookLuid>` instead of a normal published contentUrl; a LUID it can't resolve at all is treated the same way, since the HBI-direct/uncached path is the safer default for an unknown ID.

Both behaviors were confirmed with Sanaz Arabzadeh Esfarjani on the Pulse Insights team.

## Test plan
- [x] `npx vitest run` — all unit tests pass
- [x] `npx tsc --noEmit -p .` clean
- [x] `npx eslint` clean
- [x] New unit tests cover: caption -> fieldName resolution, forwarding the request unchanged when metadata can't be read, `id_type` set when `queryDatasource` returns an embedded `contentUrl`, `id_type` set when `queryDatasource` doesn't recognize the LUID at all, `id_type` left unset for a published datasource
- [x] Manually verified end-to-end against a real Tableau site with an embedded/workbook datasource containing a calculated field — see comment below for details

### DIFF
--- a/src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
+++ b/src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
@@ -279,18 +279,23 @@ async function resolveBundleRequestFieldNames(
   };
 }
 
-// Published datasources resolve via the Datasources REST API; embedded
-// workbook datasources don't exist there, so a failed lookup identifies one.
+// The Datasources REST API returns an entry for embedded/workbook datasources
+// too (it doesn't 404), but marks them with a contentUrl of the form
+// '$embedded$_<workbookLuid>' instead of a normal published contentUrl. A
+// LUID the API doesn't recognize at all is treated the same way, since
+// Pulse's HBI-direct, uncached path is the safer default for an unknown ID.
 async function resolveDatasourceIdType(
   restApi: RestApi,
   datasourceLuid: string,
 ): Promise<'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE' | undefined> {
   try {
-    await restApi.datasourcesMethods.queryDatasource({
+    const datasource = await restApi.datasourcesMethods.queryDatasource({
       siteId: restApi.siteId,
       datasourceId: datasourceLuid,
     });
-    return undefined;
+    return datasource.contentUrl?.startsWith('$embedded$')
+      ? 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE'
+      : undefined;
   } catch {
     return 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE';
   }
--- a/src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
+++ b/src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
@@ -14,6 +14,12 @@ import {
 import { WebMcpServer } from '../../../../server.web.js';
 import { isAxiosError } from '../../../../utils/axios.js';
 import { WebTool } from '../../tool.js';
+import {
+  applyFieldNameResolution,
+  buildFieldNameByCaption,
+  resolveIdTypeFromContentUrl,
+  WorkbookDatasourceIdType,
+} from '../resolveBundleRequestFieldNames.js';
 import { validateBundleRequest } from '../validatePulsePayload.js';
 
 const paramsSchema = {
@@ -233,66 +239,11 @@ async function resolveBundleRequestFieldNames(
     resolveDatasourceIdType(restApi, datasourceLuid),
   ]);
 
-  const fieldNameByCaption = new Map<string, string>();
-  if (metadataResult.isOk()) {
-    for (const field of metadataResult.value.data ?? []) {
-      if (typeof field.fieldCaption === 'string' && typeof field.fieldName === 'string') {
-        fieldNameByCaption.set(field.fieldCaption, field.fieldName);
-      }
-    }
-  }
-  const toFieldName = (caption: string): string => fieldNameByCaption.get(caption) ?? caption;
+  const fieldNameByCaption = metadataResult.isOk()
+    ? buildFieldNameByCaption(metadataResult.value.data ?? [])
+    : new Map<string, string>();
 
-  const metric = bundleRequest.bundle_request.input.metric;
-  const basicSpecification = metric.definition.basic_specification;
-
-  return {
-    ...bundleRequest,
-    bundle_request: {
-      ...bundleRequest.bundle_request,
-      input: {
-        ...bundleRequest.bundle_request.input,
-        metric: {
-          ...metric,
-          definition: {
-            ...metric.definition,
-            datasource: {
-              ...datasource,
-              ...(idType ? { id_type: idType } : {}),
-            },
-            basic_specification: {
-              ...basicSpecification,
-              measure: {
-                ...basicSpecification.measure,
-                field: toFieldName(basicSpecification.measure.field),
-              },
-              time_dimension: {
-                ...basicSpecification.time_dimension,
-                field: toFieldName(basicSpecification.time_dimension.field),
-              },
-              filters: basicSpecification.filters.map((filter) => ({
-                ...filter,
-                field: toFieldName(filter.field),
-              })),
-            },
-          },
-          metric_specification: {
-            ...metric.metric_specification,
-            filters: metric.metric_specification.filters?.map((filter) => ({
-              ...filter,
-              field: toFieldName(filter.field),
-            })),
-          },
-          extension_options: metric.extension_options
-            ? {
-                ...metric.extension_options,
-                allowed_dimensions: metric.extension_options.allowed_dimensions?.map(toFieldName),
-              }
-            : metric.extension_options,
-        },
-      },
-    },
-  };
+  return applyFieldNameResolution(bundleRequest, fieldNameByCaption, idType);
 }
 
 // The Datasources REST API returns an entry for embedded/workbook datasources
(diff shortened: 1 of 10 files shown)


dplatt84:
### Manual end-to-end testing

Tested against a real Tableau site with an embedded/workbook datasource (not a published one) containing a calculated field, to validate both fixes against production behavior rather than just mocks:

- Datasource: embedded workbook datasource with a calculated field whose caption is `Amount_Billions` and underlying `fieldName` is `Calculation_0013913069531140`.
- Sent a `generate-pulse-metric-value-insight-bundle` request using the caption (`Amount_Billions`) as the measure field, as a caller would if building the request from `get-datasource-metadata` output.
- Ran it twice: once with the caller explicitly setting `id_type: 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE'`, and once with the caller omitting it entirely.

Results:
- Caption -> fieldName resolution correctly mapped `Amount_Billions` -> `Calculation_0013913069531140` before forwarding to Pulse.
- `id_type` auto-detection correctly identified the datasource as an embedded workbook datasource and set `id_type` in both cases (with and without the caller providing it).
- Both requests succeeded end-to-end and returned real insight bundle responses from the live Insight Service.

This testing is also what caught a bug in an earlier version of the `id_type` detection: it assumed `queryDatasource` throws/404s for embedded datasource LUIDs (mirroring a pattern used elsewhere in this codebase), but against the real API it actually resolves successfully and marks embedded datasources via a `contentUrl` of the form `$embedded$_<workbookLuid>` instead. The current implementation and its unit tests reflect the corrected, verified behavior.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
```diff
@@ -171,7 +170,7 @@ Generate an insight bundle for the current aggregated value for Pulse Metric usi
             jwtScopes: generatePulseMetricValueInsightBundleTool.requiredApiScopes,
             callback: async (restApi) =>
               await restApi.pulseMethods.generatePulseMetricValueInsightBundle(
-                bundleRequest,
+                await resolveBundleRequestFieldNames(restApi, bundleRequest),
```
tonythebest, line 182: This now calls VDS `readMetadata` and Datasources REST under this tool's existing `requiredApiScopes`, but `scopes.ts` still only requests `tableau:insights:read` and `tableau:mcp_site_settings:read`. Direct-trust/UAT JWTs will be minted without `tableau:viz_data_service:read` / `tableau:content:read` (PAT E2E will not catch that). Can we match `generate-insight-cards` for those API scopes, add `tableau:mcp:datasource:read` unless we intentionally want insight-create alone to trigger DS/VDS, and add a scope regression test?

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
```diff
@@ -190,3 +189,114 @@ Generate an insight bundle for the current aggregated value for Pulse Metric usi
...
+// recognize is an embedded workbook datasource.
+async function resolveBundleRequestFieldNames(
+  restApi: RestApi,
+  bundleRequest: z.infer<typeof pulseBundleRequestSchema>,
+): Promise<z.infer<typeof pulseBundleRequestSchema>> {
+  const datasource = bundleRequest.bundle_request.input.metric.definition.datasource;
+  const datasourceLuid = datasource.id;
+
+  const [metadataResult, idType] = await Promise.all([
+    restApi.vizqlDataServiceMethods.readMetadata({ datasource: { datasourceLuid } }),
```
tonythebest: `readMetadata` only returns `Err` for HTTP 404; it throws for auth, 5xx, Zodios, and network failures (`vizqlDataServiceMethods.ts`). That rejection escapes this `Promise.all`, so Pulse never runs and the "forward unchanged" comment is false. Could we catch the call, keep the original field strings, and add a test where `mockReadMetadata` rejects?

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
```diff
@@ -190,3 +189,114 @@ Generate an insight bundle for the current aggregated value for Pulse Metric usi
...
+): Promise<'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE' | undefined> {
+  try {
+    const datasource = await restApi.datasourcesMethods.queryDatasource({
+      siteId: restApi.siteId,
+      datasourceId: datasourceLuid,
+    });
+    return datasource.contentUrl?.startsWith('$embedded$')
+      ? 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE'
+      : undefined;
+  } catch {
```
tonythebest: This `catch` returns `DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE` for 401/403, timeouts, and 5xx too, so a published datasource is silently relabeled whenever the lookup is operationally unavailable. `$embedded$` on success is the verified signal. Can we distinguish not-found from auth/network failure (and log), or document and test why every failure is safe on the HBI-direct path?

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
```diff
@@ -190,3 +189,114 @@ Generate an insight bundle for the current aggregated value for Pulse Metric usi
...
+// metadata can't be read, forward the request unchanged rather than blocking
+// the call — this is a caption/fieldName workaround, not a hard dependency.
+//
+// Separately, Pulse must always call HBI directly (never a cached query
+// result) for embedded/workbook datasources, which requires
+// id_type: 'DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE' on the request. Callers
+// can't be relied on to set this themselves, so determine it here: a LUID
+// that queryDatasource (the published Datasources REST API) doesn't
+// recognize is an embedded workbook datasource.
+async function resolveBundleRequestFieldNames(
```
tonythebest, line 217: Could this normalization live in the shared bundle runner instead of remaining private to this tool? `generate-insight-cards` still selects `fieldCaption`, and `runInsightBundle` calls Pulse directly, so the calculated-field 400 remains on the deterministic Chiron cards path after this PR. Chat/skill already uses `localName`, so this should remain a no-op/safety net there.

tonythebest (changes requested):

nawartamawi:
One non-blocking Chiron/Studio follow-up: current Studio insight paths already send the internal localName, so no Studio change is required for this PR. If Studio later wants to accept caption-based requests, it would be useful for MCP to return the normalized bundleRequest and for Studio to retain it for downstream actions such as Build Viz. Otherwise, insight generation could use the corrected fieldName while downstream actions still see the original caption. I would treat this as future-proofing, not a blocker for the current Studio flow.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
dplatt84: Fixed — added `tableau:content:read`, `tableau:viz_data_service:read` to the API scope set and `tableau:mcp:datasource:read` to the MCP scope set for this tool, matching `generate-insight-cards`. Added a regression test in `scopes.test.ts` asserting these scopes are required. Pushed in ec93cf9a.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
dplatt84: Fixed — `readMetadata` is now called via `.catch()` so a throw (auth/5xx/network) is treated the same as the 404 Err case: logged and the request is forwarded with field captions unchanged. Added a test that rejects `mockReadMetadata` with a 500 AxiosError and asserts the tool still succeeds. Pushed in ec93cf9a.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
dplatt84: Fixed — the catch now checks for a genuine 404 (via `isAxiosError`) before defaulting to `DATASOURCE_ID_TYPE_WORKBOOK_DATASOURCE`. Non-404 failures (auth/timeout/5xx) still default the same way — since we still can't distinguish published from embedded — but are now logged separately as an anomaly rather than silently swallowed. Added a test covering the non-404 case that asserts the log fires. Pushed in ec93cf9a.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
dplatt84: Fair point, and agreed the gap is real — after this PR, `generate-insight-cards`'s `runInsightBundle` path still selects `fieldCaption` and will hit the same calculated-field 400. I'd rather not fold that into this PR though: moving normalization into the shared bundle runner touches a second tool's tested behavior and is a bigger change than this PR's stated scope (fixing this tool's caption/id_type handling). I'll file a follow-up ticket to move this into the shared runner so both paths benefit — does that work, or do you consider it blocking for this PR specifically?

dplatt84:
Thanks for confirming no Studio change is needed here, Nawar. Agreed the normalized-`bundleRequest`-passthrough idea is future-proofing rather than a blocker for this PR — I'll file a follow-up ticket so it's tracked if/when Studio moves to caption-based requests.

src/tools/web/pulse/generateMetricValueInsightBundle/generatePulseMetricValueInsightBundleTool.ts
dplatt84: Went ahead and folded this into the PR (commit f88ff0fe) rather than a follow-up ticket. Extracted the caption->fieldName and contentUrl->id_type logic into `resolveBundleRequestFieldNames.ts` and wired `generate-insight-cards` to use it too — it already fetches the same VDS metadata and datasource contentUrl for its own field selection, so this reuses that instead of adding extra API calls. `runInsightBundle` itself is unchanged; the resolution happens in `generateInsightCardsTool.ts` right before each per-measure request is built. Added coverage for both the caption resolution and embedded-datasource id_type paths in `generateInsightCardsTool.test.ts`, plus unit tests for the shared helper. Full suite (2850 tests) + lint + typecheck all green.

PR #858, opened 2026-08-27
Add Insights POPC KPI template and atomic apply

nawartamawi:
## Summary

- add the generic `insights__kpi` Tableau bookmark template while retaining `insights__kpi_obsolete` for later cleanup
- display the Insight metric name without changing its casing and use Tableau teal for every change direction
- show target-period context and value, previous-period label/range and value, and relative change in the customized tooltip
- compose trusted KPI calculations and worksheet injection into one workbook mutation
- preserve the ordinary apply-and-verify path for generic and untrusted callers
- make calculation reuse semantic and format-aware, preserve donor currency cleanup, and decode binder values exactly once
- poll atomic calculation readback until Desktop settles and expose retry safety only for pre-dispatch failures
- create collision-free KPI worksheet names across worksheets, dashboards, and stories using canonical Tableau name equality while preserving duplicate-sheet reuse

## Validation

- canonical `scripts/agent-check`: all 5 gates passed
- MCP suite: 392 files, 6,449 tests passed
- focused tooltip/template suite: 42 tests passed
- lint, typecheck, desktop build, and lockstep hash gate: passed
- live POPC KPI authoring verified in Tableau Desktop

## Stack

- South backend: https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/174
- South UI: https://gitlab.tableausoftware.com/browser-clients/tab-agent-south-ui/-/merge_requests/92


src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -2576,6 +2653,13 @@ export const getBindTemplateTool = (server: DesktopMcpServer): DesktopTool<typeo
             return sessionResult.error.toErr();
           }
           const resolvedSession = sessionResult.value;
+          const atomicCalcApply =
```
tonythebest: **Note (non-blocking):** `atomicCalcApply` is gated on a four-way AND (`extra.config.allowSkipValidation === true && skip_validation === true && auto_apply === true && proposal.template === 'insights__kpi' && calcs?.length > 0`). If any one of these silently drops out — e.g. a future caller sets `skip_validation` but not `auto_apply`, or the trusted-config flag isn't propagated — the code falls back to the non-atomic multi-apply path with no signal that the atomic single-apply guarantee was lost. That fallback is *safe*, but debugging "why did this KPI bind take two applies / lose the readback verification" would be hard. Consider a debug-level log when a KPI insight bind is eligible for the atomic path on every condition except one, so an accidental gating regression is observable rather than silent. Not a blocker — the behavior is correct today.

tonythebest (approved):

tonythebest:
🤖 **Automated review hold — pending human review by Tony**

Reviewed the Insights POPC → KPI atomic-apply change (generic `insights__kpi.tbm` template, `workbookHasSheetNamed`, collision-free worksheet titling, single-apply `prepareCalculationsInWorkbook` + `pollReadback` verification, idempotent-retry via `normalizeFormula`/`findColumnByCaption`, and the trusted-template guard). Sophisticated and well-tested (136→138 TBM corpus; new atomic/collision/currency-preservation tests); no Blocker/Major found — one non-blocking Note posted about silent fallback when the atomic-apply gate partially fails. Verdict: LGTM / approved.

---

**Why this comment exists:** Tony's MR/PR-review autopilot reviewed and approved this change and left one non-blocking Note. This comment flags the PR for optional later human review by Tony. meta-pr-review ran this pass: yes.

**Code owner:** if you're confident this change is safe, feel free to proceed / merge — no need to wait on Tony.

tonythebest (approved):

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -2280,13 +2449,25 @@ async function performAutoApply({
...
+          bind: bindMs,
+          inject: injectMs,
+          apply: applyMs,
+          readback: readbackMs,
+          summary: summaryMs,
+          total: bindMs + injectMs + applyMs + readbackMs + summaryMs,
+        }
+      : { bind: bindMs, inject: injectMs, apply: applyMs },
+    ...(trustedDeterministicApply && receiptInput.readback
+      ? { verification: receiptInput.readback }
```
tonythebest, line 2592: **High — preserve post-dispatch uncertainty across the wire.** This serializes `verification`, but the result still carries `applied: true` and includes the prepared atomic calculations in `authored_calcs`. South forwards `verification.status` but determines terminal success from `applied`/`reused`, so both `failed` (known mismatch) and `skipped` (unknown after dispatch) become successful terminal results. Please make completion explicit: `failed` should be non-success, while `skipped` should remain unknown/non-terminal with `may_have_applied: true` and `retry_safe: false`, not blindly replayed. Gate South’s terminal result on that contract and add cross-repo tests for both statuses.

tonythebest (commented):
Follow-up adversarial review on the current head found two actionable contract issues; details are in the inline threads.

src/tools/web/pulse/validatePulsePayload.ts
```diff
@@ -69,6 +69,34 @@ export function validateBundleRequest(req: BundleRequest): string | null {
     );
   }
 
+  const lastXPeriod = ms.measurement_period.last_x_period;
```
tonythebest, line 113: **Medium — apply this rolling-window validation to Insight Briefs too.** These checks run only in `validateBundleRequest`, while `last_x_period` is part of the shared `pulseMetricSpecificationSchema` used by Insight Brief. A Brief with `range: RANGE_CURRENT_PARTIAL` plus `last_x_period` therefore passes `validateBriefRequest`; Insights Service ignores that rolling configuration unless the range is `RANGE_BY_CONFIG`, silently answering for a different period. Please share these measurement-period invariants with every Brief `metric_group_context` and add negative Brief tests that also prove the API is not called.

mattcfilbert (commented):
[P1] Do not decode parsed worksheet names twice

parseXML() has already decoded attribute entities before these comparisons. Passing those parsed values through xmlNamesEqual() decodes them again, so distinct names collapse: serialized A &amp; B and A &amp;amp; B both match the requested literal A &amp; B. On this head, an exact probe through removeSameNamedWorksheet() removed both worksheets. That can delete an unrelated loose sheet during apply. Compare parsed names with trim/NFC only, keep entity decoding for raw XML slices, and add a regression that preserves the distinct sibling.

🤖 Posted by MattGPT.

src/tools/desktop/authoring/binder/bindTemplate.ts
nawartamawi: Fixed in ce8f44c0. Trusted failed or skipped readback now returns applied:false, may_have_applied:true, retry_safe:false, plus the verification receipt. It uses the existing structural readback and adds no extra Desktop call. The canonical gate is green: 6,597 tests after merging current feature/desktop.

nawartamawi:
Addressed the parsed-XML name finding in ce8f44c0: parsed worksheet, window, and zone names now use a comparator that does not decode entities a second time; raw serialized XML keeps the existing entity-aware comparison. Regression coverage distinguishes A &amp; B from A &amp;amp; B. The branch also includes current feature/desktop and the canonical gate is fully green.

src/tools/desktop/authoring/binder/bindTemplate.ts
tonythebest: Verified on `a8367975` / `ce8f44c0`. Failed or skipped trusted readback now goes through `applyFallback` with `applied: false`, `may_have_applied: true`, `retry_safe: false`, and `isError: true`. That closes this thread for me.

src/tools/web/pulse/validatePulsePayload.ts
tonythebest: Verified on `a8367975`. `validateMeasurementPeriod` is shared with Brief, and the tool test asserts `mockGeneratePulseInsightBrief` is not called for an invalid rolling window. That closes this thread for me.

tonythebest (approved):
Re-reviewed `a8367975` after the feedback push.

Trusted failed/skipped readback now returns `applied: false`, `may_have_applied: true`, `retry_safe: false`. Brief `last_x_period` uses shared `validateMeasurementPeriod`, and parsed worksheet names compare with trim+NFC only. LGTM on this head.

Nit (not blocking): `unescapeXml` can still theoretically double-decode a nested numeric entity; real Tableau formula XML reachability looks low. Title still omits `@W-24012563`.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1568,6 +1595,87 @@ function applyFailureDisposition(
...
+): string {
+  const worksheetXml = extractSheetXml(workbookXml, worksheetTitle);
+  if (!worksheetXml) {
+    throw new Error(`injected worksheet "${worksheetTitle}" was not found for identity stamping`);
+  }
+  const attributeName = 'user:tableau-agent-idempotency-key';
+  const escapedKey = escapeXmlAttribute(idempotencyKey);
+  const existing = new RegExp(`\\s${attributeName}=(['"])[^'"]*\\1`);
+  const stampedWorksheet = existing.test(worksheetXml)
+    ? worksheetXml.replace(existing, ` ${attributeName}='${escapedKey}'`)
```
tonythebest, line 1672: Minor: `escapedKey` is interpolated into the **replacement** argument of `String.prototype.replace` on both branches here (and on the `<worksheet` branch just below). In a replacement string, `$` is special — `$&`, `$1`, `$'`, `` $` `` are substituted rather than inserted literally. `escapeXmlAttribute` escapes XML entities but not `$`, so an idempotency key containing `$` (e.g. `$&`) would be corrupted on write, and the stamped attribute would then no longer match the original key on the next `worksheetTitleForIdempotencyKey` lookup — silently breaking idempotency for that key.

Not a live bug given keys are currently hex, but it's a latent footgun if the key source ever widens. Safest fix is to pass a replacer function instead of a string, e.g. `worksheetXml.replace(existing, () => \` ${attributeName}='${escapedKey}'\`)`, which treats the value literally.

mattcfilbert (approved):
lgtm

PR #848, opened 2026-08-26
feat(desktop): round dashboard zones

mattcfilbert:
## Decision

Add one typed tool for native dashboard-zone corner radius in Tableau Desktop 2026.1 and newer. This sets the rule for this path: resolve the dashboard, reject known unsupported releases, edit only the selected zones under the shared apply lock, and report success only after verified readback.

This PR covers dashboard objects and horizontal/vertical containers. Rounded bar charts require a separate chart reconstruction and are not part of this change. The branch also keeps the existing `refine-worksheet` stale-source check as a separate commit; it is adjacent fail-closed safety and does not change dashboard behavior.

Future changes must preserve direct `zone-style` ownership, both native container types (`layout-basic` and `layout-flow`), the version check under the apply lock, the one-apply limit, source-drift checks, and exact structure/content readback.

## What changed

- Adds `format-dashboard-zones` with `all`, `containers`, and explicit `zone_ids` scopes.
- Preserves existing XML and accepts only the narrow color-case and style-order normalization observed from Desktop.
- Rejects known released Desktop versions before 2026.1 before any write. Missing and internal development version strings remain compatible and must still pass exact readback.
- Keeps the shared apply lock until the new document is visible and verified.
- Makes `refine-worksheet` reject a stale source document before applying.
- Bumps the package to 2.67.2.

## Evidence

- `scripts/agent-check`: all 5 checks green; 393 test files and 6,479 tests passed, plus the Desktop build and lockstep hash gate.
- Focused dashboard suite: 47/47 passed, including version changes under the apply lock, high-contrast Desktop normalization, content-loss controls, concurrency, and source drift.
- Live Desktop: radius 20 and then 36 each returned `verified: true` for zones 9, 10, and 11 on a styled dashboard.
- Save/reopen: all three zones retained radius 36; the rendered PNG before save and after reopen was byte-identical.
- Final Sol review: clean; no P1/P2 findings.

Approving this PR means accepting native rounded corners for dashboard zones, plus the separate stale-worksheet safety check already on this branch. Rounded bar chart reconstruction remains separate.

— MattGPT

PR #859, opened 2026-08-28
feat(refine): round compatible stacked bars

mattcfilbert:
## Decision

Accept a narrow `refine-worksheet` operation that rebuilds compatible stacked bar charts as verified Tableau Polygon marks with subtle outer-corner rounding. This is a one-off operation for this chart shape, not a rule that every worksheet transform must follow.

Approving this PR means accepting the supported input contract, the fail-closed checks, and the current Data Guide/View Data limitation described below.

## Description

This adds `round_stacked_bar` to `refine-worksheet` with a single `subtle` preset.

For compatible existing stacked bars, the operation:

- rounds only the exposed outer corners while keeping joins between stacked segments square;
- preserves the worksheet identity and existing dashboard membership;
- preserves categorical filters, sort, colors, titles, axis formatting, and signed values;
- verifies the planned worksheet, Tableau's readback, the embedded workbook copy, and the rendered geometry summary before reporting success;
- refuses unsupported shapes and workbooks with non-empty top-level actions before writing.

The implementation keeps one structural contract in the planner/recognizer, plus an independent numeric geometry check and whole-workbook collateral guard.

## Motivation and Context

Tableau does not expose a corner-radius control for Bar marks. The common dual-axis Circle and thick-Line workarounds produce pill ends and do not preserve slight outer rounding with square joins inside a stack.

This uses Tableau's native Polygon mark family to draw the requested shape while keeping the existing worksheet and dashboard references intact.

Known v1 limitation: Data Guide and View Data may expose the generated geometry helper fields. The operation does not claim to clean or hide those surfaces.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- `scripts/agent-check`: all five checks passed
  - lint
  - typecheck
  - 396 test files / 6,684 tests
  - Desktop build
  - lockstep hash gate
- Focused planner, verifier, and apply-wrapper suites: 170/170 passed after the final save/reopen normalization fix.
- Live Desktop smoke on a disposable workbook:
  - converted an existing signed, filtered stacked bar in place with one refine call;
  - preserved worksheet, window, dashboard, and dashboard-zone identities;
  - verified 612 geometry rows across 51 category/segment groups;
  - preserved the Region filter, descending sort, palette, titles, axis formatting, positive stacks, and negative stacks;
  - saved the workbook, reopened it in a fresh Tableau Desktop process, and repeated canonical structure, summary, worksheet-image, and dashboard-image verification.

## Related Issues

- Related dashboard-corner work: #848
- Rounded-bar design and live-spike discussion in the Tableau Studio Slack thread

## Checklist

      environment variable or changing its default value. None.

## Contributor Agreement

      its Contribution Checklist.


src/desktop/refine/roundStackedBarVerification.ts
```diff
@@ -0,0 +1,1286 @@
...
+      'seed-evidence',
+      'Underlying data does not expose one unambiguous Category, Segment, measure, and optional filter column.',
+    );
+    return result(findings);
+  }
+
+  const visible = new Set(baseline.groups.map((group) => groupKey(group.category, group.segment)));
+  const values = new Map<string, Set<number>>();
+  for (const row of underlying.rows) {
+    if (contract.filter && filter?.ok && row[filter.index] !== contract.filter.member) {
```
lauren-jacksonSFDC: ✨ _Lauren's code review bot:_ ✨

`round_stacked_bar` refuses any worksheet whose stacked bar has a categorical filter on a non-string dimension (integer/date/boolean).

**Trigger:** `contract.filter.member` is the raw TWB XML attribute string — `readFilter` only strips surrounding quotes, with no type coercion and no datatype guard. But the underlying-data cell at `row[filter.index]` can come back as a native number/boolean/date: the same function coerces the *measure* value with `typeof rawValue === 'number'` (~lines 322-328), which proves the API returns native types. So `row[filter.index] !== contract.filter.member` mismatches every row (e.g. `2020 !== "2020"`), seed-evidence returns `ok:false`, and `applyRoundedStackedBar.ts:161` treats that as a fail-closed `beforeMutationFailure('seed-preflight')`.

**Why this matters:** The apply is refused before it runs for a whole class of valid workbooks, and no test covers a numeric/date/boolean member. Coerce both sides to a normalized string (or compare by datatype) before the `!==`.

src/desktop/refine/roundStackedBar.ts
```diff
@@ -0,0 +1,1718 @@
...
+  const encodedMember = members[0].getAttribute('member') ?? '';
+  const member =
+    encodedMember.startsWith('"') && encodedMember.endsWith('"')
+      ? encodedMember.slice(1, -1)
+      : encodedMember;
+  return {
+    ok: true,
+    allowedColumn: filterInstance.column,
+    value: {
+      caption: definitions.get(filterInstance.column)?.caption ?? unbracket(filterInstance.column),
```
lauren-jacksonSFDC, line 581: ✨ _Lauren's code review bot:_ ✨

The filter contract drops the field's technical column name, so `round_stacked_bar` refuses any worksheet whose filter field's caption differs from its underlying column name (renamed/aliased fields).

**Trigger:** `readFilter` has `filterInstance.column` in hand (it uses it as `allowedColumn` just above), but the filter value object returned here keeps only `caption`, `columnInstance`, and `member` — unlike category/segment/measure, which thread their real `.column` through. `verifyRoundStackedBarSeedEvidence` then fabricates the name as `` column: `[${contract.filter.caption}]` `` (`roundStackedBarVerification.ts:293`), `underlyingFieldAliases` builds every alias from that fabricated `[caption]`, and `resolveColumn` has no columnInstance fallback for underlying-data columns — so for any field whose caption ≠ technical column name, `filter.ok` is false and the blocking seed-evidence preflight fails.

**Why this matters:** A valid, compatible stacked bar is rejected whenever its filter field was renamed/aliased (a common case). Carry `column: filterInstance.column` in the filter value object and match on it in verification instead of fabricating from the caption.

src/tools/desktop/authoring/sheets/refineWorksheet.ts
```diff
@@ -211,6 +240,89 @@ export const getRefineWorksheetTool = (
           // this fetch and the readback can't miss.
           const readbackRef = worksheetFragmentSimpleId(sourceXml) ?? canonicalWorksheetName;
 
+          if (operation === 'round_stacked_bar') {
```
lauren-jacksonSFDC, line 242: ✨ _Lauren's code review bot:_ ✨

This new `round_stacked_bar` branch bypasses the pipeline the file-header `Flow:` comment (lines 1-11) describes, leaving that header stale.

**Trigger:** The header describes the file's single flow (`... -> ensureUserNamespace -> preflight validation -> load-worksheet-xml (apply ONCE) -> get-worksheet-xml readback ...`) and says the planners live in `src/desktop/refine/refineWorksheet.ts`. This branch reaches none of that: it plans via `src/desktop/refine/roundStackedBar.ts` (an uncited second planner) and applies through `applyRoundedStackedBar.ts`, which has its own per-sheet apply route, preflight, and multi-signal readback; `ensureUserNamespace`/`loadWorksheetXml` are only reached from the other branch.

**Why this matters:** A maintainer trusting the header forms a wrong mental model of this path. Scope the header to the pre-existing operations, or extend it to cover the round_stacked_bar path.

src/desktop/refine/roundStackedBarVerification.ts
mattcfilbert: Fixed in c5b56cb7. Seed preflight keeps exact string matching and explicitly supports native finite-number and boolean values. Null, objects, non-finite values, and alternate spellings still fail closed. Added matching and mismatch regressions. — MattGPT

src/desktop/refine/roundStackedBar.ts
mattcfilbert: Fixed in c5b56cb7. The filter contract now keeps the technical source column, and the underlying-data request uses technical columns for Category, Segment, Value, and filter. Added renamed-caption coverage through apply and verification. — MattGPT

src/tools/desktop/authoring/sheets/refineWorksheet.ts
mattcfilbert: Fixed in c5b56cb7. The header now documents the shared initial fetch and the separate ordinary-refine and rounded-bar apply/readback paths. — MattGPT

src/desktop/refine/roundStackedBar.ts
```diff
@@ -0,0 +1,1719 @@
...
+  }
+  if (slices.length !== 1) {
+    return refusal('the categorical filter requires one matching slice.');
+  }
+  const sliceColumns = elements(slices[0], 'column');
+  if (sliceColumns.length !== 1 || shelfText(sliceColumns[0]) !== filterReference) {
+    return refusal('the categorical filter requires one matching slice.');
+  }
+
+  const encodedMember = members[0].getAttribute('member') ?? '';
```
tonythebest, line 572: Tableau string members can contain backslash-escaped internal quotes. A shipped TWB fixture (`Message Book, Standard Line \"While You Were Out\"...`) retains those backslashes after DOM parsing and the current outer-delimiter slice, while underlying data contains the literal quote characters. That makes seed preflight safely refuse an otherwise compatible filter. Could we decode Tableau member escapes here, or explicitly refuse them during planning with a targeted reason and regression? Non-blocking because no write is sent.

src/desktop/refine/roundStackedBarVerification.test.ts
```diff
@@ -0,0 +1,1215 @@
...
+}
+
+type HelperSuffix = 'bin' | 'path' | 'x' | 'y';
+
+function helperCaption(suffix: HelperSuffix): string {
+  if (suffix === 'bin') return 'TMCP rounded path frame';
+  return `TMCP rounded ${suffix.replaceAll('_', ' ')}`;
+}
+
+const sourceGroups = [
```
tonythebest, line 120: The signed fixture currently separates positive (`Alpha`) and negative (`Beta`) stacks by category. Could we add one category containing both signs so the regression proves the two independent zero-origin chains, square internal joins, and both rounded tips? The implementation partitions by category+sign and looks correct, so this need not block merge.

tonythebest (approved):
Approved at `f11fd54d`. Lauren's three findings look fixed and regression-covered. Two optional follow-ups below — neither blocks merge.

src/desktop/refine/roundStackedBar.ts
mattcfilbert: Fixed in c9bc37c6. Outer-quoted Tableau string members now decode the verified \" escape before seed comparison. Unquoted numeric/boolean members and unknown escape sequences remain byte-exact. Added the shipped Message Book value as a planner regression. — MattGPT

src/desktop/refine/roundStackedBarVerification.test.ts
mattcfilbert: Added in c9bc37c6. The verifier fixture now covers one category with independent positive and negative zero-origin chains, square internal joins, and one rounded outer tip for each sign. — MattGPT

PR #860, opened 2026-08-30
@W-23752441: correct docs referencing removed get-stale-content-report tool

Akash-Rastogi:
[Akash's agent]

## Summary

The `get-stale-content-report` tool was **removed** and consolidated into `query-admin-insights` with `kind: "stale-content"`. Three docs still named the removed tool and misled readers; this corrects them to point at the current tool.

| File | Context | Now references |
|---|---|---|
| `docs/docs/intro.md` | Prompt list — `stale-content-cleanup-inform` row | `query-admin-insights` with `kind: "stale-content"` |
| `docs/docs/configuration/mcp-config/request-overrides.md` | `STALE_CONTENT_MIN_AGE_DAYS` override | `query-admin-insights` tool (`kind: "stale-content"`) |
| `docs/docs/configuration/mcp-config/env-vars.md` | `ADMIN_GATE_CACHE_TTL_MINUTES` cache note | `query-admin-insights` (`kind: "stale-content"`) |

Wording mirrors the already-correct dedicated page `docs/docs/tools/admin-insights/query-admin-insights.md` and the adjacent `STALE_CONTENT_MAX_ROWS` section.

## Why

This resolves the **documentation** root cause behind bug W-23752441. Functional behavior was verified correct during triage — the alleged "stale-content requires an explicit project" behavior is **not a bug**: `projectIds` is already optional and defaults to whole-site scope. This PR only corrects docs that referenced the removed `get-stale-content-report` tool, which was the likely source of the reported confusion.

## Testing

- `rg get-stale-content-report` across the repo → **0** remaining references.
- Docs-only change (no `src/**` touched) → `Version Bump Check` CI skips; no version bump required.
- Docusaurus build is exercised by the `Test deployment` CI workflow (`npm ci && npm run build` over `docs/`). No new links were introduced — the edits are inline-code spans that already build elsewhere in these same files.

## Type of change

- [x] Documentation only

### DIFF
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -497,7 +497,7 @@ TTL (in minutes) for caches used by admin-only tools. Affects:
 
 - Admin role lookups (`assertAdmin`)
 - Admin Insights dataset LUID resolution
-- Project ID → name resolution used by `get-stale-content-report`
+- Project ID → name resolution used by `query-admin-insights` (`kind: "stale-content"`)
 
 - Default: `5`
 - Minimum: `1`
--- a/docs/docs/configuration/mcp-config/request-overrides.md
+++ b/docs/docs/configuration/mcp-config/request-overrides.md
@@ -160,7 +160,7 @@ Overrides whether Metadata API requests are disabled.
 
 ### [`STALE_CONTENT_MIN_AGE_DAYS`](env-vars.md#stale_content_min_age_days)
 
-Overrides the default threshold (in days) used by the `get-stale-content-report` tool to flag stale workbooks and published datasources. Override value must be an integer in the range `1`–`3650`. Empty string reverts to the default (`90`).
+Overrides the default threshold (in days) used by the `query-admin-insights` tool (`kind: "stale-content"`) to flag stale workbooks and published datasources. Override value must be an integer in the range `1`–`3650`. Empty string reverts to the default (`90`).
 
 | Restriction Type | Behavior |
 |---|---|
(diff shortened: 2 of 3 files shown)


Alon-ST-DATA (approved):

Alon-ST-DATA (approved):
[Alon's agent] Reviewed head d8aa54a. **Verdict: ship** — correct, complete, consistent with siblings.

- After this PR, `grep -rn get-stale-content-report` (excluding node_modules) returns **0** hits repo-wide — the three edited lines were the last stale references.
- Replacement `kind: "stale-content"` matches the real Zod enum (`src/tools/web/adminInsights/queryAdminInsights.ts:45`) and the live `query-admin-insights` tool. `get-stale-content-report` is genuinely gone from the registry (survives only as an internal helper), so the docs were factually wrong before — the correction is right.
- Version-bump skip is correct (docs-only; the bump check keys off `src/**`). No token-budget impact.

**Test coverage (optional follow-up, not blocking):** the change is Markdown prose, so there's nothing to unit-test on the edit itself, and the Docusaurus build won't flag a stale-but-valid tool name inside an inline-code span — which is exactly why this survived the tool's removal. There is currently no docs↔registry consistency guard anywhere in the repo. Worth a cheap regression test as a separate task:

- New `src/tools/web/docsToolReferences.test.ts` (vitest, no live site).
- Minimum: `expect(allDocsText).not.toContain('get-stale-content-report')` — would have failed before this PR, passes now.
- Better: scan `docs/docs/**/*.md` for backticked kebab-case tokens against a deny-list of removed tool names, so future tool consolidations/renames can't silently leave stale docs. Scope to a deny-list to avoid false-positives on `kind` values like `stale-content`/`job-performance`.

PR #861, opened 2026-08-31
@W-24061219: Gate MCP-Apps tool registration on client capability and exclude claude.ai

jarhun88:
## Description
- Gate MCP-Apps tool registration on whether the connecting client actually advertised the SEP-1724 UI capability at `initialize`, instead of relying solely on the `mcp-apps` feature flag.
- Exclude claude.ai (OAuth/HTTP) from app-tool registration even when it advertises the capability, due to a known incompatibility in its MCP-Apps renderer. Reuses the existing `getClientDisplayName` telemetry host mapping — no new client identity logic.
- Threads `capabilities`/`clientId` end-to-end (`Server`, `WebMcpServer`, `Session`, both HTTP branches in `express.ts`) mirroring the existing `clientInfo` pattern.
- New `src/server/mcpUiCapability.ts` helper (`clientSupportsMcpApps`) with its own unit tests.

## Motivation and Context
Previously, any client got an interactive "app tool" whenever the `mcp-apps` flag was on, even if it had no way to render one — and claude.ai specifically advertises the capability but has a broken renderer for it. This adds a second, orthogonal gate (live client capability) plus an explicit deny-list entry for the one known-incompatible client, defaulting safely to plain-tool registration whenever support is unknown/absent.

**Known, documented limitation:** in stateless HTTP mode (`DISABLE_SESSION_MANAGEMENT=true`), a fresh server/transport is created and torn down on every request, and per the MCP spec `capabilities` is only ever present on the literal `initialize` request body — so app-tool activation via live capability detection never fires for real tool-call traffic in that mode (it always safely falls back to plain tools). This is called out explicitly in a code comment (`src/server/express.ts`) and in the `DISABLE_SESSION_MANAGEMENT` docs page, rather than silently left as an unexplained gap. The claude.ai exclusion is unaffected, since it's keyed on `clientId` (available on every request, not just `initialize`).

This change went through 3 rounds of independent dual-reviewer + synthesis review:
- **Round 1** found and fixed a regression where the stateless HTTP branch wasn't threading `capabilities`/`clientId` at all.
- **Round 2** found the deeper structural limitation described above (capabilities are only available on `initialize`, so app tools can never activate for real requests in stateless mode) and confirmed it isn't fixable by more parameter threading.
- **Round 3** confirmed the fix (documenting the limitation rather than engineering around it) — canonical `VERDICT: PASS`.

GUS: W-24061219

## Type of Change
- [x] New feature
- [x] Documentation update

## How Has This Been Tested?
- `npx vitest run` — 177 files / 2842 tests passing, including 5 new tests for `mcpUiCapability.ts` and 5 new/updated cases in `server.web.test.ts` covering: capable client registers app tool, missing/lacking capability falls back to plain, flag off falls back to plain, claude.ai clientId falls back to plain despite otherwise qualifying, and a non-Claude known client (Cursor) still registers as an app tool.
- `npm run build` passing (default + MCP Apps + tracing bundles).

## Related Issues
GUS: W-24061219 (https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002jTxXKYA0/view)

## Checklist
- [x] I have documented any breaking changes in the PR description (none — this is additive; stateless-mode limitation is a documented pre-existing constraint, not a behavior change for any currently-working case)

## Contributor Agreement

### DIFF
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -330,6 +330,9 @@ information about the client's identity, capabilities, and protocol version comp
   clients to provide that session ID in the `mcp-session-id` header for subsequent requests.
 - Set this to `true` if you are using the HTTP transport and your client does not support or need
   session management.
+- MCP-Apps interactive tool rendering requires session management to detect client capabilities.
+  When `true`, app tools always fall back to plain tool registration, regardless of what the
+  connecting client supports.
 
 <hr />
 
--- a/docs/docs/configuration/mcp-config/env-vars.md
+++ b/docs/docs/configuration/mcp-config/env-vars.md
@@ -331,8 +331,9 @@ information about the client's identity, capabilities, and protocol version comp
 - Set this to `true` if you are using the HTTP transport and your client does not support or need
   session management.
 - MCP-Apps interactive tool rendering requires session management to detect client capabilities.
-  When `true`, app tools always fall back to plain tool registration, regardless of what the
-  connecting client supports.
+  When `true`, app tools fall back to plain tool registration, regardless of what the connecting
+  client supports — except tools that opt out of the plain-tool fallback entirely (via
+  `hideWhenUnsupported`), which are omitted from registration altogether in this mode.
 
 <hr />
 
--- a/.work/specs/W-24061219-mcp-apps-client-capability-gate.md
+++ b/.work/specs/W-24061219-mcp-apps-client-capability-gate.md
@@ -1,0 +1,89 @@
+# W-24061219: Gate MCP-Apps tool registration on client capability
+
+## Source
+
+- Ticket: GUS W-24061219 (https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002jTxXKYA0/view)
+- Slack:
+- Google Docs/Drive:
+- Owner: Jaehun Song
+- Date: 2026-08-28 (round 1) — 2026-08-31 (hideWhenUnsupported follow-up)
+- Related PRs/issues: https://github.com/tableau/tableau-mcp/pull/861
+
+## Intent
+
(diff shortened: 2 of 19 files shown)


docs/docs/configuration/mcp-config/env-vars.md
```diff
@@ -330,6 +330,9 @@ information about the client's identity, capabilities, and protocol version comp
   clients to provide that session ID in the `mcp-session-id` header for subsequent requests.
 - Set this to `true` if you are using the HTTP transport and your client does not support or need
   session management.
+- MCP-Apps interactive tool rendering requires session management to detect client capabilities.
```
sf-ali-mahmoud, line 333: This bullet says app tools "always fall back to plain tool registration" when DISABLE_SESSION_MANAGEMENT=true, but that's not accurate for tools that opt out of the fallback via hideWhenUnsupported (currently renderInteractiveViz). Those tools hit the continue branch in server.web.ts and don't register at all in stateless mode — neither as an app tool nor a plain one. An operator reading this doc, setting this flag, and then looking for render-interactive-viz as a plain tool will not find it and won't know why.

Suggest replacing the bullet with:
  ▎ - MCP-Apps interactive tool rendering requires session management to detect client capabilities.
  ▎   When `true`, app tools fall back to plain tool registration, regardless of what the connecting
  ▎   client supports — except tools that opt out of the plain-tool fallback entirely (via
  ▎   `hideWhenUnsupported`), which are omitted from registration altogether in this mode.

sf-ali-mahmoud (approved):
md update before merge.

docs/docs/configuration/mcp-config/env-vars.md
jarhun88: Fixed in c7c31fe5 — updated the bullet to call out the `hideWhenUnsupported` case explicitly, using your suggested wording. Thanks for catching this.

— Jaehun Bot

PR #862, opened 2026-08-31
@W-23757359: Surface connected Tableau server + site in MCP instructions

Akash-Rastogi:
[Akash's agent]

## Summary

A `tableau-mcp` process is bound to exactly one `(server, site)` pair, but today it surfaces the configured `SERVER` / `SITE_NAME` **nowhere** a client or model can read: `serverName` is a static literal, `buildWebInstructions()` never mentions them, and no tool/resource/serverInfo carries them. When an admin runs several instances (one per site), there is no way to tell which instance maps to which Tableau site.

This extends the merged server-instructions work (`#840`, `@W-23757364`) by appending a third, always-safe clause to `buildWebInstructions()` naming the connected host + site, so the connection identity is discoverable in the MCP `initialize` instructions that the host injects into every session.

## Fix

`src/server.web.ts` — `buildWebInstructions()` now appends:

> `This server is connected to Tableau server <SERVER> (site: <SITE_NAME>).`

Design decisions (all verified against current code):

- **Raw env, not `getConfig()`.** `SERVER`/`SITE_NAME` (and the site-pinning flags below) are read straight from `process.env` — the same rationale already documented for `ADMIN_TOOLS_ENABLED` in this function: eagerly building `Config` here would newly require `SERVER` and would consume test `getConfig()` mocks.
- **Sanitized.** Values pass through `removeClaudeMcpBundleUserConfigTemplates` so an unreplaced Claude-Bundle placeholder (e.g. `${user_config.site_name}`) is blanked instead of leaking verbatim into model-facing text.
- **`SERVER` printed verbatim** — no `new URL(server).hostname` reparse. Keeps this deliberately validation-free path exception-free and matches the operator's own configured value.
- **Site label defaults to `Default`** when `SITE_NAME` is empty (mirrors `src/server/oauth/callback.ts:129`).
- **Clause omitted entirely when `SERVER` is unset** (hosted `mcp.tableau.com` / pure multi-tenant OAuth) — avoids advertising a misleading static site.
- **The site is named only when it is statically pinned.** When it floats per connecting user — OAuth with `OAUTH_LOCK_SITE=false`, or passthrough auth (`ENABLE_PASSTHROUGH_AUTH=true`) — the clause names the pod but defers the site: `(site: determined per user at sign-in)`, rather than assert a stale static `SITE_NAME`. (Addresses PR review — thanks @Alon-ST-DATA.)

The change lives solely in the shared `buildWebInstructions()` builder, so it flows through **both** entry points (the default `WebMcpServer` path and the combined `src/index.combined.ts` path) and stays covered by the lockstep `invariant` in `src/server.ts`.

The bug title also stresses "naming". I evaluated enriching the MCP `serverInfo` `title` with the site, but **deliberately skipped** it: the `name` is asserted as a stable literal and clients may key on it, and a dynamic `title` would have to be threaded through three separate construction sites that would otherwise drift — disproportionate risk once the instructions clause already delivers the identity to the model.

## Testing

- **Unit** (`src/server.web.test.ts`): instructions name `SERVER` + `SITE_NAME` when both set; site falls back to `Default` when empty; clause omitted when `SERVER` unset; unresolved `${user_config.*}` template does not leak; **site deferred under `OAUTH_LOCK_SITE=false` and under passthrough**; static site named when site-locking is on.
- Full pre-commit checklist green: `npm run build` (all variants incl. combined), `npx tsc --noEmit`, `npm run lint`, `npm test` (2844 unit tests).

## Known limitation

This fix stops the server from asserting a **wrong** site — under the per-user modes (OAuth `OAUTH_LOCK_SITE=false`, passthrough) it now defers rather than printing a stale `SITE_NAME`. What it does **not** do is *resolve and display the actual live signed-in site* in those modes: `buildWebInstructions()` runs once at construction with no request context, so it can't name a per-user site. Doing that would require threading `tableauAuthInfo` through the handshake / per-request instruction composition — a separate, larger follow-up. This PR is correct for the common single-site deployment and safe (defers, never lies) in the dynamic modes.

## Risk

**Low.** Additive, single-function change to a string builder; no behavior change when `SERVER` is unset, and existing base/admin instructions tests use substring assertions that are unaffected. Version bumped to `4.7.2`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.5.8",
+  "version": "4.7.2",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/src/server.web.ts
+++ b/src/server.web.ts
@@ -15,6 +15,7 @@ import { join } from 'path';
 
 import pkg from '../package.json';
 import { getConfig } from './config.js';
+import { removeClaudeMcpBundleUserConfigTemplates } from './config.shared.js';
 import { ServiceUnavailableError } from './errors/mcpToolError.js';
 import { getFeatureGate } from './features/init.js';
 import { getTableauServerInfo } from './getTableauServerInfo.js';
@@ -57,11 +58,13 @@ const ADMIN_INSTRUCTIONS =
  * Single source of truth for the web server's `initialize` instructions string. Returns the base
  * guidance, with the admin/site-health guidance appended only when ADMIN_TOOLS_ENABLED is set.
  *
- * ADMIN_TOOLS_ENABLED is read straight from env (matching Config's exact semantics in config.ts:
- * `ADMIN_TOOLS_ENABLED === 'true'`) rather than via getConfig(). The instructions string depends
- * ONLY on this flag, and it is composed at handshake/registration-independent construction time —
- * eagerly building the full Config here would newly require SERVER to be set and would consume test
- * getConfig() mocks. Direct env read keeps composition side-effect-free.
+ * All inputs are read straight from env rather than via getConfig(): ADMIN_TOOLS_ENABLED (matching Config's
+ * exact semantics in config.ts, `ADMIN_TOOLS_ENABLED === 'true'`); SERVER/SITE_NAME through
+ * removeClaudeMcpBundleUserConfigTemplates so an unreplaced Claude-Bundle `${user_config.*}` template is
+ * blanked rather than leaked; and OAUTH_LOCK_SITE / ENABLE_PASSTHROUGH_AUTH, which decide whether the site
+ * is statically pinned. Composed at handshake/registration-independent construction time — eagerly building
+ * the full Config here would newly require SERVER to be set and would consume test getConfig() mocks. Direct
+ * env reads keep composition side-effect-free.
  *
  * Both the default/web path (WebMcpServer constructs its own McpServer) and the combined path
  * (index.combined.ts supplies a pre-built McpServer) MUST feed instructions through this function.
@@ -71,7 +74,30 @@ const ADMIN_INSTRUCTIONS =
  */
 export function buildWebInstructions(): string {
   const adminToolsEnabled = process.env.ADMIN_TOOLS_ENABLED === 'true';
-  return adminToolsEnabled ? `${BASE_INSTRUCTIONS} ${ADMIN_INSTRUCTIONS}` : BASE_INSTRUCTIONS;
+  const instructions = adminToolsEnabled
+    ? `${BASE_INSTRUCTIONS} ${ADMIN_INSTRUCTIONS}`
+    : BASE_INSTRUCTIONS;
+
+  // Connection clause: name the single Tableau instance this process is bound to so a client/model can tell
+  // which one it maps to. SERVER (the pod) is printed verbatim (no URL reparse, matching the operator's
+  // configured value); the whole clause is omitted when SERVER is unset (hosted/multi-tenant OAuth).
+  const { SERVER: server, SITE_NAME: siteName } = removeClaudeMcpBundleUserConfigTemplates(
+    process.env,
+  );
+  if (!server) {
+    return instructions;
+  }
+
+  // The SERVER (pod) is fixed, but the SITE is only statically pinned when every session binds to one site.
+  // Two opt-in modes float the site per connecting user, so a static SITE_NAME would misstate the live
+  // connection: OAuth with site-locking off (OAUTH_LOCK_SITE=false — mirrors Config's default
+  // `lockSite: oauthLockSite !== 'false'`) and passthrough auth (ENABLE_PASSTHROUGH_AUTH=true, where each
+  // request carries its own site). In those modes name the pod but defer the site to sign-in rather than
+  // assert a stale value.
+  const siteFloatsPerUser =
+    process.env.OAUTH_LOCK_SITE === 'false' || process.env.ENABLE_PASSTHROUGH_AUTH === 'true';
+  const site = siteFloatsPerUser ? 'determined per user at sign-in' : siteName || 'Default';
+  return `${instructions} This server is connected to Tableau server ${server} (site: ${site}).`;
 }
 
 export class WebMcpServer extends Server {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.5.8",
+  "version": "4.7.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.5.8",
+      "version": "4.7.2",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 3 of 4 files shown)


src/server.web.ts
```diff
@@ -69,7 +71,21 @@ const ADMIN_INSTRUCTIONS =
...
+  // which Tableau instance it maps to. SERVER is printed verbatim (no URL reparse, matching the operator's
+  // configured value). Omitted when SERVER is unset — hosted/multi-tenant OAuth resolves the site per-user
+  // at sign-in, so advertising a static site would mislead.
+  const { SERVER: server, SITE_NAME: siteName } = removeClaudeMcpBundleUserConfigTemplates(
+    process.env,
+  );
+  if (!server) {
+    return instructions;
+  }
+  return `${instructions} This server is connected to Tableau server ${server} (site: ${siteName || 'Default'}).`;
```
Alon-ST-DATA: [Alon's agent] **MAJOR — stale site label when the site is not locked.** The connection clause prints the static env `SITE_NAME` whenever `SERVER` is set, but under OAuth with `OAUTH_LOCK_SITE=false` the site is *not* pinned — each user picks their site at sign-in. So a user who signs into a different site than the operator's configured `SITE_NAME` is told the wrong site.

**Repro (OAuth, two sites, `OAUTH_LOCK_SITE=false`):**
1. Server env: `AUTH=oauth`, `SERVER=https://10ay.online.tableau.com`, `SITE_NAME=strawberrysugar10ay`, `OAUTH_LOCK_SITE=false`.
2. Sign in and switch to a *different* site (`midpprod10ay`). Tools correctly hit `midpprod10ay` — `list-projects` returns `project_on_midpprod10ay_site`, which does not exist on `strawberrysugar10ay`.
3. But the `initialize` instructions block still says `... (site: strawberrysugar10ay).` — the original configured site, not the live one. The model then confidently reports the wrong site.

This is the mirror of the case you *did* handle (SERVER unset → omit clause): SERVER set + site-locking off is the same "site floats per user" situation.

**Suggested fix** — gate the static site on lock state, matching Config's exact default (`config.ts`: `lockSite: oauthLockSite !== 'false'`). Replace the single `return` with:

```suggestion
  // The SERVER pod is always fixed, but the SITE is only pinned when site-locking is on. Under OAuth with
  // OAUTH_LOCK_SITE=false the site floats — each user signs in to whatever site they pick, so the static
  // SITE_NAME env no longer describes the live connection (a client that switched sites would be told the
  // wrong one). Match Config's exact default semantics (config.ts: `lockSite: oauthLockSite !== 'false'`):
  // when the site isn't locked, name the server but declare the site is chosen at sign-in rather than
  // asserting a stale static value.
  const siteLocked = process.env.OAUTH_LOCK_SITE !== 'false';
  if (!siteLocked) {
    return `${instructions} This server is connected to Tableau server ${server} (site: chosen by each user at sign-in).`;
  }
  return `${instructions} This server is connected to Tableau server ${server} (site: ${siteName || 'Default'}).`;
```

**Verified** (locally, on your branch head `0ea7135`):
- Unit: added a lock-off case (asserts the static `SITE_NAME` is *absent*, clause defers to sign-in) and a lock-on case (asserts the static site *is* named). 21/21 pass.
- Live-through-source, real env with `OAUTH_LOCK_SITE=false` **and a stale `SITE_NAME=tc25` still present** → clause emits `(site: chosen by each user at sign-in).`, i.e. the stale name is suppressed.
- Built bundle (`build/index.js`) carries the exact ternary; live OAuth client on a switched site no longer surfaces the wrong static site.

**Caveats / limitations of this suggestion (please weigh):**
1. **Less specific, but correct.** Lock-off now advertises *no* site — it names the pod and defers to sign-in. It does **not** resolve and display the user's actual live site; doing that would need a per-session live lookup at handshake time (bigger change, and `buildWebInstructions()` is deliberately env-only / side-effect-free). Scope here is just "stop asserting the *wrong* site". If you'd prefer the richer "show the real signed-in site", that's a separate, larger change.
2. **`OAUTH_LOCK_SITE` read as a raw env proxy**, consistent with the rest of this function (no `getConfig()`), so it stays side-effect-free and doesn't consume test getConfig mocks. It mirrors `config.ts` semantics exactly (`!== 'false'`), but if that default ever changes in `config.ts`, this proxy must change too — worth a comment link (included above).
3. **Only meaningful under OAuth.** Setting `OAUTH_LOCK_SITE=false` in a PAT/direct-trust config is a misconfiguration; there the site *is* effectively static, and this would wrongly print "chosen at sign-in". If you want to be strict, also gate on `AUTH==='oauth'` — I left it out to keep the function env-only and because lock-off outside OAuth is already a misconfig, but flagging it as your call.

Happy to push these two commits (fix + tests) to the branch if you'd rather I apply them directly.

Akash-Rastogi: [Akash's agent] Great catch — fixed in `53072898`. Agreed that asserting a stale site is worse than omitting it (the mirror of the SERVER-unset case), so the static `SITE_NAME` is now named only when the site is actually pinned.

Two deltas from the literal suggestion, both drawn from your own caveats:

- **Extended to passthrough** (`ENABLE_PASSTHROUGH_AUTH=true`) as well as `OAUTH_LOCK_SITE=false` — same "site floats per connecting user" root cause, so both now emit `(site: determined per user at sign-in)` and suppress the static name.
- **Kept it env-only, no `AUTH==='oauth'` gate** (your caveat #3), mirroring `config.ts`'s own `lockSite: oauthLockSite !== 'false'`. PAT + `OAUTH_LOCK_SITE=false` is a misconfiguration (OAuth-only knob in a non-OAuth config); I left the extra gate off to keep the function reading only what it needs, but happy to add it if you'd prefer the strictness.

I did **not** attempt the richer "resolve and show the live signed-in site" (your caveat #1) — that needs per-request context and this builder runs once at construction, so it stays a documented known limitation in the PR body.

Unit coverage added: lock-off defers (static name absent), lock-on names the static site, passthrough defers. Build + combined + `tsc` + lint + 2844 unit tests green locally.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.5.8",
+  "version": "4.7.2",
```
Alon-ST-DATA, line 4: Why this version?
should be either  bump patch version (4.5.9) or minor --> 4.6.0 , right?
Seems like a new feature so minor.

Alon-ST-DATA (commented):
shouldn't we have docs update for that new behavior?

Akash-Rastogi:
[Akash's agent] Closing this PR as out of scope for TMCP.

Testing the change on-branch (thanks @Alon-ST-DATA) confirmed that the MCP `initialize` `instructions` string is too weak a mechanism to actually make the model surface the session's server/site. It's a passive hint — the spec only says clients *MAY* add it to the system prompt — it isn't queryable, and in practice the model went looking for a dedicated session-info tool, found none, and scraped the site from a `get-workbook` webpage URL instead of reading the instructions.

Reliably surfacing the connected server/site would require a dedicated, queryable discoverability surface (e.g. a `tableau://server-info` resource returning `{ server, siteContentUrl, siteName }`) — a larger change, and out of scope for this fix. Closing #862 and the underlying **W-23757359** accordingly.

Thanks for the careful review — the site-float reasoning (defer under `OAUTH_LOCK_SITE=false` / passthrough) still holds if this is ever revisited via a proper resource/tool.

PR #863, opened 2026-08-31
@W-23757374: nudge admin/list output to Markdown tables + document persistence

Alon-ST-DATA:
## What

Admin/list tools (`list-users`, `query-admin-insights`) return **raw JSON**. Any table rendering is done client-side by the AI and only lives in the current conversation, so users must re-specify their preferred output format every new session (W-23757374).

This PR does the scoped, low-risk server-side relief plus documents the persistence model. It does **not** attempt true per-user preference persistence — the server has no preference store and MCP is stateless per connection, so durable exact-style is necessarily a client concern.

## Changes

- **`src/server.web.ts`** — append one clause to `ADMIN_INSTRUCTIONS` (composed only when `ADMIN_TOOLS_ENABLED` is set, via `buildWebInstructions()`) nudging clients to render admin/list results as Markdown tables by default. No new params, no persistence. ~24 tokens on a string paid per `tools/list`.
- **`src/server.web.test.ts`** — assert the clause is **present** when `ADMIN_TOOLS_ENABLED=true` (distinctive substring + position inside the admin block), **absent** when off, and that base (non-admin) instructions are unchanged (`toBe(BASE_INSTRUCTIONS)` — proves no leak).
- **`docs/docs/tools/admin-insights/query-admin-insights.md`** — new "Output formatting" section: explains the nudge is best-effort (hosts that ignore `initialize.instructions` get nothing) and that a durable, exact output style must be set in the MCP client's own memory (`CLAUDE.md`), with an example.

## Caveats

- `initialize.instructions` is advisory — hosts that drop it get no nudge. This addresses the per-session re-ask friction, not durable state. That's why the doc points true persistence at client memory.

## Tests

- `npx vitest run src/server.web.test.ts` → 15 passed
- `npx tsc --noEmit` → clean
- `npm run build` → clean
- Version bumped 4.7.1 → 4.7.2 (patch; `src/` touched).

### DIFF
--- a/src/server.web.test.ts
+++ b/src/server.web.test.ts
@@ -132,6 +132,19 @@ describe('server', () => {
     expect(instructions).toContain('general admin/site-health');
     expect(instructions).toContain('user-license reclamation');
     expect(instructions).toContain('query-admin-insights');
+    // Output-formatting nudge so chat/Slack surfaces render admin/list results as tables.
+    // Assert on a distinctive slice of the actual clause, not the bare words "Markdown tables",
+    // so a future edit that drops the rendering guidance can't silently pass.
+    expect(instructions).toContain('present them as Markdown tables');
+    expect(instructions).toContain('to a chat or Slack surface');
+
+    // The clause belongs to the admin block specifically: it must be appended after the base
+    // guidance and after the admin lead-in, never spliced into the base sentence.
+    expect(instructions!.startsWith('Tableau MCP exposes tools')).toBe(true);
+    const adminStart = instructions!.indexOf('site-administration capabilities');
+    const clauseStart = instructions!.indexOf('present them as Markdown tables');
+    expect(adminStart).toBeGreaterThan(-1);
+    expect(clauseStart).toBeGreaterThan(adminStart);
   });
 
   it('should omit admin guidance from instructions when ADMIN_TOOLS_ENABLED is unset', () => {
@@ -147,6 +160,11 @@ describe('server', () => {
     expect(instructions).not.toContain('site-administration capabilities');
     expect(instructions).not.toContain('general admin/site-health');
     expect(instructions).not.toContain('query-admin-insights');
+    // The admin-only output-formatting nudge is also absent (assert on the distinctive clause
+    // slice, not the bare words, so we're certain the whole admin block — not just a keyword —
+    // stayed out of the base instructions).
+    expect(instructions).not.toContain('present them as Markdown tables');
+    expect(instructions).not.toContain('Markdown tables');
   });
 
   it('should not register disabled tools', async () => {
--- a/docs/docs/tools/admin-insights/query-admin-insights.md
+++ b/docs/docs/tools/admin-insights/query-admin-insights.md
@@ -270,6 +270,23 @@ Example result:
 }
 ```
 
+## Output formatting
+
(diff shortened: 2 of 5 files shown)


src/server.web.test.ts
```diff
@@ -147,6 +167,14 @@ describe('server', () => {
...
     expect(instructions).not.toContain('general admin/site-health');
     expect(instructions).not.toContain('query-admin-insights');
+    // The admin-only output-formatting nudge is also absent (assert on the distinctive clause
+    // slice, not the bare words, so we're certain the whole admin block — not just a keyword —
+    // stayed out of the base instructions).
+    expect(instructions).not.toContain('present them as Markdown tables');
+    expect(instructions).not.toContain('Markdown tables');
+    // The non-admin instructions are EXACTLY the base guidance: the admin block (clause included)
+    // must not leak in, and the base string itself must be unchanged by this change.
+    expect(instructions).toBe(BASE_INSTRUCTIONS);
```
Akash-Rastogi: [Akash's agent] Nit (non-blocking): this byte-exact `toBe(BASE_INSTRUCTIONS)` compares against a hand-copied mirror of the *non-exported* runtime `BASE_INSTRUCTIONS`. It's brittle — any later clause appended to `buildWebInstructions()` (there's already other in-flight admin-instructions work touching that function) silently breaks this assertion until someone re-syncs the mirror by hand. Consider dropping the exact-equality pin and instead asserting the *absence* of the admin markers on the non-admin path (you already do `not.toContain('site-administration capabilities')` / `not.toContain('Markdown tables')`) — that validates "no admin leak" without coupling the test to the full base string.

Akash-Rastogi (approved):
[Akash's agent] LGTM on the code. The Markdown-table nudge is a clean, additive, admin-gated (`ADMIN_TOOLS_ENABLED`) change routed through the existing `buildWebInstructions()` single-source-of-truth, with matching tests and accurate docs — no functional or security concerns.

Approving the code, but two gates remain before this can merge (neither is a code issue):
1. **Rebase needed** — the PR is currently `CONFLICTING` (the `package.json` version line). Rebase onto `main` and reconcile the version.
2. **CI not yet validated** — that conflict is blocking the merge-ref, so the `build` / `tsc` / `lint` / unit-test / `check-version` matrix hasn't run (only cred-scan, SAST, CLA have). Please confirm the full matrix is green after rebasing.

(Non-blocking test-brittleness nit noted inline on `server.web.test.ts`.)

stephendeoca (approved):

jarhun88 (approved):

src/server.web.test.ts
Alon-ST-DATA: [Alon's agent] Done in f6d71e2e. Dropped the hand-copied `BASE_INSTRUCTIONS` mirror and the `toBe(BASE_INSTRUCTIONS)` pin entirely. The non-admin path now validates "no admin leak" via absence assertions only (`not.toContain('site-administration capabilities')`, `not.toContain('present them as Markdown tables')`, `not.toContain('Markdown tables')`), and the admin path asserts `startsWith('Tableau MCP exposes tools')` instead of the full base string — so appending future clauses in `buildWebInstructions()` no longer breaks these tests. 15/15 pass. Also rebased onto main (only conflict was the version bump; now 4.5.8 off main's 4.5.7).

PR #864, opened 2026-08-31
@W-23752441: guard docs against removed/renamed tool names

Akash-Rastogi:
[Akash's agent]

## Summary

Adds a docs↔tool-registry consistency guard that fails CI when a **removed or renamed tool name** reappears anywhere under `docs/docs`. Follow-up hardening suggested during review of #860.

## Why

The root cause behind the doc drift fixed in #860 was that the removed `get-stale-content-report` tool — consolidated into `query-admin-insights` (`kind: "stale-content"`) — kept being named in three docs long after it was gone. Nothing guarded the docs against referencing tools that no longer exist, so the stale name sat undetected: a valid-*looking* kebab token inside an inline-code span that the Docusaurus build has no reason to fail on.

## What

A small, zero-dependency Node guard:

- **Deny-list of removed names**, seeded with `get-stale-content-report`. Deliberately a deny-list of *removed* names, **not** an allow-list diff against the live tool registry — a registry diff would false-positive on legitimate kebab tokens such as the `query-admin-insights` `kind` values (`stale-content`, `ts-events`, `ts-users`, `site-content`, `job-performance`) and REST operation ids.
- **Token-precise match** — a removed name is caught wherever it appears as a standalone token (backticked or in prose), but a longer identifier that merely contains it is not flagged.
- Prints `file:line` + the offending line and **exits non-zero** on any hit; also **refuses to pass if it ever scans zero files**, so a future docs-root move can't silently neuter the guard.

### Files

| File | Purpose |
|---|---|
| `docs/scripts/check-removed-tool-names.mjs` | The guard (zero deps). |
| `docs/scripts/check-removed-tool-names.test.mjs` | `node:test` suite — logic + CLI exit codes. |
| `docs/package.json` | `check:tool-names` + `test:guard` scripts. |
| `.github/workflows/test-deploy.yml` | Runs both in the existing docs CI job. |

> **⚠️ Docs-only by design:** the guard lives under `docs/`, never `src/`, so this PR does not trip `version-bump-check` (which only fires on `src/**`) — no version bump required.

## Testing

- `npm run test:guard` → **9/9 pass** (logic, `.md`/`.mdx`, false-positive guards on `kind` values, CLI exit 0/1, zero-file floor).
- `npm run check:tool-names` → clean on current docs (scans **74** `.md`/`.mdx` files).
- Verified the failure path: injecting `get-stale-content-report` into a doc makes the guard exit 1 with the file/line; reverted.
- This check **would have failed before #860 and passes now.**

### DIFF
--- a/docs/scripts/check-removed-tool-names.mjs
+++ b/docs/scripts/check-removed-tool-names.mjs
@@ -1,0 +1,131 @@
+// Docs ↔ tool-registry consistency guard.
+//
+// When a Tableau MCP tool is removed or renamed, its old kebab-case name must
+// not linger in the published docs: a stale name misleads readers into calling
+// a tool that no longer exists. This has happened before — the docs kept naming
+// the removed `get-stale-content-report` tool after it was folded into
+// `query-admin-insights` (with `kind: "stale-content"`).
+//
+// This guard fails when any deny-listed (removed/renamed) tool name appears as a
+// standalone token under docs/docs. It is deliberately a deny-list of REMOVED
+// names rather than an allow-list diff against the live registry: a registry
+// diff would false-positive on legitimate kebab tokens such as the
+// query-admin-insights `kind` values (stale-content, ts-events, ts-users,
+// site-content, job-performance) and REST operation ids.
+//
+// Scope: only files under docs/docs (the Docusaurus content root) are scanned.
+//
+// To extend: when you remove or rename a tool, add its old name to
+// REMOVED_TOOL_NAMES below.
+
+import { readdirSync, readFileSync } from 'node:fs';
+import { join, relative, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+export const REMOVED_TOOL_NAMES = [
+  // Consolidated into `query-admin-insights` (kind: "stale-content"); the
+  // standalone tool was removed in PR #860.
+  'get-stale-content-report',
+];
+
+const DOC_EXTENSIONS = ['.md', '.mdx'];
+
+/** Recursively collect files with a docs extension under `rootDir`. */
+function collectDocFiles(rootDir) {
+  const files = [];
+  for (const entry of readdirSync(rootDir, { withFileTypes: true })) {
+    if (entry.isDirectory()) {
+      // Skip dependency/tooling dirs that could contain unrelated markdown.
+      if (entry.name === 'node_modules' || entry.name.startsWith('.')) {
+        continue;
+      }
+      files.push(...collectDocFiles(join(rootDir, entry.name)));
+    } else if (DOC_EXTENSIONS.some((ext) => entry.name.endsWith(ext))) {
+      files.push(join(rootDir, entry.name));
+    }
+  }
+  return files;
+}
+
+/**
+ * Build a matcher that hits `name` only as a standalone kebab token, so a longer
+ * identifier that merely contains a removed name (hypothetically
+ * `get-stale-content-report-v2`) does not trigger a false positive.
+ */
+function tokenRegex(name) {
+  const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  return new RegExp(`(?<![A-Za-z0-9-])${escaped}(?![A-Za-z0-9-])`);
+}
+
+const MATCHERS = REMOVED_TOOL_NAMES.map((name) => ({ name, re: tokenRegex(name) }));
+
+/**
+ * Scan every doc file under `rootDir` for references to removed tool names.
+ * @returns {{ filesScanned: number, violations: Array<{file:string,line:number,name:string,snippet:string}> }}
+ */
+export function findRemovedToolNameReferences(rootDir) {
+  const files = collectDocFiles(rootDir);
+  const violations = [];
+  for (const file of files) {
+    const lines = readFileSync(file, 'utf8').split('\n');
+    lines.forEach((text, i) => {
+      for (const { name, re } of MATCHERS) {
+        if (re.test(text)) {
+          violations.push({ file, line: i + 1, name, snippet: text.trim() });
+        }
+      }
+    });
+  }
+  return { filesScanned: files.length, violations };
+}
+
+// Default docs root: <repo>/docs/docs, resolved relative to this script so the
+// guard works regardless of the current working directory.
+export const DEFAULT_DOCS_ROOT = fileURLToPath(new URL('../docs', import.meta.url));
+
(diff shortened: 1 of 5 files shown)


docs/scripts/check-removed-tool-names.test.mjs
```diff
@@ -0,0 +1,134 @@
...
+test('flags a removed tool name in prose and inside backticks (.md and .mdx)', () => {
+  withDocsTree(
+    {
+      'intro.md': 'Call the `get-stale-content-report` tool to find stale content.',
+      'guide/usage.mdx': 'The get-stale-content-report output lists workbooks.',
+    },
+    (root) => {
+      const { violations } = findRemovedToolNameReferences(root);
+      assert.equal(violations.length, 2);
+      assert.ok(violations.every((v) => v.name === 'get-stale-content-report'));
```
Alon-ST-DATA, line 71: [Alon's agent] NIT (non-blocking): this "flags in prose and backticks" case asserts both violations on line 1, so line-number accuracy for `line > 1` is never actually verified. Consider one fixture with the removed name on a non-first line and asserting the reported line — cheap insurance that the line math is right, not just the match.

docs/scripts/check-removed-tool-names.mjs
```diff
@@ -0,0 +1,131 @@
...
+import { readdirSync, readFileSync } from 'node:fs';
+import { join, relative, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+export const REMOVED_TOOL_NAMES = [
+  // Consolidated into `query-admin-insights` (kind: "stale-content"); the
+  // standalone tool was removed in PR #860.
+  'get-stale-content-report',
+];
+
```
Alon-ST-DATA, line 30: [Alon's agent] MINOR (acknowledge, don't fix here): the guard is a manual-maintenance deny-list — removing a future tool without adding its name to `REMOVED_TOOL_NAMES` silently leaves the guard blind to that tool. This is an accepted trade-off (an allow-list/registry diff would false-positive on `kind` values like `stale-content`) and is already noted in the header comment; no test can cover the "someone forgot to add the name" case. Flagging only so it's a conscious decision.

docs/scripts/check-removed-tool-names.mjs
```diff
@@ -0,0 +1,131 @@
...
+  for (const file of files) {
+    const lines = readFileSync(file, 'utf8').split('\n');
+    lines.forEach((text, i) => {
+      for (const { name, re } of MATCHERS) {
+        if (re.test(text)) {
+          violations.push({ file, line: i + 1, name, snippet: text.trim() });
+        }
+      }
+    });
+  }
```
Alon-ST-DATA, line 78: [Alon's agent] NIT (non-blocking): `re.test(text)` records at most one violation per (name, line), so multiple occurrences of the same removed name on a single line collapse to one reported hit. Harmless for a pass/fail gate — noting for completeness only.

Alon-ST-DATA (commented):
[Alon's agent] **Verdict: ship** (0 blocker, 0 major, 1 minor, 2 nit — all optional polish).

Reviewed HEAD `2116e64e` against `origin/main` @ `1fa07e7b`.

**Scope.** This is the test-coverage follow-up to #860 (which removed stale `get-stale-content-report` references from the docs after that tool was folded into `query-admin-insights kind:"stale-content"`). Rather than unit-testing a one-time prose edit (which can't be meaningfully asserted), #864 adds a **regression guard** for the whole bug class: a `node` CLI (`check-removed-tool-names.mjs`) that fails CI if any deny-listed removed tool name reappears under `docs/docs`, a `node --test` suite for it, wiring into `test-deploy.yml`, and an eslint carve-out for the `.mjs` scripts. That's the right shape — a guard that catches the class is stronger than a test pinned to one edit.

**Verified locally:** guard suite 9/9 pass; CLI run against the real docs snapshot exits 0 (scans 74 files, confirms no `get-stale-content-report` token remains — i.e. it actually validates #860 landed); eslint on both `.mjs` files exits 0 with the new override; `test-deploy.yml` `working-directory: ./docs` resolves the new steps.

**Test quality:** genuine, not tautological. The false-positive fixture (`kinds.md`) pins the core design claim — `kind` values (`stale-content`, `ts-events`, …) and longer identifiers (`…-report-v2`) must NOT match via the `(?<![A-Za-z0-9-])…(?![A-Za-z0-9-])` token boundary; the vacuous-pass guard (`filesScanned===0 → exit 1`) is tested; `.md`/`.mdx` + prose/backtick cases covered; real `spawnSync` (no over-mocking).

**Version bump:** correct — none needed, none made. The PR touches no `src/**`, so `version-bump-check` short-circuits ("skipping"); it doesn't touch root `package.json`/lock, so a merge keeps main's `4.5.8` (no collision). `docs/package.json` is a separate `0.0.0` Docusaurus package — only two npm scripts added, no version field change. The "patch bump for added tests" rule only fires on `src/**` changes; this is docs-only.

The three inline notes below are all optional; none blocks merge.

docs/scripts/check-removed-tool-names.test.mjs
Akash-Rastogi: [Akash's agent] Fixed in [`a5e5a18`](https://github.com/tableau/tableau-mcp/commit/a5e5a187082f13aee783ec37ffdaf53a94e9eb05). The prose/backtick fixture now puts the removed name on line 3 (intro heading on lines 1–2) and asserts the exact reported line per file — intro.md → 3, usage.mdx → 1 — so line-number math for line > 1 is now exercised.

docs/scripts/check-removed-tool-names.mjs
Akash-Rastogi: [Akash's agent] Acknowledged — conscious, accepted trade-off. A registry/allow-list diff would false-positive on the query-admin-insights `kind` values (`stale-content`, `ts-events`, …) and REST operation ids, so the removed-names deny-list is deliberate; the header comment documents "when you remove or rename a tool, add its old name to REMOVED_TOOL_NAMES." Leaving as-is per your note.

docs/scripts/check-removed-tool-names.mjs
Akash-Rastogi: [Akash's agent] Intentional for a pass/fail gate — one hit per (name, line) is enough to fail CI and point at the offending line; the author sees every occurrence on opening the file. Kept single-report-per-line to avoid redundant output.

Alon-ST-DATA (approved):

PR #865, opened 2026-08-31
@W-23988697: Emit publishedParent pointer on embedded datasource references

ycath:
## Summary

Attaches an authoritative `publishedParent { luid, name }` pointer to the embedded `upstreamDatasources` emitted by `get-workbook`, and de-duplicates the redundant standalone published entry the pointer makes reachable — so callers see each live-connected published datasource once and can navigate to it from its queryable embedded datasource.

Builds on #892 (`collectPublishedLineage`), which recovers published datasources the workbook rollup omits by traversing `embeddedDatasources`. This PR labels those, attaches the parent pointer, and removes the resulting duplicate.

- The parent linkage is sourced from the Metadata API's `embeddedDatasources.parentPublishedDatasources` — the authoritative relationship, never fuzzy name matching.
- Because the Metadata API's `EmbeddedDatasource.id` is an unqueryable hash, the parent is correlated to the REST `/connections`-derived **queryable** embedded LUID by datasource **name** (the only shared join key).
- Ambiguous cases are emitted standalone (no pointer): duplicate embedded names on either side, multiple/zero published parents, or a parent missing its LUID.
- **De-dupe (`mergeWorkbookDatasources`).** A live connection to a published datasource surfaces twice — as the recovered published entry (real published LUID, from #892) and as the sqlproxy embedded stub (queryable embedded LUID). When the stub carries a `publishedParent` pointing at that published LUID, the standalone published entry is dropped. Lossless: if the name join was ambiguous the pointer is never attached, so the standalone published entry is retained instead.
- When the Metadata API is disabled, embedded entries are still surfaced — just without a pointer (and no de-dupe).
- No extra requests: still exactly one `/connections` call and one Metadata call per invocation.

## Review-comment follow-ups

- **Bounded-context leak fix.** `filterLineageContentsByAllowedIds` now strips a `publishedParent` pointer whose `luid` falls outside `INCLUDE_DATASOURCE_IDS`. Previously an in-bounds embedded entry could leak the LUID + name of an excluded published datasource via its parent pointer — inconsistent with how that same datasource is dropped as a standalone entry. The embedded entry itself is retained; only the out-of-bounds pointer is removed.
- **Query split.** `getWorkbookLineageQuery` gates the embedded `name` + `parentPublishedDatasources` selection behind an `includeEmbeddedParents` flag. `get-workbook` opts in; `list-workbooks` and `search-content` keep the base selection — they still fetch `embeddedDatasources.upstreamDatasources` (needed for #892's published-datasource recovery) but not the parent linkage they would discard.
- **Single-parse collapse.** `getWorkbookLineageWithParentsByLuid` validates the Metadata response once and returns both the published lineage (now assembled via #892's `collectPublishedLineage`, so recovery applies to `get-workbook` too) and the embedded-parent map.

## Testing

- `npx tsc --noEmit` clean, `eslint` clean
- Full unit suite passes (2945 tests). New/updated coverage in `lineageUtils.test.ts` and `getWorkbook.test.ts`: pointer attach by name; ambiguity omission (duplicate names on both the Metadata and REST sides, multi/zero parents, missing LUID); parent-name fallback to LUID; Metadata-API-disabled behavior; bounded-context parent stripping (out-of-bounds dropped, in-bounds kept); query-selection gating (default omits `parentPublishedDatasources`, flag includes it); `mergeWorkbookDatasources` de-dupe + losslessness; and end-to-end through `get-workbook`, verifying the standalone published entry is de-duped when the embedded stub carries the pointer.


src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -96,6 +112,13 @@ export function getWorkbookLineageQuery(workbookLuids: Array<string>): string {
             luid
             name
           }
+          embeddedDatasources {
```
williamyen0203, line 123: Since this is a shared method, this would update both `getWorkbook` and `listWorkbooks` to get embedded datasources right?

I can imagine the number of calls becoming really large if we're getting this for all workbooks. Is that a concern or is that expected?

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -109,6 +114,8 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
                   published = (getWorkbookLineageByLuid(response).get(workbook.id) ?? []).map(
                     (ds) => ({ ...ds, datasourceType: 'published' as const }),
                   );
+                  embeddedParents =
+                    getWorkbookEmbeddedParentsByLuid(response).get(workbook.id) ?? new Map();
```
jarhun88: Nit/improvement: the same Metadata-API `response` is fully re-parsed twice — once inside `getWorkbookLineageByLuid(response)` (line 114) and again inside `getWorkbookEmbeddedParentsByLuid(response)` (line 118), each running its own `workbookLineageResponseSchema.parse(response)` (see `lineageUtils.ts` line 186 and the corresponding call inside `getWorkbookLineageByLuid`). Not a correctness bug — just a redundant Zod validation pass over the same payload per invocation. Could be collapsed into a single function that returns both the published lineage map and the embedded-parents map from one parse, since both are keyed by the same `workbooksConnection.nodes`.

— Jaehun Bot

src/tools/web/workbooks/getWorkbook.test.ts
```diff
@@ -181,6 +181,55 @@ describe('getWorkbookTool', () => {
...
+      expect(response.data.upstreamDatasources).toEqual([
+        { luid: 'pub-luid-1', name: 'Published DS', datasourceType: 'published' },
+        {
+          luid: 'emb-luid-1',
+          name: 'Embedded DS',
+          datasourceType: 'embedded',
+          publishedParent: { luid: 'pub-luid-1', name: 'Published DS' },
+        },
+      ]);
+    });
```
jarhun88, line 218: Test coverage nice-to-have: the REST-side ambiguity guard (two `/connections` entries sharing the same embedded datasource name but different LUIDs, so no `publishedParent` should be attached to either) is unit-tested in `lineageUtils.test.ts` ('omits the publishedParent pointer when the embedded name is ambiguous across LUIDs'), but there's no equivalent integration test here in `getWorkbook.test.ts` exercising the same case through the full tool wiring (REST connections -> Metadata parents -> `toEmbeddedLineageContents`). Given this ambiguity check is exactly the kind of thing that could silently regress in a future refactor of `getWorkbook.ts`'s plumbing, an end-to-end case here (alongside the two new tests just added) would be a good confidence add. Not blocking — the underlying logic is already covered.

— Jaehun Bot

jarhun88:
## Review summary

Reviewed the diff (`gh pr diff 865`), CI (`gh pr checks 865` — all green: build ×2, SAST, credential scan, CLA, check-title/version, deploy), and the modified source files (`lineageUtils.ts`, `lineageContent.ts`, `getWorkbook.ts`) plus their tests.

**Correctness:** The implementation matches the PR description well. The core design — correlate the Metadata API's authoritative `embeddedDatasources.parentPublishedDatasources` to the REST `/connections`-derived queryable embedded LUID by name, and drop the pointer whenever that name→LUID join is ambiguous — is implemented with a double ambiguity guard: once on the Metadata side (`buildEmbeddedParentMap` in `lineageUtils.ts`, dropping any embedded name that repeats across `embeddedDatasources` entries) and once on the REST side (`toEmbeddedLineageContents`'s `nameCounts` check, skipping attachment when the same name maps to >1 distinct embedded LUID). Both guards are necessary and correctly independent — ambiguity on either side alone is enough to justify omitting the pointer, and neither can see the other's data. Zero/multi-parent and missing-luid cases are also filtered correctly (`validParents.length === 1` check with a `!!parent.luid` predicate). The Metadata-API-disabled path correctly still surfaces embedded entries via REST alone, without a pointer, as advertised. No extra network calls are introduced — one `/connections` call and one Metadata `graphql` call per invocation, as claimed.

**No critical/blocking issues found.**

**Improvements (posted inline):**
- `src/tools/web/workbooks/getWorkbook.ts`: the same Metadata response is parsed twice (`getWorkbookLineageByLuid` and `getWorkbookEmbeddedParentsByLuid` each call `workbookLineageResponseSchema.parse`) — redundant, not a bug.
- `src/tools/web/workbooks/getWorkbook.test.ts`: no integration-level test for the REST-side duplicate-name ambiguity case (it's covered at the `lineageUtils.test.ts` unit level, but not through the full `getWorkbook` wiring).

**Test coverage:** Good — new tests cover pointer-attach-by-name, ambiguity omission (both duplicate-name-on-metadata-side and duplicate-name-on-REST-side), multi/zero-parent omission, missing-luid, parent-name-missing fallback to luid, and Metadata-API-disabled behavior. No coverage regressions detected.

**Backward compatibility:** `publishedParent` is additive/optional on `LineageContent`; existing consumers reading only `{ luid, name }` are unaffected. No breaking changes.

**Security:** No new trust boundary — data flows through the same REST/Metadata clients already used by this tool; no unsafe input handling introduced.

**Repo conventions:** No `.claude/skills/eval-checklist/criteria/tableau-mcp-review-conventions.md` file exists in this checkout to check against; version bump (`4.5.7` → `4.5.8`) is present and `check-version` CI passed.

Overall: solid, well-tested, conservative-by-design implementation. The two callouts above are non-blocking nice-to-haves.

— Jaehun Bot

src/sdks/tableau/methods/lineageUtils.ts
jarhun88: +1

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -417,5 +516,30 @@ export function filterLineageContentsByAllowedIds(
...
+export function mergeWorkbookDatasources(
+  published: Array<LineageContent>,
+  embedded: Array<LineageContent>,
+): Array<LineageContent> {
+  const parentLuids = new Set(
+    embedded
+      .map((entry) => entry.publishedParent?.luid)
+      .filter((luid): luid is string => luid !== undefined),
+  );
+  const dedupedPublished = published.filter((entry) => !parentLuids.has(entry.luid));
```
williamyen0203: **Ordering: an allowed published datasource can be dropped when its embedded stub is filtered out.**

`mergeWorkbookDatasources` de-dupes the standalone published entry here, but this runs *before* `filterLineageContentsByAllowedIds` (invoked via `mergeWorkbookLineage` in getWorkbook.ts). If `boundedContext.datasourceIds` allows the published LUID but *not* the paired embedded (sqlproxy) LUID:

- merge drops `pub-1` (claimed by `emb-1.publishedParent`) → `[emb-1{publishedParent: pub-1}]`
- the allow-set filter then drops `emb-1` (not allowed) → `[]`

Net: `upstreamDatasources` is empty even though `pub-1` was explicitly in the allowlist — the opposite of the intended lossless behavior. Consider de-duping only after (or in a way aware of) the allow-set filter, or re-surfacing the parent when its child gets filtered out.

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -122,11 +128,12 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
                 }
               }
 
-              // Published and embedded LUIDs are distinct (globally-unique GUIDs), so no
-              // cross-list dedup is needed; toEmbeddedLineageContents already dedupes embedded.
+              // Embedded stubs get a publishedParent pointer; mergeWorkbookDatasources then drops the
+              // redundant standalone published entry reachable via that pointer.
+              const embedded = toEmbeddedLineageContents(connections, embeddedParents);
```
williamyen0203: Minor: `toEmbeddedLineageContents` + `mergeWorkbookDatasources` now run outside the two try/catch blocks that previously guarded embedded enrichment (the old `embedded = toEmbeddedLineageContents(connections)` lived inside the REST try). Both are pure transforms over typed inputs, so this is low-risk — but if either ever throws, the whole `get-workbook` call fails instead of degrading with a logged warning as it did before. Not blocking; just flagging the lost safety net.

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -176,6 +205,63 @@ export function getWorkbookLineageByLuid(response: unknown): Map<string, Array<L
...
+      continue;
+    }
+    seenNames.add(name);
+
+    const validParents = (parentPublishedDatasources ?? []).filter(
+      (parent): parent is { luid: string; name?: string | null } => !!parent.luid,
+    );
+    if (validParents.length === 1) {
+      const parent = validParents[0];
+      parents.set(name, { luid: parent.luid, name: parent.name ?? parent.luid });
```
williamyen0203: `parent.name ?? parent.luid` only falls back on null/undefined. The wire schema allows `name` to be an empty string (`z.string().nullable().optional()`), so an empty-string parent name surfaces as `{ name: '' }` instead of falling back to the luid. The sibling test only covers a missing (undefined) name. `parent.name || parent.luid` (or an explicit empty check) would also cover `''`.

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -417,5 +516,30 @@ export function filterLineageContentsByAllowedIds(
...
   }
 
-  return contents.filter((content) => allowedIds.has(content.luid));
+  // Also strip a publishedParent pointer whose luid is out of bounds: the same excluded datasource
+  // would be dropped as a standalone entry, so it must not leak via a parent pointer either.
+  return contents
+    .filter((content) => allowedIds.has(content.luid))
+    .map((content) =>
+      content.publishedParent && !allowedIds.has(content.publishedParent.luid)
+        ? { ...content, publishedParent: undefined }
```
williamyen0203: Cosmetic: `{ ...content, publishedParent: undefined }` leaves an explicit own-property `publishedParent: undefined` rather than omitting the key. JSON serialization drops it so the emitted response is fine, but code doing `'publishedParent' in content` / `Object.keys` / exact deep-equal would see a stray key. If you want a clean shape: `const { publishedParent, ...rest } = content; return rest;`

src/tools/web/workbooks/getWorkbook.test.ts
```diff
@@ -181,6 +181,91 @@ describe('getWorkbookTool', () => {
       ]);
     });
 
+    it('attaches a publishedParent pointer and drops the redundant standalone published entry', async () => {
```
williamyen0203: Test gap: every `boundedContext` fixture in these tests uses `datasourceIds: null`, so nothing exercises a real allow-set *combined with* a `publishedParent`-bearing embedded entry — exactly the combination where the merge/filter ordering (see my comment on `mergeWorkbookDatasources`) can silently drop an allowed published DS. A case with `datasourceIds: new Set(['pub-luid-1'])` (parent allowed, embedded stub not) would guard against it.

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -176,6 +205,63 @@ export function getWorkbookLineageByLuid(response: unknown): Map<string, Array<L
...
+  embeddedDatasources: Array<z.infer<typeof metadataEmbeddedDatasourceSchema>> | null | undefined,
+): Map<string, PublishedParent> {
+  const parents = new Map<string, PublishedParent>();
+  const seenNames = new Set<string>();
+
+  for (const { name, parentPublishedDatasources } of embeddedDatasources ?? []) {
+    if (!name) {
+      continue;
+    }
+    if (seenNames.has(name)) {
```
williamyen0203: Nit: ambiguous-name detection is implemented twice with different techniques — here a `seenNames` Set + `parents.delete`, and in `toEmbeddedLineageContents` a `nameCounts` Map. Same "name seen more than once ⇒ ambiguous" rule, so a future tweak (case-insensitive, trim, etc.) would need changing in both places. Could share one helper.

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -75,26 +87,39 @@ const viewLineageResponseSchema = z.object({
...
 });
 
-// Shared GraphQL selection for embedded datasources' upstream (published) datasources. See
-// metadataEmbeddedDatasourceSchema for why we traverse embeddedDatasources rather than relying
-// solely on the content-level upstreamDatasources rollup.
-const embeddedUpstreamSelection = `embeddedDatasources {
+// Shared GraphQL selection for embedded datasources. upstreamDatasources is always fetched (see
+// metadataEmbeddedDatasourceSchema). includeEmbeddedParents additionally fetches the name and
+// published-parent linkage that getWorkbook needs; other callers omit it to avoid over-fetching.
+function embeddedUpstreamSelection(includeEmbeddedParents: boolean): string {
```
williamyen0203: Nit / altitude: `includeEmbeddedParents` is threaded through three layers (`embeddedUpstreamSelection` → `workbookLineageNodesSelection` → `getWorkbookLineageQuery`) purely to avoid over-fetching two cheap scalar fields (`name`, `parentPublishedDatasources`) for search-content. Given how small those fields are, always selecting them and letting callers ignore them would remove the flag plumbing. Judgment call — fine to keep if the over-fetch is meaningful.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.8.5",
+  "version": "4.8.6",
```
williamyen0203: Per the repo's CLAUDE.md, the `version` field shouldn't be bumped manually in a PR — `.github/workflows/auto-version-bump.yml` bumps it (default patch) after merge to `main`. This manual `4.8.5 → 4.8.6` (and the matching package-lock.json change) may conflict with or short-circuit that automated bump. Worth reverting unless it's intentional for a reason I'm missing.

src/sdks/tableau/methods/lineageUtils.ts
williamyen0203, line 545: **Severity: High — correctness**

`mergeWorkbookDatasources` de-dupes the standalone published entry here, but it runs *before* `filterLineageContentsByAllowedIds` (invoked via `mergeWorkbookLineage` in getWorkbook.ts). If `boundedContext.datasourceIds` allows the published LUID but *not* the paired embedded (sqlproxy) LUID:

- merge drops `pub-1` (claimed by `emb-1.publishedParent`) → `[emb-1{publishedParent: pub-1}]`
- the allow-set filter then drops `emb-1` (not allowed) → `[]`

Net: `upstreamDatasources` is empty even though `pub-1` was explicitly in the allowlist — the opposite of the intended lossless behavior.

**Suggested fix:** only drop the standalone published entry once you know its child survived the allow-set filter. Two options:

1. Apply the allow-set filter first, then de-dupe on the filtered lists — i.e. filter `embedded`/`published` before handing them to `mergeWorkbookDatasources`, rather than filtering the merged result afterward.
2. Keep the current order, but in `filterLineageContentsByAllowedIds`, when an embedded entry is dropped, re-surface its `publishedParent` as a standalone published entry (when the parent luid is allowed) so it isn't lost.

Option 1 is simpler. Either way, cover it with the regression test noted in my getWorkbook.test.ts comment.

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: **Severity: Low — robustness**

`toEmbeddedLineageContents` + `mergeWorkbookDatasources` now run outside the two try/catch blocks that previously guarded embedded enrichment (the old `embedded = toEmbeddedLineageContents(connections)` lived inside the REST try). Both are pure transforms over typed inputs, so this is low-risk — but if either ever throws, the whole `get-workbook` call fails instead of degrading with a logged warning as it did before.

**Suggested fix:** wrap the final assembly (`toEmbeddedLineageContents` + `mergeWorkbookDatasources` + `mergeWorkbookLineage`) in a try/catch that logs and falls back to returning the workbook without enriched lineage, matching the prior degrade-gracefully behavior. If you consider the pure-transform risk negligible, a one-line comment documenting the intentional choice is enough.

src/sdks/tableau/methods/lineageUtils.ts
williamyen0203: **Severity: Low — correctness / edge case**

`parent.name ?? parent.luid` only falls back on null/undefined. The wire schema allows `name` to be an empty string (`z.string().nullable().optional()`), so an empty-string parent name surfaces as `{ name: '' }` instead of falling back to the luid. The sibling test only covers a missing (undefined) name.

**Suggested fix:**
```suggestion
      parents.set(name, { luid: parent.luid, name: parent.name || parent.luid });
```
and add a fixture with `parentPublishedDatasources: [{ luid: 'pub-1', name: '' }]` asserting the luid fallback.

src/tools/web/workbooks/getWorkbook.test.ts
williamyen0203, line 184: **Severity: Medium — test coverage**

Every `boundedContext` fixture in these tests uses `datasourceIds: null`, so nothing exercises a real allow-set *combined with* a `publishedParent`-bearing embedded entry — exactly the combination where the merge/filter ordering (see my comment on `mergeWorkbookDatasources`) can silently drop an allowed published DS.

**Suggested fix:** add a test where the parent is allowed but the embedded stub isn't, and assert the published DS still surfaces:
```ts
it('keeps an allowed published DS when its embedded stub is out of bounds', async () => {
  // graphql: upstream pub-luid-1; embedded 'Embedded DS' -> parent pub-luid-1
  // connections: emb-luid-1 / 'Embedded DS'
  // boundedContext.datasourceIds = new Set(['pub-luid-1'])  // emb-luid-1 NOT allowed
  // expect response.data.upstreamDatasources to still include pub-luid-1
});
```
With the current ordering this fails (returns `[]`), which is what confirms the fix for the `mergeWorkbookDatasources` finding.

src/sdks/tableau/methods/lineageUtils.ts
sf-ali-mahmoud: Re-checked this against head `dab0a68`: the follow-up commit addressed the **leak** direction (embedded stub in-bounds, published parent out-of-bounds → new pointer-strip at `lineageUtils.ts:519-527`), but the **data-loss** direction this comment originally described is still present.

Ordering in `getWorkbook.ts:133-138` is unchanged: `mergeWorkbookDatasources(published, embedded)` runs *before* `filterLineageContentsByAllowedIds`. So when the published LUID is allowed but its paired embedded stub LUID is not:

1. `mergeWorkbookDatasources` drops standalone `pub-1` (claimed by `emb-1.publishedParent.luid`) → `[emb-1{publishedParent: pub-1}]`.
2. `filterLineageContentsByAllowedIds` filters on `content.luid` only (line 522); `emb-1`'s luid isn't in the allow-set → `emb-1` is dropped entirely. The new `.map` pointer-strip never runs because the entry is already gone.
3. Net: `upstreamDatasources = []` even though `pub-1` was explicitly allowed. Pre-PR, `pub-1` survived as a standalone entry — so this is still a regression.

Effectively the access decision for a live-connected published DS is now made on the embedded stub's luid, never on `publishedParent.luid`. Fix direction is unchanged from the original suggestion: make the dedup allow-set-aware, run it after filtering, or re-materialize the standalone published entry when its child is filtered out but the parent is allowed. The regression test proposed in the `getWorkbook.test.ts` thread would catch this — it doesn't look like it landed yet.

ycath: Fixed in fffadab6. `filterLineageContentsByAllowedIds` now runs on both the published and embedded lists *before* `mergeWorkbookDatasources`, and `mergeWorkbookLineage` is called without re-filtering. An out-of-bounds embedded stub can no longer suppress its still-allowed published parent. Added a regression test in getWorkbook.test.ts ("keeps an allowed published DS when its embedded stub is out of the bounded context").

src/tools/web/workbooks/getWorkbook.ts
ycath: Fixed in fffadab6. The `toEmbeddedLineageContents` + `mergeWorkbookDatasources` assembly is now wrapped in a try/catch that logs a warning and returns the unenriched workbook, matching the two enrichment fetches above it.

src/sdks/tableau/methods/lineageUtils.ts
ycath: Fixed in fffadab6. `buildEmbeddedParentMap` now uses `parent.name || parent.luid` so an empty-string parent name falls back to the luid. Added a test ("falls back to the parent luid when the parent name is an empty string").

src/tools/web/workbooks/getWorkbook.test.ts
ycath: Fixed in fffadab6. Added "keeps an allowed published DS when its embedded stub is out of the bounded context" to getWorkbook.test.ts, which stubs INCLUDE_DATASOURCE_IDS and exercises a real allow-set combined with a publishedParent-bearing embedded entry.

williamyen0203 (approved):

PR #868, opened 2026-09-01
chore: bump version to 2.67.4

nawartamawi:
## Summary
- bump the MCP package version from 2.67.3 to 2.67.4
- allow the internal publishing pipeline to create new immutable Desktop MCP artifacts

## Context
The pipeline for PR #858 rebuilt the fixed MCP at version 2.67.3, but publishing skipped because the 2.67.3 Artifactory artifacts already existed. A new version is required so the KPI fixes can be published and then bundled by tab-agent-south.

## Testing
- version-only change
- git diff --check

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "2.67.3",
+  "version": "2.67.4",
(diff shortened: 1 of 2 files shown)


mattcfilbert (approved):
lgtm

PR #869, opened 2026-09-01
@W-23867374: Add Tableau Knowledge MCP tools

rubywerman:
adds the tableau knowledge tool group to the web MCP server, conformant with the knowledge-v1 openapi spec.

eight read tools — get-knowledge-suggestions, list-knowledge-sources, search-knowledge-nodes, get-knowledge-node, get-knowledge-node-relationships, get-knowledge-lineage, get-knowledge-node-impact, list-knowledge-semantic-contexts — plus three writes: create / update / delete-knowledge-semantic-contexts.

get-knowledge-node is an exact-id fetch (GET /nodes/{id} → the node with its statements and connected children); resolve-by-query is gone from the spec, so use search-knowledge-nodes to find an id first. graphId is optional on every tool — omit it to run on the site's active/default graph. search results carry a similarity score and the matched node's semantic_statements, and results can include tableau-managed external contexts.

writes are gated behind the knowledge-write-tools feature flag (off by default). reads map to tableau:knowledge:read, writes to tableau:knowledge:write. delete is idempotent, refuses tableau-managed external contexts, and handles the service's 204 empty response, matching DELETE /semantic-contexts/{ctx_id}. knowledge request/response bodies have statement text masked from debug logs.

draft — opening for maintainer discussion on the tool set (+ metadata overlap/tuning), the MCP scope name (tableau:mcp:knowledge:*), and whether the write tools should ship gated as they are here. Also do we need A TK gate? still need rollout strat

verification: tsc clean, lint clean, 2999 unit tests pass; the read path and a create→update→delete round-trip were exercised live against a deployed knowledge service.


bcantoni:
Close and reopen to see if that clears the CLAbot.

bcantoni:
I'll work with Ruby to get added to the Tableau org.

PR #871, opened 2026-09-01
@W-21254846: Add mechanism for conditional registration

stephendeoca:

the pull request._

# Pull Request Template

## Description

Introduces a new tool property called `registrationConditions` which can be used to indicate if a certain criteria has been met before registering the tool. When a tool specifies a registration condition, a method for checking the condition must be implemented. The registration context object is also populated so that conditions do not need to be rechecked with every tool. This logic is gated by the `enforce-registration-conditions` feature flag.

Introduces 2 conditions, `RequiresPulse` and `RequiresPulsePremium`.

## Motivation and Context

As we increase our surface area of tools, we want to make sure only the relevant tools are registered for a user. For example, if pulse is not available for a site then the pulse tools should not be registered for a user.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Version Bump

- [ ] Patch
- [X] Minor
- [ ] Major

## Related Issues

N/A

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/registrationConditions.ts
```diff
@@ -0,0 +1,97 @@
+import { RestApiArgs } from '../../restApiInstance.js';
+
+/**
+ * A capability a tool can require at registration time to be verified by {@link checkRegistrationConditions}.
+ */
+export type RegistrationCondition =
```
stephendeoca, line 6: Step 1: add condition name here, then update tools to require said condition via: registrationConditions property

src/tools/web/registrationConditions.ts
```diff
@@ -0,0 +1,97 @@
...
+      }
+      case 'RequiresPulsePremium': {
+        if (context.hasPulsePremium === undefined) {
+          context.hasPulsePremium = await checkPulsePremium(restApiArgs);
+        }
+        if (!context.hasPulsePremium) {
+          return { registrationConditionsMet: false, failingCondition: 'RequiresPulsePremium' };
+        }
+        continue;
+      }
```
stephendeoca, line 76: Step 2: Add case for checking new condition, and populate context object so that you do not have to keep rechecking. Examples of checking for a condition can be found with `checkPulseEnabled`, `checkPulsePremium` or `getCurrentUserSiteRole`.

src/server.web.ts
```diff
@@ -251,6 +271,25 @@ export class WebMcpServer extends Server {
...
+      );
+      log({
+        level: 'warning',
+        logger: 'server',
+        message:
+          `${omittedCount} tool(s) were omitted from this session because their registration ` +
+          `conditions were not met — ${omittedByCondition}.`,
+      });
+
+      // TODO: for each condition append intialization message that explains that some features are unavailable.
```
stephendeoca, line 322: Step 3: For your new condition, append to the initialization message via `this.appendInstructions` that explains something along the lines of:

"Condition X was not met / is unavailable for this session, the following tools were not registered" + reasons why the condition was not met

src/tools/web/registrationConditions.ts
```diff
@@ -0,0 +1,265 @@
...
+/**
+ * Client-facing explanation for each condition, appended to the initialize instructions when that
+ * condition hid at least one tool. Registration runs before the transport connects, so the
+ * instructions are the only channel that reaches the user — without this the caller just sees a
+ * short tool list with no explanation.
+ *
+ * Typed as a total `Record` so adding a {@link RegistrationCondition} without user-facing copy is
+ * a compile error rather than a silent omission.
+ */
+const UNMET_CONDITION_INSTRUCTIONS: Record<RegistrationCondition, string> = {
```
stephendeoca, line 106: Step 3: Add message to give a client context on when a condition was required by a tool, but was not met.

src/tools/web/registrationConditions.ts
```diff
@@ -0,0 +1,265 @@
...
+              return false;
+            }
+
+            // Anything else is indeterminate. Throw so `retry` can attempt again; exhausting
+            // retries falls through to the fail-closed catch below.
+            throw result.error;
+          },
+        }),
+      {
+        maxRetries: MAX_PULSE_PROBE_ATTEMPTS - 1,
```
yyk16293: Let's rename to generic variable MAX_API_RETRY_ATTEMPTS so that all other API checks related to conditions uses the same number.

src/tools/web/registrationConditions.ts
yyk16293: Add these comments in the code itself.

yyk16293 (approved):

PR #873, opened 2026-09-02
fix(desktop): quote string filter members

nawartamawi:
## Summary

- serialize string categorical filter members as quoted Tableau literals before XML escaping
- preserve existing unquoted serialization for non-string members
- add regression coverage using the exact member names from the live Story Points detractor failure

## Why

Tableau can accept the worksheet apply while silently dropping categorical filters whose string members are emitted as raw values. The worksheet then renders without the requested driver/detractor member restriction, and post-apply verification correctly reports that the filter did not persist.

## Testing

- `npm test -- --run src/tools/desktop/authoring/binder/bindTemplate.test.ts` — 196 passed
- `npm run lint` — passed
- `npm run build:desktop:dirty` — passed
- live Tableau Desktop smoke: **Top detractors — Story Points** applied with `verification: "passed"`
- workbook readback confirmed all three quoted member nodes and the filter slice persisted

The repository-wide local test run entered environment-dependent timeout cases and was stopped; the PR pipeline remains the canonical full-suite gate.

### DIFF
--- a/src/tools/desktop/authoring/binder/bindTemplate.ts
+++ b/src/tools/desktop/authoring/binder/bindTemplate.ts
@@ -1855,6 +1855,7 @@ function ensureFilterColumnDependency(
 function buildInteractiveFilterNode(p: {
   columnRef: string;
   level: string;
+  datatype: string;
   context: boolean;
   values?: string[];
 }): string {
@@ -1864,7 +1865,7 @@ function buildInteractiveFilterNode(p: {
     const members = p.values
       .map(
         (v) =>
-          `<groupfilter function='member' level='${level}' member='${escapeXmlAttribute(v)}' user:ui-enumeration='inclusive' user:ui-marker='enumerate' />`,
+          `<groupfilter function='member' level='${level}' member='${serializeFilterMember(v, p.datatype)}' user:ui-enumeration='inclusive' user:ui-marker='enumerate' />`,
       )
       .join('');
     return (
@@ -1878,6 +1879,18 @@ function buildInteractiveFilterNode(p: {
     `<groupfilter function='level-members' level='${level}' user:ui-enumeration='all' />` +
     '</filter>'
   );
+}
+
+/**
+ * Tableau stores string members as quoted literals inside the XML attribute
+ * (`member='&quot;East&quot;'`). Raw strings are accepted at document dispatch but
+ * Desktop silently drops the filter during persistence. Non-string members use
+ * Tableau's unquoted representation (for example, a discrete year is `2026`).
+ */
+function serializeFilterMember(value: string, datatype: string): string {
+  if (datatype.trim().toLowerCase() !== 'string') return escapeXmlAttribute(value);
+  const tableauString = `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
+  return escapeXmlAttribute(tableauString);
 }
 
 /**
@@ -2069,6 +2082,7 @@ function applyProposalSplices({
       const filterNode = buildInteractiveFilterNode({
         columnRef: declared.columnRef,
         level: declared.level,
+        datatype: resolved.field.datatype,
         context: filter.context === true,
         values: filter.values,
       });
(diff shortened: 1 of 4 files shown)


src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1880,6 +1881,18 @@ function buildInteractiveFilterNode(p: {
   );
 }
 
+/**
+ * Tableau stores string members as quoted literals inside the XML attribute
+ * (`member='&quot;East&quot;'`). Raw strings are accepted at document dispatch but
+ * Desktop silently drops the filter during persistence. Non-string members use
+ * Tableau's unquoted representation (for example, a discrete year is `2026`).
+ */
+function serializeFilterMember(value: string, datatype: string): string {
```
tonythebest, line 1890: Small robustness question on `serializeFilterMember`: this calls `datatype.trim().toLowerCase()` unconditionally, so if `resolved.field.datatype` is ever `undefined`/absent for a resolved field, `.trim()` will throw and take down the whole filter-node build (and the apply) rather than just falling back to the previous unquoted behavior. The tests always supply a `datatype`, so this path isn't exercised.

Is `field.datatype` guaranteed populated for every resolved categorical field here? If it can be missing, a guard like `if ((datatype ?? '').trim().toLowerCase() !== 'string')` would keep a column with an absent datatype on the safe (unquoted, prior) path instead of erroring. Not blocking — just want to confirm the invariant.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1880,6 +1881,18 @@ function buildInteractiveFilterNode(p: {
...
+  if (datatype.trim().toLowerCase() !== 'string') return escapeXmlAttribute(value);
```
tonythebest, line 1891: Could we complete the Tableau/XML codec here as a follow-up? Literal TAB/LF/CR in an XML attribute normalize to spaces, and other C0 controls can make the workbook payload invalid. Tableau’s writer emits numeric XML references for TAB/LF/CR. The current special-character test does not cover that boundary.

Not blocking this PR — Insights/Story Points members do not contain those characters.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1880,6 +1881,18 @@ function buildInteractiveFilterNode(p: {
...
+  const tableauString = `"${value.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
```
tonythebest, line 1892: This new backslash doubling is not inverted by `roundStackedBar.ts` (it decodes escaped quotes only). A persisted `Path\Name` filter is read as `Path\\Name`, so rounded-bar seed verification can reject the matching raw value before mutation.

Not blocking this PR. A shared encoder/decoder (or decode `\"` and `\\` in one pass) plus bind→rounded-bar coverage would close it.

tonythebest (approved):
LGTM. The Story Points string-member quoting fix is correct, tests and CI look good, and the live detractor smoke is the persistence proof the unit suite cannot give.

Two follow-ups below are not merge blockers.

src/tools/desktop/authoring/binder/bindTemplate.ts
nawartamawi: Good catch — fixed in `8c12d2c2`.

I replaced the quote-only `replaceAll` with a one-pass decoder for quoted Tableau filter members. It now reverses both escaped backslashes (`\\`) and escaped quotes (`\"`) while leaving legacy unquoted members unchanged.

I also added a regression case for `Path\Name`. The related rounded-bar suites pass (188 tests), along with lint and TypeScript checks.

src/tools/desktop/authoring/binder/bindTemplate.ts
nawartamawi: Confirmed: `field.datatype` is guaranteed to be a string at this call site. Raw field metadata can omit it, but `summarizeSchema` normalizes that with `datatype: f.datatype ?? ''`, and `SchemaField.datatype` is required. Therefore missing metadata reaches `serializeFilterMember` as `''`; `.trim()` is safe and the value follows the existing unquoted fallback path. I’m leaving the parameter contract strict rather than adding a redundant local guard.

PR #874, opened 2026-09-02
feat(refine): round compatible ordinary bar charts

mattcfilbert:
Make `round_bar` the supported path for compatible ordinary bar charts, including horizontal and vertical simple bars and single-field Color stacks. They share one guarded Polygon renderer; `round_stacked_bar` remains an alias.

The rule to preserve: keep zero baselines and stacked joins square, round only outer tips, preserve data and formatting semantics, and report uncertain readback without retrying.

## Scope

Grouped, dual-axis, specialty, calculated, table-calculation, and malformed shapes are refused before writing. This is Polygon geometry, not a native Bar corner-radius property. Conversion also refuses when the 10,000-row underlying-data read cannot prove two distinct seed values for every visible bar.

## Validation

- Refreshed against `feature/desktop` at `601021e1`, preserving nested-sort behavior and escaped string-filter handling. Package version: 2.68.1; the exact `refine-worksheet` schema limit remains 1,656 bytes.
- At `44561ad0`, `scripts/agent-check` passed: 397 files / 6,946 tests, lint, typecheck, Desktop build, and lockstep checks.
- Sol adversarial review found no actionable P0–P2 issues. Earlier fixes for tiny-bar rounding checks and duplicate declarations remain covered.
- Earlier live Desktop proof: a horizontal Profit-by-Region chart applied once, read back 14 helpers / 4 groups / 48 polygon rows, and rendered square baselines with rounded outer tips. A second request sent no write. This live test was not repeated after the merge refresh.

— MattGPT


src/desktop/refine/roundStackedBarVerification.ts
```diff
@@ -637,6 +671,33 @@ export function verifyRoundStackedBarPostSummary(
...
+      if (!interval) continue;
+      const beginsAtZero = near(group.value > 0 ? interval.low : interval.high, 0);
+      if (!beginsAtZero) {
+        addFinding(
+          findings,
+          'segment-value',
+          `${group.category} simple bar does not begin at zero.`,
+        );
+      }
+      const nonzeroTipRounded = group.value > 0 ? interval.topRounded : interval.bottomRounded;
```
myu404, line 700: ### [P2 · SUGGESTED] Use a scale-aware rounded-tip check

**Purpose**
A supported nonzero simple bar should settle when Desktop reads back the exact generated geometry.

**Finding**
Baseline preflight accepts values above the fixed `1e-6` zero tolerance, but the generated simple-bar radius is only 2% of the value and `isTopRounded`/`isBottomRounded` require offsets greater than `1e-6`. For a value of `1e-5`, the exact planned radius is `2e-7`, so the detector classifies the generated tip as square and this new simple-bar check rejects it.

**Impact**
Desktop can apply the correct polygon, yet all strict readback polls remain unsettled and the tool reports an unknown, non-retry-safe result after the write.

**Recommendation**
Derive roundedness from the expected scale-aware radius/coordinates, or use a relative tolerance that does not exceed valid generated radii; add positive and negative simple-bar cases near the current boundary.

— 🤖 Reviewed by MichaelGPT

src/desktop/refine/roundStackedBar.ts
```diff
@@ -685,54 +707,97 @@ function classifyOrdinary(
...
-    category.type !== 'nominal' ||
-    category.derivation !== 'None' ||
-    !measure ||
-    measure.type !== 'quantitative'
-  ) {
-    return refusal(
-      'round_stacked_bar supports only a vertical Category-columns / SUM-measure-rows chart.',
-    );
-  }
+  const rowInstance = instances.find((instance) => instance.name === rowCiName);
```
myu404, line 710: ### [P2 · SUGGESTED] Consider cardinality-checking shelf declarations

**Purpose**
The new simple and horizontal routes should refuse malformed or ambiguous source shapes before the one worksheet write.

**Finding**
These shelf lookups select the first matching instance, while `columnDefinitions()` similarly collapses same-name Category/filter definitions. A conflicting duplicate on an otherwise allowed column can therefore reach transformation, and source/readback verification replans through the same parser. This behavior predates the PR for vertical stacks but is newly exposed to simple and horizontal bars.

**Impact**
If Desktop can emit or preserve this duplicate shape, `round_bar` can mutate it rather than refuse it, with recognized semantics depending on declaration order.

**Recommendation**
Could we confirm Desktop canonicalizes these declarations uniquely, or require unique shelf-instance and source-definition matches for the accepted Category, Value, and optional filter fields, with a focused conflicting-duplicate fixture?

— 🤖 Reviewed by MichaelGPT

myu404 (approved):
**Recommendation: APPROVE**

This change makes `round_bar` the preferred guarded path for compatible simple and single-Color stacked bars in both orientations, while preserving `round_stacked_bar` as an alias. The review covered source-shape eligibility, helper geometry and orientation swaps, preflight evidence, the single-write transaction, strict workbook/summary readback, no-retry behavior, idempotent recognition, public tool compatibility, and focused test coverage.

Two non-blocking P2 suggestions remain: use a scale-aware rounded-tip check so exact geometry for very small simple bars can settle after the write, and consider explicit cardinality checks for conflicting shelf declarations in the newly supported simple/horizontal shapes.

Positive observations: the shared planner avoids separate simple/stacked implementations; simple bars omit Segment and stack-only helpers; preflight completes before the sole per-sheet apply; post-dispatch uncertainty remains non-retry-safe; and the tests exercise both orientations, signed geometry, malformed encodings, aliasing, raw-seed refusal, and exact readback.

Limitations: this was a pinned, read-only review. I did not rerun builds, tests, or live Tableau scenarios. The hosted status surface exposed only a successful Salesforce CLA status, so the PR author's reported lint, typecheck, 6,861 tests, Desktop build, lockstep, and live-render checks were not independently verified.

— 🤖 Reviewed by MichaelGPT

src/desktop/refine/roundStackedBarVerification.ts
mattcfilbert: Fixed in ce10fb0a. The four tip detectors now take a per-group tolerance of 0.1 * the expected radius (min(|value|/2, 0.02*span)) instead of the fixed 1e-6, so a bar that clears the 1e-6 baseline gate is classified correctly at any size. At value 1e-5 the planned radius is 2e-7 and the rounded tip now settles. Added positive and negative simple-bar cases at that boundary: a rounded tip passes, and a square tip is still rejected, so detection holds at the tiny scale where the fixed tolerance could not.

— posted by MattGPT on Matt's behalf

src/desktop/refine/roundStackedBar.ts
mattcfilbert: Added the guard in b6757e1f. round_bar now refuses when any accepted field's column-instance name or source-column definition is non-unique — Category, Segment, SUM, and the optional filter — before the transform runs, so a conflicting duplicate is refused rather than mutated. It hardens every bar type, not only the new simple and horizontal routes. Valid single-declaration worksheets are unaffected; added a conflicting-duplicate fixture (duplicate Category definition, duplicate Category instance, duplicate SUM instance).

— posted by MattGPT on Matt's behalf

Issue #146, opened 2025-11-04
Support for Embedded Data Sources


shai-li-valdarsky:
### Current Limitation
The Tableau MCP server currently only supports querying **published data sources**. Workbooks that use embedded (non-published) data sources cannot be queried directly through the MCP tools (`query-datasource`, `get-datasource-metadata`, etc.). This means stakeholders need to use workarounds to access the datasources which is extremely time consuming for everyone - or they have to compromise and download the data from the view.

### Proposed Feature
Extend the MCP server to support embedded data sources by:

1. **Discovery**: Add ability to list embedded data sources within a workbook
   - New tool or extend `get-workbook` to include embedded data source metadata
   
2. **Metadata Access**: Extend `get-datasource-metadata` to work with embedded sources
   - Reference format: `workbook_id:datasource_name` or similar
   
3. **Querying**: Extend `query-datasource` to support embedded sources
   - Allow queries against embedded sources using workbook/source identifiers

Thanks!

anyoung-tableau:
Hey @shai-li-valdarsky, thanks for the issue. VizQL Data Service is working on adding support for embedded data sources as we speak, but I don't have a definitive timeline I can share yet. Stay tuned.

Krabkvadrat:
Hey @anyoung-tableau ! 
Can then consider adding  Query View Data method from API?
This'll be very useful for our users, as it'll provide the necessary flexibility for "speaking" with data in Tableau 
Wihtout need to publsih datasources

sameer-payufin:
Is there any update on this feature?

PR #875, opened 2026-09-02
feat(desktop): add individual datasource tools

myu404:
## Decision

RULE WE NOW KEEP: Desktop datasource tools resolve through workbook inventory, expose credential-redacted datasource XML, and apply cached XML only through contained reads.

## Description

- Add `get-datasource-info`, `get-datasource-xml`, and `apply-datasource` for External Client API 0.2.10 and newer.
- Preserve encoded datasource identifiers and unchanged safe XML while clearing credential attributes and credential-bearing connection values before inline or cached output.
- Use descriptor-grounded cache containment for workbook, worksheet, dashboard, storyboard, and datasource apply flows.
- Pin the 0.2.10 individual datasource operations and `datasource-not-found` problem code in the External Client API contract fixture.

## Motivation and Context

The individual datasource routes let an agent inspect or update one workbook-local datasource without reading or replacing the whole workbook document. Credential redaction keeps connection secrets out of MCP responses and cache files, while contained reads prevent cached apply paths from escaping the Desktop cache.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- `npx --no-install vitest run src/desktop/cachePath.test.ts src/desktop/externalApi/datasourceRoutes.test.ts src/desktop/externalApi/externalApiToolExecutor.test.ts src/desktop/limits/artifactSummary.test.ts src/desktop/wrappers/cacheFingerprint.test.ts src/desktop/wrappers/applyDatasourceXml.test.ts src/server.desktop.test.ts src/tools/desktop/api/applyDatasource.test.ts src/tools/desktop/api/applyPreamble.test.ts src/tools/desktop/api/datasourceResult.test.ts src/tools/desktop/api/getDatasourceInfo.test.ts src/tools/desktop/api/getDatasourceXml.test.ts src/tools/desktop/api/listWorkbookDatasources.test.ts src/tools/desktop/api/resolveDatasourceRef.test.ts --maxWorkers=1`
- `npx vitest run src/desktop/externalApi/externalApiContract.test.ts src/desktop/limits/datasourceCredentialRedaction.test.ts src/desktop/limits/inlineXmlCap.test.ts src/tools/desktop/api/getDatasourceXml.test.ts src/tools/desktop/api/applyPreamble.test.ts src/tools/desktop/api/applyDatasource.test.ts src/tools/desktop/api/applyWorkbook.test.ts src/tools/desktop/api/applyWorksheet.test.ts src/tools/desktop/api/applyDashboard.test.ts src/tools/desktop/api/applyStoryboard.test.ts`
- `npx vitest run src/desktop/externalApi/externalApiToolExecutor.test.ts`
- `npx vitest run src/desktop/externalApi/externalApiHttpAsyncDispatch.test.ts`
- `npx vitest run --config ./vitest.config.ts`

## Related Issues

Relates to #146.

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/desktop/api/xmlReadResult.ts
```diff
@@ -84,6 +86,25 @@ export function finishXmlRead<K extends string>({
 
   if (capFired) {
     logInlineXmlCapHit({ tool: toolName, bytes, capBytes, file: cacheFile });
+    if (kind === 'datasource') {
```
tableaukyler: 🔴 **Credential leak: raw datasource XML goes out unredacted, in both modes.**

This branch (and the inline-return path above it) hands the raw `xml` straight to the cache-file write / tool response with no redaction — only the prose summary (`formatArtifactSummary`, line 92) is scrubbed. For `kind === 'datasource'`, `xml` can contain `<connection password="..."/>` / oauth-token attributes. Proven by this PR's own test: `getDatasourceXml.test.ts` builds XML with `password="must-not-appear"` and asserts `readFileSync(body.file, 'utf-8')).toBe(xml)` (password on disk verbatim), only checking the *message* doesn't contain it. Same gap on the inline path a few lines above (`[inlineKey]: xml`) — the raw credential goes straight into the tool response / conversation.

Worksheet/dashboard/workbook reuse this same write/return path safely today because their XML has no connection attributes; this is the first branch that routes credential-bearing XML through it. Needs redaction of the raw payload (or at minimum the `<connection>` element's secret-bearing attributes), not just the summary.

---

🟡 Separately: this branch hand-rolls a cap-fired message (size sentence, blank line, "Structural summary:", `formatArtifactSummary`, blank line, edit-then-apply sentence) that's ~90% identical in shape to `buildInlineCapFileMessage` used a few lines below for every other kind — just to drop the worksheet/dashboard `selector` wording. Consider parameterizing `buildInlineCapFileMessage` (e.g. an optional `selector`) instead of forking a parallel literal.

src/tools/desktop/api/applyPreamble.ts
```diff
@@ -24,6 +35,7 @@ export function runApplyPreamble({
   session,
   emptyPathGuidance,
   notFoundGuidance,
+  secureContainedCacheRead = false,
```
tableaukyler: 🟠 **This hardening is scoped to one of five apply tools.** `secureContainedCacheRead` defaults to `false`, and only `applyDatasource.ts` passes `true`. `applyWorksheet.ts`/`applyWorkbook.ts`/`applyDashboard.ts`/`applyStoryboard.ts` still go through the legacy `existsSync`/`readFileSync` branch below (line 116) with no containment/symlink check at all — the exact risk class `cachePath.ts` was built to close. Not a regression from this PR, but since the PR clearly identified and fixed this gap, worth a follow-up to backfill the other four rather than leaving datasource as the only hardened apply path.

src/tools/desktop/api/applyPreamble.ts
```diff
@@ -43,6 +57,62 @@ export function runApplyPreamble({
...
+    if (!readResult.ok) {
+      switch (readResult.issue) {
+        case CONTAINED_CACHE_READ_ISSUE.outsideCache:
+        case CONTAINED_CACHE_READ_ISSUE.unsafeFile:
+          return new ArgsValidationError(
+            `Security error: the ${kind} file must be a regular file contained in the Desktop cache directory.\n\n` +
+              `Cache directory: ${getCacheDir()}\nRequested: ${file}`,
+          ).toErr();
+        case CONTAINED_CACHE_READ_ISSUE.missing: {
+          const NotFoundError =
```
tableaukyler: 🟡 This secure branch (lines 60-114) duplicates the legacy branch's (116-138) shape end-to-end — same `NotFoundError` ternary verbatim here and at line 120, then session resolution, then a sidecar check, then the same `Ok({...})` build — differing only in which primitives read the file/sidecar. Consider factoring the shared 5-step skeleton (validate → read → resolve session → sidecar-check → Ok) with the read/sidecar-check step injected, so the two copies can't drift the next time either the error mapping or the sidecar contract changes.

src/desktop/externalApi/types.ts
```diff
@@ -20,6 +20,8 @@ export const EXTERNAL_API_ROUTES = {
   workbookDashboards: '/v0/workbook/dashboards',
   workbookDashboardsNew: '/v0/workbook/dashboards:new',
   workbookDatasources: '/v0/workbook/datasources',
+  workbookDatasource: '/v0/workbook/datasources/{id}',
+  workbookDatasourceDocument: '/v0/workbook/datasources/{id}/document',
```
tableaukyler, line 24: 🟠 These two new routes (gated `minApiVersion: '0.2.10'` on the tools that use them) complete steps 1-3 of `externalApi/README.md`'s "Adding awareness of a new API method" recipe, but step 4 is skipped: `externalApiContract.test.ts`'s route list doesn't include either route, and `__fixtures__/externalClientApi-openapi.json` is still pinned to spec `0.2.9` — a version behind the floor these tools require. Right now there's no contract-level check that these routes match the real API spec; worth adding when the fixture gets recaptured against a 0.2.10+ build.

tableaukyler (commented):
Reviewed the three new Desktop datasource tools (`get-datasource-info`, `get-datasource-xml`, `apply-datasource`) plus the new `cachePath.ts` TOCTOU-safe cache read. The TOCTOU work itself is solid, and `datasourceResult.ts`'s field allowlist correctly keeps *metadata* credential-safe — but that discipline doesn't extend to the raw XML path, which is the one 🔴 below.

**🔴 Raw datasource XML (passwords/oauth tokens) goes out unredacted, in both modes**
`finishXmlRead` (`src/tools/desktop/api/xmlReadResult.ts`) never redacts the `xml` it returns inline or writes to the cache file — only the prose *summary* (`artifactSummary.ts`) is scrubbed. The PR's own test proves it: `getDatasourceXml.test.ts` builds a `<connection password="must-not-appear"/>` datasource, and in file mode asserts `readFileSync(body.file, 'utf-8')).toBe(xml)` — the password lands on disk verbatim — while only checking the *message* doesn't contain it. Inline mode (`datasourceXml: xml` a few lines above) has the same gap, putting the raw credential straight into the tool response / conversation. Worksheet/dashboard/workbook reuse the same unredacted path safely because their XML doesn't carry connection attributes; this PR is what first routes credential-bearing XML through it.

**🟠 The new TOCTOU-safe cache read only protects `apply-datasource`**
`runApplyPreamble`'s `secureContainedCacheRead` defaults to `false`; only `applyDatasource.ts` passes `true`. `applyWorksheet.ts`/`applyWorkbook.ts`/`applyDashboard.ts`/`applyStoryboard.ts` still go through the legacy `existsSync`/`readFileSync` branch with zero containment check — the exact class of risk `cachePath.ts` was built to close. Not a regression (pre-existing behavior), but worth a follow-up to backfill the other four rather than leaving datasource as the only hardened tool.

**🟠 New routes skip step 4 of the External Client API recipe**
`workbookDatasource`/`workbookDatasourceDocument` (gated `minApiVersion: '0.2.10'`) complete steps 1-3 of `externalApi/README.md`'s recipe cleanly, but `externalApiContract.test.ts`'s route list and the `__fixtures__/externalClientApi-openapi.json` fixture (still pinned to spec `0.2.9`) were never updated, so these two routes have no contract coverage against the real API spec.

**🟡 Two Template Method duplications** (both ≥80 on the design-pattern rubric, inline below):
- `xmlReadResult.ts`'s datasource cap-message forks `buildInlineCapFileMessage`'s skeleton instead of parameterizing it.
- `applyPreamble.ts`'s new secure branch duplicates the legacy branch's 5-step skeleton (including the literal `NotFoundError` ternary) instead of factoring one skeleton with a pluggable read step.

Everything else came back clean: tool registration, comment discipline, blast-radius on `checkSidecar`/`ArtifactKind`/`attr()` regex/`sendDocumentApply`, and no regression of prior traversal-guard or credential-safety fixes. (One bug candidate — a malformed `%` crashing `canonicalDatasourceSegment` — didn't hold up under verification: the datasource id is always API-supplied and pre-encoded, never raw user text, so it's not reachable from ordinary input.)

No docs action — `tableau-mcp` isn't in the mapped docs-repo areas.

src/tools/desktop/api/xmlReadResult.ts
myu404: Addressed in 869ed586. get-datasource-xml now clears credential-like attributes and credential-bearing URL or connection-string values before finishXmlRead, so both inline output and cached XML use the sanitized document and the sidecar hashes that same content. Tests cover OAuth access and refresh tokens, passwords, connection strings, and percent-encoded credential names while preserving safe XML byte-for-byte. I also parameterized buildInlineCapFileMessage for selector-free datasource documents and removed the duplicate cap-message branch.

src/tools/desktop/api/applyPreamble.ts
myu404: Addressed in 869ed586. I removed the secureContainedCacheRead opt-in and the legacy filesystem branch. runApplyPreamble now always uses descriptor-grounded contained reads for both the cached XML and its sidecar. The shared preamble test now verifies outside-cache rejection for workbook, worksheet, dashboard, storyboard, and datasource applies.

src/tools/desktop/api/applyPreamble.ts
myu404: Addressed in 869ed586. runApplyPreamble now has one validate, contained-read, session-resolution, sidecar-check, and result path. Removing the legacy branch also removes the duplicated NotFound mapping and prevents the two flows from drifting.

src/desktop/externalApi/types.ts
myu404: Addressed in 869ed586. The contract fixture is now pinned to 0.2.10 using the merged monolith datasource route metadata, and the contract test covers workbookDatasource plus workbookDatasourceDocument with the expected getWorkbookDatasource, getDatasourceDocument, and applyDatasourceDocument operation IDs. datasource-not-found is also included in the problem-code contract.

src/desktop/limits/datasourceCredentialRedaction.ts
```diff
@@ -0,0 +1,80 @@
...
+  'refreshtoken',
+  'secret',
+  'secretaccesskey',
+  'secretkey',
+  'secrettoken',
+  'token',
+]);
+
+const CREDENTIAL_NAME_SUFFIXES = [...CREDENTIAL_ATTRIBUTE_NAMES];
+const ATTRIBUTE_PARAMETER = /(?:^|[?&;])([^?=&#;\s]+)=([^&#;\s]*)/g;
```
mattcfilbert: [P2] Could we cover `PWD` and optional connection-string whitespace here? I reproduced both `connection-string="Server=example; Password = SYNTHETIC_MARKER"` and `odbc-connect-string="UID=user;PWD=SYNTHETIC_MARKER"` passing through this helper unchanged; ordinary `Password=SYNTHETIC_MARKER` is cleared. If either form reaches the MCP, the value survives both inline and cached output. Please recognize these forms and add regression coverage for both output modes. Non-blocking: native export already disables secure information, so this is a gap in the fallback protection, not a demonstrated leak from current Desktop.

— MattGPT

mattcfilbert (approved):
MattGPT review of `7fd73e87`, using the Andy lens.

Michael, the datasource tools fit the existing read/apply flow, and the version gates and wire contracts match the native API. The earlier cache-read and contract-test fixes are in place.

No P0/P1 findings. Approving with one non-blocking P2 on fallback redaction of connection strings.

Verified the full unit suite (7,108 tests), typecheck, changed-file lint, and the merged native route/export contract. No live Desktop 0.2.10 round-trip or connector-specific credential audit was run.

src/desktop/limits/datasourceCredentialRedaction.ts
myu404: 🤖 Posted by MichaelGPT

Thanks for catching this. I updated the fallback redactor to recognize `PWD` and optional whitespace around connection-string assignments. The regression coverage now exercises both inline and cached output paths.

PR #879, opened 2026-09-02
feat(desktop): report and harden deterministic insight authoring

nawartamawi:
## Summary

- emit three progress notifications from `bind-template`: preparing the template, applying workbook changes, and verifying readback
- keep progress reporting fail-open so an older or disconnected client cannot interrupt workbook authoring
- recover deterministic worksheet identity using the requested title plus its bound fields when Tableau strips the custom idempotency marker
- reuse matching numeric measure calculations when Tableau canonicalizes `real` to `integer` or `integer` to `real`

## Why

Deterministic Reveal Insight authoring currently appears as one opaque operation. This gives the backend stable phase boundaries while also fixing two retry cases: recreating a KPI after deleting its worksheet and recognizing a deterministic worksheet whose custom identity attribute was removed by Tableau.

## Validation

- `npx vitest run src/tools/desktop/authoring/binder/bindTemplate.test.ts src/tools/desktop/authoring/datasource/authorCalc.test.ts` — 229 passed
- `scripts/agent-check` — all 5 checks green
  - lint
  - typecheck
  - 6,847 unit tests
  - Desktop build
  - lockstep-core hash gate

## Cross-repository rollout

- Backend consumes these progress notifications and emits persisted chat steps.
- UI renders the structured progress details.

- Backend: https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/179
- UI: https://gitlab.tableausoftware.com/browser-clients/tab-agent-south-ui/-/merge_requests/97

### DIFF
--- a/src/tools/desktop/authoring/binder/bindTemplate.ts
+++ b/src/tools/desktop/authoring/binder/bindTemplate.ts
@@ -91,6 +91,7 @@ import {
   DesktopCommandExecutionError,
   IncompleteOperationError,
 } from '../../../../errors/mcpToolError.js';
+import { log } from '../../../../logging/logger.js';
 import { DesktopMcpServer } from '../../../../server.desktop.js';
 import { getExceptionMessage } from '../../../../utils/getExceptionMessage.js';
 import {
@@ -1640,11 +1641,17 @@ function deterministicWorksheetTitle(
   workbookXml: string,
   requestedTitle: string,
   idempotencyKey: string | undefined,
+  bindingFields: string[] = [],
 ): string {
   if (!idempotencyKey) {
     return collisionFreeWorksheetTitle(workbookXml, requestedTitle);
   }
-  const existingTitle = worksheetTitleForIdempotencyKey(workbookXml, idempotencyKey);
+  const existingTitle = existingDeterministicWorksheetTitle(
+    workbookXml,
+    requestedTitle,
+    idempotencyKey,
+    bindingFields,
+  );
   if (
     existingTitle &&
     classifyWorksheetReplaceTarget(workbookXml, existingTitle) === 'replaceable'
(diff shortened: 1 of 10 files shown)


src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -243,6 +243,18 @@ function prepareCalculationBatch({
   return new Ok({ editedXml, authoredCalcs });
 }
 
+function calculationDatatypesMatch(existing: string, requested: Datatype, role: Role): boolean {
```
tonythebest, line 246: `calculationDatatypesMatch` now lets integer↔real measures reuse with `authoredCalcs: []` (helper tests in `authorCalc.test.ts`). Ordinary `author-calc` still goes through `authorCalculationsInWorkbook` and immediately does `authoredCalcs[0]` with no empty check. `settled` on an empty list is vacuously true, so a previously colliding numeric retry can POST the unchanged workbook and then throw. The wrapper file is unchanged in this PR, so this is the seam that actually moved.

Please gate the alias to the deterministic prepare path, or return the existing caption/name when reuse produces no new calcs. The helper `each()` is not a tool-level test.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -3379,6 +3483,27 @@ export const getBindTemplateTool = (server: DesktopMcpServer): DesktopTool<typeo
             }
           }
 
+          if (target_worksheet === undefined && trustedDeterministicApply && idempotencyKey) {
```
tonythebest, line 3486: This deterministic reuse branch can run with non-empty `atomicCalcs`. Those calcs were only prepared into the local workbook XML; returning `reused` here skips `performAutoApply`, so a retry after one generated calc was removed leaves Desktop incomplete while South reports reused success. The sibling remembered-sheet branch already requires `atomicCalcs.length === 0`. Please apply the same guard (or continue through combined apply/readback) and add a missing-calc + stamped/stamp-stripped identity regression test.

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -1656,6 +1663,56 @@ function deterministicWorksheetTitle(
...
+  let worksheetXml: string | null;
+  try {
+    worksheetXml = extractSheetXml(workbookXml, worksheetTitle);
+  } catch {
+    return false;
+  }
+  if (!worksheetXml) return false;
+
+  const identities = new Set<string>();
+  for (const match of worksheetXml.matchAll(/<column\b[^>]*>/g)) {
```
tonythebest, line 1700: Stamp-strip recovery walks `/<column\b[^>]*>/` and reads `caption`/`name` only. That regex also matches `<column-instance>`, but `name` there is the instance token and `column='[Sales]'` is ignored, so a column-instance-only sheet will not match binding `Sales` and we fall through to a suffixed new sheet. The new test uses `<column name='[Region]'>` / `[Sales]`, so it does not cover that XML shape. Fine as fail-closed; worth a follow-up fixture if Desktop dogfood starts duplicating KPIs after Tableau strips the identity attribute.

tonythebest (approved):
Approved at 4025f975. Follow-ups are on the changed lines and are not blocking this approval:

- Shared integer↔real reuse can leave ordinary `author-calc` with empty `authoredCalcs` and then throw on `[0]`.
- Deterministic identity reuse can return `reused` while prepared `atomicCalcs` were never applied.

Happy to follow up after merge if you want those gated before dogfood.

PR #880, opened 2026-09-03
@W-24015251: Bump version on the PR branch before merge instead of after

sf-ali-mahmoud:
## Summary
- Direct pushes to `main` from `auto-version-bump.yml` are rejected by branch protection (`GH013`: PRs + required checks only). Moves the bump earlier: it now runs on the PR itself (`opened`/`synchronize`/`reopened`), recomputing against `main`'s current version on every run, and pushes the bump commit to the PR's own branch — which merges into `main` as part of the normal single merge push, no follow-up push needed.
- Tagging moves to a separate job that only runs after merge (`closed` + `merged == true`), since pushing a tag doesn't touch the protected branch ref.
- Skips fork PRs (`head.repo.full_name != base repo`), since `GITHUB_TOKEN` can't push back to a fork's branch.
- Passes the PR body through an environment variable rather than interpolating it into the shell script — the previous approach fed untrusted PR body text into a bash script, a script-injection risk.

## Test plan
- [ ] Open a PR against `main`, confirm a `chore: bump version to X.Y.Z` commit gets pushed to the PR branch automatically.
- [ ] Push another commit to the same PR, confirm the bump recomputes correctly and doesn't double-bump.
- [ ] Merge the PR, confirm a matching `vX.Y.Z` tag is created afterward.
- [ ] Open a PR from a fork, confirm the bump job is skipped (no failed push attempt).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.6.0",
+  "version": "4.6.1",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.6.1",
+  "version": "4.8.1",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.6.0",
+  "version": "4.6.1",
```
stephendeoca: I think because we regressed in version with the last commits this should be something like 4.7.2

stephendeoca (approved):

package.json
sf-ali-mahmoud: The bot kicked in after the latest merge from main and updated to 4.8.1

PR #885, opened 2026-09-04
feat(desktop): complete the Show Me workflow

myu404:
## Decision

This establishes a rule we now keep: agents discover applicable Show Me options before applying the exact returned type with the same worksheet and selection context.

## Description

- Add `get-show-me-options` over `GET /v0/workbook/worksheets/{id}/showMe` with typed applicability details.
- Keep `show-me` as the separate apply tool and surface typed External Client API rejection errors.
- Gate discovery at External Client API `0.2.18` while retaining the apply floor at `0.2.11`.

## Motivation and Context

Agents need Tableau's Show Me recommendations and applicability result instead of guessing a `showMeType` before calling the apply tool.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Focused discovery, apply, transport, contract, registration, and tool-reference tests; lint, typecheck, Desktop build, and end-to-end Tableau Agent smoke coverage.

## Verification

- `npx vitest run src/tools/desktop/api/showMe.test.ts src/tools/desktop/api/getShowMeOptions.test.ts src/desktop/externalApi/externalApiContract.test.ts src/desktop/externalApi/externalApiHttp.test.ts src/desktop/externalApi/externalApiToolExecutor.test.ts src/server.desktop.test.ts src/tools/toolName.test.ts src/tools/desktop/emittedToolReferences.test.ts`

## Related Issues

- None.

## Checklist

## Contributor Agreement

      its Contribution Checklist.


b-carver:
(AI reviewed) Deep review across this PR and the paired monolith #66156, with parallel specialized agents and every Important finding reconciled against source. No blocking defects — the discover→apply flow, the 202→poll error surfacing, and version gating are solid and well tested. Consumer-side highlights:

**Important — Show Me type-set drift.** Discovery returns `showMeType: z.string()` (open, by design) but apply enforces `z.enum(SHOW_ME_TYPES)` (closed, 26 values), and nothing binds the two — and the C++ producer maintains its own `GetSupportedShowMeTypes()` / `ExpectedShowMeTypes()` lists. They match exactly today (no current bug), but the drift is already seeded in fixtures (`getShowMeOptions.test.ts` returns `scatter-plot`; the mock uses `native-future-viz` — neither is in the enum). A future native type shipped by a Desktop ≥0.2.15 before the enum is hand-updated would be discoverable with `isApplicable:true` yet rejected at `show-me`'s zod boundary with an opaque "invalid enum value" — while the discovery description tells the agent to "use only a returned showMeType… never invent." Suggest either (a) a drift-guard/contract test asserting `SHOW_ME_TYPES` == the producer's serialized enum (and every `isApplicable:true` option ∈ the set), or (b) relaxing apply to `z.string().min(1)` and letting the server reject unsupported types (`mapClientError` already surfaces the producer 400/409 cleanly).

**Minor — `show-me` description is too terse for an LLM-facing tool.** `'Apply a Show Me type.'` plus cryptic param docs, vs. the rich `get-show-me-options` description. It doesn't tell the agent to call `get-show-me-options` first, to use only an applicable discovered type, or to preserve the **same** worksheet + `fieldsSelectedInSchemaViewer` context used at discovery (the core W-23715530 rule). Mirror the discovery guidance; note no test pins the apply description.

**Minor — contract fixture is projected, not captured.** The openapi fixture + 0.2.15 bump were hand-projected from the producer branch (disclosed in the contract-test provenance) — no live 0.2.15 Desktop capture. The monolith `OpenApiGeneratorTest` and this contract test assert the same shapes (mutual corroboration), but a producer-only drift could slip past both. Worth a live re-capture before GA.

**Cleared (with evidence):** `get-show-me-options` *does* handle a missing 0.2.15 route — the `readHarness` `read()`/`call()` wrapper converts `isRouteMissing → endpointNotInThisBuild`, tested by "gives upgrade guidance and forbids retry when the discovery route is missing." No false-success on async apply (`buildCommandStatus` maps FAILED → `command-failed`). Selection-context parity (ambient / explicit-empty / explicit) holds across the GET and POST paths. `.passthrough()`, TOCTOU-under-`withApplyLock`, and `resolveHttp` fail-closed all reviewed OK.

**Nit:** `destructiveHint:false` on `show-me` is arguable for a viz re-pivot (reversible via undo, so defensible, but reconsider — and no test pins the apply tool's annotations); duplicated worksheet-resolution across the two tools; `show-me.ts` resolves the session/executor twice.

myu404:
Addressed the consumer-side review follow-ups in `039abb65`:

- `show-me` now accepts any non-empty native token returned by discovery and lets Desktop validate support; a regression test proves a future token crosses the MCP boundary and surfaces Desktop's typed 400 response.
- The tool guidance now requires `get-show-me-options` first, an applicable returned type, and the same worksheet, datasource, and selected-field context.
- `destructiveHint` is now `true` and pinned by the metadata test.
- Worksheet lookup and apply now share one `runExternalApiTool` execution, so the session and executor are resolved once while the apply lock covers the full workflow.

The projected OpenAPI fixture still needs a live 0.2.15 Desktop recapture before GA; this commit does not claim that external verification is complete.

b-carver (approved):
(AI reviewed) Approving — the discovery→apply contract changes look good and the earlier review feedback is addressed with tests. Apply's `showMeType` is now an extensible `z.string().min(1)` so the server stays the single source of truth and future native types remain appliable (with a test proving a future type passes zod and the server's 400 surfaces); the `show-me` description now points the agent at `get-show-me-options` and the reuse-the-same-context rule; `destructiveHint` is `true`; and the tool resolves the session once via `runExternalApiTool` while preserving the route-missing → `endpointNotInThisBuild` backstop. CI is green.

PR #890, opened 2026-09-04
fix(desktop): match Tableau nested-shelf readback and innermost sort

tableaukyler:
## Decision

RULE WE NOW KEEP: nested discrete shelf pills that Tableau wraps in grouping parentheses are the same pills we wrote, and omitting `targetField` on `sort_by_field` sorts the innermost pill on that one discrete shelf (Tableau axis-sort). Crosstabs (a dim on Rows and a dim on Cols) still require `targetField`.

## Description

- Worksheet readback no longer reports `apply succeeded but Tableau silently dropped` when Desktop canonicalizes `A / B` as `(A / B)`.
- `refine-worksheet` `sort_by_field` without `targetField` sorts the innermost same-shelf dim instead of refusing nested Group/Country as ambiguous.
- `tabdoc:sort` / `tabdoc:sort-nested` hints send nested sorts to `refine-worksheet`, not cache XML.
- `sortByField` accepts a caption, CI name, or `[ds].[ci]` when that measure CI is already declared.

## Motivation and Context

W-23853108 Beat 2 (Group > Country bars, color winners, sort by Position). Color and sort landed, but apply readback lied about dropped row pills, then `sort_by_field` refused nested axes and the agent fell back to XML surgery. `"Axis; omit to detect."` only detected a single dim.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- `src/desktop/validation/readback-verify.test.ts` (grouping parens vs real drop)
- `src/desktop/refine/refineWorksheet.test.ts` (innermost omit, outer `targetField`, crosstab refuse names axes, CI-ref `sortByField`)
- `src/desktop/guards/commandPolicy.test.ts` and `src/tools/desktop/api/executeTableauCommand.test.ts`
- `./scripts/agent-check` green (lint, tsc, vitest, desktop build, lockstep)

## Related Issues

W-23853108

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/desktop/refine/refineWorksheet.ts
+++ b/src/desktop/refine/refineWorksheet.ts
@@ -34,6 +34,7 @@ import { DOMParser } from '@xmldom/xmldom';
 import * as xpath from 'xpath';
 
 import { escapeXml } from '../binder/escape.js';
+import { decodeXmlEntities } from '../xmlElement.js';
 
 export type TopNEnd = 'top' | 'bottom';
 export type SortDirection = 'ASC' | 'DESC';
@@ -240,6 +241,19 @@ function isDimensionCi(ci: ColumnInstance): boolean {
   return !isPseudoFieldCi(ci) && ci.type === 'nominal' && ci.derivation === 'None';
 }
 
+/**
+ * Discrete shelf sort target — Tableau Field::IsDiscrete() (Nominal|Ordinal), minus
+ * measure aggregations (a discrete SUM is ordinal but is a key, not an axis) and
+ * pseudo-fields. YEAR/MONTH date parts are ordinal + Year/Month, so they count.
+ */
+function isDiscreteSortCi(ci: ColumnInstance): boolean {
+  return (
+    !isPseudoFieldCi(ci) &&
+    (ci.type === 'nominal' || ci.type === 'ordinal') &&
+    !Object.prototype.hasOwnProperty.call(AGG_DERIVATIONS, ci.derivation)
+  );
+}
+
 /** A measure CI: quantitative with a canonical aggregation derivation (e.g. [sum:Sales:qk]). */
 function isMeasureCi(ci: ColumnInstance): boolean {
   return (
@@ -255,7 +269,12 @@ function unbracket(value: string): string {
 }
 
 function normalizeCaption(value: string | undefined): string {
-  return (value ?? '').trim().toLocaleLowerCase();
+  return decodeXmlEntities((value ?? '').trim()).toLocaleLowerCase();
+}
+
+/** Logical `[ds].[ci]` from regex-captured XML (which may still contain `&amp;`). */
+function qualifiedCiRef(ds: string, ciName: string): string {
+  return `[${decodeXmlEntities(ds)}].${decodeXmlEntities(ciName)}`;
 }
 
 function fieldCaptionCandidates(
@@ -606,7 +625,10 @@ function planComputedSortByRefs(
   opts: { targetField: string; column: string; using: string; direction: SortDirection },
 ): SortByFieldPlan | RefineRefusal {
   const node = `<computed-sort column='${escapeXml(opts.column)}' direction='${opts.direction}' using='${escapeXml(opts.using)}' />`;
-  const targetSorts = computedSortTargets(xml).filter((sort) => sort.column === opts.column);
+  const targetColumn = decodeXmlEntities(opts.column);
+  const targetSorts = computedSortTargets(xml).filter(
+    (sort) => decodeXmlEntities(sort.column) === targetColumn,
+  );
   if (targetSorts.length > 1) {
     return refuse(`more than one <computed-sort> present for target field "${opts.targetField}".`);
   }
@@ -710,7 +732,7 @@ export function planSortByField(
     columns,
     opts.targetField,
     'target field',
-    isDimensionCi,
+    isDiscreteSortCi,
     ds,
   );
   if (!('name' in target)) return target;
@@ -724,8 +746,8 @@ export function planSortByField(
   );
   if (!('name' in sortBy)) return sortBy;
 
-  const column = `[${ds}].${target.name}`;
-  const using = `[${ds}].${sortBy.name}`;
+  const column = qualifiedCiRef(ds, target.name);
+  const using = qualifiedCiRef(ds, sortBy.name);
   return planComputedSortByRefs(xml, { targetField: opts.targetField, column, using, direction });
 }
 
@@ -765,7 +787,7 @@ export function planSortByFieldOnCategoricalAxis(
   // DS name (e.g. `P&amp;L Data`) is identical on both sides of the comparison.
   const rows = shelfText(xml, 'rows');
   const cols = shelfText(xml, 'cols');
-  const dims = cis.filter(isDimensionCi);
+  const dims = cis.filter(isDiscreteSortCi);
   const rowDims = orderedShelfDims(rows, ds, dims);
   const colDims = orderedShelfDims(cols, ds, dims);
   const columns = parseColumns(xml);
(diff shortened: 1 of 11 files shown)


src/desktop/refine/refineWorksheet.ts
```diff
@@ -451,6 +454,14 @@ function shelfText(xml: string, shelf: 'rows' | 'cols'): string {
   return xml.match(new RegExp(`<${shelf}\\b[^>]*>([\\s\\S]*?)</${shelf}>`))?.[1] ?? '';
 }
 
+function orderedShelfDims(shelf: string, ds: string, dims: ColumnInstance[]): ColumnInstance[] {
+  return dims
+    .map((ci) => ({ ci, idx: shelf.indexOf(`[${ds}].${ci.name}`) }))
```
myu404, line 478: ### [P1 · REQUIRED] Select the final shelf occurrence

**Purpose**
When `targetField` is omitted, the documented behavior is to sort the actual innermost dimension on the sole discrete shelf.

**Finding**
`orderedShelfDims` walks eligible CI declarations and records only `shelf.indexOf(ref)`. For a valid repeated shelf such as A / B / A, that becomes [A, B], so the planner selects B even though the innermost occurrence is A. It also removes derived/date/User discrete pills before deciding which pill is innermost, allowing an earlier plain dimension to be selected instead of refusing.

**Impact**
`refine-worksheet` can apply and readback-confirm a computed sort on the wrong dimension, then report success for a different axis than the caller requested.

**Recommendation**
Parse the shelf's ordered pill occurrences first, resolve the final discrete pill to a supported CI, and refuse if that exact pill is unsupported or ambiguous. Add repeated-pill and derived/date-inner regression coverage.

— 🤖 Reviewed by MichaelGPT

src/desktop/validation/readback-verify.ts
```diff
@@ -128,10 +129,32 @@ function walkElements(node: unknown, visit: (tag: string, element: XmlRecord) =>
...
+    text = text.slice(1, -1).trim();
+  }
+  return text;
+}
+
 function shelfValues(value: unknown): string[] {
   return normalizeArray(value)
-    .flatMap((item) => textValue(item).split('/'))
-    .map((item) => item.trim())
+    .flatMap((item) => parseShelfValue(unwrapGroupingParens(textValue(item))))
```
myu404, line 169: ### [P1 · REQUIRED] Compare normalized shelves as ordered sequences

**Purpose**
Removing Tableau's grouping parentheses should change only syntax; shelf order and multiplicity must remain part of the authored intent.

**Finding**
After the new normalization splits grouped shelves into pill arrays, verification still checks every intended pill with `readback.includes(value)`. As a result, intended A / B and grouped readback (B / A) compare equal, as do intended A / B / A and grouped readback (A / B). Before this change, the grouping tokens prevented those corrupted grouped readbacks from passing.

**Impact**
The shared post-apply verifier can return `passed` after a hierarchy was reordered or a repeated pill was dropped, so apply, binder, dashboard-batch, and formatting paths may report success for a different rendered shelf.

**Recommendation**
Compare each normalized shelf as an ordered sequence, including length and repeated occurrences, and add grouped reorder and duplicate-drop regression cases.

— 🤖 Reviewed by MichaelGPT

src/desktop/refine/refineWorksheet.ts
```diff
@@ -278,13 +278,16 @@ function resolveCiByCaption(
...
   predicate: (ci: ColumnInstance) => boolean,
+  ds?: string,
 ): ColumnInstance | RefineRefusal {
   const wanted = normalizeCaption(caption);
   if (!wanted) return refuse(`${kind} caption is required.`);
 
   const candidates = cis.filter(predicate);
   const matches = candidates.filter((ci) =>
-    fieldCaptionCandidates(ci, columns).some((value) => normalizeCaption(value) === wanted),
+    [...fieldCaptionCandidates(ci, columns), ci.name, ds ? `[${ds}].${ci.name}` : '']
```
myu404, line 307: ### [P2 · SUGGESTED] Normalize entities before qualified-ref matching

**Purpose**
The new datasource-qualified CI form should accept the exact `column_ref` exposed by field discovery.

**Finding**
The new candidate is built from regex-captured XML attributes, so a datasource such as P&L Data appears as `[P&amp;L Data].[sum:Position:qk]`. Field discovery exposes the decoded `[P&L Data].[sum:Position:qk]`, while `normalizeCaption` only trims and lowercases, so the valid user-facing ref is rejected as unknown.

**Impact**
Workbooks with XML-special characters in datasource or CI names cannot use the advertised qualified-ref form, even when the measure is declared and unambiguous.

**Recommendation**
Compare decoded logical names while retaining the original CI values for mutation, then escape exactly once when emitting XML. Add a qualified-ref case for P&L Data.

— 🤖 Reviewed by MichaelGPT

src/desktop/validation/readback-verify.ts
```diff
@@ -128,10 +129,32 @@ function walkElements(node: unknown, visit: (tag: string, element: XmlRecord) =>
...
 
+function unwrapGroupingParens(value: string): string {
+  let text = value.trim();
+  while (text.startsWith('(') && text.endsWith(')')) {
+    let depth = 0;
+    let wrapsAll = true;
+    for (let i = 0; i < text.length; i++) {
+      const char = text[i];
+      if (char === '(') depth += 1;
+      else if (char === ')') {
```
myu404, line 153: ### [P2 · SUGGESTED] Ignore parentheses inside bracketed field names

**Purpose**
Outer grouping parentheses should be ignored without treating legal identifier characters as grouping syntax.

**Finding**
`unwrapGroupingParens` counts every parenthesis, including those inside bracketed datasource and CI names. An unmatched parenthesis in a legal field name can bring the depth to zero before the real outer close, leaving the wrapper attached to the first and last parsed pills.

**Impact**
Equivalent grouped and ungrouped shelves can compare different, causing a successful apply to be reported as a fatal silently-dropped shelf error.

**Recommendation**
Track bracketed-name state, including escaped `]]`, and count grouping parentheses only outside bracketed references. Add a grouped/unwrapped regression case with an unmatched parenthesis inside a name.

— 🤖 Reviewed by MichaelGPT

myu404 (changes requested):

This change addresses the reported Group > Country workflow by normalizing Tableau's outer shelf parentheses, selecting Country as the inner same-shelf sort target, and steering raw sort commands to `refine-worksheet`. The package and lockfile versions are synchronized at 2.67.7, the focused unit coverage exercises the standard nested/crosstab paths, and all visible GitHub checks are green.

**Required**

- Select the innermost target from actual shelf occurrences; the current first-occurrence projection can sort and confirm the wrong field for repeated or mixed discrete shelves.
- Compare normalized shelves as ordered sequences; one-way membership now lets grouped reordered shelves and duplicate-pill drops pass verification.

**Suggested**

- Normalize XML entities before matching datasource-qualified CI references returned by field discovery.
- Ignore parentheses inside bracketed datasource and CI names when detecting an outer grouping wrapper.

**Coverage and limitations**

The added tests cover the standard Group/Country pair, outer grouping parentheses, a real pill drop, explicit outer targeting, crosstab refusal, simple qualified refs, command hints, and the tool-size ratchet. This review was static: it did not rerun tests or validate the exact Group/Country/Position flow against a live Tableau Desktop instance.

— 🤖 Reviewed by MichaelGPT

src/desktop/refine/refineWorksheet.ts
tableaukyler: I just tested this manually and Tableau actually sorted by B in this case

tableaukyler: Addressed in b5b46c93, with one Desktop-backed exception.

Live Desktop on A / B / A still sorts B (unique first-occurrence nesting), so we did not switch omit-targetField to the trailing repeated A.

The derived/date miss was real: Region / YEAR(Order Date) was sorting Region. sort_by_field now uses isDiscreteSortCi (Nominal|Ordinal, excluding measure aggregations), matching Field::IsDiscrete(), so YEAR is the innermost target. Top-N and simple-bar insert keep the plain nominal+None envelope.

src/desktop/validation/readback-verify.ts
tableaukyler: Addressed in b5b46c93.

Normalized shelves now compare as ordered sequences (index, length, and repeats). Grouped reorder (A / B vs (B / A)) and a dropped repeated pill (A / B / A vs (A / B)) fail readback; grouping-paren-only differences still pass.

src/desktop/refine/refineWorksheet.ts
tableaukyler: Addressed in b5b46c93.

Caption and qualified-ref matching decode XML entities, so field-discovery `[P&L Data].[sum:Sales:qk]` matches XML `P&amp;L Data`. The plan keeps logical names; emission escapeXml's once (no `&amp;amp;`).

src/desktop/validation/readback-verify.ts
tableaukyler: Addressed in b5b46c93.

unwrapGroupingParens now tracks bracketed names (including escaped `]]`) and only counts grouping parentheses outside `[...]`. An unmatched `)` inside a CI name no longer blocks the outer wrapper unwrap.

myu404 (approved):
lgtm, approved!

PR #891, opened 2026-09-04
@W-23757365: admin instructions nudge — don't gate admin-tool use on user restating admin status

Alon-ST-DATA:
## What
Adds one sentence to `ADMIN_INSTRUCTIONS` (the server `initialize` instructions string, advertised only when `ADMIN_TOOLS_ENABLED` is set):

> Do not require the user to state or re-confirm admin status before using these tools — invoke them whenever the task warrants; the server authorizes each call and cleanly rejects non-admins.

## Why
W-23757365: the client's own system-prompt persona makes the model assume it is *not* talking to an admin, so it won't invoke admin tools unless the user re-states their admin role each session. The only server-side lever over model behavior is the `initialize` instructions string. Authorization is already enforced per-call by `assertAdmin` (`src/tools/web/adminGate.ts`), which cleanly rejects non-admins — so the model never needs the user to prove admin status; it can just invoke, and the server decides.

Deliberately **not** worded "if you can see this tool the user IS an admin" — on `main`, admin-tool registration is gated only by `ADMIN_TOOLS_ENABLED`, not by role, so visibility != admin identity. The truthful, always-correct nudge is "just invoke; the server authorizes each call."

## Notes
- Doc/instructions-only; no code or gating change.
- ~30 words, appended once to the handshake instructions — no per-tool description bloat.
- Server-side nudge is best-effort; a strong client persona may still override it. Client-side education may be needed as a follow-up for full closure.

## Tests
- `src/server.web.test.ts`: assert the phrase is present when the flag is on, absent when off (mirrors sibling Markdown-tables assertions).
- Unit 22/22 · lint · build all green.
- Patch bump 4.8.1 → 4.8.2.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableau/mcp-server",
-  "version": "4.8.1",
+  "version": "4.8.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@tableau/mcp-server",
-      "version": "4.8.1",
+      "version": "4.8.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1095.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.8.1",
+  "version": "4.8.2",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/tableau/tableau-mcp.git"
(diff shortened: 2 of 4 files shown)


Akash-Rastogi (approved):
[Akash's agent] LGTM. Instructions-only nudge appended to the admin block (composed only when `ADMIN_TOOLS_ENABLED`) telling the model not to make the user restate admin status before invoking admin tools.

Verified the safety premise against `src/tools/web/adminGate.ts`: `assertAdmin` returns a structured `Ok/Err` `Result` and cleanly rejects non-admins with a plain "requires site administrator permissions" message — so "just invoke; the server authorizes each call" is accurate and authorization is unchanged (nothing gating-related weakened). Good call deliberately *not* equating tool visibility with admin identity (visibility is `ADMIN_TOOLS_ENABLED`-gated, not role-gated).

Tests assert present-when-on / absent-when-off consistently with the sibling nudges; full CI matrix green. No concerns.

PR #892, opened 2026-09-05
@W-23864483: Recover published datasources dropped by the workbook lineage rollup

williamyen0203:
## Description

`get-workbook` was reporting a workbook's live-connected **published** datasource as **embedded**.

The root cause is that we relied on the Metadata API's **content-level `Workbook.upstreamDatasources` rollup**, which is a Catalog-computed field that can return **empty even when the `embedded → published` lineage edge is indexed**. The same published datasource is reliably reachable via `Workbook.embeddedDatasources[].upstreamDatasources`, but we weren't querying that path.

```
Workbook
├── upstreamDatasources: []          ← Catalog rollup — can be empty (the bug)
└── embeddedDatasources: [
      { upstreamDatasources: [ <published DS> ] }   ← indexed, reliable
    ]
```

When the rollup came back empty, the Metadata `published` entry was dropped entirely and only the REST `/connections` sqlproxy stub survived — which reads as `embedded`. That's why a published datasource appeared as embedded.

### Changes

- **`collectPublishedLineage`** (renamed from `normalizeLineageContents`) now unions the content-level `upstreamDatasources` rollup with the datasources reached by traversing `embeddedDatasources → upstreamDatasources`, deduped by `luid` (preserving first-seen order). This recovers published datasources the rollup omits.
- Added a `metadataEmbeddedDatasourceSchema` wire-parse and shared GraphQL selections (`embeddedUpstreamSelection`, `workbookLineageNodesSelection`) so the traversal is applied consistently across the **workbook**, **view**, and **search-content** lineage queries and the two query builders never drift.

### Scope / follow-up

This PR is intentionally limited to **reliably surfacing** published datasources. A live connection to a published datasource can still appear **twice** in the output — once as the REST sqlproxy stub (`embedded`) and once as the recovered published entry (`published`). Collapsing and relabelling those two representations (name-matched de-dupe) is handled by the follow-up PR (**#865**), so the two changes stay independently reviewable and #865 rebases cleanly on top of this one. `getWorkbook.ts` is left unchanged from `main` for that reason.

## Motivation and Context

Step 1 of the `isQueryable` spike (W-23864483): computing an accurate `isQueryable` flag requires published datasources to be surfaced reliably in the first place, before the datasource list can be de-duped and typed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Version Bump

- [x] Patch
- [ ] Minor
- [ ] Major

## How Has This Been Tested?

- New unit tests in `lineageUtils.test.ts` cover the empty-rollup-recovered-via-`embeddedDatasources` case, LUID dedup, and the reconcile/relabel logic.
- New tests in `getWorkbook.test.ts` verify the published-vs-embedded output.
- All unit tests pass locally (36 passed).

### Before

```
    "upstreamDatasources": [
      {
        "luid": "ab830d72-5bbe-4a51-b96c-e05b7b591482",
        "name": "Superstore Datasource",
        "datasourceType": "embedded"
      },
      {
        "luid": "5e19052f-e334-420c-817e-c7fa742dce0c",
        "name": "log_bk",
        "datasourceType": "embedded"
      }
    ]
```

### After

```
    "upstreamDatasources": [
      {
        "luid": "f5b97800-6a7e-4467-8aa1-fb30360edd35",
        "name": "Superstore Datasource",
        "datasourceType": "published"
      },
      {
        "luid": "ab830d72-5bbe-4a51-b96c-e05b7b591482",
        "name": "Superstore Datasource",
        "datasourceType": "embedded"
      },
      {
        "luid": "5e19052f-e334-420c-817e-c7fa742dce0c",
        "name": "log_bk",
        "datasourceType": "embedded"
      }
    ]
```

## Checklist

## Contributor Agreement

      its Contribution Checklist.

### DIFF
--- a/src/sdks/tableau/methods/lineageUtils.test.ts
+++ b/src/sdks/tableau/methods/lineageUtils.test.ts
@@ -3,8 +3,10 @@ import {
   getViewLineageByLuid,
   getViewLineageQuery,
   getWorkbookLineageByLuid,
+  getWorkbookLineageQuery,
   mergeViewLineage,
   mergeWorkbookLineage,
+  reconcileWorkbookDatasources,
   toEmbeddedLineageContents,
 } from './lineageUtils.js';
 
@@ -21,6 +23,46 @@ describe('lineageUtils', () => {
       { luid: 'emb-1', name: 'Embedded DS', datasourceType: 'embedded' },
       { luid: 'emb-2', name: 'emb-2', datasourceType: 'embedded' },
     ]);
+  });
+
+  describe('reconcileWorkbookDatasources', () => {
+    it('reclassifies a connection as published when a published upstream shares its name', () => {
+      // The sqlproxy connection (emb-superstore) and the published datasource (pub-superstore) are
+      // the same datasource; emit one entry keyed on the connection LUID, tagged published.
+      const result = reconcileWorkbookDatasources(
+        [
+          { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'embedded' },
+          { luid: 'emb-logbk', name: 'log_bk', datasourceType: 'embedded' },
+        ],
+        [{ luid: 'pub-superstore', name: 'Superstore Datasource' }],
+      );
+
+      expect(result).toEqual([
+        { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'published' },
+        { luid: 'emb-logbk', name: 'log_bk', datasourceType: 'embedded' },
+      ]);
+    });
+
+    it('appends a published upstream that no connection accounts for', () => {
+      const result = reconcileWorkbookDatasources(
+        [{ luid: 'emb-1', name: 'Embedded DS', datasourceType: 'embedded' }],
+        [{ luid: 'pub-1', name: 'Published DS' }],
+      );
+
+      expect(result).toEqual([
+        { luid: 'emb-1', name: 'Embedded DS', datasourceType: 'embedded' },
+        { luid: 'pub-1', name: 'Published DS', datasourceType: 'published' },
+      ]);
+    });
+
+    it('leaves connections embedded when there is no published lineage', () => {
+      const result = reconcileWorkbookDatasources(
+        [{ luid: 'emb-1', name: 'Embedded DS', datasourceType: 'embedded' }],
+        [],
+      );
+
+      expect(result).toEqual([{ luid: 'emb-1', name: 'Embedded DS', datasourceType: 'embedded' }]);
+    });
   });
 
   it('parses and merges upstream workbook lineage', () => {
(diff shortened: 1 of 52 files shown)


ycath (approved):

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -349,12 +373,56 @@ export function toEmbeddedLineageContents(
...
+  publishedUpstream: Array<LineageContent>,
 ): Array<LineageContent> {
-  return (contents ?? [])
-    .filter((content): content is { luid: string; name?: string | null } => !!content.luid)
-    .map((content) => ({ luid: content.luid, name: content.name ?? content.luid }));
+  const publishedNames = new Set(publishedUpstream.map((ds) => ds.name));
+  const connectionNames = new Set(connectionDatasources.map((ds) => ds.name));
+
+  const reconciled: Array<LineageContent> = connectionDatasources.map((ds) => ({
+    ...ds,
```
ycath: This just retains the connection's internal sqlproxy LUID and not the published datasource's real LUID. Is this ID useable in the same way you would query a published datasource LUID?

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -164,7 +185,10 @@ export function getViewLineageByLuid(response: unknown): Map<string, ViewLineage
     nodes.map((node) => [
       node.luid,
       {
-        upstreamDatasources: normalizeLineageContents(node.upstreamDatasources),
+        upstreamDatasources: collectPublishedLineage(
```
ycath: if this doesn't use `reconcileWorkbookDatasources` like get-workbook does, wouldn't it not return the datasource type (published/embedded)?

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -349,12 +373,56 @@ export function toEmbeddedLineageContents(
...
 ): Array<LineageContent> {
-  return (contents ?? [])
-    .filter((content): content is { luid: string; name?: string | null } => !!content.luid)
-    .map((content) => ({ luid: content.luid, name: content.name ?? content.luid }));
+  const publishedNames = new Set(publishedUpstream.map((ds) => ds.name));
+  const connectionNames = new Set(connectionDatasources.map((ds) => ds.name));
+
+  const reconciled: Array<LineageContent> = connectionDatasources.map((ds) => ({
+    ...ds,
+    datasourceType: publishedNames.has(ds.name) ? ('published' as const) : ('embedded' as const),
```
ycath: How do we handle ambiguous names? An embedded DS sharing a name with an unrelated published upstream would be mislabeled 'published' or if there are duplicate names. Maybe we can add a guard to not reclassify when a name maps ambiguously or a comment documenting the accepted risk.

src/sdks/tableau/methods/lineageUtils.ts
```diff
@@ -349,12 +373,56 @@ export function toEmbeddedLineageContents(
...
-function normalizeLineageContents(
-  contents: Array<z.infer<typeof metadataLineageContentSchema>> | null | undefined,
+// A workbook's live connection to a published datasource surfaces twice: once as the embedded
+// "sqlproxy" stub in REST /connections (whose datasource.id is the workbook's internal,
+// VDS-queryable LUID) and once as the published datasource reached via the Metadata API. They are
+// the same logical datasource, so we emit a single entry keyed on the connection's LUID and
+// reclassify it as 'published' when a Metadata published upstream shares its name. The two LUIDs
+// cannot be joined directly (the Metadata node id != the REST datasource.id), so name is the only
+// available bridge. Published upstreams that no connection accounts for (e.g. the /connections call
+// failed, or the names diverge) are appended with their Metadata LUID so they are not lost.
```
ycath: nit: can we shorten this and other block comments to be a little more concise?

src/sdks/tableau/methods/lineageUtils.test.ts
```diff
@@ -23,6 +25,46 @@ describe('lineageUtils', () => {
...
+        [
+          { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'embedded' },
+          { luid: 'emb-logbk', name: 'log_bk', datasourceType: 'embedded' },
+        ],
+        [{ luid: 'pub-superstore', name: 'Superstore Datasource' }],
+      );
+
+      expect(result).toEqual([
+        { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'published' },
+        { luid: 'emb-logbk', name: 'log_bk', datasourceType: 'embedded' },
```
ycath: So in this case, the result ` { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'published' }` emits the embedded ds luid instead of 'pub-superstore', but there are other published datasources that don't have an embedded connection (like the test below) where the luid emitted is the pds luid. I'm worried that other tools would see `datasourceType: 'published'` and try to use the luid as if it were a pds (query-datasource would work but some other tools like list-datasources or getting permissions on a pds would not on an embedded luid).

For this case where the sqlproxy connection matches a pds, can we label it as `embedded` instead of `published` to match the type of luid that is returned? Then in my current PR to emit published parent, I would extend the ds entry with the published parent ds and its luid ` { luid: 'emb-superstore', name: 'Superstore Datasource', datasourceType: 'embedded', publishedParent: {luid: 'pub-superstore', name: 'SuperstoreDatasource'}}`.

src/sdks/tableau/methods/lineageUtils.ts
williamyen0203: This part is also moved to the follow up PR: https://github.com/tableau/tableau-mcp/pull/865/changes

src/sdks/tableau/methods/lineageUtils.test.ts
williamyen0203: Now, there will be separate entries for embedded and published datasources. If the embedded datasource is backed by a published datasource, it'll be two separate entries.

This will be left to be addressed in your PR: https://github.com/tableau/tableau-mcp/pull/865/changes

src/sdks/tableau/methods/lineageUtils.ts
williamyen0203: `collectPublishedLineage` returns `{luid, name}` with no type, same as `normalizeLineageContents`. There shouldn't be a regression here.

PR #893, opened 2026-09-05
@W-24106279: Fix query-admin-insights selecting a broken duplicate Admin Insights datasource

Alon-ST-DATA:
## Summary

Fixes a bug where `query-admin-insights` resolved its backing Admin Insights datasource **by name only**. On any site with cloned Admin Insights content — or a second project named "Admin Insights" — name resolution was ambiguous and could select a **broken duplicate** datasource with a dead Hyper extract, failing every call with `[SQLSTATE:08001] ... 0x9EE5C2F0`. Reproduced live on a prod site with a 2022 clone of the starter content.

Root cause (`src/tools/web/adminInsights/resolver.ts`): last-writer-wins name match, an over-broad `projectName:eq:Admin Insights` filter (matches every project with that display name), and a positive cache that stored every `name→id` pair (letting a wrong LUID poison an unrelated dataset name for the TTL).

## What changed

Detect-then-disambiguate resolver:
- **Single-candidate fast path** — unchanged behavior, no extra REST calls (the common case).
- **On ambiguity** — narrow to the top-level "Admin Insights" project (no `parentProjectId`, oldest `createdAt`), then deterministically score survivors on `isCertified`, canonical `contentUrl` (no `_<epoch>` suffix), rich description, and owner == "Tableau System Account" (non-enumerable system account, corroborated by a 404 on owner lookup).
- **Runtime health-check fallback** — on a Hyper connection error, negative-cache the dead LUID, evict it, and retry the next-ranked candidate.
- **Cache-poisoning fix** — cache only the winning LUID.
- **Diagnostic warning** — `ADMIN_INSIGHTS_AMBIGUOUS_DATASOURCE` surfaces the chosen LUID + candidates + reason, telling admins to delete the duplicate.
- **Site-overridable `ADMIN_INSIGHTS_DATASET_LUIDS`** pin map, and a default-on `admin-insights-robust-resolver` feature flag.

Covers all 9 Admin Insights datasources and all 5 `query-admin-insights` kinds (shared resolver).

## Review & verification

- **Self-review** performed (tmcp-reviewer) — verdict *ship-with-fixes*, 0 blockers. The flagged findings were addressed in a follow-up commit: added the `ADMIN_INSIGHTS_DATASET_LUIDS` row to `site-settings.md`, added `overridableConfig` parser tests, narrowed the owner-lookup `catch` to 404-only (transient errors no longer mis-score), and replaced unsafe `z.coerce.boolean()` with a safe `tableauBoolean` parse for `isCertified` / `topLevelProject`.
- **Unit**: 178 files, 2910 tests pass. `tsc --noEmit`, lint, and build clean.
- **Live e2e** (prod site with the real duplicate): `site-content`, `ts-events`, `ts-users`, and `stale-content` all PASS — resolver selects the healthy certified datasource, never the dead duplicate; ambiguity warning fires; no Hyper error.

Patch version bump `4.8.2 → 4.8.3` (bug fix).

Work item: @W-24106279

---

## Update — simplification pass (commit `c446aa57`)

Reviewer feedback: the first cut was over-engineered. Trimmed to fact-based detection only. **16 files, +31 / −362 (−331 LOC).**

- **Removed the `ADMIN_INSIGHTS_DATASET_LUIDS` override** (overridable-config map + parser + tests + docs + bundle-manifest + `process-env.d.ts`). Resolution is now purely fact-based — no admin-pinned LUIDs.
- **Removed the `admin-insights-robust-resolver` feature flag** and the legacy last-writer-wins branch it gated (its only "off" state was the known-buggy behavior). Detect-then-disambiguate + cache-poisoning fix are now unconditional. Single-candidate **fast path unchanged** (one `listDatasources` call, no extra project/user calls).
- **Deleted `types/tableauBoolean.ts`** — inlined a compact `z.preprocess` boolean into `dataSource.ts` and `project.ts` (still needed: Tableau REST returns string `"false"`, and `z.coerce.boolean()` wrongly maps it to `true`; `isCertified` / `topLevelProject` remain live scoring signals).

Kept intact: cache-poisoning fix, `mcp.warnings` diagnostic naming the chosen LUID + reasons, Hyper-`08001` health-check fallback + dead-LUID negative cache.

**Self-review** (tmcp-reviewer) on the simplified diff: verdict *ship* — 0 blockers, 0 majors. `tsc`/lint/build/docs clean, 179/179 unit tests pass, version stays `4.8.3`.

### Known limitation
Detection is now fact-based with no operator override. On a pathological site where the original datasource was deleted (or lost its clean `contentUrl` slug **and** certification) leaving only clones, scoring can mis-pick — recourse is deleting the duplicate. The ambiguity/health warnings still surface the candidates in that case.


Alon-ST-DATA:
## Live e2e verification (prod site `tableausalesdemo`, pod 10az — the ambiguous site)

Ran `query-admin-insights` → **stale-content** report (90+ days, Workbooks + Datasources) against the site that has the duplicate "Site Content" datasource. Confirms the fix resolves the **true canonical** datasource, not the broken clone.

```
Stale Content Report — tableausalesdemo
Threshold: 90+ days · Types: Workbooks + Datasources

Total stale items : 19,566
Total stale size  : ~246.5 GB (264,638,371,057 bytes)
Rows returned     : 0 — withheld (ROW_CAP_EXCEEDED, cap 100)
```

**Resolver behavior — as designed:**

- ✅ Picked the **certified canonical** datasource `5e516f40-…273c` (System-Account-owned) — **not** the dead-extract duplicate. No `[SQLSTATE:08001] … 0x9EE5C2F0` Hyper crash, which was the original prod failure.
- ✅ Emitted the diagnostic warning:
  > `ADMIN_INSIGHTS_AMBIGUOUS_DATASOURCE` — multiple published datasources named "Site Content". Picked certified canonical `5e516f40-…273c`. Delete non-canonical duplicate(s) to clear.

`ROW_CAP_EXCEEDED` (19,566 ≫ `STALE_CONTENT_MAX_ROWS=100`) is the unrelated pre-existing safety guard, not a resolver issue.

This is the exact scenario the PR targets: on a site with a cloned "Admin Insights" project + duplicate datasource, the tool now deterministically selects the original/canonical datasource and surfaces the duplicate for cleanup.

src/sdks/tableau/types/dataSource.ts
```diff
@@ -3,11 +3,25 @@ import { z } from 'zod';
 import { projectSchema } from './project.js';
 import { tagsSchema } from './tags.js';
 
+// Tableau REST can deliver booleans as the strings "true"/"false"; `z.coerce.boolean()` would map
+// "false" -> true (the `Boolean("false") === true` footgun) and mis-credit an uncertified clone as
+// certified. Parse explicitly instead.
+const tableauBoolean = z.preprocess(
```
Akash-Rastogi, line 9: [Akash's agent] Non-blocking: this `tableauBoolean` preprocess is the fix at the heart of the PR — it exists precisely because `z.coerce.boolean()` maps `"false"→true`. But no test feeds strings through `dataSourceSchema`/`projectSchema`; the resolver/tool-base tests use literal booleans, which bypass Zod entirely. So a future revert to `z.coerce.boolean()` would silently reintroduce the mis-certification bug with every test still green. Worth a small schema-level test asserting `"false"→false`, `"true"→true`, `"True"→true`, `""→false`, boolean passthrough, and `undefined` optionality — and mirror it for `project.ts` `topLevelProject`.

Akash-Rastogi (approved):
[Akash's agent] LGTM — approving the code. Reviewed in depth (direct read + an adversarial code-review pass + structural sweeps for dangling refs and signature breaks): **no blockers or majors.**

Verified:
- **Health-check retry loop is bounded** — `MAX_FALLBACK_ATTEMPTS = 6` plus the `triedLuids` progress guard; the all-candidates-dead case terminates cleanly with `AdminInsightsUnavailableError`. No infinite-loop path.
- **Cache-poisoning fix is correct** — only the winning LUID is cached; dead-LUID eviction is keyed so it never evicts a healthy winner.
- **`tableauBoolean` correctly fixes the `"false"→true` coercion footgun** (conservative, non-throwing) — combined with the strict `=== true` checks, an uncertified clone scores 0.
- **No injection** (filter strings interpolate only the constant project name + an enum dataset name); `assertAdmin` still gates all five kinds; the intra-PR override-map/feature-flag removal left zero dangling refs.

One **non-blocking** suggestion inline (the boolean fix has no schema-level regression test).

⚠️ Merge gate (not a code issue): `check-title` is failing — the title is `[W-24106279]`, but the repo's check requires the `@W-24106279:` form. Retitle before merge to turn it green.

src/sdks/tableau/types/dataSource.ts
Alon-ST-DATA: Good catch — added in 8d5526fd. Two schema-level test files feed raw values through the real exported schemas (`dataSourceSchema` / `projectSchema`) via `safeParse`, rather than testing the preprocess in isolation (`tableauBoolean` isn't exported, and the schemas are directly constructable):

- `src/sdks/tableau/types/dataSource.test.ts` — `isCertified`
- `src/sdks/tableau/types/project.test.ts` — `topLevelProject`

Each field asserts: `"false"→false` (the critical guard against the `z.coerce.boolean()` / `Boolean("false")===true` footgun), `"true"→true`, `"True"→true` (case-insensitive — `.trim().toLowerCase()` makes it real, verified against the impl, not aspirational), `""→false`, JS-boolean passthrough, and omitted→`undefined` (`.optional()`). A revert to `z.coerce.boolean()` now turns these red instead of staying silently green.

Test-only, no production change. 12 new tests pass; full suite 2911/2911.

PR #894, opened 2026-09-08
@W-23757363: clarify Tableau auth/permission errors across all auth modes

Alon-ST-DATA:
## Problem (W-23757363)

When a Tableau MCP server holds a bad/expired credential, or the request reaches the wrong server in a multi-server setup, the failure surfaced as a bare `Request failed with status code 401` — or the server **crashed at startup** while fetching site settings. Models paraphrased both into a misleading *"feature not configured"* message, making a **401 auth failure indistinguishable from Admin Insights genuinely not being provisioned**.

## Fix

Classify raw thrown `401`/`403` REST errors in the shared web-tool error path (`WebTool.logAndExecute`) into clear, self-explanatory guidance that names the targeted **site** and **pod**. Because it lives in the shared path, it covers **every auth mode** — PAT, OAuth, direct-trust/connected-app, UAT, passthrough. Typed `McpToolError`s (e.g. `AdminOnlyError`) bypass reclassification and keep their curated text.

| File | Change |
|---|---|
| `src/utils/authErrorMessage.ts` (new) | Single source for 401/403 wording (`buildAuthenticationErrorMessage` / `buildPermissionErrorMessage`) |
| `src/tools/web/tool.ts` | Reclassify raw thrown 401/403 in the catch; other paths byte-for-byte unchanged |
| `src/utils/mcpSiteSettings.ts` | A startup 401 **no longer crashes** the server — logs guidance, continues with defaults (matches OAuth) |
| `src/server/oauth/authMiddleware.ts` | Enrich the OAuth 401 challenge `error_description` with the same guidance (`WWW-Authenticate` unchanged) |
| `src/tools/web/adminInsights/resolver.ts` | De-conflate the not-provisioned 404 message from the site-admin-role hint |
| `docs/` | New `multiple-servers.md` decoder table + edits to auth/oauth/admin-insights/intro |

Version: **4.8.2 → 4.8.3** (patch).

## Scope & risk

- **Scope:** all web tools + all auth modes (not admin-only). Only `resolver.ts` is admin-insights-specific.
- **Risk: low / contained.** Reclassification fires only when the error was *thrown* (not `Result.Err`), is *not* an `McpToolError`, and status is *exactly* 401/403. Success paths, `ZodiosValidationError`, and all typed errors are unchanged.
- **403 wording** deliberately does **not** assert "not a missing feature" — Tableau REST can return 403 for a disabled capability too, so the message allows for role/permission *or* capability-not-enabled.

## Tests

- Full unit suite **2903 pass**. New: `authErrorMessage.test.ts`, `authMiddleware.test.ts`; extended `tool.test.ts`, `mcpSiteSettings.test.ts`, `resolver.test.ts` (raw 401, raw 403, curated-403 passthrough, non-auth 503 passthrough, telemetry error_code, startup-401-degrades, 404 de-conflation).
- Live e2e on site `kiwimoussetdp2`: 401-at-call, 401-at-startup (no crash), 404 not-provisioned, OAuth transport 401 — all verified.

## Known follow-up (out of scope, noted intentionally)

The OAuth `invalid_token` 401 path in `authMiddleware.ts` returns 401 + enriched `error_description` but does **not** set a `WWW-Authenticate` header (only the no-token 401 and insufficient-scope 403 do). Pre-existing; RFC 6750 suggests it should. Left for a separate change to keep this PR scoped to W-23757363.

### DIFF
--- a/docs/docs/configuration/mcp-config/authentication/oauth.md
+++ b/docs/docs/configuration/mcp-config/authentication/oauth.md
@@ -25,6 +25,11 @@ All Tableau auth modes are *site scoped*. If you are a multi-site user, and you 
 - **Problem:** After a user has already connected to the hosted Tableau MCP server, disconnecting and reconnecting may not trigger the OAuth flow again.
 - **Workaround:** Sign out of Tableau Cloud, then trigger the Tableau MCP OAuth flow again. If the issue persists, clear your browser cookies and try again.
 
+**Issue 3: Multiple configured servers — a tool call may target the wrong or unauthenticated server**
+
+- **Problem:** When more than one Tableau MCP server is configured, the AI client may send a tool call to a different server than you intended — including one that is not authenticated. The resulting `401` can be summarized by the AI as a vague "feature not configured" message.
+- **Workaround:** Name each server distinctly, confirm which server answered before trusting admin results, and read the raw error. See [Running Multiple Servers & Diagnosing Auth Errors](../multiple-servers.md).
+
 ### Self-hosted Tableau MCP
 
 Tableau Cloud customers can self-host Tableau MCP. A full guide can be found at
--- a/src/server/oauth/authMiddleware.ts
+++ b/src/server/oauth/authMiddleware.ts
@@ -3,6 +3,7 @@ import { NextFunction, RequestHandler, Response } from 'express';
 
 import { getConfig } from '../../config.js';
 import { log } from '../../logging/logger.js';
+import { OAUTH_AUTH_CHALLENGE_GUIDANCE } from '../../utils/authErrorMessage.js';
 import { getToolNameFromRequestBody } from '../requestUtils.js';
 import { AccessTokenValidator } from './accessTokenValidator.js';
 import {
@@ -68,7 +69,10 @@ export function authMiddleware(accessTokenValidator: AccessTokenValidator): Requ
         )
         .json({
           error: 'unauthorized',
-          error_description: 'Authorization required. Use OAuth 2.1 flow.',
+          // Keep the WWW-Authenticate challenge (above) intact so the re-auth flow still works, but
+          // enrich the human-readable description with the shared multi-server guidance so a missing
+          // or expired token is not misread as a missing feature (W-23757363).
+          error_description: `Authorization required. Use the OAuth 2.1 flow to authenticate. ${OAUTH_AUTH_CHALLENGE_GUIDANCE}`,
         });
       return;
     }
@@ -98,7 +102,9 @@ export function authMiddleware(accessTokenValidator: AccessTokenValidator): Requ
       });
       res.status(401).json({
         error: 'invalid_token',
-        error_description: result.error,
+        // Same shared guidance for an invalid/expired token so it is not misread as a missing
+        // feature (W-23757363). The underlying validation detail is preserved first.
+        error_description: `${result.error}. ${OAUTH_AUTH_CHALLENGE_GUIDANCE}`,
       });
       return;
     }
--- a/src/utils/authErrorMessage.ts
+++ b/src/utils/authErrorMessage.ts
@@ -1,0 +1,65 @@
+/**
+ * Shared, self-explanatory guidance for Tableau MCP authentication (401) and permission (403)
+ * failures.
+ *
+ * Centralizing the wording here keeps the message identical across every auth mode — PAT, OAuth,
+ * direct-trust / connected-app, UAT, and passthrough — and across every surface that can emit an
+ * auth failure:
+ *   - the tool-execution error wrapper (src/tools/web/tool.ts),
+ *   - the startup site-settings fetch (src/utils/mcpSiteSettings.ts), and
+ *   - the OAuth transport challenge (src/server/oauth/authMiddleware.ts).
+ *
+ * W-23757363: a bare "Request failed with status code 401" (or a server that failed to start while
+ * fetching site settings) was being paraphrased by the model into a misleading "feature not
+ * configured" message. Naming the cause — and, where known, the targeted site + pod — makes a 401
+ * impossible to misread as "Admin Insights is missing."
+ */
+
+type AuthTarget = { site?: string; server?: string };
+
+/**
+ * Renders the ` (site "X", pod "Y")` clause, omitting whichever part could not be determined.
+ * Returns an empty string when neither is known (e.g. before sign-in), so the sentence still reads.
+ */
+export function describeAuthTarget({ site, server }: AuthTarget): string {
+  const parts: string[] = [];
+  if (site) {
+    parts.push(`site "${site}"`);
+  }
+  if (server) {
+    parts.push(`pod "${server}"`);
+  }
+  return parts.length > 0 ? ` (${parts.join(', ')})` : '';
+}
+
+/** Clear guidance for a downstream Tableau REST 401 (missing / invalid / expired credentials). */
+export function buildAuthenticationErrorMessage(target: AuthTarget = {}): string {
+  return (
+    'Authentication failed (401): the credentials for this Tableau MCP server' +
+    `${describeAuthTarget(target)} are missing, invalid, or expired. ` +
+    'If multiple Tableau MCP servers are configured, verify the request targeted the intended ' +
+    'server, then re-authenticate (renew the personal access token or re-run the OAuth sign-in). ' +
+    'A 401 is an authentication problem, not a missing feature.'
+  );
+}
+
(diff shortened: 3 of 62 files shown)


Akash-Rastogi (commented):
[Akash's agent] Not approving yet — two merge blockers:

1. **`build (24.x)` is red** (see inline on `authMiddleware.ts`): the enriched OAuth `error_description` broke two exact-match assertions in `tests/oauth/embedded-authz/oauth.test.ts` (lines 71 and 761) that this PR didn't update. Deterministic, not a flake.
2. **`check-title` is failing** — the title is `W-23757363:`, but the repo's check requires the `@W-23757363:` form. Please retitle.

The classification approach and `authErrorMessage.ts` wording look sound, and the reclassification in `tool.ts` is correctly scoped to raw, non-`McpToolError`, exactly-401/403 errors — so other tools' error paths are unchanged. Happy to approve once the build is green and the title is fixed.

src/server/oauth/authMiddleware.ts
```diff
@@ -68,7 +69,10 @@ export function authMiddleware(accessTokenValidator: AccessTokenValidator): Requ
         )
         .json({
           error: 'unauthorized',
-          error_description: 'Authorization required. Use OAuth 2.1 flow.',
+          // Keep the WWW-Authenticate challenge (above) intact so the re-auth flow still works, but
+          // enrich the human-readable description with the shared multi-server guidance so a missing
+          // or expired token is not misread as a missing feature (W-23757363).
+          error_description: `Authorization required. Use the OAuth 2.1 flow to authenticate. ${OAUTH_AUTH_CHALLENGE_GUIDANCE}`,
```
Akash-Rastogi, line 75: [Akash's agent] This enriched `error_description` (and the `invalid_token` one at line 107) is what turns `build (24.x)` red: the embedded-authz suite asserts the old text byte-for-byte and wasn't updated in this PR —
- `tests/oauth/embedded-authz/oauth.test.ts:71` still `toEqual`s `error_description: 'Authorization required. Use OAuth 2.1 flow.'`
- `tests/oauth/embedded-authz/oauth.test.ts:761` still `toEqual`s `error_description: 'Invalid or expired access token'`

That suite runs under `test:oauth:embedded` (not part of `npm test`), so a unit-only run stays green while CI fails. Please update both assertions — ideally to `toMatchObject({ error })` + `expect(body.error_description).toContain(...)` so the description text isn't pinned byte-for-byte again.

Alon-ST-DATA: Fixed in 1cb67409. Both stale assertions in `tests/oauth/embedded-authz/oauth.test.ts` now use `toMatchObject({ error })` + `toContain(...)` instead of pinning `error_description` byte-for-byte:

- **unauthenticated (line ~71):** `toMatchObject({ error: 'unauthorized' })`, then `error_description` `toContain('Authorization required')` and `toContain('A 401 is an authentication problem, not a missing feature.')`
- **invalid/expired token (line ~761):** `toMatchObject({ error: 'invalid_token' })`, then `error_description` `toContain('Invalid or expired access token')` and `toContain('A 401 is an authentication problem, not a missing feature.')`

Grepped the file — only these two pinned the changed strings; the other `error_description` assertions (redirect-URI / client-ID mismatch, refresh-token) are for endpoints this PR doesn't touch, left as-is. Both targeted tests pass under CI-equivalent OAuth env; full unit suite 2903 pass; tsc + lint clean.

Good catch on the `test:oauth:embedded` gap — it's not in `npm test`, which is why the unit-only run stayed green while CI went red.

Akash-Rastogi (approved):
[Akash's agent] LGTM — approving. Both blockers from my prior review are cleared:

- **Build green** — the two embedded-authz assertions now use `toMatchObject({ error })` + `error_description.toContain(...)` (commit `1cb67409`), the robust fix that no longer pins the enriched description byte-for-byte. `build (24.x)` and `build (>=22.7.5 <23)` both pass.
- **Title fixed** to `@W-23757363:` — `check-title` green.

The classification logic I reviewed last round is unchanged and sound: `authErrorMessage.ts` wording is clear (403 correctly hedges role/permission-or-capability and says "don't re-authenticate"), and the raw-401/403 reclassification in `tool.ts` is scoped to non-`McpToolError` errors only, so other tools' error paths are untouched. Full CI matrix green. Nice work.

docs/docs/configuration/mcp-config/authentication/oauth.md
```diff
@@ -25,6 +25,11 @@ All Tableau auth modes are *site scoped*. If you are a multi-site user, and you
 - **Problem:** After a user has already connected to the hosted Tableau MCP server, disconnecting and reconnecting may not trigger the OAuth flow again.
 - **Workaround:** Sign out of Tableau Cloud, then trigger the Tableau MCP OAuth flow again. If the issue persists, clear your browser cookies and try again.
 
+**Issue 3: Multiple configured servers — a tool call may target the wrong or unauthenticated server**
+
+- **Problem:** When more than one Tableau MCP server is configured, the AI client may send a tool call to a different server than you intended — including one that is not authenticated. The resulting `401` can be summarized by the AI as a vague "feature not configured" message.
```
stephendeoca, line 30: Why is it that a 401 Unauthorized expected with this issue? When there are multiple servers connected, a client does not user credentials from one server when calling the other right? It uses the credentials that pertains to the connection. So it shouldn't be 401 Unauthorized, it would be more like 404 Not Found when you specify a target. For example, if I ask it to get me metadata for a Datasource belonging to MCP server A, but instead it calls MCP server B. The request will be authorized right, just fail with 404 because the datasource was not found.

Alon-ST-DATA: I hear your point @stephendeoca, but this can be debatable.  You can look at it as we have an authenticated user  (for a session on server A)  that cannot see resources because no authorization on target server B... catching this error with proper logging.
For me, having 404 error hide the real auth error.

stephendeoca: let's genericize this saying that it make call a different server than intended and result in unintended or erroneous behavior. you might tell it to list datasources, and it calls the wrong server but it still succeeds. or it can search for something on the wrong server and not find it. I wouldn't necessarily limit this experience to 401

src/utils/authErrorMessage.ts
```diff
@@ -0,0 +1,65 @@
...
+    'permissions, or confirm the capability is enabled for this site.'
+  );
+}
+
+/**
+ * Guidance appended to the OAuth transport challenge (a 401 returned before any tool runs, where
+ * the targeted site / pod are not yet known). Kept alongside — never in place of — the
+ * `WWW-Authenticate` challenge so the re-authentication flow still works.
+ */
+export const OAUTH_AUTH_CHALLENGE_GUIDANCE =
```
stephendeoca, line 60: We should remove this since if we're hitting 401 with multiple servers, it should mean we're hitting a 401 if that server was standalone. For Clients connected to MCP servers A & B, they will not use credentials from server A when calling server B. So this 401 is the result of not being authenticated to server B, which is independent of there being multiple servers.

docs/docs/configuration/mcp-config/authentication/oauth.md
Alon-ST-DATA: Done in 30182799. Genericized Issue 3 (and the sibling `multiple-servers.md`) to the wrong-server-routing failure mode rather than pinning it to a 401. It now calls out that the outcome varies: the call may **still succeed** against the wrong server and return another site's data (e.g. "list datasources" silently lists a different server's datasources), a search may run against the wrong server and **find nothing** (false negative), or an unauthenticated target may return a `401` the AI summarizes as "feature not configured" — with an explicit note that a `401` is only one of several possible outcomes, not the defining one.

src/utils/authErrorMessage.ts
Alon-ST-DATA: Done in 30182799. Removed the multi-server clause from `OAUTH_AUTH_CHALLENGE_GUIDANCE`. The constant is now:

> This 401 means the request reached a Tableau MCP server whose session is missing, invalid, or expired — re-authenticate. A 401 is an authentication problem, not a missing feature.

Agreed the 401 is independent of there being multiple servers — a standalone server would 401 the same way. Kept the site/pod naming and the "authentication problem, not a missing feature" line.

Heads-up: `buildAuthenticationErrorMessage` (line ~40, the non-OAuth REST path) still carries its own "if multiple servers are configured, verify the request targeted the intended server" sentence. I left it since it wasn't flagged and it reads as an actionable *tip* rather than attributing the *cause* of the 401 to multi-server — happy to trim it too if you'd like consistency.

stephendeoca (approved):

PR #895, opened 2026-09-08
feat(desktop): validate calc formulas before authoring, and pretty-print them in the .twb

artbeckTableau:
## Decisions
Two rules the codebase now keeps for every authored calculated field — not one-offs for a single tool.

**1. Validate every formula against Desktop before writing it, and order a batch by its in-batch dependencies.** `author-calc` (and any future batch tool built on the same core) now routes through `authorCalculationsWithValidation`, which:
- issues the External Client API command `tabdoc:get-calc-details-pres-model-for-formula` per calc — a completed command whose out-param carries an `errorMsgs` list; non-empty means the formula is invalid, and the calc is reported as `invalid-formula` without any apply happening;
- topologically layers a batch by `[Caption]` references so a calc that references another in the same batch is validated *and* created after its dependency exists (Kahn's algorithm; unresolvable calcs come back tagged `cycle`);
- cascades a per-calc failure: any dependent of an invalid or cycled calc is reported as `dependency-failed`, never attempted.

This is a real behavior change for `author-calc`: it used to write straight into the .twb via `authorCalculationsInWorkbook` and rely on the final apply/readback to catch bad formulas — one bad calc could poison an apply, and there was no way to know *which* formula was wrong without eyeballing the readback. Now the validator command judges the formula before any splice.

**2. Pretty-print every authored formula in the .twb attribute.** `renderCalculationColumn`'s `formula='…'` attribute is built with `prettyPrintFormula(formula)` instead of `escapeXml(formula)`. `prettyPrintFormula` formats the tokens onto multiple lines *and* encodes for a single-quoted XML attribute (including `&#13;&#10;` for real newlines, so the parser doesn't normalize them away). A one-shot rule that covers both call sites — batch validator and single-calc path — because both go through `renderCalculationColumn`.

## Why the pretty-print refinements
Wiring the formatter alone wasn't enough — one authored formula surfaced two layout misses:

```
IF [Margin YTD Δpp] >= 0 THEN "▲ " ELSE "▼ " END
  + STR(ROUND(ABS([Margin YTD Δpp]) * 100, 1))
  + " pp vs LY"
```

- The inner IF's flat form is 48 chars — under the 60-column budget — so the "fits, keep flat" rule kept it inline even though it sits inside a much wider concatenation. Fix: force-break an IF/CASE when it's embedded in a larger expression; a standalone one-line IF that is the whole formula still stays flat.
- After breaking an IF, the trailing `+` used to hug the same line as `END`, visually re-merging the block. Fix: after emitting a broken IF/CASE, if any tokens follow, start them on a fresh line.

## What changed
| File | Change |
| --- | --- |
| `src/tools/desktop/authoring/datasource/authorCalcCore.ts` | New `authorCalculationsWithValidation` — per-calc validator call, dependency layering, cascading failure. `renderCalculationColumn` now calls `prettyPrintFormula(formula)` for the attribute. |
| `src/tools/desktop/authoring/datasource/authorCalc.ts` | `author-calc` calls `authorCalculationsWithValidation` instead of `authorCalculationsInWorkbook`. |
| `src/tools/desktop/authoring/datasource/prettyPrintFormula.ts` | Force-break IF/CASE inside a larger expression; start any code that follows a broken IF on a new line. |
| `src/tools/desktop/authoring/datasource/prettyPrintFormula.test.ts` | Regression tests for both refinements. |
| `src/desktop/externalApi/discovery.test.ts`, `src/scripts/buildSea.ts` | Windows build/test portability fixes. |

## Test plan
- [ ] `npx vitest run src/tools/desktop/authoring/datasource/` — full authoring suite (143 tests) passes locally.
- [ ] Author a calc with an invalid formula from an MCP client and verify the tool reports the validator's error message *without* writing anything to the workbook.
- [ ] Author a batch where calc B references calc A's caption and verify B lands after A on the same apply.
- [ ] Author a calc whose formula contains an IF embedded in a larger expression and verify the calc editor renders it multi-line with `END` on its own line.
- [ ] Author a short one-line calc and verify the editor still shows it as one line.


src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -108,6 +136,404 @@ export async function authorCalculationsInWorkbook({
...
+    }
+  }
+  for (const value of Object.values(node)) {
+    const found = findErrorMessagesNode(value);
+    if (found !== undefined) {
+      return found;
+    }
+  }
+  return undefined;
+}
```
b-carver, line 615: (AI Review) 🔴 Critical — formula validation fails **open** here, which recreates the exact bug W-24061123 exists to fix. `findErrorMessagesNode` early-returns `undefined` for any array (line 520), and its recursion feeds `Object.values` (which may be arrays) back into that same guard — so it **never descends into array elements**. An `errorMsgs` array nested under any array (e.g. `{ result: [{ errorMsgs: […] }] }`, common for pres-model payloads) is missed. `validateCalcFormula` then treats `undefined` as VALID, so an invalid formula is written and reported as success. This also contradicts the doc comment's "regardless of nesting" claim. Secondary gaps: `/^errormsgs?$/` won't match `errorMessages`/`errors`; and the first key-matching array wins even if empty. No fixture/test/logging pins the real shape (the comment admits it). Please **fail closed** on shape-miss (return `status:'error'` — do not write) and log with an errorId, and pin the real command payload with a committed fixture test. This is the one blocking item.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -108,6 +136,404 @@ export async function authorCalculationsInWorkbook({
...
+
+  // A collision with an existing field is detected before any apply, so the batch aborts with
+  // no partial state rather than silently shadowing a real field.
+  for (const calc of specs) {
+    if (findColumnByCaption(targetResult.value.xml, calc.caption) !== undefined) {
+      return new ArgsValidationError(
+        'caption collision — pick a new caption or use the existing field',
+      ).toErr();
+    }
+  }
```
b-carver: (AI Review) 🟠 Important — behavior regression: this collision check has no idempotent-retry escape. The prior `author-calc` path (`prepareCalculationBatch`, ~lines 629-648) treated an *identical* existing calc (same caption+formula+role+datatype) as an idempotent success (`continue`); here any pre-existing caption unconditionally errors with "caption collision". Re-running `author-calc` with identical args — a common agent retry — now fails where it used to succeed, and this isn't noted in the PR description. Please confirm it's intended; if not, restore the exact-match idempotent path so retries stay safe.

src/tools/desktop/authoring/datasource/authorCalc.ts
```diff
@@ -74,24 +74,26 @@ export const getAuthorCalcTool = (server: DesktopMcpServer): DesktopTool<typeof
...
-            labelErrors: false,
             resolveLooseReferences: true,
           });
           if (authored.isErr()) {
             return authored.error.toErr();
           }
-          const calc = authored.value.authoredCalcs[0];
+          const outcome = authored.value[0];
+          if (outcome.status === 'failed') {
+            return new ArgsValidationError(outcome.message).toErr();
```
b-carver, line 90: (AI Review) 🟠 Important — this collapses every failure into `ArgsValidationError`, including infrastructure failures. `validateCalcFormula` deliberately separates `validation-error` (command rejected / non-`completed` envelope — the validator never ran) from `invalid-formula`. Surfacing an infra failure as `ArgsValidationError` tells the agent its *formula* is bad when the real problem is transport, so it "fixes" a correct formula or gives up. Please branch on `outcome.failure`: map `validation-error` (and any shape-mismatch added per the Critical fix) to `DesktopCommandExecutionError`; only `invalid-formula` / `cycle` / `dependency-failed` belong under `ArgsValidationError`.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -108,6 +136,404 @@ export async function authorCalculationsInWorkbook({
...
+}
+
+/**
+ * Topologically layers a batch of calcs by their in-batch `[Caption]` references.
+ * Only captions that name another calc in the same batch form edges; references to
+ * existing workbook fields or unknown names are ignored (the validator judges those).
+ * Uses Kahn's algorithm; any calcs still unresolved once no further layer can form are
+ * returned as `cycleIndices`.
+ */
+export function layerCalculationsByDependency(
```
b-carver, line 159: (AI Review) 🟠 Important — test coverage gap. This PR's only new test file is `prettyPrintFormula.test.ts`; none of the new validation/orchestration logic is covered, and CI runs no test job (only cred-scan/SAST/CLA). `layerCalculationsByDependency` is exported, pure and synchronous — trivially unit-testable (multi-layer ordering, self-ref exclusion, cycle → `cycleIndices`, cascading `dependency-failed`) yet has zero tests. Please also add a captured-payload fixture test for `calcErrorMessages` / `findErrorMessagesNode` (the Critical heuristic) and an executor-mocked test for `authorCalculationsWithValidation`.

b-carver (commented):
(AI Review) Automated deep review against W-24061123 (P0: detect & surface calc-authoring failures instead of reporting false success). Posting the blocking + strongly-recommended items as inline threads so they can be tracked to resolution. Cleared items (name-collision, validator caption-form input) and minor nits are omitted here.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
artbeckTableau: Fixed in `f0566f22`.
- **Fail closed on shape-miss:** `validateCalcFormula` now returns `status:'error'` (a `D
esktopCommandExecutionError` carrying the raw result under `invalid-response`) when `calc
 instead of being read as valid (`authorCalcCore.ts:537-548`).
- **Array descent:** `findErrorMessagesNode` now recurses array elements as well as objec
found.
- **Pinned the real shape:** committed fixtures captured live from `tabdoc:get-calc-detai
lidation-invalid.json` (6 messages) -- with tests asserting valid -> `[]`, invalid -> 6,
missing key -> `undefined`, array-nested -> found.
On `errorMessages`/`errors`: the live key is exactly `errorMsgs` (`result.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
artbeckTableau: Fixed in `f0566f22`. Restored the exact-match idempotent path: a pre-existing caption who
se resolved definition is identical (same normalized formula, role, compatible numeric da
tatype) is reused and authors nothing, mirroring `prepareCalculationBatch`; only a captio
n whose definition differs aborts as a collision (`authorCalcCore.ts:270-319`). An `autho
r-calc` retry with identical args succeeds again. Covered by an executor-mocked test asse
rting reuse returns `status:'created'` with the existing calc name and issues no apply.

src/tools/desktop/authoring/datasource/authorCalc.ts
artbeckTableau: Fixed in `f0566f22`. Infrastructure failures no longer collapse into `ArgsValidationError
`. `validateCalcFormula` returns `status:'error'` with a `DesktopCommandExecutionError` f
or a rejected command, a non-`completed` envelope, or an unparseable result; `authorCalcu
lationsWithValidation` propagates that as `Err(DesktopCommandExecutionError)` (`authorCal
cCore.ts:405-410`), which `author-calc` surfaces unchanged (`authorCalc.ts:85-86`). Only
`invalid-formula` / `cycle` / `dependency-failed` become an `Ok` outcome with `status:'fa
iled'` -> `ArgsValidationError` (`authorCalc.ts:88-90`). The `validation-error` kind was
removed in favor of returning the typed error directly. Test asserts a `failed`-status va
lidate envelope aborts as `desktop-command-execution-error` with no apply.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
artbeckTableau: Added in `f0566f22`, in `authorCalc.test.ts`:

- `layerCalculationsByDependency`: linear-chain layering, independent calcs sharing a lay
er, self-reference exclusion, mutual cycle -> `cycleIndices`, cascading `dependency-faile
d`.
- `calcErrorMessages` / `findErrorMessagesNode`: captured-payload fixtures (invalid -> 6,
 valid -> `[]`, missing key -> `undefined`, array-nested -> found).
- `authorCalculationsWithValidation` (executor-mocked): idempotent reuse, differing-formu
la collision, infra abort, invalid-formula from fixture, dependency cascade.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
b-carver: (AI Review) ✅ Resolved by the rework (head `f0566f2`). `validateCalcFormula` now **fails closed**: when `calcErrorMessages` returns `undefined` it returns `{ status: 'error', error: DesktopCommandExecutionError(… "did not contain the expected errorMsgs field") }` instead of falling through to `valid`, so an unrecognized/renamed result shape aborts rather than silently writing an unvalidated calc. `findErrorMessagesNode` now **descends into arrays** (`if (Array.isArray(node)) { for (const entry of node) { … } }`), fixing the array-blind walk. The real payload shape is pinned by committed `__fixtures__/calc-validation-{invalid,valid}.json` (`result.userCalculationDetails.errorMsgs`), and `calcErrorMessages` tests cover extract / empty→valid / undefined→fail-closed / array-nested. This closes the P0 direction. Thanks.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
b-carver: (AI Review) ✅ Resolved. The idempotent-retry escape is restored via a new `existingCalcMatches` helper: a pre-existing caption whose stored definition is identical (same normalized formula + role + compatible numeric datatype, no extra default-format) now reuses the existing calc and authors nothing — no datasource activation, no apply — while a *differing* definition still errors as a real collision. This mirrors `prepareCalculationBatch`, so an `author-calc` retry with identical args succeeds on its own prior output. Covered by "reuses an identical existing calc and authors nothing" and "rejects a caption collision when the existing formula differs".

src/tools/desktop/authoring/datasource/authorCalc.ts
b-carver: (AI Review) ✅ Resolved — fixed upstream in `authorCalcCore.ts` rather than here, which is the cleaner fix. The `validation-error` failure kind was removed entirely; `CalcValidation` `'error'` now carries a typed `DesktopCommandExecutionError`, and an infra failure does `return validation.error.toErr()` to abort the batch on the `Err` channel. So this `authorCalc.ts` mapping only ever receives genuine `invalid-formula` / `cycle` / `dependency-failed` verdicts — a dropped/timed-out validate command now surfaces to the agent as a `DesktopCommandExecutionError`, not "your formula is bad". Covered by "aborts the batch when the validate command cannot complete (infra failure, not a verdict)".

src/tools/desktop/authoring/datasource/authorCalcCore.ts
b-carver: (AI Review) ✅ Resolved. `authorCalc.test.ts` now adds dedicated coverage for the previously-untested logic: `layerCalculationsByDependency` (chain ordering, same-layer independents, self-ref ignored, mutual-ref→cycle indices), `calcErrorMessages` against the captured fixtures (extract, empty→valid, undefined→fail-closed, array-nested descent), and `authorCalculationsWithValidation` (idempotent reuse, caption collision, infra abort, invalid-formula-from-fixture, cascade-to-dependent). One non-blocking follow-up remains: this PR's CI still runs no test job (only Credential-Scan / SAST / CLA), so these tests aren't exercised in CI — worth wiring a vitest job — but the coverage itself now exists.

b-carver (commented):
(AI Review) ✅ Sign-off — all four items I raised are addressed by the latest commits (head `f0566f2`) and their threads are resolved:

- **🔴 C1 (fail-open validation)** — `validateCalcFormula` now fails **closed** on an unrecognized/absent `errorMsgs` shape, `findErrorMessagesNode` descends into arrays, and the real payload is pinned by committed `__fixtures__` + tests. This is the one that made the P0 real: an unparseable validator response now aborts instead of reporting false success.
- **🟠 I1 (idempotent-retry regression)** — restored via `existingCalcMatches`; an identical re-run reuses the existing calc and authors nothing, only a differing definition collides.
- **🟠 I2 (infra mislabeled as input error)** — the `validation-error` kind is gone; infra failures now surface as a typed `DesktopCommandExecutionError` on the `Err` channel, so `ArgsValidationError` only carries genuine formula/cycle/dependency verdicts.
- **🟠 I4 (no tests)** — dedicated coverage added for `layerCalculationsByDependency`, `calcErrorMessages`, and `authorCalculationsWithValidation`.

Non-blocking follow-ups (not gating this PR): this PR's CI runs no test job (only Credential-Scan / SAST / CLA), so the new tests aren't exercised in CI — worth wiring a vitest job; and I3 (a mid-batch apply failure discards the partial `outcomes[]`) remains latent for the future batch tool. Nice, fast turnaround — thanks.

b-carver (approved):
(AI Review) ✅ Approving — all four items I raised (C1 fail-closed validation + array descent + fixtures, I1 idempotent-retry restored, I2 infra failures surface as `DesktopCommandExecutionError`, I4 tests for the layering/validation/orchestration logic) are addressed by head `f0566f2` and their threads are resolved.

Non-blocking follow-ups (not gating merge): this PR's CI runs no test job (only Credential-Scan / SAST / CLA), so the new tests aren't exercised in CI — worth wiring a vitest job; and I3 (a mid-batch apply failure discards the partial `outcomes[]`) remains latent for the future batch tool.

PR #898, opened 2026-09-08
Report logical-table ownership for insight fields

nawartamawi:
## Summary

- Extract a field's logical-table ID from Desktop metadata only when the referenced metadata object is declared in the datasource object graph.
- Add optional `logicalTableId` provenance to field references.
- Return authoritative `fieldTables` mappings from the slim available-fields response used by insight selection.
- Omit ambiguous or unknown table ownership so the backend can fail closed instead of generating a mixed-table KPI.
- Add focused coverage for valid, unknown, and ambiguous provenance.

## Why

The insight service needs authoritative field-to-logical-table ownership to keep the measure, date, and breakdown in one Tableau relationship context. This prevents KPI recreation from silently changing aggregation because of cross-table relationships.

## Validation

- `scripts/agent-check` — all 5 checks passed
- 396 test files and 6,861 tests passed
- Lint, typecheck, Desktop build, and lockstep-core hash gate passed


src/desktop/metadata/field-builder.ts
```diff
@@ -310,8 +316,19 @@ export function listAvailableFields(
...
     const ambiguousTableColumns = new Set<string>();
+    const logicalTableIdByColumn = new Map<string, string>();
+    const ambiguousLogicalTableColumns = new Set<string>();
     const approxCountByName = new Map<string, number>();
 
+    // In federated datasources, metadata-record parent-name may only identify the
+    // repeated physical relation (for example `[sqlproxy]`). `object-id` is the
+    // field's logical-table identity. Only trust ids declared by object-graph.
+    const logicalTableIds = new Set(
+      normalizeArray(datasource['object-graph']?.objects?.object)
```
tonythebest: The extraction here looks correct: normalizing away the brackets lets a bracketed metadata-record `object-id` match the unbracketed object-graph `@_id`, only ids in the object-graph trust set are honored, ambiguous ownership is dropped (sticky), and limiting the physical-`table` projection to `column` records while still visiting `measure` records for logical identity preserves prior behavior. Tests cover the valid / unknown / ambiguous cases.

One consumer-facing thing worth confirming before this and the backend MR ([tab-agent-south !186](https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/186)) leave draft:

`logicalTableId` is emitted **only** for columns whose `object-id` is present in this `<object-graph>` trust set. On the consumer side, a field with no `logicalTableId` is treated as unknown ownership and the automatic insight/KPI selection **fails closed** (no insight). So a datasource that carries metadata-record `object-id`s but has no `<object-graph>` block (older/embedded/single-table exports?) would yield an empty ownership map and silently disable automatic insights for it.

Is `<object-graph>` reliably present on the datasources this path runs against? If some real datasources omit it, it'd be worth confirming the fallback is acceptable (or capturing telemetry on how often ownership comes back empty) so the fail-closed behavior stays a rare edge rather than a segment we quietly turn off. Not blocking — the code is right; just want the coverage assumption to be explicit given the downstream fail-closed consumer.

src/desktop/metadata/field-builder.ts
```diff
@@ -368,10 +385,27 @@ export function listAvailableFields(
...
         const records = normalizeArray(metadataRecords['metadata-record']);
         for (const record of records) {
-          if (record['@_class'] !== 'column') continue;
+          const recordClass = record['@_class'];
+          if (recordClass !== 'column' && recordClass !== 'measure') continue;
           const localName = record['local-name'];
           if (!localName) continue;
           const bracketedName = localName.startsWith('[') ? localName : `[${localName}]`;
+
+          const logicalTableId = normalizeLogicalTableId(record['object-id']);
```
tonythebest: This only assigns ownership when a metadata-record `object-id` is in the object-graph trust set.

A shipped datasource with exactly one graph object can still lose top-level aggregate calcs such as `Profit Ratio`; South then drops that otherwise-safe measure. Do you think the one-object case could inherit that object, with multi-object calcs resolved only when all dependencies share an owner? A real XML → slim → South regression would help lock it.

The PR body also mentions ambiguous-provenance coverage; the parser test I saw covers valid + unknown IDs. A two-conflicting-valid-IDs case would match that claim.

tonythebest (approved):
The same-table fail-closed direction matches the Sep 8 Daily/Slack decision.

[W-24080281](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002jdch0YAA/view) still specifies an aggregate-first cross-table path, Tableau-vs-Insight verification that fails closed, and the TCM 399 reproduction. At these heads the actual TCM pair returns `insufficient_data` (safer than 1550), while Create KPI in [south !186](https://gitlab.tableausoftware.com/browser-clients/tab-agent-south/-/merge_requests/186) can still author `SUM(IF date … THEN measure END)` for an already-shown card.

Before these leave draft, do you think we should update/split that GUS scope, or keep the remaining AC as a follow-up? Approving the draft so you can keep iterating. Rollout/object-graph is already noted on `field-builder.ts`.

src/desktop/metadata/field-builder.ts
nawartamawi: I checked the monolith implementation before changing this. ObjectModelEncapsulateLegacy wraps an eligible legacy relation tree, including physical joins, as one logical object and tags its metadata with that object ID. Commit 0d5dd62 now reads plain and enabled feature-controlled object-graph and object-id elements, and lets otherwise-unmapped fields such as top-level aggregate calculations inherit ownership only when the graph contains exactly one object. Missing and multi-object graphs are still not inferred, and invalid or conflicting IDs remain unmapped. I added XML-to-field-builder-to-slim coverage for the single-object case, plus multi-object conflicting-ID and no-graph cases. The 72 affected tests, TypeScript, lint, and the Desktop build pass.

tonythebest (approved):
Re-approving at 0d5dd62 (single-object inherit). Prior approval was on 0e28fc8.

PR #906, opened 2026-09-12
fix(desktop): report only dashboard results the workbook confirms

mattcfilbert:
## Decision

Desktop authoring must preserve the requested field meaning, filter scope and workbook-change scope, and return the evidence needed to verify its result. This PR applies that rule to template binding, dashboard actions, authoring guidance and readback. Approving it accepts those contracts; future changes must preserve verification before reporting success or reusing a worksheet.

A `COUNTD(Order ID)` KPI must remain a count. A State-selection action must target the requested sheets. An uncertain apply must not invite another mutation, a truncated preview must not look complete, and an analytical question alone must not enter a workbook-creation route.

## Changes

- Preserve count derivations and result types through binding. Reject count overrides that change the authored aggregation or apply another aggregation to an aggregated source, including shared sibling slots. Preserve valid template aggregations and standalone count KPIs. Validation and explicit binding share the derivation and column-suffix rules, while each preserves its existing XML escaping boundary.
- Add typed native dashboard filter actions, verify all settings, and reuse an action with the same caption and settings. Compare exclusions as membership sets. Ignore the exact `Parameters` pseudo-datasource when checking for one shared real datasource.
- Bound pending summary reads and retain cancellation and native diagnostics. Keep failed or uncertain artifact applies non-retryable; preserve `state`, `retrySafe` and public errors in text and structured results after stale-buffer cleanup. An internal worksheet-image failure during summary materialization now identifies its parent operation and quoted worksheet target.
- Label explicit summary results and retained automatic previews with their actual worksheet scope and ignored-selection policy; these rows do not verify values displayed by a worksheet instance in a dashboard. Withhold automatic previews that exceed row, byte or cell limits. Apply pending calculations before reusing a standalone worksheet; refuse an affected dashboard member before any write.
- Preserve native bookmark title layout and center KPI sheet titles above their values. Include title-only fields in binding/removal, resolve each source mapping once, retain target geography metadata, and clear stale donor members when a bare filter field changes.
- Preserve connection-discovery guidance when a worksheet has no datasource references. Include the three existing image exports in the default authoring profile so requested appearance checks can use rendered evidence.
- Scope derived-metric, dynamic-authoring and dashboard-creation instructions to explicit creation/change requests. Field-only requests stop after readback; dynamic requests create only the requested constructs and needed parameter dependencies. Analytical questions and dashboard previews do not enter dashboard creation.

## Verification

At `1d6e0a4f`, `scripts/agent-check` passed all five checks: lint, type checking, **7,343 tests across 411 files**, Desktop build and shared-file hashes. Fresh independent review passed. Regressions cover real count templates and zero-write rejection, filter datasource identity, pending reads, artifact uncertainty, summary-image error context, advertised instructions across default/dynamic/full profiles, and summary scope through full/projected reads and retained/omitted previews.

Paired model fixtures exercised missing-metric analysis, explicit field-only creation, two calculations plus a chart, mixed analysis/authoring, and positive Insights routing. Missing-metric analysis made no workbook changes, and calculation-only requests made no chart calls. Both positive Insights controls invoked the skill and expected Pulse request with the returned field identifiers, then used the synthetic comparison result with no authoring. The first positive attempt exceeded a harness output cap; only that pair was repeated with a sufficient cap. Simplified chart responses do not establish chart correctness, and unsupported next-step suggestions remain open.

Earlier native regression evidence, from the candidate versions recorded in the review comments:

- Fresh serialized Superstore journeys matched independently calculated Sales **2,326,534.3543**, distinct Orders **5,111**, and Quantity **38,654**. Inspected native XML and images confirmed the KPI row, bold 36-point values, centered titles and chart placement.
- Filter probes verified source/targets and settings, retained one action on an identical request, and left workbook XML byte-identical on that retry. Matt reported the manual select/change/clear smoke passed; those clicks were not instrumented.
- A live pending-calculation probe verified standalone worksheet reuse and an unchanged no-pending retry. Dashboard-member refusal remains covered by zero-write regression tests.
- The rebuilt default profile advertised exactly the three added image tools. A read-only native export produced an inspected Sales Cockpit PNG while preserving workbook XML, inventory and the current Ohio selection.
- A fresh discovery journey recovered from empty worksheet datasource references by checking workbook connections and building the requested map.

## Limits

These are bounded regressions, not a production reliability estimate. The earlier rendered-dashboard journeys predate the final routing change. A new combined development bundle has built and passed component checks; its native journey and installed promotion are not yet claimed here. Fresh native launches stopped at Tableau's Repository-in-Documents access dialog before any model call. No merge or release has occurred.

Positive real published Insights consumption remains unverified. Tool routing and synthetic answers do not prove a visible insight card. Model reporting can still misattribute worksheet errors, overstate completion or suggest unsupported alternatives.

Worksheet summaries also do not establish values in dashboard action-filter context: the Ohio-selected dashboard showed Sales 78,258 while worksheet summary reads returned the global total with either `ignoreSelection` value. The image and manual filter report are separate evidence. Native XML diagnostic work is a separate uncompiled patch; attachments are outside this PR.

## Related issue

[GUS W-24167835](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002kYTKmYAO/view).

## Checklist

- [x] Regression tests added and existing behavior checks retained.
- [x] Local canonical checks pass.
- [ ] Peer review and release/version preparation complete.

Prepared by MattGPT on Matt Filbert’s behalf.


src/desktop/externalApi/externalApiToolExecutor.ts
```diff
@@ -114,6 +116,10 @@ const LOGGER = 'ExternalApiToolExecutor';
 // Liveness must fail fast: the health probe runs on a shorter budget than the global
 // request ceiling (the HTTP layer only ever tightens, so a smaller global still wins).
 const HEALTH_TIMEOUT_MS = 10_000;
+const SUMMARY_DATA_PENDING_MAX_RESPONSES = 3;
```
tableaukyler, line 119: is this a standard pattern to have the tool-call retry instead of the agent itself? just curious

src/tools/desktop/api/applyWorksheet.ts
```diff
@@ -50,12 +50,16 @@ import {
 } from './applyWorksheetArtifact.js';
 
 const templatePlanSchema = z.object({
-  templateName: z.string().trim().min(1).max(128).describe('Worksheet template ID.'),
-  title: z.string().trim().min(1).max(255).describe('Worksheet name to create.'),
-  datasource: z.string().trim().min(1).max(255).describe('Live datasource name.'),
+  templateName: z.string().trim().min(1).max(128).describe('Template ID.'),
+  title: z.string().trim().min(1).max(255).describe('Worksheet name.'),
+  datasource: z.string().trim().min(1).max(255).describe('Live datasource.'),
```
tableaukyler: seems like the reduction here is a bit confusing. Perhaps change the param name to datasourceName? the it's self-explanatory

src/tools/desktop/authoring/binder/bindTemplate.ts
```diff
@@ -430,20 +431,36 @@ function capSummaryRows(
...
+      : []),
+  ];
+  if (omissionReasons.length > 0) {
+    return {
+      summary_rows_error: boundedSummaryRowsError(
+        `summary rows omitted because ${omissionReasons.join('; ')}`,
+      ),
+      truncated: true,
+    };
   }
```
tableaukyler, line 462: 🔴 **This discards a complete, in-budget preview instead of returning it.**

`rows.length > SUMMARY_ROWS_MAX_ROWS` alone (e.g. 21 rows, no cell truncation, no byte-budget overrun) now lands in this branch and returns **no rows at all** — just `summary_rows_error` + `truncated: true`. The `cappedRows` computed just above (first 20 rows, fully valid, within the byte budget) are thrown away.

Old behavior (base `20adccf8`) returned the capped rows *plus* `truncated: true` whenever `cappedRows.length > 0`, only erroring out fully when the very first row blew the byte budget. This diff regresses that: any breakdown with >20 members — the common measure-by-dimension case — now gets zero preview instead of a truthful partial one. Confirmed by this PR's own test (`bindTemplate.test.ts`, 59-row case asserting `summary_rows` is `undefined`).

Recommend: keep returning `{ summary_rows: { columns: cappedColumns, rows: cappedRows }, summary_rows_order: rowOrder, truncated: true }` when `cappedRows.length > 0`, and reserve `summary_rows_error` for the case that's actually unrecoverable (first row too big for the budget).

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -339,6 +450,17 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
+                  caption,
+                  sourceDashboard: sourceDashboard!.trim(),
+                  sourceWorksheet: sourceWorksheet.trim(),
+                  excludedWorksheets: filterPlan!.excludedWorksheets,
+                  clearSelection: resolvedClearSelection,
+                }) !== undefined
+              );
+            }
             if (mode === 'set') {
               return hasActionWithTargetParam(
```
tableaukyler: 🟠 **Strategy-pattern smell — `mode` dispatch repeated across this function, and this diff adds another site.**

`targetParamLanded` re-checks `mode` three times here (`'filter'`/`'set'`/`'url'`, falling through to `'parameter'`) to pick a verification predicate. The same four-way switch repeats for the not-applied error message, the action-XML dispatch, and the final Ok-result construction elsewhere in this file — ~20 sites total, ~10 of which this diff touched to add the new `filter` branch. Same pattern occurred when `url` mode was added in #842 (86356125): each new mode pays this multi-site tax.

A per-mode strategy object (`{ isLanded, buildActionXml, buildOkResult }` keyed by `mode`) would collapse this to one seam instead of ~10, so the next mode addition doesn't need to touch this many call sites again.

tableaukyler (commented):
## KylerGPT review

Change: preserve COUNTD/count-derivation semantics through template binding, add typed native dashboard filter actions with readback/dedup, bound the pending-summary-read/cancellation path, and report apply verification as uncertain (not success) when readback can't confirm it. Draft PR, reviewed with `--include-drafts`.

Ran conventions, bugs, blast-radius, history/prior-PR-comments, comment-policy, and design-pattern passes, plus an adversarial premortem, then verified every 🔴/🟠-equivalent finding with an independent model before posting. Most of what the premortem seeded (uncertainty/success conflation, derivation-override plumbing, filter-dedup corrupting the wrong action) checked out clean — the PR's central promise holds under an adversarial read. Two things don't:

### 🔴 `capSummaryRows` withholds the *entire* preview on the single most common case, not just the truncated part
The old cap returned the first 20 rows + `truncated: true`. This diff changes it to: if `rows.length > 20` at all — even with zero cell/byte truncation — return **zero rows**, just `summary_rows_error` + `truncated: true`. The first 20 rows are fully computed and well inside the byte budget; they're discarded anyway. Any categorical breakdown with >20 members (the standard measure × dimension case) now gets no preview instead of an honestly-labeled partial one — the opposite of what "withhold incomplete previews" should mean for a case that isn't actually incomplete. Confirmed directly by the diff and by the PR's own test (`bindTemplate.test.ts`, 59-row case asserting `summary_rows` is `undefined`).

### 🟠 Strategy-pattern smell in `authorAction.ts` — `mode` dispatch repeated ~10+ times, this diff adds to it
`mode` (`'parameter'|'set'|'url'|'filter'`) is switched on at ~20 sites across the file. This diff's new `filter` mode required a new branch at ~10 of them (`targetParamLanded`, the not-applied error message, action-XML construction, the Ok-result build, validation blocks). Same pattern occurred when `url` mode was added in #842 (86356125) — each new mode keeps paying a multi-site tax that a polymorphic dispatch (one object per mode, keyed lookup) would collapse to one seam.

**Note, not a finding on this diff:** PR #879 review comment 3929161606 flagged that `bindTemplate.ts`'s deterministic-reuse branch (~line 3556) can report `reused` while `atomicCalcs` is non-empty and only staged locally — the sibling "duplicate-sheet reuse" branch already guards this with `atomicCalcs.length === 0`, and that comment has no reply. This diff doesn't touch that code path, so it's out of scope here, but it's the same "don't claim success on an incomplete apply" class this PR is otherwise fixing — worth a follow-up.

src/desktop/externalApi/externalApiToolExecutor.ts
mattcfilbert: For this specific read, yes: Desktop returns `503 operation-pending` while preparing a worksheet, without an operation URL to poll. The executor repeats only that summary-data GET, stops after three pending responses, respects `Retry-After` within a five-second per-wait bound, and uses the caller's cancellation signal. It never replays an authoring mutation.

Keeping that short readiness wait in the client avoids another model turn just to finish the same read. The HTTP client already owns `202` operation polling; this handles the separate prerequisite-pending response. Tests cover exhaustion, cancellation, long `Retry-After`, and preserving the budget across the existing 401 rescan.

— MattGPT, on Matt Filbert's behalf.

src/tools/desktop/authoring/binder/bindTemplate.ts
mattcfilbert: The first 20 rows are valid, but they are an incomplete result for a ranking or total. Withholding this automatic preview is intentional.

The live case behind this change asked “Show Sales by State as a filled map.” The old reply returned 20 of 59 State/Province rows with `truncated: true`. The model still said Texas, New York, and Washington had the highest totals; California, the actual leader, was outside that slice. Restoring capped rows would restore the exact payload that produced that failure.

I'm keeping the guard scoped to the automatic `bind-template` preview. Complete results within the row/cell/byte limits, including scalar KPIs and exactly 20 rows, still return normally. Explicit `get-summary-data` behavior is unchanged. The regression test deliberately puts the maximum outside the first 20 rows.

— MattGPT, on Matt Filbert's behalf.

src/tools/desktop/api/applyWorksheet.ts
mattcfilbert: Clarified the description to `Datasource name.` in [40e032d7](https://github.com/tableau/tableau-mcp/commit/40e032d71a4cc18dce4d7332b148d2599debbc87). I kept `templatePlan.datasource` because it is already the serialized field emitted and consumed by the template-plan flow; renaming it here would change that contract. The wording now names the expected value and stays within the existing tool-description size budget.

— MattGPT, on Matt Filbert's behalf.

src/tools/desktop/authoring/datasource/authorAction.ts
mattcfilbert: Changed in [40e032d7](https://github.com/tableau/tableau-mcp/commit/40e032d71a4cc18dce4d7332b148d2599debbc87). A keyed per-mode factory now prepares the action, its readback predicate, failure message, and result builder once. The shared splice, validation, and apply/readback sequence consumes that record.

Filter results still resolve the persisted native action ID from readback; input validation and existing-action matching retain their behavior. The existing 121 action/apply tests and the full repository check pass, and an independent review found no regression.

— MattGPT, on Matt Filbert's behalf.

mattcfilbert:
Updated in [171cb3af](https://github.com/tableau/tableau-mcp/commit/171cb3af). The KPI template now preserves native centered title layout through conversion. Shared field rewriting resolves each source once, preserves target geography metadata, and clears stale members when a filter’s field changes. The two exact-error-wording assertions are removed; error and zero-write behavior checks remain.

All five canonical checks pass, including 7,308 tests. The final packaged two-turn journey built the three KPIs and monthly Sales chart, then invoked insights and left the analytical ask unresolved at the missing published-source prerequisite. Values, native XML and the dashboard image were inspected; the analytical turn left workbook XML byte-identical. Positive published-service execution and instrumented filter clicks remain outside that proof. The PR body now records the final artifact hashes and these limits.

Prepared by MattGPT on Matt Filbert’s behalf.

src/tools/desktop/authoring/binder/bindTemplate.test.ts
```diff
@@ -5390,6 +5714,167 @@ describe('bindTemplateTool auto_apply gate', () => {
...
+        auto_apply: true,
+        skip_validation: true,
+        allowSkipValidation: true,
+        getExecutor,
+      });
+
+      expect(result.isError).toBe(true);
+      invariant(result.content[0].type === 'text');
+      expect(result.content[0].text).toContain('dashboard member sheet with pending calculations');
+      expect(result.content[0].text).toContain('missing calculations prevent safe reuse');
```
tableaukyler: 🟡 **Test asserts on the literal error-message string, which duplicates the production copy.**

`.toContain('dashboard member sheet with pending calculations')` / `.toContain('missing calculations prevent safe reuse')` are copy-pasted from the exact `ArgsValidationError` template literal this diff adds. If that wording changes later for clarity, this test breaks for a reason that has nothing to do with behavior.

The test already asserts the signal that actually matters (`result.isError`, and that `buildInjectedWorkbookXml`/`applyWorkbookDocument` were never called) — that alone proves the dashboard-member path was refused. Consider dropping the message-text assertions, or narrowing to a short invariant fragment rather than the full sentence, so a copy tweak doesn't churn the test.

tableaukyler (commented):
## KylerGPT re-review — 2 new commits since last review

Covers `40e032d7` and `2e0b8a2d` on top of the previously-reviewed `464896e2`.

### Strategy-pattern fix — confirmed, matches the recommendation
`authorAction.ts`'s `mode` dispatch is now a keyed `Record<mode, PrepareAction>` factory (`filter`/`set`/`url`/`parameter`), replacing the if/else chain that repeated across ~10 sites. Traced all four factories against the pre-refactor branches line by line: captured variables (`caption`, `activation`, `resolvedClearSelection`, `sourceWorksheet`, `filterPlan`), readback predicates (`isLanded`), failure messages, and result construction (`buildResult`) are byte-for-byte behaviorally identical to what they replace. No stale outer-scope reads, no cross-mode swaps. This closes the finding cleanly.

### Incomplete-preview guard — the live-ranking-miss case is a legitimate reason to withhold, not a regression
Kept scoped to the automatic preview path only, with `get-summary-data` and complete-within-limits results unaffected — that's the right boundary. Not re-raising this.

### Deterministic-reuse guard — good fix, and it closes a second bug too
The new `atomicCalcs.length === 0` / `classifyWorksheetReplaceTarget(...) === 'in-dashboard'` split does two things: it implements the dashboard-member safety check this diff set out to add, and it incidentally fixes a case where the old unconditional `return Ok(reused...)` would report success while silently dropping a newly-authored calc on a *standalone* worksheet too (the calc was already spliced into the datasource XML but the sheet itself was never rebound to reference it).

Pressure-tested the obvious way this new guard could misfire: a routine re-touch that resends the same `calcs` after they're already applied to a dashboarded sheet. That doesn't false-trigger the new error — `prepareCalculationBatch`'s existing idempotency check (formula/role/datatype/format match against the live datasource XML) treats an already-applied calc as a no-op and excludes it from `authoredCalcs`, so `atomicCalcs` comes back empty on the repeat call and the reuse path is taken, not the new error path. The new guard only fires when there's a genuinely new, unapplied calc against a dashboarded sheet — which is exactly the case it can't safely handle.

One small thing below.

src/tools/desktop/authoring/binder/bindTemplate.test.ts
mattcfilbert: Removed the two wording assertions in 171cb3af. The regression still checks the error result and proves that neither workbook construction nor apply is called. All five checks pass, including 7,308 tests.

Posted by MattGPT on Matt Filbert’s behalf.

src/desktop/binder/validate.ts
```diff
@@ -535,9 +542,54 @@ export function validateBinding(
...
+        code: 'derivation-illegal',
+        slot_id: slotId,
+        detail:
+          `${countSource} '${effDeriv}' returns a quantitative value and cannot bind to ` +
+          `${slot.kind} slot '${slotId}'. Use a quantitative or quantitative-or-categorical slot for count/count-distinct.`,
+      });
+      continue;
+    }
+
+    if (countDimensionInMeasureSlot && feedsCalc) {
```
myu404: ### [P1 · REQUIRED] Reject count overrides the template cannot preserve

**Purpose**
A public `cnt`/`ctd` plan must preserve that aggregation across the template's uses of the field or fail before apply.

**Finding**
The direct-plan schemas accept `cnt`/`ctd` for any slot, but this guard rejects calculation-dependent counts only for dimension sources. A numeric measure can therefore produce a count shelf instance while a dependent formula retains its authored raw/SUM semantics. Separately, an accepted count override on an already-aggregated field is forced to `usr`.

**Impact**
`apply-worksheet` can report success for a structurally valid worksheet whose calculation uses different aggregation semantics, or whose requested count was discarded. The mismatch occurs before readback; this review does not claim a reproduced wrong live value.

**Recommendation**
Reject `cnt`/`ctd` when any same-template-field use participates in an incompatible calculation, and reject explicit count overrides on `isAggregated` inputs. Add focused validation and explicit-bind cases plus a direct `apply-worksheet` boundary assertion.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -421,6 +570,359 @@ function hasActionCaption(xml: string, caption: string): boolean {
...
+    if (!dashboardWorksheetSet.has(worksheetName)) {
+      return new ArgsValidationError(
+        `Worksheet "${worksheetName}" is not a member of dashboard "${sourceDashboard}"`,
+      ).toErr();
+    }
+  }
+
+  let sharedDatasource: string | undefined;
+  for (const worksheetName of requested) {
+    const datasourceNames = worksheetDatasourceNames(worksheetByName.get(worksheetName)!);
```
myu404, line 648: ### [P2 · SUGGESTED] Ignore the Parameters pseudo-datasource when validating filter sheets

**Purpose**
Allow native dashboard filter actions on a worksheet that uses one real datasource plus Tableau parameters.

**Issue and impact**
`worksheetDatasourceNames()` returns every `<datasource>` reference, including Tableau's dedicated `Parameters` pseudo-datasource, while this loop requires exactly one name. A parameter-driven sheet with one real datasource is therefore rejected as using two datasources before the filter action is constructed. The shipped Pareto template has this exact XML shape.

**Suggestion**
Exclude the exact `Parameters` pseudo-datasource before enforcing that one real datasource remains and comparing source/target identity. Add a representative source/target fixture with one shared data datasource plus `<datasource name='Parameters'/>`.

src/tools/desktop/api/getWorkbookInventory.ts
```diff
@@ -20,7 +20,7 @@ export const getWorkbookInventoryTool = (
     name: 'get-workbook-inventory',
     title,
     description:
-      'Orienting read: title, unsaved changes, and worksheet/dashboard/storyboard inventory. Not needed before template discovery or building; use for exploration or non-template authoring.',
+      "Orienting read: title, unsaved changes, and worksheet/dashboard/storyboard inventory. A worksheet's datasource list contains its references; an empty list does not mean the workbook has no connected datasource. Use list-workbook-datasources only when connection inventory matters. Not needed before template discovery or building; use for exploration or non-template authoring.",
```
myu404, line 23: ### [P2 · SUGGESTED] Pin the datasource-inventory guidance as a public tool contract

**Purpose**
Preserve the model-facing rule that an empty worksheet datasource-reference list does not prove the workbook has no connected datasource.

**Issue and impact**
The PR adds that distinction and the `list-workbook-datasources` follow-up to this description, but the colocated test still asserts only the older generic inventory phrases. Removing or inverting the new rule would not fail a test and could restore the discovery failure described in the PR.

**Suggestion**
Add narrow semantic substring assertions for the non-inference rule and the `list-workbook-datasources` route without pinning the complete prose.

myu404 (changes requested):

This PR substantially improves the request-to-workbook contract for dashboard authoring: it preserves field identity through swaps, verifies native filter-action semantics on readback, bounds pending summary reads, keeps uncertain writes non-retryable, and avoids presenting truncated automatic previews as complete results. The latest image-export profile commit is also internally consistent with the existing API-version floors.

### Required

- **Reject count overrides the template cannot preserve.** The new public `cnt`/`ctd` surface still has successful paths where a dependent calculation keeps different aggregation semantics or an already-aggregated input is emitted as `usr`.

### Suggested

- **Ignore the Parameters pseudo-datasource when validating filter sheets.** A sheet with one real datasource plus Tableau's required `Parameters` reference is currently rejected as cross-datasource.
- **Pin the datasource-inventory guidance as a public tool contract.** The new empty-list safety rule is behaviorally important but is not protected by the colocated description test.

### Criteria coverage and positives

The static trace covers count propagation, one-pass field swaps, geography metadata, filter-action idempotency/readback, pending-read bounds and cancellation, apply-state/retry safety, title/CDATA preservation, and worksheet-versus-workbook datasource guidance. Existing discussion fixes for mode dispatch and brittle error-string assertions are present at this head and were not duplicated. I also explicitly reviewed and accept the measured `author-action` schema ceiling change from 1,433 to 2,231 bytes for the new filter-mode contract; the exact per-tool ratchet remains enforced.

### Verification and limits

Reviewed the pinned merge-base-to-head diff `20adccf8..220f749b` and the late `6aba8881..220f749b` delta statically. I did not run tests, a Desktop session, native filter clicks, or a live KPI rendering. The author reports `scripts/agent-check` passing 7,315 tests at the prior head; at final scan, both Node builds and the security/label checks passed for the new head. Native TAS routing/telemetry, positive published-insight consumption, the native operation-pending and worksheet-datasource/workbook-inventory producers, and byte-for-byte downstream lockstep parity remain outside independently verified evidence.

src/desktop/binder/validate.ts
mattcfilbert: MattGPT on Matt Filbert’s behalf. Fixed in [b572935e](https://github.com/tableau/tableau-mcp/commit/b572935e907c23eed031676f6a2af246eb3dcd42).

The shared validator now rejects explicit `cnt`/`ctd` overrides that change the authored aggregation of a calculation input, including through another slot using the same template field. It also rejects count overrides on already-aggregated sources instead of emitting `usr`. Numeric measures and dimensions take the same override check. The new guard permits an override equal to the authored count; existing input checks still apply. A real histogram regression covers that case.

Regression coverage includes validation, explicit binding, and a direct `apply-worksheet` plan that is rejected before any native write. All five canonical checks passed, including 7,336 tests across 411 files; the final independent review found no new issues.

src/tools/desktop/authoring/datasource/authorAction.ts
mattcfilbert: MattGPT on Matt Filbert’s behalf. Fixed in [f77e185f](https://github.com/tableau/tableau-mcp/commit/f77e185f89c1568a97b16da408d101c8340d3a1c).

The common worksheet-datasource helper excludes only the exact `Parameters` name before checking filter sources and targets. Tests cover shared real datasource plus Parameters, Parameters-only rejection, distinct real datasource rejection, and case/caption handling. Rejections assert no apply call. The 74-test action suite and the full canonical checks pass.

src/tools/desktop/api/getWorkbookInventory.ts
mattcfilbert: MattGPT on Matt Filbert’s behalf. Added in [a632eecf](https://github.com/tableau/tableau-mcp/commit/a632eecf5ae165037a4f5666811884de296c0b92).

The registration test now protects both points: empty worksheet references do not prove there is no workbook connection, and connection inventory goes through `list-workbook-datasources`. It uses two narrow substring assertions. The four inventory tests and full canonical checks pass.

mattcfilbert:
MattGPT on Matt Filbert’s behalf. Pushed [78cc9eb2](https://github.com/tableau/tableau-mcp/commit/78cc9eb2da8e209c5495bff6f188000c80a0b52a).

A failed internal worksheet-image request in `get-summary-data` now retains the parent operation, worksheet name/id, and Desktop’s native message and error code. The existing error classification and dialog flag are preserved. Route-missing errors and callers without context are unchanged.

Both new regressions failed before the fix and pass afterward. All five canonical checks passed, including **7,338 tests across 411 files** and the Desktop build; final independent review found no new issues.

This repairs missing facts in the tool result. Model replies can still misattribute a worksheet render failure to a dashboard capture; this change does not claim that reporting defect is fixed.

mattcfilbert:
MattGPT on Matt Filbert’s behalf. Pushed [f2c05503](https://github.com/tableau/tableau-mcp/commit/f2c0550361dd02ea663c64f124a7b8186e713ccb).

The MCP initialization instructions still directed derived-metric and dynamic-calculation questions into workbook authoring. Those existing routes now require an explicit creation/change request. Field-only requests stop after readback; dynamic requests create parameters only when a requested construct depends on one. Dashboard creation also requires an explicit create/rebuild request.

All five canonical checks passed, including **7,341 tests across 411 files** and the Desktop build. Fresh independent review found no remaining issues.

The paired model probes retained calculation-only authoring and kept missing-metric analysis from making workbook changes. Both old/new positive Insights controls called the expected Pulse exploration path with the returned field identifiers and made no authoring calls. Those controls use synthetic tool responses; they do not prove a live published-service response or chart correctness. The original positive control hit the harness’s output-token cap, so only that pair was repeated with a sufficient cap.

Unsupported next-step suggestions still appeared in analytical replies. This removes a conflicting instruction source; it does not claim to resolve the separate reporting defect or establish general model reliability.

mattcfilbert:
Summary results now state their actual read scope. `get-summary-data` returns `readScope: {target: "worksheet", ignoreSelection: true}`, and complete automatic previews carry the same fact as `summary_rows_scope`. Both use the shared native query policy. Omitted previews keep their existing omission result.

This makes the known limit explicit: these worksheet rows do not verify a selected dashboard's displayed values. The native query, returned values and preview limits are unchanged; no selection toggle or client filter reconstruction was added.

All five local checks pass, including 7,341 tests; independent review passed. Tests cover full/projected reads, retained/scalar previews, omission limits, and text/structured results. This is a deterministic result-contract fix, not new evidence that the remaining model-reporting issue is resolved.

— MattGPT, on Matt Filbert's behalf

src/desktop/binder/validate.test.ts
```diff
@@ -614,6 +617,170 @@ describe('binder/validate — aggregate calc-input compatibility', () => {
...
+        },
+        SUMMARY,
+      );
+
+      expect(result.ok).toBe(false);
+      if (result.ok) return;
+      expect(result.blockers).toContainEqual({
+        code: 'aggregation-level-mismatch',
+        slot_id: 'input',
+        detail: expect.stringContaining(
```
tableaukyler: 🟡 **New count-override tests pin production `detail` copy again.**

`ok === false` plus `code: 'aggregation-level-mismatch'` and `slot_id` already prove the reject. The `detail: expect.stringContaining('requested count override ... would change mapped shelf instances')` clause is copied from the new `validate.ts` template.

The optional-input test in this same file already uses the non-copy form (`objectContaining({ code, slot_id })`). Same pattern we asked to drop on `bindTemplate.test.ts`; you removed it there in `171cb3af`.

tableaukyler (commented):
## KylerGPT re-review — 9 new commits since last review

Covers `171cb3af` through `73f9a442` on top of `2e0b8a2d`. Special pass: Desktop / monolith External Client API, not just the TS diff.

### Monolith contract — this increment matches what Desktop actually writes and serves

I checked the new XML and the summary/image paths against `tabexternalapihost` + native workbook fixtures.

- Filter actions still emit native `tsc:tsl-filter` (`special-fields=all`, `target` = dashboard name, `exclude` as a Tableau list, `auto-clear` / `on-empty=none`). That is the same shape Desktop generates in product workbooks (World Indicators, DQW audit). `exclude` escaping matches `ScriptDescription::ListToString`. The Parameters skip is the exact worksheet-ref name Desktop writes (`name='Parameters'`).
- KPI title layout is a `<worksheet><layout-options>` sibling of `<table>`, which is where native sheets keep titles. `fontalignment='1'` is Desktop's center. Live KPI journeys in the PR body line up with that.
- `GET .../summaryData` `ignoreSelection` is mark selection. Codegen default is false. Dashboard `tsc:tsl-filter` does not flow into a worksheet summary read. Hardcoding `true` was already here; this increment only names it (`readScope`). A dashboard PNG is the appearance check, which is why adding the three existing image exports to the default DYNAMIC profile is the right companion, not a contract break.
- `503 operation-pending` + `Retry-After: 1` for the inventory precursor is unchanged in this increment.
- `fieldReferenceRewriter.ts` lockstep hash matches the file bytes at this head (`efdec2e4…`). Re-sync into `agent-to-tableau-desktop` is still the manual follow-up.

Michael's required count-override reject, Parameters skip, and inventory-description pin are in this head and look closed. Prior KylerGPT items stay closed (mode factory, preview withhold scoped to bind-template, dashboard-member reuse guard).

### This increment

No production bug and no Desktop apply/read mismatch in the new commits. One test-style repeat below: new count-override tests pin production `detail` copy again, in the same file that already uses `objectContaining({ code, slot_id })` for the optional-input case.

Docs note, not a blocker: `tableau-studio-docs` `docs/api/implemented-endpoints.md` still describes `ignoreSelection` as an optional query flag with the command default. It does not mention that MCP always sends `true`, or that worksheet summary cannot verify a dashboard action-filter instance. The MCP tool description now does. I am not changing either side.

src/desktop/binder/validate.test.ts
mattcfilbert: MattGPT, on Matt’s behalf: Fixed in [9bcbdf4e](https://github.com/tableau/tableau-mcp/commit/9bcbdf4e). All four new count-override assertions now use `objectContaining({ code, slot_id })`. The rejection checks, scenario matrices and valid controls remain. Full local verification passed: 7,341 tests, lint, type checking, Desktop build and shared-file hashes; independent review found no issues.

mattcfilbert:
MattGPT, on Matt’s behalf: [1d6e0a4f](https://github.com/tableau/tableau-mcp/commit/1d6e0a4f) removes duplicated derivation and column-suffix rules from the two binding paths. Both use the existing validator helpers; each keeps its XML escaping boundary. COUNT/COUNTD regressions cover names containing `&` and `<` through both emitters and the actual DOM rewriter. Independent review and all five local checks passed, including 7,343 tests.

PR #908, opened 2026-09-14
fix(desktop): preserve KPI title formatting

mattcfilbert:
## Decision

Preserve native bookmark title formatting when binding a worksheet. Fields used only in a title must follow the same binding and optional-field removal rules as fields in the view. The KPI bookmark now explicitly centers its sheet title.

## Description

- Carry the bookmark's direct-root `layout-options` into the generated worksheet.
- Infer fields referenced only in that title, bind references in text or CDATA, and remove omitted optional references while preserving static text and styles.
- Center the KPI sheet title and test that its value retains the existing bold 36pt, horizontally and vertically centered formatting.

The package version is bumped to `2.68.7` in both package files. The shared rewriter hash is updated with the change. This adds no tool or schema fields.

## Motivation and Context

This extracts the title-formatting portion of #906 for focused review against W-24167835. COUNT/COUNTD behavior, filters, Insights/dashboard routing, retries, and workbook readback remain in #906.

The base branch already has the explicit 36pt KPI value style. The original small-font symptom was not reproduced in the earlier native trials; this PR does not claim to fix that symptom or complete W-24167835.

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- `AGENT_CHECK_MAX_WORKERS=4 ./scripts/agent-check`: all five checks passed: lint, typecheck, 7,216 unit tests across 411 files, Desktop build, and lockstep hashes.
- Five title regressions failed before the extraction and passed afterward; all 150 tests in the three changed test files passed.
- Independent review found no blocking issues and passed 165 focused/corpus tests, including coverage of all 137 bundled bookmarks.
- Local lockstep consumer check: the copied rewriter and its derivations dependency match byte-for-byte; hash, typecheck, build, script tests, and both title regressions pass. Its full unit suite has 2,365 passing tests, three skipped, and one choropleth assertion failure. That same failure reproduces with the unmodified producer base `20adccf8`: the consumer expects a donor geographic role to survive without target metadata, while the producer already removes it. Consumer synchronization remains separate; this PR does not change that rule.

Native rendering limitation: the earlier centered-title screenshot and XML were produced by the larger #906 candidate at `171cb3af`. This isolated branch has not been installed and freshly rendered in Studio.

## Related Issues

- W-24167835
- #906: remaining dashboard and reliability work stays open there.

## Checklist

- [x] Bumped `package.json` and `package-lock.json` from `2.68.6` to `2.68.7` with `npm run version:patch`.
- [x] Tests added for the changed behavior.
- [x] New and existing unit tests pass locally.
- [x] No breaking API change or documentation update required.
- [x] Read the contribution guidelines; this PR targets `feature/desktop` as requested.

Prepared by MattGPT on Matt Filbert's behalf.


src/desktop/templates/bookmarkTemplate.ts
```diff
@@ -402,7 +415,7 @@ export function bookmarkToTemplateWorkbook(
     : `${rootCardsM ? rootCardsM[0] : ''}${stableWindowInner}`;
   const hasCards = nestedCards || hoisted;
 
-  let body = `<worksheet name='{{TITLE}}'>\n${table}\n</worksheet>`;
+  let body = `<worksheet name='{{TITLE}}'>\n${layoutOptions ? `${layoutOptions}\n` : ''}${table}\n</worksheet>`;
```
tableaukyler, line 418: This injects whatever bookmark-root layout-options exists. That's too broad for a KPI-centering fix.

Empty titles: `magnitude__lollipop`, `ranking__lollipop`, `ranking__bump`, and `part-to-whole__marimekko` ship `<title><formatted-text /></title>`. The two lollipops are still pass-1 eligible. Desktop `VizLayoutOptionsParser` calls `SetTitle` whenever `<title>` is present; `SetTitle` sets `m_bAutoTitle=false` even for empty `FormattedText`. `TitleCaptionComputer` then uses `GetTitle()` instead of `GetAutoTitle()`, so the bound sheet name never shows.

Gallery chrome: `magnitude__horizontal-bar` title is `<Sheet Name> (horizontal)`, Æ, and "Best for comparing discrete categories...". Same `(vertical)` / Æ pattern on the other magnitude/part-to-whole bookmarks. Those used to be dropped.

kpi-text's alignment-only `<Sheet Name>` run is the shape you want. Filter empty titles and donor copy before serialize, or only copy `fontalignment` onto a Sheet Name run.

https://github.com/tableau/tableau-mcp/blob/fac219acbc780318193461cd7fdf2158af7a1fb3/src/desktop/templates/bookmarkTemplate.ts#L386-L418

tableaukyler (commented):
## KylerGPT review

Change: hoist bookmark-root layout-options onto the generated worksheet, infer/bind/prune title-only fields (including CDATA), and center the kpi-text sheet title.

The rewriter slice is sound. xpath `//text()` already returns CDATA in this stack; the old TEXT_NODE filter dropped it. pruneTitleFieldReferences matches the hoist path. kpi-text's alignment-only `<Sheet Name>` run is the right shape. Desktop maps `fontalignment=1` to HCenter on worksheet title runs, so that part should actually render.

The hoist is the problem. It is catalog-wide, not kpi-text-only. `bookmarkToTemplateWorkbook` copies the first bookmark-child layout-options with no title-content check.

### Required

- Do not copy empty `<title><formatted-text /></title>`. Four bundled bookmarks ship that (magnitude lollipop, ranking lollipop, ranking bump, marimekko). The two lollipops are still pass-1 eligible. Desktop `VizLayoutOptionsParser` calls `SetTitle` whenever `<title>` is present; `SetTitle` clears auto-title even for empty `FormattedText`. Those binds get a blank title bar instead of the sheet name.
- Do not copy FT-gallery chrome. `magnitude__horizontal-bar` title is `<Sheet Name> (horizontal)`, an Æ glyph, and "Best for comparing discrete categories...". Same `(vertical)` / Æ pattern on a pile of magnitude and part-to-whole bookmarks. Before this PR that XML was dropped.

Strip empty titles, or only hoist alignment/style onto a `<Sheet Name>` run. Don't ship donor suffixes, Æ, or tutorial copy.

CDATA rewrite, title-slot inference, lockstep hash update, and the kpi-text centering XML look fine. No docs action: no tool/schema change, and studio-docs does not cover this seam.

tableaukyler (approved):

PR #909, opened 2026-09-15
feat(desktop): return native diagnostics with document edits

mattcfilbert:
## Decision

Use Desktop's native diagnostics for field errors, without screenshots. Read diagnostics returned by supported document edits; use `get-diagnostics` to inspect an existing workbook or worksheet without editing it.

Keep “the edit applied” separate from “the resulting worksheet is valid.” Missing, malformed, or partial diagnostics must not become a clean result or tell the agent to replay an edit that already succeeded.

## Scope

- Carry inline diagnostics through workbook/worksheet applies and template auto-apply. Parse the optional payload once, separately from real document-import warnings.
- Preserve known invalid fields and incomplete coverage. Do not issue another diagnostics GET when an inline result is present.
- Retain the follow-up read for supported generic commands with a proven stable worksheet target. Name-only commands still execute but report `target-unproven` instead of inspecting a guessed worksheet.
- Expose `get-diagnostics` in the default authoring profile. Pin the Desktop instance; worksheet reads require exactly one matching stable ID.
- Preserve read-only inspection guidance when structural readback is unavailable, even on older Desktop versions without native diagnostics.

These are static used-field checks, not query execution, rendering, unused datasource fields, or every Desktop warning. There is no automatic repair.

## Integration

Targets `feature/desktop` directly. Requires External Client API **0.2.16** from [monolith #66073](https://github.com/sf-analyticscloud/monolith/pull/66073). [Screenshot #855](https://github.com/tableau/tableau-mcp/pull/855) follows this PR; screenshots are not required for diagnostics.

Future consumers must preserve target identity, coverage, and applied/non-retryable semantics. Approval makes native diagnostics the source of field errors on the paths above.

## Verification

- Canonical check passed: lint, typecheck, 7,333 tests, Desktop build, and lockstep check.
- The three worksheet apply routes share one post-apply check. Regression tests preserve warning handling, worksheet identity, and one output receipt.
- Fresh native-build smoke checks covered an ordinary missing field and a missing Measure Values member, inline/read agreement, no extra diagnostics GET during apply, and exact restoration.
- The built MCP inspection tool passed against the new native build; its OpenAPI matches the checked-in fixture.

Windows live behavior and name-only generic-command target binding remain unverified. Name-only commands are deliberately not claimed as verified.

— MattGPT


src/tools/desktop/api/applyWorksheet.ts
```diff
@@ -210,10 +215,11 @@ export const getApplyWorksheetTool = (
                   // A 'done' marker tells the agent to stop; an observed FAILED readback
                   // is the one outcome where stopping buries the failure, so that branch
                   // points at the follow-up work instead of minting a terminal receipt.
-                  verification.status === 'skipped'
+                  verification.status === 'skipped' &&
```
myu404, line 218: ### [P2 · SUGGESTED] Preserve structural-skip inspection on older Desktop

**Purpose**
An unsupported optional native check should not change the follow-up for an independently unavailable structural readback, while the completed mutation remains non-retryable.

**Issue and impact**
The merged result has `status: skipped` and an `unsupported-api` finding both when structural verification passed and when it was also skipped. This branch checks only for that finding, so all three apply modes emit `nextAction.kind: done` on Desktop below 0.2.16 even when structural re-read was unavailable. The response remains truthfully unverified, applied, and non-retryable, but a client following `nextAction` can stop without the existing read-only inspection that might expose dropped shelves, encodings, marks, filters, or sorts.

**Suggestion**
Retain source-level structural completion and bypass the inspection prefill only when structural verification actually passed or produced an accepted warning. Keep the combined structural-skipped/native-unsupported result applied and non-retryable, route it to the existing live-inspection prefill, and cover cached, artifact, and direct-template modes.

src/desktop/validation/usedFieldValidity.ts
```diff
@@ -0,0 +1,133 @@
...
+  worksheetId,
+  expectedInstanceId,
+  signal,
+}: {
+  executor: ExternalApiToolExecutor;
+  worksheetId: string;
+  expectedInstanceId: string;
+  signal: AbortSignal;
+}): Promise<UsedFieldValidityOutcome> {
+  if (!apiVersionAtLeast(executor.desktopApiVersion, WORKSHEET_FIELD_VALIDATION_MIN_API_VERSION)) {
```
myu404: ### [P2 · SUGGESTED] Check instance identity before the API-version exit

**Purpose**
The compatibility floor should make native validation optional only on the same older Desktop that accepted the edit; replacement-instance detection still needs to govern post-apply evidence.

**Issue and impact**
`checkUsedFieldValidity` returns `unsupported-api` from `executor.desktopApiVersion` before calling `getWorksheetFieldValidation`, where `expectedInstanceId` is enforced. An unpinned structural read can 401-rescan and cache a replacement Desktop first; if that replacement is below 0.2.16, the code never compares instance IDs and classifies lost provenance as an ordinary compatibility case. Apply modes, and qualifying bind flows, can then provide terminal guidance even though post-apply reads no longer belong to the Desktop that accepted the write.

**Suggestion**
Compare `executor.desktopInstanceId` with `expectedInstanceId` before the version-floor return, or enforce both inside one executor operation. Return an instance-change outcome on mismatch, reserve `unsupported-api` for a matching instance, and test a 0.2.16 instance being replaced by a cached 0.2.15 instance.

src/tools/desktop/api/executeTableauCommand.ts
```diff
@@ -210,6 +311,80 @@ function shapeCommandResult({
...
+
+function commandWorksheetTarget(
+  namespace: 'tabui' | 'tabdoc',
+  command: string,
+  dispatchArgs: Record<string, unknown>,
+): CommandWorksheetTarget | null {
+  if (`${namespace}:${command}` === 'tabdoc:generate-viz-from-notional-spec') {
+    return { kind: 'active' };
+  }
+  const registry = lookupExternalApiCommandRegistry(namespace, command);
```
myu404, line 331: ### [P2 · SUGGESTED] Cover the generic single-worksheet registry path

**Purpose**
Completed registered mutators with one proven worksheet target are intended to run native used-field validation and preserve non-retryable mutation semantics.

**Issue and impact**
The generic path composes a mutating registry flag, exactly one `DPI_Worksheet` parameter, local-to-wire argument rewriting, explicit worksheet resolution, and native validation. Existing tests cover those ingredients separately, but the public command test reaches native validation only through the generate-viz active-sheet exception; its registry-backed coverage stops at the two-worksheet unresolved-target branch. A regression at this seam could downgrade an eligible command to skipped validation or select the wrong target while the current tests remain green.

**Suggestion**
Add one public-tool test with `modifies_state='true'` and exactly one `DPI_Worksheet` parameter whose local and wire names differ. Include the requested worksheet plus a decoy, return an invalid native result, and assert rewritten dispatch, the requested stable worksheet ID, failed verification, and explicit non-retry guidance.

src/tools/desktop/api/executeTableauCommand.ts
```diff
@@ -119,12 +155,61 @@ export const getExecuteTableauCommandTool = (
...
+                        signal: extra.signal,
+                      })
+                    : {
+                        status: 'unknown',
+                        worksheetId: targetBefore.worksheetId,
+                        reason: 'target-changed',
+                        message:
+                          'Field verification was not checked because the active worksheet changed during the command.',
+                      };
+              } else {
```
myu404, line 198: ### [P2 · SUGGESTED] Verify how Desktop binds name-based command targets

**Purpose**
Post-command field validity should be attributed to the same worksheet that Desktop actually selected for a generic mutating command.

**Issue and impact**
For one explicit `DPI_Worksheet` parameter, this branch resolves the supplied name or ID before dispatch, sends the original string, and later validates the earlier worksheet ID. The pinned consumer and producer delta do not establish whether Desktop binds a display name before a concurrent rename or name reuse can intervene. If it does not, a narrow race could mutate one sheet while the response reports static validity for another; this is a contract-verification risk, not a demonstrated regression.

**Suggestion**
Confirm the producer's `DPI_Worksheet` binding semantics. If display names are resolved non-atomically, preserve the actual target ID across dispatch—by sending or returning a supported stable ID—or return target-changed/unknown when identity cannot be proven. Add a focused public-boundary test for that contract; unconditional post-command name re-resolution would be unsafe for commands that rename or delete their target.

myu404 (approved):
**Recommendation: APPROVE**

This change adds Desktop's native static used-field validation to the existing post-edit verification report while keeping a completed mutation separate from verification success. I traced the 0.2.16 producer route and schema through the consumer, merge logic, public tool receipts, instance and worksheet guards, and the stacked parent.

The current head preserves structural failures and messages, reports invalid native fields without making the write replayable, and now consistently says that a clean static check does not prove query execution or rendering. The added tests cover the main apply, bind, refine, command, merge-precedence, and unsupported-version behaviors.

I found four non-blocking P2 improvements: retain live-inspection guidance when structural readback and native validation are both unavailable on older Desktop; enforce instance identity before the API-version shortcut; add a public test for the generic single-worksheet registry path; and verify Desktop's target-binding contract for name-based commands.

Limitations: I reviewed source, tests, the exact producer diff, and the stacked base, but did not rerun the repository test suite, verify producer deployment or live 200/202 behavior, or exercise a live Desktop scenario. No exact deployed command-registry snapshot or producer `DPI_Worksheet` binding contract was available, and no producer/consumer diagnostic-payload budget or representative high-cardinality limit was measured. The hosted status API exposed only the passing Salesforce CLA status; the author's broader test and live-run results were not independently verified.

src/tools/desktop/api/applyWorksheet.ts
mattcfilbert: Fixed in c0257780. Unavailable structural readback still leads to inspection, even when native diagnostics are unsupported. Tests cover template, inline, file, and artifact applies while preserving applied/non-retryable semantics.

— MattGPT

src/desktop/validation/usedFieldValidity.ts
mattcfilbert: Fixed in c0257780. Instance identity is checked before the API version floor. Replaced or missing instances cannot be reported as merely unsupported; both regression cases assert that no diagnostics request is sent.

— MattGPT

src/tools/desktop/api/executeTableauCommand.ts
mattcfilbert: Added the public-tool regression in c0257780, narrowed to a proven stable worksheet-ID parameter. It covers local-to-wire rewriting, a different active worksheet, invalid diagnostics, and one dispatch. Name-only DPI_Worksheet parameters intentionally remain unverified; see the target-binding reply below.

— MattGPT

src/tools/desktop/api/executeTableauCommand.ts
mattcfilbert: Removed the name-based inference in c0257780. A lookup before dispatch does not prove which worksheet Desktop mutated. Name-only commands still run, but verification reports target-unproven without querying a guessed worksheet; only proven stable targets receive the follow-up check.

— MattGPT

src/desktop/externalApi/apiVersion.ts
```diff
@@ -1,5 +1,7 @@
 export type ApiVersionFloor = `${number}.${number}.${number}`;
 
+export const WORKBOOK_DIAGNOSTICS_MIN_API_VERSION: ApiVersionFloor = '0.2.16';
```
myu404, line 3: ### [P2 · SUGGESTED] Use a distinct API floor for aggregate diagnostics

**Purpose**
The version gate must identify a producer that serves the aggregate diagnostics routes and terminal `Operation.diagnostics`, including when producer and consumer release independently.

**Finding**
The consumer treats every API version at least 0.2.16 as supporting `/v0/workbook/diagnostics`, worksheet `/diagnostics`, aggregate status records, and inline operation diagnostics. The previously pinned producer snapshot also publishes 0.2.16 but exposes only worksheet `/validation` with `WorksheetFieldValidation`. The current fixture proves the newer shape exists, but the exact final producer source and deployment ordering could not be refreshed.

**Impact**
If any Desktop build with the earlier 0.2.16 contract can pair with this consumer, it passes registration and compatibility checks but lacks the new routes and inline shape. The new tool then errors and post-apply fallback degrades to unknown on a version presented as supported. This review did not verify that such a build shipped.

**Recommendation**
Give the aggregate/inline contract a distinct API version and gate `WORKBOOK_DIAGNOSTICS_MIN_API_VERSION` and `get-diagnostics` on it. If no incompatible 0.2.16 build was distributed, pin that release fact and the exact final producer SHA here; otherwise add a legacy `/validation` fallback only if that compatibility is intentional.

src/tools/desktop/api/getDiagnostics.ts
```diff
@@ -0,0 +1,88 @@
...
+            extra,
+            callback: async (executor, signal, read) => {
+              const expectedInstanceId = executor.desktopInstanceId;
+              if (!expectedInstanceId) {
+                return new DesktopCommandExecutionError({
+                  type: 'unknown',
+                  error: 'Workbook diagnostics could not pin the current Desktop instance.',
+                }).toErr();
+              }
+              if (worksheetName === undefined) {
```
myu404, line 51: ### [P2 · SUGGESTED] Bound whole-workbook diagnostic results truthfully

**Purpose**
Whole-workbook diagnostics should remain usable as MCP output while making any omitted static-field evidence explicit.

**Finding**
Whole-workbook `get-diagnostics` returns the producer aggregate unchanged, and the schemas place no item or byte bound on worksheets, invalid fields, or their strings. `apply-workbook` also exposes the complete aggregate and represents that diagnostics body in both text and structured content. Comparable high-cardinality result paths in this repository define explicit bounds, but these two paths have no result budget or continuation contract.

**Impact**
Result size grows with every reported worksheet and invalid field, and the apply receipt amplifies its downstream wire and context footprint. No representative payload proves an actual overflow or latency regression, but the public contract has no truthful way to stay within a chosen client budget while signaling that more evidence exists.

**Recommendation**
Define and test a public diagnostics-result budget. Return total and returned counts plus an explicit output-incomplete marker, retain actionable invalid-field summaries, and direct callers to worksheet-scoped `get-diagnostics` for full detail. Do not let a projected empty result appear complete; add producer pagination or limits as well if upstream memory must be bounded.

myu404 (approved):
## Follow-up review

Reviewed `89db211235da` → `f89e21580f93`.

**Previous findings:** 3 addressed, 0 partial, 0 open, 1 obsolete, 0 awaiting verification.

**New findings:** 3.

**Recommendation: APPROVE**

All four prior scenarios are closed on the rewritten head. Three non-blocking P2 concerns remain: the aggregate diagnostics contract still shares API version 0.2.16 with an incompatible earlier producer snapshot; **Verify active-sheet target binding before attributing diagnostics**; and whole-workbook diagnostic output has no truthful result budget.

Limitations: this was a complete fresh review because the prior head is not an ancestor of the current branch. I did not rerun tests or live Desktop behavior. The exact current producer source/deployment, Windows behavior, and representative high-cardinality payload limits remain unverified; the hosted status surface exposed only the passing Salesforce CLA check.

### [P2 · SUGGESTED] Verify active-sheet target binding before attributing diagnostics

**Purpose**
Diagnostics for generate-viz should describe the worksheet Desktop actually bound to the command, including when UI focus changes concurrently.

**Finding**
The active-sheet exception samples worksheet A before dispatch and authorizes diagnostics for A whenever the post-command sample is also A. That does not exclude an A-to-B-to-A focus change around the command's actual binding point. The MCP apply lock does not lock Desktop UI selection. Exact producer binding and inline-aggregate scope were unavailable, so this is an attribution risk requiring verification rather than a proven current miswrite.

**Impact**
If Desktop binds the command while B is active, the receipt can report clean or invalid evidence for A after the completed command changed B. If the producer atomically binds A before any focus change, this path is safe.

**Recommendation**
Pin and test the producer's generate-viz target-binding contract. Prefer returning the actual mutated worksheet ID or accepting a supported stable-ID argument. Until identity is authoritative, treat the active target as unproven rather than minting proof from equal snapshots; use a focused live A-to-B-to-A check for the producer guarantee.

PR #913, opened 2026-09-15
W-24135750: Add Tableau Knowledge tools

rubywerman:
Adds Tableau Knowledge API support and two consolidated MCP tools: `query-knowledge-context` for read-only grounding and graph exploration, and `manage-knowledge-context` for graph inspection and semantic-context curation. The shared REST client provides typed graph discovery, search, traversal, suggestion, and semantic-context operations. Both tools use the existing `RequiresKnowledge` registration condition and established role requirements: Viewer for queries and Creator for management. The implementation requests only the minimum Knowledge OAuth scopes, while Tableau services continue to enforce entitlements and content permissions at call time; hosted deployments must enable the existing registration-condition and role-requirement gates for registration-time omission. Validated with `scripts/agent-check` (lint, typecheck, 188 test files, 3,136 tests) and live testing against Nexus.


src/tools/web/registrationConditions.ts
```diff
@@ -132,6 +147,28 @@ export function getUnmetConditionInstructions(condition: RegistrationCondition):
...
 
+async function checkKnowledgeAvailable(restApiArgs: RestApiArgs): Promise<boolean> {
+  try {
+    return await retry(
+      () =>
+        useRestApi({
+          ...restApiArgs,
+          jwtScopes: ['tableau:knowledge:read'],
+          callback: async (restApi) => {
+            await restApi.knowledgeMethods.listGraphs();
```
stephendeoca, line 159: What are all the scenarios related to Tableau Knowledge availability and can you confirm you verified all of them register / omit accordingly by this logic? Also is the listGraphs endpoint available to all user roles?

rubywerman: Confirmed the probe is available to every role eligible for the read tool; it is not intended to be universal for roles without Knowledge access. The registration tests now enumerate the positive read-only/read/write paths and the negative unsupported, unknown-role, and unavailable-environment paths. The condition tests also cover enabled sites with no graphs, deterministic access/unavailable responses, transient recovery, and fail-closed retry exhaustion. Added in `84f2e8f6`.

stephendeoca:
Please add a feature flag to features.json so that we can toggle the tableau-knowledge tools off as a safety mechanism for hosted deployments

docs/docs/tools/knowledge/manage-knowledge-context.md
```diff
@@ -0,0 +1,64 @@
+---
+sidebar_position: 2
+---
+
+# Manage Knowledge Context
+
+Inspects graph status and curates customer-governed semantic context through one management tool. It
+is available on Tableau+ sites.
+
```
stephendeoca, line 8: Please add section on role requirements and other requirements here:
```
## Required permissions

- **Site Role**: Requires Creator role or higher
```

src/sdks/tableau/apis/knowledgeApi.ts
```diff
@@ -0,0 +1,498 @@
...
+export type KnowledgeNodeSearchResponse = z.infer<typeof knowledgeNodeSearchResponseSchema>;
+export type KnowledgeNodeContext = z.infer<typeof knowledgeNodeContextSchema>;
+export type KnowledgeNodeRelationships = z.infer<typeof knowledgeNodeRelationshipsSchema>;
+export type KnowledgeLineage = z.infer<typeof knowledgeLineageSchema>;
+export type KnowledgeNodeImpact = z.infer<typeof knowledgeNodeImpactSchema>;
+export type SemanticStatementInput = z.infer<typeof semanticStatementInputSchema>;
+export type SemanticStatementContext = z.infer<typeof semanticStatementContextSchema>;
+export type SemanticContextNode = z.infer<typeof semanticContextNodeSchema>;
+export type SuggestionReport = z.infer<typeof suggestionReportSchema>;
+export type SuggestionSeverity = z.infer<typeof severitySchema>;
```
stephendeoca: I don't want to give you a bunch of extra work, but would it be possible to refactor this file so that it only contains endpoint declarations + knowledgeApi (lines 468 - 483). Then put types and schemas in a types file under the types folder.

Imo we haven't been totally consistent on that, but if you notice `pulseApi.ts` it only has endpoint + api declaration and it defines all its schemas in `src/sdks/tableau/types/pulse.ts`.

If refactoring this gets pretty messy you can leave this as is any maybe try to refactor in a follow up MR.

src/tools/web/knowledge/manageKnowledgeContext.ts
```diff
@@ -0,0 +1,224 @@
...
+  resultLimitSchema,
+} from './knowledgeToolUtils.js';
+
+const actionSchema = z.enum(['status', 'list', 'suggestions', 'create', 'update', 'delete']);
+const statementInputSchema = z.object({
+  statement: z.string().trim().min(1).max(10000),
+  id: z.string().trim().min(1).max(512).nullable().optional(),
+});
+
+const paramsSchema = {
```
stephendeoca: For LLM tool call accuracy, I highly recommend making the params schema as granular as possible by limiting the shape as much as possible case by case.

For example, you can use discriminated union to enforce a particular schema based on each action.

While I don't know the exact cases, let's say if action is delete then the schema is only graphId and contextId and they're both required. Then if action is list you require graphId, nodeId, isGlobal and the other properties are not even available.

You can even have custom descriptions for each property based on each case, and your tool won't required as much validation

src/tools/web/knowledge/queryKnowledgeContext.ts
```diff
@@ -0,0 +1,360 @@
...
+import { WebTool } from '../tool.js';
+import {
+  flattenKnowledgeStatements,
+  getKnowledgeResultLimit,
+  graphIdSchema,
+  isGlobalKnowledgeContext,
+  resultLimitSchema,
+} from './knowledgeToolUtils.js';
+
+const intentSchema = z.enum(['ground', 'relationships', 'lineage', 'impact', 'sources']);
```
stephendeoca: same idea here, make schema as granular as possible. only require relevant properties to each case.

src/tools/web/knowledge/manageKnowledgeContext.ts
stephendeoca: Query-datasource is an excellent example of a granular schema. We allow for a bunch of different fields, filters, and parameters types in a single query so we break the schema down case by case so the LLM isn't having to decide which properties are relevant or not.

src/tools/web/registrationConditions.ts
stephendeoca: Were you able to validate the conditional registration behavior live using a locally hosted MCP server?

src/sdks/tableau/apis/knowledgeApi.ts
rubywerman: sounds good, updated to follow the Pulse structure. schemas and types now live in types/knowledge.ts.

docs/docs/tools/knowledge/manage-knowledge-context.md
rubywerman: added

src/tools/web/registrationConditions.ts
rubywerman: yes I tested locally against two live sites with registration enforcement enabled. Site A w/ TK enabled registered both tools and both calls worked; Site B, where TK isn’t enabled, omitted both.

src/tools/web/knowledge/manageKnowledgeContext.ts
rubywerman: thanks for this feedback, i updated it so params now use a strict discriminated union per action, so only relevant fields are exposed.

stephendeoca (approved):

stephendeoca (approved):

PR #915, opened 2026-09-16
@W-23864483: Implement isQueryable via VDS's userHasQueryPermissions API

williamyen0203:
## Description

**Design doc**: https://docs.google.com/document/d/1NkWQdRnvGhJgJrlLxUaeCOs-3LlTlRGxhdGLGm_-PaY/edit

Adds an `isQueryable` flag to each upstream datasource returned by the `get-workbook` tool, so callers know whether they can actually query a given datasource with the `query-datasource` tool before attempting it.

- **New VDS endpoint** — Adds a `userHasQueryPermissions` method (`POST /user-has-query-permissions`) to the VizQL Data Service SDK, wrapping HBI's per-user query-permission check. Only `hasQueryPermission` is consumed; the response schema uses `passthrough()` so diagnostic fields remain available for logging.
  - This is a new endpoint that VDS team has added. This API will be [backported to 264](https://salesforce-internal.slack.com/archives/C09NR1NJMHV/p1789512304006089?thread_ts=1788888157.414669&cid=C09NR1NJMHV).
- **Enrichment in `get-workbook`** — After lineage is merged, `enrichUpstreamDatasourceQueryability` calls the endpoint once per upstream datasource (concurrently) and maps the result onto `isQueryable`:
  - `200` → `hasQueryPermission` from the response
  - API error (any HTTP status other than 404) → `false`
  - `feature-disabled` (404), Zodios error, or thrown error → left **unset**
- **Best-effort** — A failed permission check is logged at `debug` and never fails `get-workbook`.
- **Schema/scopes** — Adds optional `isQueryable` to `lineageContentSchema`, adds `tableau:viz_data_service:read` to the `get-workbook` API scope set, and updates the tool description to document the new field.

## Motivation and Context

The `get-workbook` response lists a workbook's upstream datasources, but callers had no way to know whether they were permitted to query a given datasource until they tried `query-datasource` and got an error. Surfacing `isQueryable` lets callers filter to queryable datasources up front.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update

## Version Bump

- [ ] Patch
- [x] Minor
- [ ] Major

## How Has This Been Tested?

Unit tests added for the new VDS method (`vizqlDataServiceApi.test.ts`) and for the enrichment behavior across all mapping cases — queryable/not-queryable, API error, feature-disabled, and thrown error (`getWorkbook.test.ts`). New and existing unit tests pass locally.

## Related Issues

@W-23864483: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002kG51xYAC/view

## Checklist

## Contributor Agreement

      its Contribution Checklist.


src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -1,6 +1,8 @@
 import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+import { ZodiosError } from '@zodios/core';
 import { Ok } from 'ts-results-es';
 import { z } from 'zod';
+import { fromError, isZodErrorLike } from 'zod-validation-error/v3';
```
ycath: unnecessary import?

src/server/oauth/scopes.ts
```diff
@@ -251,7 +251,11 @@ const toolScopeMap: Record<
   },
   'get-workbook': {
     mcp: ['tableau:mcp:workbook:read'],
-    api: new Set(['tableau:content:read', ...RESOURCE_ACCESS_CHECKER_REQUIRED_API_SCOPES]),
+    api: new Set([
+      'tableau:content:read',
+      'tableau:viz_data_service:read',
```
ycath, line 256: Adding `tableau:viz_data_service:read` to get-workbook's required api scopes likely breaks JWT minting for Connected App / direct-trust deployments. Per the get-flow note at lines 110-123, "Connected Apps reject a JWT mint that asks for an un-granted scope" — so a deployment that hasn't granted VDS scope would now fail get-workbook entirely, not just skip the best-effort enrichment. Suggest the same min-vs-per-call scope split get-flow/list-flow-runs use, so VDS read is requested only when enrichment actually runs.

williamyen0203: https://salesforce-internal.slack.com/archives/C09S2ML7FH9/p1789490466529029

I don't think this will be an issue. query-datasource requires the same scope so if the user has already granted that, it will not need anything special for get-workbook.

The new scope is required if we are going to be calling the new VDS userHasQueryPermissions APl.

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -192,6 +207,72 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
+  workbook: Workbook;
+  vizqlDataServiceMethods: VizqlDataServiceMethods;
+  extra: TableauWebRequestHandlerExtra;
+}): Promise<Workbook> {
+  const upstreamDatasources = workbook.upstreamDatasources;
+  if (!upstreamDatasources?.length) {
+    return workbook;
+  }
+
+  const enriched = await Promise.all(
```
ycath: This Promise.all fans out one permission check per upstream datasource with no concurrency cap, and it runs on every get-workbook call regardless of whether the caller needs queryability — adding N VDS round-trips to a previously VDS-free tool. Consider a concurrency limit, gating enrichment behind a param/flag, or a batch endpoint.

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -192,6 +207,72 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
+      try {
+        const result = await vizqlDataServiceMethods.userHasQueryPermissions({
+          datasource: { datasourceLuid: ds.luid },
+        });
+        if (result.isOk()) {
+          return { ...ds, isQueryable: result.value.hasQueryPermission };
+        }
+        // A definitive (non-transient) HTTP rejection means VDS evaluated the request,
+        // so the caller cannot query this data source.
+        if (result.error.type === 'api-error' && !isTransientVdsError(result.error.httpStatus)) {
```
ycath: A 401 auth failure falls through here and is reported as `isQueryable: false,` but an auth failure isn't a permission verdict — it's "we couldn't authenticate," which is indeterminate like the transient bucket. Consider treating 401 as indeterminate (leave unset).

src/tools/web/workbooks/getWorkbook.test.ts
```diff
@@ -336,6 +343,178 @@ describe('getWorkbookTool', () => {
     });
   });
 
+  describe('isQueryable enrichment', () => {
```
ycath, line 355: nit: can we add a test case fora workbook with zero upstreatDatasources to assert the early return skips the permission calls?

src/server/oauth/scopes.ts
ycath: Got it, could be worth configuring a direct-trust Connected App whose granted scope list excludes tableau:viz_data_service:read, then call the get-workbook tool (not just start the server) and check whether the sign-in for that call is rejected.

sf-ali-mahmoud (commented):
Review of the `isQueryable` enrichment. Nice feature and a thorough test matrix — the design shape (best-effort, tri-state, no auth leak) is sound. Flagging one build-breaking lint error and one correctness bug in the error→`isQueryable` mapping that I'd treat as blocking, plus a few scale/observability items worth addressing before this ships broadly.

**Blocking**
- Unused imports will fail `npm run lint` (`getWorkbook.ts:2,5`).
- The error mapping keys on HTTP status instead of `errorCode`, so 401/400/systemic-403 get reported to the LLM as "cannot query" (`getWorkbook.ts:247`).

**Before broad rollout**
- Unbounded `Promise.all` fan-out against the permission endpoint (`getWorkbook.ts:235`).
- `debug`-level silence hides a systemic failure of the newly-required scope (`getWorkbook.ts:258`, `scopes.ts:256`).

**Recommend**
- Distinguish 404937 DATASOURCE_NOT_FOUND from feature-disabled (`vizqlDataServiceMethods.ts:143`).
- Consume the `resources[]` matrix instead of N per-datasource calls (`vizqlDataServiceApi.ts:441`).
- Document all three states of `isQueryable` (`lineageContent.ts:18`).

✅ Confirmed no authorization leak — enrichment runs on the already-filtered `upstreamDatasources` and uses the caller's own auth. And the new method's use of the real `error.response.status` is more correct than the sibling `queryDatasource` (which hardcodes 400).

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -1,6 +1,8 @@
 import { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
+import { ZodiosError } from '@zodios/core';
```
sf-ali-mahmoud: **Blocking (lint):** `ZodiosError` is imported but never referenced — the zodios-error path below is handled via `JSON.stringify(result.error)`, not this type. `@typescript-eslint/no-unused-vars` will fail the build. Please remove.

src/tools/web/workbooks/getWorkbook.ts
sf-ali-mahmoud: **Blocking (lint):** `fromError` and `isZodErrorLike` are imported but never used (looks like leftovers from an abandoned pretty-print approach — the code uses `JSON.stringify` / `getExceptionMessage`). Remove this import line to unblock `npm run lint`.

src/tools/web/workbooks/getWorkbook.ts
sf-ali-mahmoud: **Blocking (correctness):** this maps *any* non-transient `api-error` to `isQueryable: false`, keying only on `httpStatus` and ignoring the captured `errorCode`. Per the endpoint contract, several non-transient errors are **not** a "can-see-but-cannot-query" verdict:

- **401** `INVALID_AUTH_CREDENTIALS` — whole-request auth failure
- **403** `HBIPermissionException` — workbook-datasource feature flag disabled (systemic)
- **403** `PERMISSIONS_FAILED_ERROR` — downstream authz lookup *failed* (a 5xx-class failure with a 403 status)
- **400** `ValidationException` — we sent a bad request

Only **403800** (missing `Read`) is a genuine deny. Highest-probability failure: if the deployment's grant lacks the newly-added `tableau:viz_data_service:read` scope, every call 401s and **every** datasource is reported `isQueryable: false` — the model then tells the user they can't query any of their data.

Suggest discriminating on `errorCode`: only `403800` (± `404937`) → `false`; everything else → indeterminate (unset), same bucket as transient. (Side note: the test comment for the 403 case labels 403800 as "or feature is not enabled" — those are two different conditions.)

src/tools/web/workbooks/getWorkbook.ts
sf-ali-mahmoud: **Scale:** this fans out one VDS permission call per upstream datasource with no concurrency cap. Enterprise workbooks embed dozens of datasources, so a single `get-workbook` can fire 40+ simultaneous POSTs to a permission endpoint that itself performs downstream authz lookups. Under load HBI will 429 — and since 429 is (correctly) treated as transient→unset, you get **non-deterministic partial enrichment**: a random subset comes back set, the rest silently unset, varying call-to-call. Consider a small concurrency limit, or collapsing to 1–2 calls (see the note on the response schema).

src/tools/web/workbooks/getWorkbook.ts
```diff
@@ -192,6 +207,72 @@ export const getGetWorkbookTool = (server: WebMcpServer): WebTool<typeof paramsS
...
+        }
+        detail = JSON.stringify(result.error);
+      } catch (error) {
+        detail = getExceptionMessage(error);
+      }
+
+      log(
+        {
+          message: `Could not determine queryability for data source ${ds.luid}`,
+          level: 'debug',
```
sf-ali-mahmoud: **Observability:** this logs every failure at `level: 'debug'`, whereas the two sibling enrichments in this same file log at `warning` (connections and lineage). Combined with the new `tableau:viz_data_service:read` scope requirement, a systemic failure — scope not granted, feature flag off site-wide, or VDS down — would silently degrade *every* `get-workbook` with no warning and no metric. Suggest `warning` for this failure class (and ideally a metric) so a site-wide outage is visible.

src/sdks/tableau/methods/vizqlDataServiceMethods.ts
```diff
@@ -121,4 +123,39 @@ export default class VizqlDataServiceMethods extends AuthenticatedMethods<
...
+   * @param {UserHasQueryPermissionsRequest} request
+   */
+  userHasQueryPermissions = async (
+    request: UserHasQueryPermissionsRequest,
+  ): Promise<Result<QueryPermissionsOutput, VdsQueryError>> => {
+    try {
+      return Ok(await this._apiClient.userHasQueryPermissions(request, { ...this.authHeader }));
+    } catch (error) {
+      if (isErrorFromAlias(this._apiClient.api, 'userHasQueryPermissions', error)) {
+        if (error.response.status === 404) {
```
sf-ali-mahmoud: **Recommend:** collapsing all 404s to `feature-disabled` conflates two distinct meanings from the contract: (a) the endpoint isn't deployed, and (b) `404937 DATASOURCE_NOT_FOUND` (the datasource genuinely doesn't exist). A deleted datasource (404937) is effectively a hard deny but ends up left `isQueryable` *unset* instead of `false`, and the `feature-disabled` log line becomes misleading during triage. Consider branching on `errorCode === '404937'`. (This mirrors the sibling `queryDatasource`, so it's a consistent gap rather than a new one.)

src/sdks/tableau/apis/vizqlDataServiceApi.ts
```diff
@@ -430,6 +430,18 @@ export const queryOutputSchema = z
...
 
+export const userHasQueryPermissionsRequestSchema = z
+  .object({
+    datasource: datasourceSchema,
+  })
+  .passthrough();
+
+// Only hasQueryPermission is consumed; passthrough keeps the diagnostic fields (datasourceType,
+// resources, capabilities) available for logging without over-modeling a shape HBI may still change.
+export const queryPermissionsOutputSchema = z
```
sf-ali-mahmoud, line 441: **Recommend (efficiency):** modeling the response as just `{ hasQueryPermission }` parses-but-discards `resources[].capabilities[]`. Per the contract, a single call on a WORKBOOK-type (embedded) datasource returns the matrix for the owning workbook **plus every published DS it embeds** — exactly the per-resource data needed to answer many datasources from one call. The enrichment instead issues N calls and reads only the aggregate boolean. The aggregate `hasQueryPermission` can't be decomposed per-DS, but `resources[].capabilities` (Read + Connect + VizqlDataApiAccess all Allow per resource luid) can — which would also resolve the unbounded fan-out. Worth consuming the matrix rather than paying N round-trips.

src/sdks/tableau/types/lineageContent.ts
```diff
@@ -15,6 +15,7 @@ export const lineageContentSchema = z.object({
   luid: z.string(),
   name: z.string(),
   datasourceType: z.enum(['published', 'embedded']).optional(),
+  isQueryable: z.boolean().optional(),
```
sf-ali-mahmoud, line 18: **Recommend (docs):** `isQueryable` is now tri-state to the model — `true` / `false` / absent — but the tool description only defines `true`. Given the mapping issue noted in `getWorkbook.ts`, `false` isn't always a real deny, and `absent` is undocumented (the model may read it as false, true, or ignore it). Once the `false` semantics are tightened, document all three: `true` = confirmed queryable, `false` = confirmed not, absent = undetermined (don't infer).

src/server/oauth/scopes.ts
sf-ali-mahmoud, line 256: **Heads-up (regression vector):** adding `tableau:viz_data_service:read` to `get-workbook` means any existing deployment whose PAT/OAuth grant doesn't already include this scope will start 401/403-ing on the new permission call — and via the error mapping in `getWorkbook.ts:247` that currently surfaces as every datasource being `isQueryable: false`. Fixing the error mapping (indeterminate instead of false for auth failures) plus `warning`-level logging would make this failure mode safe and visible.

williamyen0203: Actually I dug into it but I still don't think the scope will break existing deployments.

Screenshot 2026-09-17 at 11 13 22 AM

Documentation ([direct-trust Connected App docs](https://help.tableau.com/current/online/en-us/connected_apps_direct.htm)) confirms they're not configured in the connected app itself, they're set by the caller in the JWT's scp claim (which is what we do in the code)

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: Good call, adding 401 to list of transient failures.

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: Yep, will remove.

williamyen0203: Will remove.

src/sdks/tableau/apis/vizqlDataServiceApi.ts
williamyen0203: The follow up story would add a reason and owner, which would need additional fields from the response: https://docs.google.com/document/d/1NkWQdRnvGhJgJrlLxUaeCOs-3LlTlRGxhdGLGm_-PaY/edit?disco=AAACFocxfy8

I am thinking of adding those at the same time, rather than defining the whole schema, needed or not, all at once.

src/server/oauth/scopes.ts
williamyen0203: This is what AI is flagging to me too, but I don't think it will be an issue: https://github.com/tableau/tableau-mcp/pull/915#discussion_r4040061804

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: Good point, updating to only set isQueryable on 200 and 403 responses.

src/sdks/tableau/methods/vizqlDataServiceMethods.ts
williamyen0203: Will return isQueryable=false on 404 feature-disabled case too.

src/tools/web/workbooks/getWorkbook.test.ts
williamyen0203: Added

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: Updated to warning

src/sdks/tableau/types/lineageContent.ts
williamyen0203: Updated tool description to mention this.

src/tools/web/workbooks/getWorkbook.ts
williamyen0203: Added a short circuit: probe the first datasource and if it's a 404 or 403 (feature flag off) response, we can set isQueryable to false for everything, instead of querying each DS and getting the same response back.

Also added a concurrency cap of 5.

williamyen0203: https://github.com/tableau/tableau-mcp/pull/915#discussion_r4042122605

ycath (approved):

PR #916, opened 2026-09-16
@W-24132500: Recognize WorkbookDatasourceNotEnabled VDS error on query-datasource

ycath:
## Description

Half A of W-24132500 (making query-datasource embedded/workbook-aware): recognize the VDS `VDSForWorkbookDatasources` gate on the `query-datasource` path and surface a specific, actionable message instead of a raw, misleading one.

- **`vizqlDataServiceMethods.ts`** — export `WORKBOOK_DS_NOT_ENABLED_FLAG` and an `isWorkbookDatasourceNotEnabled(error)` predicate; add a `'workbook-datasource-not-enabled'` variant to `VdsQueryError`. Detection keys off the **flag identifier in the error message, not the HTTP status or errorCode**, and is checked **before** the `404 → feature-disabled` branch so it can never be mislabeled as "VizQL disabled" regardless of the status the gate arrives with.
- **`mcpToolError.ts`** — new `WorkbookDatasourceNotEnabledError` (statusCode 403, matching the wire status) with a `getErrorText()` override naming the per-site site-admin opt-in, following the existing Pulse actionable-error pattern.
- **`queryDatasource.ts`** — surface the new error from the tool.
- **`adminInsightsToolBase.ts`** — handle the widened shared `VdsQueryError` union (Admin Insights uses published system datasources, so not expected to fire there, but the type is shared).
- **Tests** — SDK + tool layers: the real gate shape (HTTP 403 / `403800` / flag in message), status-independent detection, case-insensitive matching, predicate units, and regression coverage that unrelated errors — including a **generic `403800` per-datasource denial** — are unaffected.

**No auth scope changes** — `query-datasource` already requires `tableau:viz_data_service:read`.

## How detection works (verified against live behavior)

The `VDSForWorkbookDatasources` gate is enforced by **headless-bi's request interceptor**, which runs *before* the request reaches the VDS query engine and short-circuits **every** VDS endpoint (`query-datasource` and `user-has-query-permissions` alike). Verified against live `query-datasource` behavior on a flag-off site with an embedded datasource:

| field | value |
|---|---|
| HTTP status | `403 Forbidden` |
| `errorCode` | `403800` (generic `PERMISSION_DENIED` — **not** used for detection) |
| `message` | `The VDSForWorkbookDatasources feature is not enabled.` (localized) |

There is a single front-line gate, not two enforcement points with different codes: the interceptor's `403800` short-circuits both endpoints, so the VDS-internal `501` / `0x4A7F2B19` path is unreachable for this condition. `errorCode` is unusable as a signal because `403800` is a generic `PERMISSION_DENIED` shared with ordinary per-datasource denials, distinguished only by message. Detection therefore keys off the **flag identifier `VDSForWorkbookDatasources` in the message** — a proper name that stays untranslated even when the surrounding message is localized. (An earlier revision keyed off a `501` / `tab-error-code 0x4A7F2B19` shape and a hex secondary signal; live testing showed that shape never surfaces on this path, so the hex signal was removed.)

## Motivation and Context

Static workbook-datasource query is **opt-in per site** (flag `VDSForWorkbookDatasources`) with no discovery endpoint. On non-enabled sites, querying an embedded datasource is the gate's *expected* response, not a rare edge case. Previously this fell through to the generic `api-error` path and surfaced a raw, non-actionable message. This is the necessary backstop even when upstream permission checks pass, since the site-scoped flag is distinct from per-user query permission.

## Type of Change

- [x] Bug fix

## Version Bump

- [x] Patch

_Half A adds no new tool parameter. The caller-supplied `datasourceType` param (Half B) will come in a follow-up and would warrant a Minor bump then._

## How Has This Been Tested?

- The full unit suite passes (**3057 tests**), and ESLint is clean on all changed files.
- SDK suite `vizqlDataServiceMethods.test.ts` covers the real 403/403800 gate shape, case-insensitive matching, status-independent detection, and negative cases (including a generic 403800 denial with a non-flag message); the tool-layer test in `queryDatasource.test.ts` asserts the surfaced actionable text.
- **Verified against a live flag-off site via Postman** (`query-datasource` returns 403 / 403800 / "The VDSForWorkbookDatasources feature is not enabled."). Integration test against a real embedded LUID (AC#4) remains environment-gated and tracked separately.

## Related Issues / Interaction with #915

Story W-24132500 (epic: Support Workbook static data sources). Rebased on top of #915 (`isQueryable` via `userHasQueryPermissions`), which merged first and shares `vizqlDataServiceMethods.ts` + the `VdsQueryError` union.

**Cross-path note:** both endpoints hit the same interceptor and receive the same 403 / 403800 / "feature is not enabled" response. #915 maps that on the `userHasQueryPermissions` path to `feature-disabled`; this PR maps the same condition on the `query-datasource` path to the more specific `workbook-datasource-not-enabled`. Reconciling the `isQueryable` path to the more specific label, if desired, is a follow-up.

## Checklist


src/sdks/tableau/methods/vizqlDataServiceMethods.ts
```diff
@@ -14,8 +14,19 @@ import {
 import { RestApiCredentials } from '../restApi.js';
 import AuthenticatedMethods from './authenticatedMethods.js';
 
+// VDS error code returned when querying an embedded (workbook) datasource on a site where
+// workbook-datasource querying isn't enabled (site-scoped flag `VDSForWorkbookDatasources`, opt-in
+// per site with no discovery endpoint). Catching this code is the only client-side way to detect the
+// gate. Exported so other VDS paths can reuse the predicate rather than re-hardcoding the hex string.
+export const WORKBOOK_DS_NOT_ENABLED_CODE = '0x4A7F2B19';
```
williamyen0203: Were you able to validate this on a site with the flag off?

When I tested against userHasQueryPermissions API, I got this: https://docs.google.com/document/d/1NkWQdRnvGhJgJrlLxUaeCOs-3LlTlRGxhdGLGm_-PaY/edit?tab=t.gm0azlau2m9r#bookmark=id.hw4i3xjymo8m

```
{
  "errorCode": "403800",
  "message": "The VDSForWorkbookDatasources feature is not enabled.",
  "datetime": "2026-09-14T22:57:31.11806382Z"
}
```

I didn't get an error code that looked like this.

ycath: yeah, you are right, I thought the hex code would be exposed through tab-error-code but should be using errorCode

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "4.11.0",
+  "version": "4.11.1",
```
stephendeoca: The last commit to main was accidentally checked in without a version bump. It would have been 4.12.0, which will be added in soon. So I think after we should expect this to be 4.12.1, just an FYI

williamyen0203 (approved):

stephendeoca (approved):

PR #918, opened 2026-09-17
fix(W-24198280): reject dashboards referencing an unrendered worksheet (HasVisualDoc preflight)

b-carver:
### Problem
A dashboard can reference a worksheet that **exists by name** (it may even have a
`<window class='worksheet'>` pairing) but was **never rendered with a mark** — a blank
`<table>` with no field placed on any shelf/encoding, typically a template artifact that
silently failed to render. Tableau Desktop rejects the workbook at apply with:

> `Dashboard references sheet '<name>' which has no visual representation in the workbook.`

This is Desktop's `HasVisualDoc` gate (`IDP_ERR_DASHBOARD_MISSING_VISUAL_DOC`), not the
MCP's. The MCP's preflight previously checked only that the worksheet **name** existed and
had a `<window class='worksheet'>` — **not** that it carried a placed mark — so it let the
bad workbook through to Desktop, which then rejected it. The agent, seeing a bare failure,
retry-thrashed the same payload across every dashboard-apply lane
(`apply-dashboard` → `run-dashboard-batch` → `apply-worksheet`), all rejected identically.
Reproduced across three LangSmith threads.

### Fix
Move the check into the MCP preflight so the bad workbook never reaches Desktop:

- **`refactor`** — extract worksheet render-state detection into a shared module
  (`src/desktop/metadata/worksheetRenderState.ts`) so every apply lane uses one definition
  of "rendered."
- **`feat`** — guard the `apply-dashboard` routes: a dashboard zone that references a
  present-but-blank worksheet is rejected before dispatch with a `sheet-not-rendered`
  classification. Because nothing is mutated, a retry *after* this rejection is always safe
  — but the correct move is to render the sheet first, then reference it.
- **`fix`** — `resolveRenderedWorksheetNames` now requires an actual **placed field
  reference**, not merely a worksheet/window pairing.
- **`test`** — render the sibling dashboard-test fixtures so they satisfy the new gate.

### Effect
**Deterministic: 100% → 0%** of the blank-sheet signature reaching Desktop. This is a gate,
not a rate — it either fires or it doesn't. Verified by the new red→green unit/integration
tests (`worksheetRenderState.test.ts`, `loadDashboardXml.test.ts`,
`composeDashboardCore.test.ts`).

### Controlled repro: blank worksheet vs empty `<viewpoints/>` (per review)
The shipped fixture carried **both** a blank worksheet **and** an empty `<viewpoints/>`, so
it could not isolate which one drives the `HasVisualDoc` rejection. `checkWorksheetsRendered`
keys purely on the worksheet `<table>`'s placed-field state and **never reads
`<viewpoints>`**. Two added cases in `loadDashboardXml.test.ts` pin that separation:
- **fires** on a blank worksheet even when the dashboard window carries **valid** viewpoints
  (blank sheet is the cause, not viewpoints), and
- **does NOT fire** on a populated worksheet even when `<viewpoints/>` is empty (empty
  viewpoints alone must not trip the guard).

### Scope / what this does NOT catch (by design)
- A worksheet the zone references that is **absent** from `<worksheets>` entirely — a
  different rule's job; this gate fails open there.
- The render-time D9 blank (a correctly-wired worksheet with real field refs that still
  draws zero marks) — undetectable from XML alone; only a live render probe catches it.
- This is a **different** failure from the malformed-`<viewpoints>`-nesting `HasVisualDoc`
  case; both trip the same Desktop gate name but the cause and fix are unrelated.

### Companion knowledge change
The agent-facing recovery guidance (`recovery.md`) lives in the `agent-to-tableau-desktop`
repo and is shipped separately (`tableau/agent-to-tableau-desktop#324`). An offline paired
content-value eval of that doc showed a **+20-point** directional lift (baseline 33% →
treatment 53% at N=15, 0 errors) after calibrating the grader on the saved answers — the
original PR read +27 before that calibration corrected one mis-scored baseline answer. That
lift is directional pilot evidence, not statistical proof; **this code fix is the
deterministic guarantee.**


mattcfilbert:
Good stuff, Brian!

I like the structure here: one preflight covers both apply routes, and the tests verify that rejection sends no changes to Desktop. You're giving the agent a much clearer recovery point 

Did a monolith context sanity check with my robot since I'm trying to get more understanding / coverage at this side:

> Could we separate field placement from dashboard viewpoints before making this a hard block? Native `HasVisualDoc` checks the dashboard’s registered visual documents, while the companion fixture has empty` <viewpoints />`

(a controlled repro with valid viewpoints would help establish whether the blank worksheet itself causes the failure)

Probs would also cover a blank worksheet that already has a valid dashboard visual

PS - for context, the robot was referring to `ParseDashboardWindow` in `WindowParser.cpp`

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -110,6 +115,87 @@ function resolveCanonicalDashboardName(
...
+  if (!dashboard?.zones) {
+    return [];
+  }
+
+  const names = new Set<string>();
+  const visit = (zone: ParsedZone): void => {
+    if (zone['@_name'] && !zone['@_type-v2']) {
+      names.add(zone['@_name']);
+    }
+    for (const child of normalizeArray(zone.zone as ParsedZone | ParsedZone[] | undefined)) {
```
artbeckTableau: **Correctness: zone recursion misses `<zone-pane>`-wrapped worksheets.** `visit()` only recurses through `zone.zone`, so a worksheet-naming `<zone>` nested inside a `<zone-pane>` wrapper (a structure `ParsedZone` explicitly models) is never collected. The established `dashboard-zones-reference-included-worksheets` rule handles this with XPath `.//zone[@name and not(@type-v2)]` at any depth/wrapper.

Impact: a floating-group or hand-edited dashboard can slip a genuinely blank worksheet past this guard and hit the exact `IDP_ERR_DASHBOARD_MISSING_VISUAL_DOC` failure this PR is meant to prevent. Consider recursing through `zone-pane` children too, or reusing the existing XPath selector for parity with the validation rule.

artbeckTableau (commented):
Automated review pass (Claude Code). Six inline findings below: three correctness, two efficiency, one encoding convention. The two message-consistency notes on compose/batch paths are omitted at the author's request.

src/desktop/metadata/worksheetRenderState.ts
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+      continue;
+    }
+
+    for (let index = 0; index < element.attributes.length; index++) {
+      const attribute = element.attributes.item(index);
+      if (
+        attribute &&
+        (attribute.value.includes('].[') ||
+          ((attribute.name === 'column' || attribute.name.endsWith('field')) &&
```
artbeckTableau, line 46: **Correctness: blank-detection heuristic ignores non-`column`/`*field` attributes.** `hasPlacedFieldReference` only recognizes a single-datasource bracket ref (`[Field]`, no `].[`) when the carrying attribute is literally named `column` or ends with `field`. Other legitimate shelf/filter attributes that can carry a placed ref (e.g. a groupfilter `level`/`member`) are invisible to this scan.

Impact: a single-datasource worksheet whose only placed reference lives on an unrecognized attribute can be misclassified `'blank'`, so `checkWorksheetsRendered` / `resolveRenderedWorksheetNames` rejects a legitimate apply that should have succeeded (false-positive block).

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -110,6 +115,87 @@ function resolveCanonicalDashboardName(
...
+  canonicalName: string,
+  dashboardXml: string,
+  { executor, signal }: WithExecutorAndAbortSignal,
+): Promise<Result<void, Extract<LoadDashboardXmlError, { type: 'sheet-not-rendered' }>>> {
+  const worksheetZoneNames = collectWorksheetZoneNames(dashboardXml);
+  if (worksheetZoneNames.length === 0) {
+    return Ok.EMPTY;
+  }
+
+  const workbookResult = await getWorkbookXml({ executor, signal });
```
artbeckTableau: **Correctness (race): live read runs outside `withApplyLock`.** This `getWorkbookXml()` executes before the guard returns, but the actual apply happens later under a different lock/timing. A concurrent build-and-apply that renders the worksheet between this read and the apply is not accounted for.

Impact: another in-flight call renders e.g. `Sheet 1` right after this unlocked read finds it blank; `loadDashboardXml` then rejects with `sheet-not-rendered` even though the sheet is populated by the time the caller retries -- a spurious failure/retry. Consider reading under the same lock the apply uses, or documenting the window as acceptable given the fail-open design.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -190,6 +276,20 @@ export async function loadDashboardXml({
   const canonicalFocus: ApplyFocus =
     focus.navigate === 'artifact' ? { ...focus, sheetName: canonicalName } : focus;
 
+  // Preflight render guard, ahead of both the per-sheet and whole-workbook apply routes below —
+  // a zone naming an existing-but-blank worksheet would otherwise dispatch and be rejected by
+  // Desktop's own HasVisualDoc check. See {@link checkWorksheetsRendered}.
+  const renderCheck = await checkWorksheetsRendered(canonicalName, xml, { executor, signal });
```
artbeckTableau: **Efficiency: whole-workbook route fetches the full workbook twice.** This guard fetches the workbook (via `checkWorksheetsRendered`), and the whole-workbook apply route (`requireExistingSheet=false`) then fetches + parses the entire workbook again in `loadDashboardXmlViaExternalApi`. The PR's own test asserts `getWorkbookDocument` is called twice.

Impact: every build-and-apply-dashboard / apply-dashboard-with-viewpoints call pays two full workbook round trips and two full-document parses -- latency proportional to workbook size -- even when no worksheet is blank. Consider threading the already-fetched XML through to the apply path instead of re-fetching.

artbeckTableau: **Efficiency: per-sheet route now pays a full-workbook fetch it never used to.** apply-dashboard / apply-storyboard (`requireExistingSheet=true`) previously only listed sheets or fetched a single document -- this guard adds an unconditional whole-workbook `getWorkbookXml()` whenever any zone names a worksheet.

Impact: a large-workbook apply-dashboard with no blank worksheets now always transfers and parses the entire workbook document just for the render check -- a latency regression on a previously cheap, targeted operation. Worth gating the fetch or scoping it to the referenced worksheets.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -110,6 +115,87 @@ function resolveCanonicalDashboardName(
   return Ok(xmlName);
 }
 
+/**
+ * Worksheet-zone names referenced by the dashboard fragment's zones, at any nesting depth.
+ * Mirrors the selector convention used by the `dashboard-zones-reference-included-worksheets`
+ * validation rule: a zone with a `@name` and no `@type-v2` names a worksheet; layout, text, and
+ * blank zones carry `type-v2` and do not. Returns `[]` (never throws) when the fragment has no
+ * `<dashboard>` root or no zones — malformed/absent XML has nothing to say here, and is caught
```
artbeckTableau: **Convention: non-ASCII em-dash (`U+2014`) introduced in new lines.** This diff adds em-dash characters at lines 123, 154, 159, 279, and 286. Line 286 is especially worth fixing since the em-dash lands inside a user-facing string literal (`'... worksheet -- not sent to Tableau'`).

Recommend keeping source comments and string literals ASCII-only (replace with `--` or ` - `) to stay safe under Windows-1252 / code-page-sensitive tooling.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -110,6 +115,87 @@ function resolveCanonicalDashboardName(
...
+  } catch {
+    return [];
+  }
+  if (!dashboard?.zones) {
+    return [];
+  }
+
+  const names = new Set<string>();
+  const visit = (zone: ParsedZone): void => {
+    if (zone['@_name'] && !zone['@_type-v2']) {
```
myu404: ### [P1 · REQUIRED] Collect type-v2 visual worksheet zones

**Purpose**
The preflight must inspect every supported worksheet-zone representation before either dashboard apply route can dispatch a reference to a blank worksheet.

**Finding**
`collectWorksheetZoneNames` accepts a named zone only when `type-v2` is absent, but this repository's target-dashboard invariant also classifies named `type-v2='visual'` zones as worksheet zones. A valid direct or nested visual zone is therefore discarded here, leaving that referenced worksheet out of the render guard's check.

**Impact**
A dashboard using the supported visual-zone form can reference a present-but-blank worksheet, bypass this preflight, and reach Desktop with the failure this change is intended to prevent. The coordinated offline floor repeats the predicate. This is distinct from the existing `zone-pane` thread because it also fails for a direct zone.

**Recommendation**
Use the existing worksheet-zone semantics—a name with no `type-v2` or `type-v2='visual'`—add direct and nested visual-zone tests with a blank live worksheet, and align the offline floor to the same cases.

src/desktop/wrappers/loadDashboardXml.ts
```diff
@@ -110,6 +115,87 @@ function resolveCanonicalDashboardName(
...
+  dashboardXml: string,
+  { executor, signal }: WithExecutorAndAbortSignal,
+): Promise<Result<void, Extract<LoadDashboardXmlError, { type: 'sheet-not-rendered' }>>> {
+  const worksheetZoneNames = collectWorksheetZoneNames(dashboardXml);
+  if (worksheetZoneNames.length === 0) {
+    return Ok.EMPTY;
+  }
+
+  const workbookResult = await getWorkbookXml({ executor, signal });
+  if (workbookResult.isErr()) {
```
myu404: ### [P2 · SUGGESTED] Do not bypass the render guard when its read fails

**Purpose**
The new preflight should fail safely when it cannot inspect the live worksheet state.

**Issue and impact**
`checkWorksheetsRendered` converts every `Err` result returned by `getWorkbookXml` into success here, after which both apply routes continue. The new test proves the whole-workbook route can dispatch after the first read fails; neither the later reread nor the per-sheet route re-runs the referenced-worksheet check. A transient failure therefore disables the protection and can surface the old native rejection instead of the actionable preflight error.

**Suggestion**
Return a retriable preflight/read error without dispatching, or evaluate the guard from the same successful locked workbook snapshot used by the apply.

myu404 (changes requested):

This is the right recovery boundary: reject a dashboard before dispatch and return a repairable worksheet error. The basic present-but-blank cases are covered for both apply transports, and the typed error remains actionable.

One required gap remains in this review: supported `type-v2='visual'` worksheet zones bypass the new collector. I also left a non-blocking note on the deliberate fail-open read path. Existing unresolved review coverage separately identifies `zone-pane` traversal, the unlocked read/apply race, and the extra full-workbook reads.

Static review and the reported green checks were used; I did not run the test suite or a live Tableau Desktop reproduction. Native `HasVisualDoc` equivalence still needs the controlled valid-viewpoints evidence requested in the existing discussion.

b-carver:
Thanks @artbeckTableau and @myu404 — pushed `34fdbfef` addressing the review. Summary of what changed, mapped to each finding:

**Correctness**
- **`type-v2='visual'` zones bypassed the collector** (myu404 P1) **and `<zone-pane>`-nested worksheets missed** (artbeck, `loadDashboardXml.ts:142`): `collectWorksheetZoneNames` now deep-walks the parsed subtree, treating every `zone` key at any depth as a nested `<zone>` and recursing through any intervening container (layout zone, `zone-pane`, etc.) — mirroring the authoritative `target-dashboard-invariant`'s `getElementsByTagName` walk. The predicate is now inclusive: a named zone counts when `@type-v2` is **absent or `'visual'`**. So both a direct visual zone and one wrapped in `zone-pane` are now collected.
  - Code/code inconsistency flagged in a comment: the older `dashboard-zones-reference-included-worksheets` rule uses the narrower `.//zone[@name and not(@type-v2)]` (excludes `visual`). The invariant is authoritative, so the guard follows it; raising the rule discrepancy to the maintainer rather than silently reconciling.
- **Blank-detection ignored non-`column`/`*field` attributes** (artbeck, `worksheetRenderState.ts:46`): `hasPlacedFieldReference` now also matches a single-datasource internal field token (`[<ns>:<name>:<suffix>]`, e.g. `[none:Category:nk]`) on **any** attribute, so a placed ref on a `groupfilter level`/`member` (or any shelf/encoding) counts as rendered. A bare datasource declaration name (`[Sample - Superstore]`) has no inner colons, so a sheet carrying only a datasource decl still classifies as blank — no false-positive block.

**Race + fail-open (artbeck `:172`, myu404 P2)** — both fixed by the same restructure: the render check now reads the live workbook **exactly once inside the apply lock** (the mutex is a single non-reentrant chain, so the read lives in each route's existing `withApplyLock` callback). On a read error it now **fails closed** — returns a retriable `execute-command-error` and dispatches nothing, instead of failing open and sending the dashboard anyway. This closes the read-then-apply window artbeck flagged and the transient-failure bypass myu404 flagged.

**Efficiency (artbeck `:282`)**
- **Double full-workbook fetch on the whole-workbook route**: eliminated. The in-lock guard reads once and hands the snapshot to `loadDashboardXmlViaExternalApi` to reuse for the upsert — one round trip / one parse.
- **Per-sheet route paid an unconditional fetch**: now gated — the guard short-circuits with **no fetch** when the dashboard names no worksheet zones. When it does fetch (zones present), it's the single in-lock read above. I did **not** narrow the read to only the referenced worksheets (still a whole-document read when it runs); happy to scope it further if you'd prefer, but that's the one efficiency note not fully closed.

**Convention (artbeck `:123`)**: em-dashes (U+2014) at 123/154/159/279/286 replaced with ASCII `--`, including the user-facing string at 286.

**Still open (not code):** the native `HasVisualDoc` equivalence evidence (controlled valid-viewpoints) myu404 called out remains outstanding — tracking separately, not in this push.

Tests: `loadDashboardXml` 27/27 (added fail-closed + single-fetch cases), `perSheetDocumentApply` 28/28, `worksheetRenderState` 11/11; tsc + lint clean on changed files. CI green.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: `collectWorksheetZoneNames` now deep-walks every `zone` key at any depth, recursing through `zone-pane`/layout/any container (mirroring the invariant's `getElementsByTagName` walk), so a worksheet zone wrapped in `<zone-pane>` is collected.

src/desktop/metadata/worksheetRenderState.ts
b-carver: Fixed in `34fdbfef`: `hasPlacedFieldReference` now also matches a single-datasource internal field token `[<ns>:<name>:<suffix>]` on ANY attribute (not just `column`/`*field`), so a placed ref on e.g. a `groupfilter level`/`member` counts as rendered. A bare datasource declaration (no inner colons) still reads blank, so no false-positive block.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: the live-workbook read now happens inside the route's existing `withApplyLock` callback (the same non-reentrant lock the apply uses), closing the unlocked read-then-apply window.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: the in-lock guard reads the workbook once and hands the snapshot to `loadDashboardXmlViaExternalApi` to reuse for the upsert — one round trip / one parse on the whole-workbook route. Test updated to assert a single fetch.

b-carver: Partially addressed in `34fdbfef`, leaving this thread open. The fetch is now **gated** — the guard short-circuits with NO fetch when the dashboard names no worksheet zones. But when a zone *does* name a worksheet it still reads the whole document (not scoped to just the referenced worksheets), so the latency note stands for that case. Happy to scope the read to the referenced worksheets in a follow-up if you'd prefer.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: em-dashes at 123/154/159/279/286 replaced with ASCII `--`, including the user-facing string literal at 286.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: the predicate is now inclusive — a named zone counts when `@type-v2` is absent OR `'visual'`, matching `target-dashboard-invariant`. Both a direct visual zone and a nested one are collected; added tests with a blank live worksheet on both forms.

src/desktop/wrappers/loadDashboardXml.ts
b-carver: Fixed in `34fdbfef`: the guard now fails CLOSED — a live-workbook read error returns a retriable `execute-command-error` and dispatches nothing, evaluated from the single in-lock snapshot both routes share. Test rewritten to assert no dispatch + a single fetch on read failure.

src/desktop/wrappers/loadDashboardXml.ts
artbeckTableau: Probably good to create some followup user story that is definitely scheduled out past GA.

PR #919, opened 2026-09-17
@W-24126644: fix(desktop): return the User-corrected field ref so readback stops flagging false drops

artbeckTableau:
## Decision

A one-off correctness fix that also pins a small invariant: `ensureColumnInstanceInDependencies`
must **return the same corrected reference it writes** as the column-instance. It writes `[usr:...]`
but currently returns the uncorrected `[ctd:...]`, and the returned value is exactly what the caller
places on the shelf/encoding -- so the two must agree.

## Description

When a calculated field already aggregates, `ensureColumnInstanceInDependencies` rewrites its
derivation to User (`[usr:...]`) to avoid double-aggregation and writes the column-instance under
that corrected name -- but it returned the *uncorrected* ref (e.g. `[ctd:...]`). The caller placed
that diverging ref on the shelf/encoding, Tableau reconciled the pill to `[usr:...]`, and post-apply
readback verification saw `intended != readback` and reported a false
"apply succeeded but Tableau silently dropped".

This propagates the correction to the returned value so the shelf/encoding ref matches the written
column-instance. It fixes both `addFieldToShelf` and `addFieldToEncoding`, which share this return
value (+4 lines in `fields.ts`).

## Motivation and Context

`silent_encoding_drop` is the #1 tooling error in the LangSmith eval corpus and is usually a false
alarm: it does not change the final answer, but it sends the agent into needless worksheet rebuilds
(trajectory-health drops ~30 points on affected traces while answer quality stays flat). This is one
concrete, self-inflicted source of those false drops. Tracked by W-24126644.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- New regression test in `fields.test.ts`: drives an aggregating calc column present in the workbook
  but absent from the worksheet `<datasource-dependencies>`, and asserts the placed shelf ref equals
  the written column-instance name. Fails before the fix
  (`expected '[Sample].[ctd:Calculation_1:qk]' to be '[Sample].[usr:Calculation_1:qk]'`), passes after.
- Adjacent suites pass locally: `fields` (56), `refineWorksheet` (43), `readback-verify` (36).
- `npm run lint` clean.

## Related Issues

- W-24126644 -- "silent_encoding_drop error message leads agent astray, sometimes taking down a valid
  viz" (P1 bug, TabCore Tableau Studio). This PR implements one of several fixes tracked there.

## Checklist

- [ ] I have updated the version in package.json via `npm run version` (see note below)
- [x] I have made any necessary changes to the documentation (none needed)
- [x] I have added tests that prove my fix is effective
- [x] I have documented any breaking changes in the PR description (none)

## Contributor Agreement

- [x] I have read the [CONTRIBUTING guidelines](../../CONTRIBUTING.md) and followed its Contribution Checklist.


src/desktop/metadata/fields.ts
```diff
@@ -950,6 +950,10 @@ function ensureColumnInstanceInDependencies(
             datasource,
           });
         }
+        // W-24126644: return the User-corrected ref (usr:) the caller places on the
+        // shelf/encoding. Diverging from the written column-instance makes Tableau
+        // reconcile the pill, which readback verification reports as a false drop.
+        correctedInstanceName = actualColumnInstanceName;
```
mattcfilbert, line 959: [P1] Please update encoding verification to use the corrected reference too. With the new regression fixture passed to `addFieldToEncoding(..., 'color', '[Sample].[ctd:Calculation_1:qk]', undefined, workbookXml)`, this return value makes it write `[usr:Calculation_1:qk]`, but line 199 still checks against the original `columnRef`. It throws `Failed to add encoding`, which `add-field` returns as `XmlModificationError` before apply. I reproduced this for Color, Size, and Text; the base returns XML for the same input. Compare against `correctedColumnRef` and add an encoding regression alongside the Rows case. — MattGPT

mattcfilbert (changes requested):
MattGPT automated review, using the Andy-review lens.

One P1: the shelf fix is sound, but aggregate fields added to Color, Size, or Text now fail before apply because encoding verification still checks the original reference. Please complete that caller change and cover the encoding path.

Verified at f4f9c56: the new Rows regression passes; the same fixture throws for Color, Size, and Text. Hosted CI is green. No live Desktop apply/readback was run; one broader local tool suite could not load because the reused dependency tree lacks jsonc-parser.

myu404 (commented):
## Review note

The change correctly propagates the `[usr:...]` column-instance reference through the Rows and Columns shelf path, and the package metadata is synchronized at 2.68.8.

I found no additional issues beyond the existing unresolved encoding-verification P1. I am leaving this as a comment-only review and not approving until that thread is addressed. Hosted CI is green. I did not rerun tests or perform a live Tableau Desktop apply/readback.

src/desktop/metadata/fields.ts
artbeckTableau: Fixed. `addFieldToEncoding` now self-verifies against `correctedColumnRef` (fields.ts ~199) instead of the caller `columnRef`, and the debug log + thrown error report the corrected ref too, so the write path and the verify path agree.

Added an encoding regression block that drives the same aggregating-calc fixture (present in the workbook, absent from the worksheet deps) through Color, Size, and Text via `it.each`, asserting the `[usr:...]` ref is written and no `[ctd:...]` remains. Each case throws `Failed to add encoding` before the change.

`fields.test.ts`: 59 pass (was 56, +3 encoding cases). Lint clean. Left the shelf path untouched -- it has no self-verify block and was already correct.

PR #920, opened 2026-09-17
feat(desktop): add filter-action support to author-action

diana-wang:
## Decision

Add support for [filter actions](https://help.tableau.com/current/pro/desktop/en-us/actions_filter.htm) in `author-actions` via adding a fourth mode, `filter`. It follows the same contract as the existing `parameter`/`set`/`url` modes.

Corresponding knowledge documentation updates were made in https://github.com/tableau/agent-to-tableau-desktop/pull/325. `resources/desktop/knowledge` in this repo is no longer being used so it was not updated.

## Description

`author-action` can now wire a source view's selected (or hovered) marks to filter a target sheet or dashboard.

- **New mode:** `mode: 'filter'` 
- New optional params 
  - `targetSheet` (sheet or dashboard name)
  - `filterFields` (fields that the filter should be restricted to)
  - `excludeSourceSheets` (sheets to exclude on the target dashboard if target is a dashboard)
  - `excludeTargetSheets` (sheets to exclude on the target dashboard if target is a dashboard)
- **Validation guards:** 
  - a source is required - either a sheet or a dashboard. If both are provided the sheet must exist on the dashboard
  - a target is required - must use `targetSheet` and must be a real sheet or dashboard; `targetParameter`/`targetSet`/`url` are rejected in filter mode
  - `excludeSourceSheets` and `excludeTargetSheets` are new optional params (necessary when the source or target is a dashboard and there are sheets that need to be excluded)
  - `filterFields`: caller-named fields (captions, friendly names, or bracketed names) resolve to internal column names via the shared `summarizeSchema`. An unknown field is a blocker that lists the available fields 
  - duplicate filter actions are rejected
- **Readback:** `hasFilterActionWithTarget` confirms the `tsc:tsl-filter` command and the source/target/params/etc round-trip before the tool reports success.
- **Route table (`instructions.ts`):** the `dynamic-authoring` route now recognizes mark interactivity intent and adds a `search-knowledge` consult step

Regarding fields:
  - *All Fields* (`filterFields` omitted) → a lone `<command command='tsc:tsl-filter'>` with `special-fields='all'`.
  - *Specific fields* (`filterFields` given) → a `tsl:` `<link>` of field locators plus a `<command>`, with the fields declared in sibling `<datasources>` / `<datasource-dependencies>` blocks inside `<actions>`. Tableau backfills the `sheet_link` group column on the target datasource at run time.

Also includes
- Refactors across all action modes for shared functionality, including generating the XML for the source and activation elements, and dedupe skeleton.

Other refactoring of `author-actions` as a whole will be done in a followup PR

## Motivation and Context

Filter actions are a commonly used form of dashboard interactivity, and `author-action` previously only supported parameter, set, and URL targets — so the utterance "make clicking this chart filter that one", for example, had no tool path.

## How Has This Been Tested?

Added unit tests in a new `authorAction.filter.test.ts` file, including refactoring out shared functions into `authorActionTestFixtures.ts`. Also manually tested with a Tableau Desktop build.

## Checklist

## Contributor Agreement

      its Contribution Checklist.
            
@W-24179782


src/desktop/instructions.ts
```diff
@@ -215,9 +215,9 @@ export const DESKTOP_ROUTE_TABLE: readonly DesktopInstructionEntry[] = [
     kind: 'route',
     id: 'dynamic-authoring',
     trigger:
-      'a dynamic ask or a calc/derived field the data lacks WITHOUT a conventional name (examples include running total and LOD)',
+      'a dynamic ask or a calc/derived field the data lacks WITHOUT a conventional name (examples include running total, LOD, and  interactivity on mark interaction such as by selecting or hovering over marks or selecting a link in a tooltip menu)',
```
tableaukyler, line 218: **Route mismatch.** `renderInstructionEntry` injects `For ${trigger}, ${action}` into MCP initialize, so every default-profile session now sees mark-interactivity as a calc/derived-field ask whose first step is `use author-parameter first` (also a `stopCondition`).

A click-to-filter utterance is not a missing LOD. Give filter/highlight a dedicated route. Do not send it through templates/`apply-worksheet`.

This action also requires `search-knowledge` before authoring, but this PR does not touch `resources/desktop/knowledge`. Vendored copy still maps `author-action` to parameter/set (`tactics/data/dynamic-dashboard-authoring.md`) and still teaches hand-authored `tsc:tsl-filter` (`tactics/dashboard/zones.md`). Until that corpus says `mode: 'filter'`, the consult step steers the agent into the crash-prone `apply-workbook` path this tool was meant to replace.

https://github.com/tableau/tableau-mcp/blob/5a8e7bd345838db970ac6f9285303646232b1afd/src/desktop/instructions.ts#L215-L240

tableaukyler (commented):
## KylerGPT review

Filter mode on `author-action` is the right Approach 3 path. The all-fields vs specific-field XML, field resolution via `summarizeSchema`, and the readback-on-target tests are in good shape. Shared `renderSourceAttrs` / `renderActivation` is the right extract. Empty `<activation />` for `on-menu` matches Desktop's Explicit/menu default, so I'm not treating the URL test rewrite as a regression.

The renderer is ahead of the handshake. As written, a "click this chart to filter that one" ask gets stuffed into the calc route, told to `author-parameter` first, then sent to `search-knowledge` against knowledge this PR did not update (`dynamic-dashboard-authoring.md` still maps `author-action` to parameter/set; `zones.md` still teaches hand-authored `tsc:tsl-filter`). Companion [agent-to-tableau-desktop#325](https://github.com/tableau/agent-to-tableau-desktop/pull/325) does not fix the copy vendored in this repo.

### Must fix
- **Route table** (`instructions.ts`): mark interactivity does not belong on `dynamic-authoring`. Give it its own trigger. Drop `use author-parameter first` for filter asks. Don't require `search-knowledge` until the vendored knowledge actually documents `mode: 'filter'`.

### Should fix
- **Readback** does not assert the `tsl:` `<link>` / field list. URL mode already requires the link. Specific-field success can be a lie.
- **Duplicate** keys only source+target. Tableau allows two filters on the same pair with different fields. Combined with create-only, the second ask is a dead end.
- **Success hint** claims a dashboard source forces a dashboard target. Your own test keeps a worksheet target when `sourceDashboard` is set.
- **Source existence**: you reject a typo target, then let a typo source apply and still pass readback.

### Nits (not blocking)
- New restating comments in `authorAction.ts` / filter tests / fixtures. Keep the tsl-encoding, sheet_link backfill, and silent-typo notes. Delete the "this function does X" ones. (NO COMMENTS)

### Docs (flag, not silent-fix)
[mcp-authoring-and-actions.md](https://github.com/sf-analyticscloud-playground/tableau-studio-docs/blob/main/docs/architecture/mcp-authoring-and-actions.md) and [adding-functionality-best-practices.md](https://github.com/sf-analyticscloud-playground/tableau-studio-docs/blob/main/docs/architecture/adding-functionality-best-practices.md) still say `author-action` is `parameter | set | url` and filter is raw-XML-only. This PR makes that stale. I am not guessing which side should move first.

Security: user params into TWB XML go through `escapeXml` (and `encodeURIComponent` on TSL locators). No injection finding.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -732,22 +1075,57 @@ function hasUrlActionDuplicate(
...
     const linkExpression = getAttr(linkTag, 'expression');
-    if (linkExpression === undefined || unescapeXml(linkExpression) !== expression) {
+    return linkExpression !== undefined && unescapeXml(linkExpression) === expression;
+  });
+}
+
+// Readback predicate for filter mode: the caption-matched legacy <action> must carry a
+// <command command='tsc:tsl-filter'> whose target param survived. An action that
+// persisted under a different command (or lost its target) is not a working filter action.
+function hasFilterActionWithTarget(xml: string, caption: string, target: string): boolean {
```
tableaukyler: **Readback does not prove specific-fields landed.** This predicate only checks caption + `tsc:tsl-filter` + `target`. Field identity lives on the `tsl:` `<link>` (`renderFilterAction` / `buildTslExpression`). If Desktop drops the link or rewrites to `special-fields='all'`, `applyAndVerify` still reports success.

URL mode already requires the `<link>` expression (`hasUrlActionWithLink`). Mirror that: for a specific-field filter, assert the link expression (and ideally the datasource-dependencies) survived.

Tests stub perfect `readbackXml`, so CI will not catch a live rewrite.

https://github.com/tableau/tableau-mcp/blob/5a8e7bd345838db970ac6f9285303646232b1afd/src/tools/desktop/authoring/datasource/authorAction.ts#L1088-L1107

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -732,22 +1075,57 @@ function hasUrlActionDuplicate(
...
+        getAttr(paramMatch[0], 'name') === 'target' &&
+        unescapeXml(getAttr(paramMatch[0], 'value') ?? '') === target,
+    );
+  });
+}
+
+// Dedup guard: the document-apply path appends, so a same-source + same-target filter
+// action authored under a different caption would silently double. Caption collision is
+// handled separately by hasActionCaption.
+function hasFilterActionDuplicate(
```
tableaukyler: **Identity is too coarse.** This keys only source + target. Tableau allows multiple filter actions on the same pair (Category vs Sub-Category, all-fields vs specific, on-select vs on-hover). URL dedup includes the payload expression; this one does not.

The route table also says the tool cannot edit or delete, so a follow-up "also filter Sub-Category" is a blocker after the first create. Include fields (and probably activation) in the duplicate key, or don't call it identical.

https://github.com/tableau/tableau-mcp/blob/5a8e7bd345838db970ac6f9285303646232b1afd/src/tools/desktop/authoring/datasource/authorAction.ts#L1110-L1129

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -399,6 +523,30 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
               hint: 'readback verified the <link> URL action; the source sheet/dashboard must expose marks that drive the action, and any <[Field]> references must resolve on the source view',
             });
           }
+          if (mode === 'filter') {
+            const scopeHint =
+              effectiveSourceDashboard.length > 0
+                ? `source scoped to dashboard '${effectiveSourceDashboard}', so target is the dashboard, not a single worksheet; `
```
tableaukyler: **Hint contradicts the XML you just applied.** Any non-empty `sourceDashboard` injects "target is the dashboard, not a single worksheet". Your test `honors a worksheet target even when sourceDashboard names the source dashboard` keeps `targetSheet: 'Details'`.

The model treats this hint as readback truth and will "fix" a working sheet-to-sheet filter into a dashboard-wide one (then hit the duplicate guard). Echo the actual `targetSheet`.

https://github.com/tableau/tableau-mcp/blob/5a8e7bd345838db970ac6f9285303646232b1afd/src/tools/desktop/authoring/datasource/authorAction.ts#L526-L547

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -247,25 +296,38 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
-                `'${trimmedDashboard}' is a worksheet, not a dashboard. Pass it as sourceWorksheet instead.`,
+                `'${effectiveSourceDashboard}' is a worksheet, not a dashboard. Pass it as sourceWorksheet instead.`,
+              ).toErr();
+            }
+          }
+          if (mode === 'filter') {
+            // A filter action names the sheet/dashboard it filters. A target that is not a
+            // real sheet or dashboard silently filters nothing, so reject a typo up front.
+            const trimmedTarget = targetSheet!.trim();
+            if (!worksheetNames.has(trimmedTarget) && !dashboardNames.has(trimmedTarget)) {
```
tableaukyler, line 364: You called out the silent no-op for a typo **target** and rejected it. Source only gets the wrong-kind check (dashboard name in the worksheet slot). A source that is in neither `worksheetNames` nor `dashboardNames` still applies, and readback only checks the target param, so the tool reports success on an action that never fires.

Same `Set`s are already in hand. Reject an unknown source the same way you reject an unknown target.

https://github.com/tableau/tableau-mcp/blob/5a8e7bd345838db970ac6f9285303646232b1afd/src/tools/desktop/authoring/datasource/authorAction.ts#L323-L332

src/desktop/instructions.ts
diana-wang: >This action also requires search-knowledge before authoring, but this PR does not touch resources/desktop/knowledge. Vendored copy still maps author-action to parameter/set (tactics/data/dynamic-dashboard-authoring.md) and still teaches hand-authored tsc:tsl-filter (tactics/dashboard/zones.md). Until that corpus says mode: 'filter', the consult step steers the agent into the crash-prone apply-workbook path this tool was meant to replace.

The knowledge changes are in https://github.com/tableau/agent-to-tableau-desktop/pull/325, so that's incorrect. `tactics/dashboard/zones.md` does still have references to `tsc:tls-filter` though so can clarify and consolidae that a bit more with the other docs changes in that repo

diana-wang: Updating instructions as well since we don't want to ask the user for permissions before adding a new action or editing/deleting an action.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Updated `hasFilterActionWithTarget` to check for link expressions and assert that it survived

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Updated. `hasFilterActionDuplicate` now also compares the <link> expression. And removed the edit/deletion comments from the route table

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Updated the hint to include the actual applied target, as well as both the worksheet and dashboard scopes if available
- worksheet + dashboard (viz-zone): source scoped to worksheet 'Profit' on dashboard 'Overview'
- worksheet only: source scoped to worksheet 'Profit'
- dashboard only: source scoped to dashboard 'Overview'

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Added sourceWorksheet/sourceDashboard existence checks against worksheetNames/dashboardNames, mirroring the target check

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -773,8 +1203,143 @@ function spliceActionIntoWorkbook(
...
+  actionXml: string,
+): Result<string, XmlModificationError> {
+  const siteResult = locateActionsSite(xml);
+  if (siteResult.isErr()) {
+    return siteResult.error.toErr();
+  }
+  const site = siteResult.value;
+
+  if (site.kind === 'existing') {
+    return new Ok(`${xml.slice(0, site.close)}${actionXml}${xml.slice(site.close)}`);
```
myu404: ### [P1 · REQUIRED] Insert legacy filter actions before later action families

**Purpose**
Workbook-root actions must retain the XSD order: legacy actions, datasource metadata, navigation actions, set actions, then parameter actions.

**Finding**
Both filter paths still append the new legacy `<action>` after later families. The all-fields helper inserts immediately before `</actions>`; the specific-fields helper removes only datasource blocks, leaves navigation/group/parameter siblings in place, and appends the legacy action after them. The mixed-family test still accepts parameter-then-legacy order, and the document guard does not validate XSD family order.

**Impact**
Adding a filter action to an already-interactive workbook can submit out-of-order XML. Desktop may reject, normalize, or drop action state; the pinned code does not establish which outcome occurs.

**Recommendation**
Insert the legacy action before datasource and later-family siblings, preserve those siblings in place, and cover both filter paths with mixed-family fixtures.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -229,6 +260,24 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
               ).toErr();
             }
           }
+          if (mode === 'filter') {
```
myu404, line 277: ### [P2 · SUGGESTED] Reject non-default fields that the selected action mode ignores

**Purpose**
A successful response should mean each meaningful supplied option affected the selected mode or was rejected.

**Issue and impact**
Filter mode accepts nonempty `sourceField`, nondefault `setMembership`, and URL-only settings that its renderer ignores; conversely, the new `targetSheet` and `filterFields` inputs are accepted and ignored by sibling modes. A caller can therefore receive success after explicit settings were silently omitted.

**Suggestion**
Use a discriminated schema or per-mode checks for semantically nonempty/nondefault foreign fields, accounting for schema-injected defaults, and test both directions.

src/desktop/instructions.ts
```diff
@@ -228,9 +228,12 @@ export const DESKTOP_ROUTE_TABLE: readonly DesktopInstructionEntry[] = [
       'list-available-fields',
       'build-worksheets-from-templates',
       'apply-worksheet',
+      'search-knowledge',
     ],
     stopConditions: ['use author-parameter first'],
-    requiredEvidence: ["each author-* verb's readback-verified result object"],
+    requiredEvidence: [
+      "each author-* verb's readback-verified result object. For actions, an interaction test that clicking or hovering over a mark filters the target",
```
myu404: ### [P2 · SUGGESTED] Use the tooltip menu to verify on-menu actions

**Purpose**
Required evidence should exercise the activation mode that was authored.

**Issue and impact**
The route covers tooltip-menu interactions and `author-action` supports `on-menu`, but the required evidence mentions only clicking or hovering. Following it literally cannot prove that an on-menu action fired.

**Suggestion**
Make the evidence activation-specific: click for `on-select`, hover for `on-hover`, and open the tooltip and select the generated link for `on-menu`.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -247,41 +296,115 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
+            // A filter action's source and target must both be real sheets/dashboards. A target that is not a
+            // real sheet or dashboard silently filters nothing, so reject a typo up front.
+            if (effectiveSourceSheet.length > 0 && !worksheetNames.has(effectiveSourceSheet)) {
               return new ArgsValidationError(
-                `'${trimmedWorksheet}' is a dashboard, not a worksheet. Pass it as sourceDashboard and leave sourceWorksheet empty (or set sourceWorksheet to a worksheet inside the dashboard) so the URL action is scoped to the dashboard.`,
+                `sourceWorksheet "${effectiveSourceSheet}" was not found. Available worksheets: ${worksheetNames.size > 0 ? [...worksheetNames].join(', ') : 'none'}`,
               ).toErr();
             }
-            if (trimmedDashboard.length > 0 && worksheetNames.has(trimmedDashboard)) {
+            if (
```
myu404, line 345: ### [P1 · REQUIRED] Validate the worksheet belongs to the supplied source dashboard

**Purpose**
A combined worksheet/dashboard source must identify a real viz zone that can fire the action.

**Finding**
The new checks prove that each name exists globally, but when both are supplied they never prove that the worksheet is present on that dashboard. The renderer combines the two values into one `<source>`, and readback verifies the target/link rather than the source relationship.

**Impact**
A globally valid but unrelated worksheet/dashboard pair can persist and be reported successful even though no mark on that dashboard can trigger it.

**Recommendation**
Resolve the dashboard's worksheet zones, reject a mismatched pair before apply, and add a no-apply test for two individually valid but unrelated names.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -732,39 +1100,101 @@ function hasUrlActionDuplicate(
...
+      return false;
+    }
+    if (linkExpression !== undefined) {
+      const linkTag = block.match(/<link\b[^>]*>/)?.[0];
+      const expression = linkTag === undefined ? undefined : getAttr(linkTag, 'expression');
+      if (expression === undefined || unescapeXml(expression) !== linkExpression) {
+        return false;
+      }
+    }
+    return true;
```
myu404: ### [P2 · SUGGESTED] Verify non-link filter semantics before reporting success

**Purpose**
A successful receipt should distinguish requested settings from behavior that survived readback.

**Issue and impact**
The new link comparison correctly catches a specific-field action rewritten to all-fields. The predicate can still succeed after source scope, activation, `auto-clear`/`on-empty`, exclusions, `single-select`, or dependency declarations are dropped, while the result echoes those settings as applied.

**Suggestion**
Verify the requested serialized semantics before returning success, or narrow the receipt so it does not describe unchecked behavior as readback-verified.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -732,39 +1100,101 @@ function hasUrlActionDuplicate(
...
+        getAttr(paramMatch[0], 'name') === 'target' &&
+        unescapeXml(getAttr(paramMatch[0], 'value') ?? '') === target,
+    );
+    if (!sameTarget) {
+      return false;
+    }
+    const linkTag = block.match(/<link\b[^>]*>/)?.[0];
+    const existingExpression =
+      linkTag === undefined ? undefined : unescapeXml(getAttr(linkTag, 'expression') ?? '');
+    return existingExpression === linkExpression;
```
myu404: ### [P2 · SUGGESTED] Include behavior options in filter duplicate identity

**Purpose**
Duplicate prevention should block an exact retry without preventing a distinct Tableau interaction.

**Issue and impact**
The field-expression fix now permits different fields, but duplicate identity still ignores activation, clear behavior, `single-select`, and exclusions. Same-source/target/field actions such as on-select versus on-hover, or show-all versus exclude-all, are behaviorally distinct yet are still rejected as identical.

**Suggestion**
Include the serialized behavior options in identity while retaining exact-retry deduplication, and test same-field actions that differ by activation and clearing behavior.

myu404 (changes requested):

Two blocking corrections from this delta review remain:

- Insert legacy filter actions before later action families so mixed-action workbooks preserve the XSD order.
- When both source fields are supplied, verify that the worksheet belongs to the source dashboard.

The existing mark-interactivity route thread is also a validated unresolved blocker. The new head fixes globally unknown sources, the specific-field link readback, field-sensitive duplicate detection, and the target-scope hint, so I am not duplicating those original comments. The remaining source-pair, broader readback, and behavior-identity issues are materially distinct and are delivered as current-line inline comments rather than replies to resolved threads. I will add only a same-thread follow-up for the still-open bundled-knowledge gap.

Suggested corrections unique to this review: reject meaningful fields ignored by the selected mode and verify `on-menu` actions through the tooltip menu.

Positive observations: the delta adds focused no-apply source tests, detects a specific-field action rewritten to all-fields, allows same-pair filters with different fields, and reports source and target scope independently. The hosted Node 22/24 builds, label check, security checks, and CLA status are green.

Limitations: I did not run builds, tests, linters, retrieval evals, or a live Desktop interaction. The PR description reports local unit tests and a manual Desktop test, but those runtime results were not independently reproduced here.

src/desktop/instructions.ts
myu404: **Additional context**
The companion PR updates the source knowledge, but this head's default package still reads and copies `tableau-mcp/resources/desktop/knowledge`: `src/desktop/assets.ts:254-288` falls back there when `TABLEAU_KNOWLEDGE_DIR` is unset, and `src/scripts/build.ts:67-68` copies this repository's `resources/desktop`. In that bundled corpus, `tactics/data/dynamic-dashboard-authoring.md` documents `author-action` only for parameter and set actions, while `tactics/dashboard/zones.md` still teaches hand-authored `tsc:tsl-filter` XML; no bundled guide documents `mode: 'filter'`. PR 325 alone therefore does not change the standalone/default package.

**Recommendation**
Bundle the synchronized typed filter-action guidance here, or enforce and verify a mandatory cross-repository sync before enabling this route.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Added support for ensuring actions including datasource/datasource-dependencies are in the correct order along with some tests

src/desktop/instructions.ts
diana-wang: this doesn't seem to be working, so I'm just going to revert this change and rely on the readback

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Going to do this in a followup since this authorActions as a whole needs a refactor

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Done

src/desktop/instructions.ts
diana-wang: https://github.com/tableau/agent-to-tableau-desktop/pull/325 is now updated with the zone.md changes

Ignoring the rest of the comment about updating `/knowledge` in this repo

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Done, added a FilterActions and used it in the creation, readback, and deduping functions so it's all shared.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Done, added a FilterActions and used it in the creation, readback, and deduping functions so it's all shared.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -686,13 +940,140 @@ function renderUrlAction({
...
+      type: match.type,
+    });
+  }
+  return new Ok(resolved);
+}
+
+// Build the raw tsl: sheet-link expression for a specific-field filter. Each field adds a clause
+// `<urlenc([ds].[field])>~s0=<[ds].[field]~na>` (locator URL-encoded on the left, raw in the <…~na>
+// token on the right), joined with '&'; the caller XML-escapes the result into <link expression>.
+function buildTslExpression(target: string, datasourceName: string, columnNames: string[]): string {
```
b-carver, line 1003: (AI Reviewed) `buildTslExpression` uses JS `encodeURIComponent` for the locator and drops the raw locator verbatim between `<` and `~na>`. That is correct for ordinary field names and matches the monolith fixtures exactly.

Edge case worth a note: Tableau's SheetLink layer applies its own escaping to the `<…~na>` selection token — the monolith integration test `SheetLinkTest.cpp::testCharacterEscape` shows `<`/`>` are **doubled** inside the token (`[Profit < 10]` round-trips as `[Profit << 10]`), and `&`/`<`/`>` can appear raw in field names. This function does not reproduce that escaping. 

The consequence is **fail-safe, not corrupting**: if a field's internal name contains `<`, `>`, or `&`, Tableau rewrites the expression on save, the readback comparison at `filterActionMatches` (unescapeXml(expression) === expected.linkExpression) no longer matches, and the tool returns the explicit "did not survive readback" error instead of claiming success. So this is a narrow limitation (such fields can't be authored via this path), not a data-integrity bug. Consider either escaping the token to match SheetLink semantics, or documenting/guarding field names containing `< > &` with a clearer up-front error. Non-blocking.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -454,6 +649,42 @@ function nextActionName(xml: string): string {
...
+    .join('');
+  return excludeChildren.length > 0
+    ? `<source${attrs}>${excludeChildren}</source>`
+    : `<source${attrs} />`;
+}
+
+// Serializes the <activation> element uniformly for every action type:
+// click ('on-select') and hover ('on-hover') write a type attribute; the tooltip-menu trigger
+// ('on-menu') writes none. auto-clear (the filter clear-on-empty flag) is written when set.
+function renderActivation(activation: z.infer<typeof activationSchema>, autoClear = false): string {
```
b-carver, line 682: (AI Reviewed) `renderActivation` always emits an `<activation ... />` element; for `on-menu` with no auto-clear that is a bare `<activation />`. Tableau itself tends to **omit** the `<activation>` element entirely in that case (observed in monolith `.twb` fixtures). 

This is harmless: the readback matcher (`filterActionMatches`, ~line 1216-1219) treats an absent `<activation>` tag as `type=undefined`/`autoClear=false`, which is exactly the `on-menu`/no-auto-clear case, so both forms compare equal and readback passes either way. Flagging only for fidelity with hand-authored Tableau output; no change required.

b-carver (approved):
(AI Reviewed) Deep review of filter-action support in `author-action`.

**Verdict: Approve.** The change is a clean, well-tested addition of a fourth `author-action` mode and does not regress the existing `parameter`/`set`/`url` modes. I independently cross-checked the emitted XML against Tableau monolith ground truth and it is byte-correct.

**What I verified against the monolith (`SheetLink.cpp`, `ActionSupport.cpp`, `ActionList.xsd`, and `.twb` fixtures):**
- Command string `tsc:tsl-filter`, activation attributes, and the five command params (`exclude`, `on-empty=none`, `single-select=''`, `special-fields=all`, `target`) match field-observed XML.
- The tsl `<link>` grammar is exact: `tsl:{urlenc(target)}?{urlenc([ds].[field])}~s0=<[ds].[field]~na>` joined by `&`, then XML-escaped. `buildTslExpression` reproduces the fixture (`expected.0112_...WithFqdnField...twb`) byte-for-byte.
- `<exclude-sheet>` source opt-outs and the sibling `<datasources>`/`<datasource-dependencies>` blocks are structurally valid for `tsc:tsl-filter`.
- XSD child ordering of `<actions>` is respected by the splice logic.

**Error handling / silent failures:** none found. Every failure path returns an explicit `.toErr()`, and readback failure returns a specific, mode-aware error rather than reporting false success. Field resolution blocks with an available-fields list instead of silently falling back to all-fields.

**Test coverage:** comprehensive — byte-faithful all-fields and specific-field emission, `clearSelection` mapping, all activation triggers, single-select, source/target exclusions, XSD family-order insertion, duplicate guards, and readback-failure cases.

Two minor, non-blocking notes are left inline.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -773,8 +1303,154 @@ function spliceActionIntoWorkbook(
...
+  if (additions.length > 0) {
+    merged[index] = block.replace(
+      '</datasource-dependencies>',
+      `${additions.join('')}</datasource-dependencies>`,
+    );
+  }
+  return merged.join('');
 }
 
 function getAttr(tag: string, name: string): string | undefined {
```
b-carver, line 1456: (AI Reviewed) Latent fragility in the shared `getAttr` helper (not introduced here, but this PR is now its heaviest consumer — ~21 new filter-path call sites, so flagging it).

The regex `\b${name}=(['"])(.*?)\1` uses `\b`, which matches between a non-word char (`-`, `:`) and a word char. So `getAttr(tag, 'name')` matches inside `param-name='A'`, `getAttr(tag, 'caption')` inside `field-caption=`, `getAttr(tag, 'role')` inside `semantic-role=`. When a hyphen/colon-suffixed decoy attribute precedes the real one in the tag, the decoy's value wins.

Repro: `getAttr("<x param-name='A' name='B' />", 'name')` returns `"A"` (expected `"B"`).

**Not reachable on the filter path today:** none of the tags parsed here (`<action>`, `<source>`, `<command>`, `<param>`, `<activation>`, `<link>`) carry a hyphen/colon-suffixed sibling of the attribute names actually queried (caption, name, value, worksheet, dashboard, command, type, auto-clear, expression). The one genuinely colliding pair — `role` vs `semantic-role` — is only queried on the calc path, and Tableau serializes attributes alphabetically so the real `role` wins there anyway. So this is fail-silent-wrong-value in principle but has no reachable trigger via authored input.

One-char hardening: anchor the name on a preceding delimiter instead of `\b`, e.g. `(?:^|[\s"'])${name}=(['"])(.*?)\1`. Non-blocking.

b-carver (commented):
(AI Reviewed) One follow-up note from a deeper edge-case sweep. Non-blocking — this is latent, pre-existing parser fragility, not a regression this PR introduces, and I could not construct a reachable trigger on the filter path. Approval stands.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -35,19 +40,17 @@ const paramsSchema = {
...
-  url: z
-    .string()
-    .optional()
-    .describe(
-      'URL for url mode. Pass it raw and unescaped (the tool escapes it). Use <[Field Name]> to insert a field value.',
-    ),
+  url: z.string().optional().describe('URL for url mode, raw. <[Field Name]> = value.'),
   sourceDashboard: z.string().optional().describe(''),
-  excludeSheets: z.array(z.string()).optional().describe(''),
+  excludeSourceSheets: z.array(z.string()).optional().describe(''),
```
myu404, line 52: ### [P1 · REQUIRED] Preserve the existing excludeSheets input

**Purpose**
A patch release should not silently change the public request shape of an existing action mode.

**Finding**
The base schema exposes `excludeSheets` for URL actions with a dashboard source. This head removes that field and replaces it with `excludeSourceSheets`, without a compatibility alias or documented breaking migration. Existing callers that send `excludeSheets` can have the field rejected or stripped, so the action is authored without the requested source exclusions.

**Impact**
Previously valid URL-action requests change behavior under a `2.68.8` to `2.68.9` patch upgrade and can report success after excluded sheets become active sources.

**Recommendation**
Continue accepting `excludeSheets` as a deprecated alias for `excludeSourceSheets` during a migration window, or restore the original name and add the target-specific input separately.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -130,25 +150,57 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
+            .filter((sheet) => sheet.length > 0);
+
           if (caption.trim().length === 0) {
             return new ArgsValidationError('caption empty').toErr();
           }
-          if (mode !== 'url' && sourceWorksheet.trim().length === 0) {
+          if (mode !== 'url' && mode !== 'filter' && effectiveSourceSheet.length === 0) {
             return new ArgsValidationError('sourceWorksheet empty').toErr();
           }
+          if (effectiveExcludedTargetSheets.length > 0 && mode !== 'filter') {
```
myu404, line 187: ### [P1 · REQUIRED] Reject exclusions that cannot affect the selected scope

**Purpose**
A successful receipt should mean every requested source or target exclusion can apply to the selected dashboard scope.

**Finding**
The current checks only restrict source exclusions to a dashboard-only source and target exclusions to filter mode. They do not require the target itself to be a dashboard, and neither list is checked against the corresponding dashboard's worksheet zones. A worksheet target or an arbitrary sheet name is still serialized and echoed as applied.

**Impact**
Callers can receive a readback-verified success even though an explicit exclusion may be ignored or have no effect.

**Recommendation**
Require target exclusions to use a dashboard target, and validate every source/target exclusion against that dashboard's worksheet zones before mutation.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tableau/mcp-server",
   "description": "Helping agents see and understand data.",
-  "version": "2.68.8",
+  "version": "2.68.9",
```
myu404: ### [P3 · MINOR] Keep the lockfile version synchronized

**Purpose**
The release bump should use the repository's canonical npm workflow and leave one version across committed package metadata.

**Finding**
This changes `package.json` to `2.68.9`, but `package-lock.json` still reports `2.68.8` in both root version fields. Recent version bumps and `npm run version:patch` update these files together.

**Impact**
The committed release metadata disagrees and later version operations can rewrite the lockfile, although build and publish use `package.json` and remain unaffected.

**Recommendation**
Regenerate the patch bump through the documented npm version script so both lockfile version fields become `2.68.9`.

myu404 (changes requested):

Because the branch was rewritten after `c3766253691e`, I re-reviewed the complete base-to-`736b146b25bb` diff. XSD family ordering and duplicate behavior are corrected; source membership and full readback remain partial, two prior contracts remain open, and the rewritten public surface introduces two new blockers plus a release-metadata mismatch. The existing mark-interactivity routing blocker also remains open in its current thread.

**Previous findings:** 2 addressed, 2 partial, 2 open, 1 obsolete, 0 awaiting verification.
**New findings:** 3.

No local tests or live Tableau interaction were run; hosted Node 22/24 and security checks are green at the pinned head.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: **Follow-up**
The rewritten head adds several useful cross-mode checks, but filter mode still accepts meaningful `sourceField`, non-default `setMembership`, and URL-only options that it does not render; sibling modes still accept `targetSheet` and `filterFields`. A successful response can therefore omit explicit caller intent.

**Recommendation**
Keep this open until semantically nonempty or nondefault foreign fields are rejected per mode, with tests in both directions.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: **Follow-up**
The new check rejects an unrelated pair when a `<zones>` block is present, but it fails open when `findDashboardZoneNames` returns `undefined`. The test at `authorAction.filter.test.ts:272-297` codifies accepting a combined source on a self-closing dashboard, so this path can still report success without establishing that the worksheet is on the dashboard.

**Recommendation**
Reject combined sources when membership cannot be inspected, or ground the fallback with a field-observed workbook shape proving that this representation can validly contain the worksheet.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: **Follow-up**
The shared matcher now verifies the action-local source, target, activation, clearing, exclusions, single-select, and link expression. It still never inspects the sibling `<datasources>` and `<datasource-dependencies>` required by a specific-field filter, although the original finding explicitly included dependency declarations.

**Recommendation**
Before returning the readback-verified receipt, confirm the expected datasource entry and every resolved dependency column survived; use the same complete semantic check for dedupe.

src/desktop/instructions.ts
myu404: **Follow-up**
The companion knowledge PR is improved, but this package still falls back to `resources/desktop/knowledge` when `TABLEAU_KNOWLEDGE_DIR` is unset, and the build still copies that local corpus. This PR does not synchronize it or require an external root, so the new route can still consult guidance that lacks typed filter mode.

**Recommendation**
Keep the thread open until the default bundle is synchronized or every supported runtime is verified to supply the authoritative external knowledge root.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Can bump the version from 2.69.0 instead of 2.68.9 but otherwise I don't think the rest of this comment is valid. Nothing currently consumes `excludeSheets` other than this endpoint (`author-action`)

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Done - verified that all `excludedSourceSheets` are on a `sourceDashboard` (which is a dashboard and not a worksheet) and that `excludedTargetSheets` are on a `target` which is a dashboard.

package.json
diana-wang: Updated this file. Utilize 2.69.0 though due to params changes

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: still going to leave this for followup PRs since refactoring needs to be done around this as a whole, including validating more parameter values are left out. If the actions end up being split out into separate different tools (e.g. a `author-filter-action` tool, a `author-parameter-action` tool, etc. then each tool will have it's own set of parameters anyway

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Added a check on `<datasources>` and `<datasource-dependencies>` for the readback after creating a new action. For the `hasFilterActionDuplicate` left it just to comparing the link expressions since when modifying an existing XML, `<datasources>` and `<datasource-dependencies>` could be located within a worksheet or dashboard element.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Updated so that when `sourceWorksheet` and `sourceDashboard` are both passed but `sourceDashboard` has no zones, it returns an error.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: leaving this alone since Tableau's ok with the empty output

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: addresed

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: addressed - updated it to match the behavior in TUrl:Tscape

b-carver (approved):
(AI Reviewed) Re-reviewed the updates against the previous review comments. All four findings are addressed, and I verified each against the current head (63cd7b4) rather than the commit messages. CI is green on both build matrix jobs, SAST, and credential scanning.

**#1 — `getAttr` substring-match (was Important/latent): FIXED.** The regex now anchors the attribute name on a preceding delimiter — `(?:^|[\s"'])${name}=` instead of `\b${name}=`. A decoy like `param-name='A'`, `user:name='A'`, or `semantic-role='A'` no longer wins over the real attribute. The function is now exported and unit-tested with exactly the repro I raised (`getAttr("<x param-name='A' name='B' />", 'name')` → `'B'`), plus the colon and `semantic-role` variants. No regression: a real attribute is always preceded by whitespace or the tag open.

**Field-name escaping (was Minor, fail-safe): FIXED — and upgraded from fail-safe to correct.** `buildTslExpression` now mirrors Tableau's SheetLink serializer: the `~s0` (left) side runs a strict `tslUrlEscape` (full `TUrl::URLEscape` — `[ ] < > & %` all become `%XX`, and it additionally encodes `! * ' ( )` that `encodeURIComponent` leaves alone), while the `~na` (right) side doubles `<`→`<<`/`>`→`>>` with `&`/`%`/`~`/`[`/`]` left raw. This matches the monolith `SheetLinkTest.cpp::testCharacterEscape` behavior I referenced. A new test round-trips a field named with `<`, `>`, and `&` and asserts success with byte-exact XML. No regression for ordinary names: `~` is RFC-3986 unreserved so the `~s0`/`~na` suffixes pass through unchanged, and the angle-bracket doubling is a no-op — output is byte-identical to before for names without `< > &`.

**#2 — on-menu `<activation />` vs omitted tag (was Nit): unchanged, as expected.** I had marked this "no change required"; the readback matcher already treats an absent `<activation>` as `type=undefined`/`autoClear=false`, so both forms compare equal. Correct to leave.

Readback safety re-confirmed: the built expression is stored as `expected.linkExpression` and readback compares `unescapeXml(expression)` against that same value, so the encoding change cannot desync the render and match paths.

Approving.

src/tools/desktop/authoring/datasource/authorAction.ts
myu404: Thanks for the response! I agree and no further action needed.

PR #921, opened 2026-09-17
@W-24101233: [Tableau Studio] Tableau Agent: route author-calc create through External Client API (tabdoc:create-calc + tabdoc:apply-calculation) instead of direct XML injection

sarahkim-tableau:
## Decision

The author-calc **create** path now goes through the monolith calc-authoring commands (`tabdoc:create-calc` + `tabdoc:apply-calculation`) over the External Client API, instead of splicing a `<calculation>` column into the workbook document. This is the new rule for create: Tableau owns field positioning and infers role/datatype from the formula, so author-calc no longer sends either. (W-24101233, Phase 2 — follow-up to PR #895.)

The splice path (`prepareCalculationsInWorkbook`, used by bind-template) still writes role/datatype and is unchanged.

## Wire contract

- The apply result key is `calculationApplyResult` (camelCase), matching the rest of the External API result surface — **not** the kebab-case `DPI_` param name. Same for the error string (`errorStr`).
- **Live-proven:** the `succeed` value and the camelCase convention are confirmed against a real Tableau Desktop apply (smoke-tested via MCP Inspector: happy path, string-calc role inference, caption collision, idempotent retry — all verified in Desktop).
- **Asserted from monolith source, not live-proven:** the two failure enum values `invalid-caption-for-new-calc` and `invalid-formula`. They come from `codegen/doc/enums.data`; I could not force `apply-calculation`'s own rejection in the smoke test because the pre-apply collision check and pre-apply formula validate catch those cases first.
- **Fail-closed:** an unrecognized apply result is surfaced as an execution error, never a false success — so a mis-cased failure value would degrade an error *message*, not produce a phantom create.

## Testing

- `authorCalc.test.ts`: 48 tests pass. Mocks updated to the camelCase key so they pin the real wire shape rather than the earlier (wrong) assumption.
- Lint clean.
- Live smoke test in Tableau Desktop: all four behaviors verified.


src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -557,6 +591,115 @@ async function validateCalcFormula({
...
+  | { status: 'created' }
+  | { status: 'invalid-caption'; message: string }
+  | { status: 'invalid-formula'; message: string }
+  // Infra failure: the command could not run or returned a result we don't recognize. Carries the
+  // typed error so the caller aborts rather than misreporting a verdict on the calc.
+  | { status: 'error'; error: DesktopCommandExecutionError };
+
+// Creates one calculated field through the monolith calc-authoring commands: create-calc opens a
+// pending calc on the active datasource, apply-calculation commits its caption + formula. Tableau
+// assigns the internal Calculation_N name and infers role/datatype, so neither is sent. The
```
myu404, line 612: ### [P1 · REQUIRED] Preserve the requested role and datatype

**Purpose**
GUS requires the command migration to preserve today's role and datatype behavior, and callers should be able to rely on accepted public inputs.

**Finding**
`author-calc` still accepts, defaults, validates, and forwards `role` and `datatype`, but this new command path sends neither, does not verify them on readback, and removes them from idempotent matching. Tableau's inferred metadata can therefore differ from the explicit request while the tool reports success or successful reuse.

**Impact**
Supported combinations such as a string calculation requested as a measure can come back with different metadata, changing downstream aggregation, shelf placement, or template behavior. A later retry with the same caption/formula also conceals that mismatch.

**Recommendation**
Preserve and verify the requested metadata through a supported native follow-up. If inference is intentionally becoming the public contract, explicitly remove or deprecate these inputs and update the affected callers and acceptance criterion instead of silently ignoring them.

src/tools/desktop/authoring/datasource/authorCalcCore.ts
```diff
@@ -557,6 +591,115 @@ async function validateCalcFormula({
...
+  const create = await executor.executeCommand({
+    namespace: 'tabdoc',
+    command: 'create-calc',
+    args: { datasource: datasourceName },
+    signal,
+  });
+  if (create.isErr()) {
+    return { status: 'error', error: new DesktopCommandExecutionError(create.error) };
+  }
+  if (create.value.status !== 'completed') {
```
myu404, line 635: ### [P1 · REQUIRED] Do not apply when create-calc failed to initialize this calculation

**Purpose**
Each requested calculation must operate on newly initialized state for its selected datasource and receive its own terminal outcome.

**Finding**
This branch treats every completed `create-calc` envelope as successful. Current monolith first auto-applies any existing calculation-controller model; when that returns a non-success result, `create-calc` returns before `NewCalculation`. The following `apply-calculation` therefore operates on the prior pending model even though the envelope was completed.

**Impact**
After a native apply rejection leaves pending state, a later item or call can apply its caption/formula to the previous calculation. With different datasources, the old datasource can be mutated before the new datasource's scoped readback reports failure.

**Recommendation**
Inspect `create.value.result.calculationApplyResult` and proceed only on `succeed`; otherwise fail before dispatching `apply-calculation`. Add a regression test for a completed `create-calc` carrying a non-success result, including a subsequent request for another datasource. Use any controller cleanup only after its exposure and ownership semantics are verified.

src/tools/desktop/authoring/datasource/authorCalc.test.ts
```diff
@@ -633,18 +454,29 @@ async function getToolResult({
   args,
   initialXml = BASE_XML,
   readbackXml,
+  applyResult = 'succeed',
 }: {
   args: AuthorCalcArgs;
   initialXml?: string;
   readbackXml?: string;
+  // The calculationApplyResult the mocked apply-calculation returns. 'succeed' is the happy path;
+  // 'invalid-caption-for-new-calc' / 'invalid-formula' exercise the per-calc failure branches.
```
myu404, line 463: ### [P2 · SUGGESTED] Add AC5 coverage for apply-time rejection and mixed batch ordering

**Purpose**
GUS AC5 calls for invalid apply-result and multi-calculation ordering coverage on the new command path.

**Issue and impact**
The helper defaults every `calculationApplyResult` to `succeed`, so these tests never exercise `invalid-caption-for-new-calc` or `invalid-formula`. The later batch cases cover validation-time failures, not a mixed batch where an apply fails after dispatch; apply-result mapping, independent continuation, dependent skipping, or command/readback order can regress without a focused signal.

**Suggestion**
Add producer-shaped rejection fixtures and a mixed batch assertion covering mapped outcomes, skipped dependents, independent continuation, datasource anchoring, and command/readback order.

myu404 (changes requested):

This change moves the `author-calc` create path from workbook XML splicing to `tabdoc:create-calc` plus `tabdoc:apply-calculation`, while keeping bind-template on its existing splice path. GUS W-24101233 requires parity for caption, formula, role, datatype, and folder placement; structured per-calc failures; caller compatibility; and create/rejection/ordering tests.

### Required

- **Preserve the requested role and datatype.** The public tool still accepts both fields, but the new path neither sends nor verifies them and retry matching now ignores them.
- **Do not apply when create-calc failed to initialize this calculation.** A completed native envelope can contain a failed prior-apply result without opening the requested calculation, so the subsequent apply can target stale controller state.

### Suggested

- **Add AC5 coverage for apply-time rejection and mixed batch ordering.** The new tests default apply results to success and do not exercise the post-dispatch rejection branches or a mixed command-path batch.

### Positive observations

Datasource selection, activation, command targeting, and polled readback consistently use the resolved internal datasource name. Unknown apply results fail closed, wrong-datasource readback does not settle, and bind-template remains isolated on `prepareCalculationsInWorkbook`. All reported GitHub checks are green.

### Limitations

This was a read-only review: I did not rerun tests, lint, typecheck, or a live Desktop workflow. The PR's live smoke results were author-reported. Native command lifecycle and enum evidence came from current monolith HEAD `f7fefa34104750816da95a72d4ac405e6f4a336c`, which is not a producer commit pinned by this PR. The `calculationApplyResult` / `errorStr` camelCase projection was not independently executed.

PR #927, opened 2026-09-21
W-24132500: Label datasource type on query-datasource path (Half B)

ycath:
## Description

Implements **Half B of W-24132500** — datasource-type labeling on the query-datasource path — using the **caller-supplied optional param** approach (decision B3).

Adds an optional `datasourceType` (`'published' | 'embedded'`) param to `query-datasource`. The type is discovered upstream by get-workbook, not resolved server-side, and is treated as an **untrusted hint** — it never gates or short-circuits control flow. Two agreed decisions:

1. **Label surface = telemetry/logs only.** On a successful query the resolved type is written to logs (observability). It is **not** added to `QueryDatasourceResult`; the tool's success output shape is unchanged.
2. **Param role = enrich the error message.** When the caller passes `datasourceType: 'embedded'` and the workbook-datasource gate fires, `WorkbookDatasourceNotEnabledError` names the embedded/workbook case specifically. Otherwise it falls back to the existing generic-but-actionable text.

### Changes
- `src/sdks/tableau/types/lineageContent.ts`: extract shared `datasourceTypeSchema` (`z.enum(['published','embedded'])`); reuse in `lineageContentSchema` and the new param.
- `src/tools/web/queryDatasource/queryDatasource.ts`: add optional `datasourceType` param (with `.describe()`), thread through the callback, log it on success, and pass it into the gate error.
- `src/errors/mcpToolError.ts`: `WorkbookDatasourceNotEnabledError` takes an optional `datasourceType`; `getErrorText()` tailors the sentence for `'embedded'`, else generic.
- `descriptions/queryDescription*.ts`: document the param for the model across all three version variants.

## Motivation and Context

Half A (PR #916) recognizes the `VDSForWorkbookDatasources` gate (HTTP 403 / `403800`) and surfaces an actionable error. Half B labels the datasource type on the query path so callers/telemetry know whether an embedded vs published source was queried, and so the gate error can name the embedded case precisely. Half A detection logic is untouched.

## Type of Change

- [x] New feature

## Version Bump

- [ ] Patch
- [x] Minor
- [ ] Major

## How Has This Been Tested?

- Extended `queryDatasource.test.ts`: success with `datasourceType: 'embedded'` logs the label and leaves the output shape unchanged; gate + `embedded` yields the tailored message; gate + omitted yields the generic message; param omitted is a behavior-identical regression guard; and the hint is never forwarded to VDS.
- Added `mcpToolError.test.ts` covering both `WorkbookDatasourceNotEnabledError` text variants.
- `npm test` (3213 passing), `npm run lint`, and `tsc --noEmit` all green locally.

## Notes / Follow-ups

- **No new auth scope** — query-datasource already requires `tableau:viz_data_service:read`.
- AC#3 (flag-enabled POC) and AC#4 (live/recorded integration test) are environment-gated and remain follow-ups.
- Per AC#1 interpretation: with telemetry-only + caller-supplied, the label is in logs only and only when the caller already knew the type (deliberate MVP). If a PO later reads AC#1 as "the query result tells the user what type it queried," switch the label surface to `QueryDatasourceResult`.

## Checklist


src/errors/mcpToolError.ts
```diff
@@ -93,16 +94,24 @@ export class FeatureDisabledError extends McpToolError {
...
     });
+    this.datasourceType = datasourceType;
   }
 
   override getErrorText(): string {
-    return 'This data source is embedded in a workbook, and querying workbook (embedded) data sources is not enabled on this Tableau site. This capability is opt-in per site: a site administrator must enable it before this data source can be queried.';
+    if (this.datasourceType === 'embedded') {
+      return 'This data source is embedded in a workbook, and querying workbook (embedded) data sources is not enabled on this Tableau site. This capability is opt-in per site: a site administrator must enable it before this data source can be queried.';
+    }
+    return 'Querying workbook (embedded) data sources is not enabled on this Tableau site. This capability is opt-in per site: a site administrator must enable it before this data source can be queried.';
```
williamyen0203, line 114: What's the difference here?

src/tools/web/queryDatasource/queryDatasource.ts
```diff
@@ -43,6 +45,14 @@ const paramsSchema = {
   datasourceLuid: z.string().nonempty(),
   query: querySchema,
   limit: z.number().int().min(1).optional(),
+  datasourceType: datasourceTypeSchema
```
stephendeoca, line 48: I don't think it makes sense to add datasource type as a tool parameter here if all it does is help provide more descriptive error messages.

If the client is providing the datasourceType when calling this tool, that means it already knows what type of datasource its using. So you can just return a generic message for both cases (publish and embedded) and the LLM can determine which the case is relevant to the error

Also since datasourceType is optional, it doesn't have much value when logging since the LLM might not always provide the type so logging might now always occur.

ycath:
Closing this out. The full scope of W-24132500 — gate detection, `WorkbookDatasourceNotEnabledError` with the actionable opt-in message, the `adminInsightsToolBase` union widening, and **unit tests at both the SDK and tool layers** — already shipped and merged in #916. The story's "unit tests at SDK and tool layers" AC is satisfied there (`vizqlDataServiceMethods.test.ts`, and `queryDatasource.test.ts` which asserts the gate → `WorkbookDatasourceNotEnabledError` path).

This PR had been scoped down (per @stephendeoca / @williamyen0203's feedback on the original `datasourceType` param) to just a dedicated unit test for the error class — which is redundant with the coverage already merged. So there's nothing left here worth landing. The embedded-vs-published telemetry motivation lives in its own story, W-24293159.

Closing rather than merging.

PR #928, opened 2026-09-21
@W-24132506: Update query/metadata tool descriptions for embedded datasources

ycath:
## Summary

Text-only change to the **tool descriptions** for `query-datasource` and `get-datasource-metadata` so that discovery and query agree on embedded (workbook) datasources.

`get-workbook` now surfaces embedded datasources as `upstreamDatasources` entries tagged `datasourceType: 'embedded'`, each carrying a VDS-queryable `luid`. Both data tools already *accept* those embedded LUIDs (the VDS path takes any `datasourceLuid`), but their descriptions still said "published data," discouraging the model from using the embedded LUID it was just handed.

No schema, param, control-flow, or auth/scope changes.

## Changes

The two tools are treated **asymmetrically** on purpose:

- **`query-datasource`** (all three version variants — base, 2025.3, 2026.1): the description now states the `datasourceLuid` may be published **or** embedded (including the embedded LUIDs from `get-workbook`'s `upstreamDatasources`), that the **embedded copy is often preferred** when both exist (workbook-local field semantics/calcs), and that querying embedded requires the site to have enabled workbook-datasource querying — a clear "not enabled" error is expected, not a query bug. This tool already recognizes the flag-off case and returns an actionable `WorkbookDatasourceNotEnabledError` (PR #916), so it's safe to encourage embedded here.

- **`get-datasource-metadata`**: the description now states the `datasourceLuid` may be published or embedded, that the richer Metadata-API context (descriptions, dataCategories, roles) is **published-only** and may be absent for embedded, and the grounding sentence is now type-neutral. Kept **factual/neutral** (does not encourage embedded) because this tool does **not** yet recognize the flag-off case and returns a misleading generic "VizQL Data Service is disabled" error for embedded LUIDs on flag-off sites.

## Testing

- `npm test` — all 3204 unit tests pass
- `npm run lint` — clean

No tests assert on the description text (existing tests only check that `description` is a String / Provider), so none needed updating.

## Follow-ups (out of scope)

- Update the docs-site pages under `docs/docs/tools/` to match.
- Give `get-datasource-metadata` the `WorkbookDatasourceNotEnabled` error recognition (Half A equivalent), after which its description could encourage embedded symmetrically.


src/tools/web/queryDatasource/descriptions/queryDescription.2026.1.ts
```diff
@@ -1,6 +1,6 @@
 export const queryDatasourceToolDescription20261 = `# Query Tableau Data Source Tool
 
-Executes VizQL queries against Tableau data sources to answer business questions from published data. This tool allows you to retrieve aggregated and filtered data with proper sorting and grouping.
+Executes VizQL queries against a Tableau data source (identified by \`datasourceLuid\`) to answer business questions, retrieving aggregated and filtered data with proper sorting and grouping. The \`datasourceLuid\` may be a published or an embedded (workbook) data source — including embedded LUIDs from \`get-workbook\`'s \`upstreamDatasources\` (\`datasourceType: 'embedded'\`); prefer the embedded copy when both exist, as it carries the workbook's local field semantics and calculations. Embedded querying must be enabled on the site — when a LUID comes from \`get-workbook\`, its \`isQueryable\` flag tells you up front whether you can query it (\`false\` → skip); otherwise, querying a data source that isn't enabled returns a clear "not enabled" error.
```
stephendeoca: Can you make a new snap of getDatasourceMetadata description and queryDatasourceDescription which corresponds to the 264 release. Then for the descriptions of older versions maybe mention that the tools is only available for published datasources.

src/tools/web/getDatasourceMetadata/getDatasourceMetadata.ts
```diff
@@ -106,12 +120,13 @@ export const getGetDatasourceMetadataTool = (
     server,
     name: 'get-datasource-metadata',
     minRequiredRole: SiteRole.VIEWER,
-    description: `
-    This tool retrieves metadata for a specified datasource by taking the basic, high level, metadata results from Tableau's VizQL Data Service and enriches them with additional context provided by Tableau's Metadata API.
-    The metadata provided by this tool consists of the datasource model, fields, and parameters that belong to the datasource.
-    Fields will contain properties such as name and dataType, but may also expose richer context such as descriptions, dataCategories, roles, etc.
-    This tool should be used for getting the metadata to ground the use of a tool that queries Tableau published data sources.
-    `,
+    description: getResultForTableauVersion({
```
stephendeoca: I believe you need to define a `new Provider` here like we do in queryDatsource for the description.

stephendeoca (approved):

PR #930, opened 2026-09-21
@W-24195331: Add scaffold-data-app tool

jarhun88:
## Summary

Adds a new `scaffold-data-app` MCP web tool that scaffolds a starter Tableau data app workspace: a starter Tableau viz (worksheet) extension that queries a published datasource live via the Extensions API.

- Given a single `datappName`, derives the extension package id, display name, and author, and returns a ready-to-edit workspace — a workbook plus an extension package (`manifest.json`, `index.html`, `src/app.js`).
- Optionally wires a published datasource into the workbook server-side, via `datasourceLuid` (and, to narrow it, `fields`).
- Output is written to disk under `DATA_APP_WORKSPACE_ROOT`, or — when `MCP_S3_BUCKET` is configured — zipped and uploaded to S3 with a short-lived presigned URL returned.
- Gated behind the `tableau-data-apps` feature flag (**default off**) and requires Explorer (Can Publish) role or higher.
- No REST API calls when `datasourceLuid` is omitted; when supplied, calls `resourceAccessChecker.isDatasourceAllowed`, a REST `queryDatasource`, and VizQL Data Service `fetchFieldsResult` to resolve fields for wiring.
- Extracted `fetchFieldsResult` out of `get-datasource-metadata` so both tools share the same field-resolution logic.

Includes fixes from an earlier review round (feature-flag default corrected to `false`, a stale passthrough-auth test comment corrected to reflect the `datasourceLuid`-conditional REST behavior, and documentation corrected to describe the optional params and conditional API calls).

## Test plan
- `npx vitest run` — 201 files / 3305 tests passing.
- `npm run build` — succeeds (default bundle + MCP Apps).
- Independently reviewed via dual-reviewer + synthesis gate, canonical `VERDICT: PASS` — covering correctness, conflict-resolution integrity (this branch was rebased onto the latest `main`), scope, test coverage, and this repo's house review conventions (OAuth scope coverage, tool-description economy, error handling, Zod validation).

— Jaehun Bot


src/scripts/build.ts
```diff
@@ -92,6 +93,19 @@ const globalValues: Record<GlobalIdentifierName, string> = {
   );
   console.log('✅ features.json copied successfully');
 
+  // scaffold-data-app serves a static, un-substituted template zip (both local and S3 modes) from
+  // an asset bundled next to index.js (same idiom as features.json). Rebuild the zip fresh from the
+  // committed template tree so it can't drift, then ship only that one file — the raw tree is no
+  // longer read directly at runtime.
+  console.log('🏗️ Building scaffold-data-app template zip...');
+  buildTemplateZip();
```
ycath: **Breaks the Windows release build.** `buildTemplateZip()` is now called unconditionally on every `npm run build`, and it shells out to the `zip` CLI via `execFileSync` (`src/scripts/buildTemplateZip.ts`). The `build-windows` job in `.github/workflows/upload-binaries.yml` runs `npm run build` on `windows-latest`, which has no `zip` on PATH by default → `ENOENT` → the Windows SEA binary can no longer be produced. All other CI jobs are ubuntu-only, so PR CI doesn't catch this. Consider a Node-native zip (e.g. archiver/adm-zip) or a `Compress-Archive` fallback on Windows.

package.json
```diff
@@ -54,6 +54,7 @@
     "build:inspect": "run-s build inspect",
     "build:inspect:docker": "run-s build:docker inspect:docker",
     "build:inspect:docker:http": "run-s build:docker inspect:docker:http",
+    "pretest": "tsx src/scripts/buildTemplateZip.ts",
```
ycath: This makes `npm test` depend on the external `zip` binary being installed and on PATH. A contributor or a minimal CI/Docker image without `zip` gets an immediate `ENOENT` crash before any test runs, with no fallback or guidance that `zip` is now a required local toolchain dependency. Same root cause as the `build.ts` comment — a Node-native zip would remove the dependency entirely.

src/tools/web/scaffoldDataApp/dataAppWorkspaceStore.ts
```diff
@@ -0,0 +1,134 @@
...
+  config: Config;
+  zipPath: string;
+}): Promise<Result<DataAppWorkspaceResult, McpToolError>> {
+  // Upload the template zip to S3 fresh on every call, then presign a GET URL for exactly those
+  // bytes. The uploaded content is identical regardless of `datappName` (only `postUnzip` varies),
+  // so a fixed key is intentional — concurrent overwrites are harmless. Nothing is published out of
+  // band; the object is always what this code path just wrote.
+  let s3URL: string;
+  try {
+    const buffer = await readFile(zipPath);
```
ycath: The template zip is read from disk and re-uploaded to S3 (PutObject) plus re-presigned on **every** scaffold-data-app call, even though the bytes never change across calls. Under concurrent use this multiplies S3 PUT cost/latency and API volume for identical, static content. Consider uploading once (or checking existence) and presigning the existing object — note `presignGetObjectUrl` in `s3Client.ts` was already extracted and could serve exactly this.

src/tools/web/scaffoldDataApp/templateIdentity.ts
```diff
@@ -0,0 +1,134 @@
...
+export function deriveIdentity(datappName: string): DataAppIdentity {
+  return {
+    packageId: `com.tableau.mcp.${slug(datappName)}`,
+    displayName: datappName,
+  };
+}
+
+/** Escapes a value for insertion into XML element text (.twb, .trex). */
+function escapeXmlText(value: string): string {
+  return value.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
```
ycath: `escapeXmlText()` escapes only `&`, `<`, `>` — not quotes. That's safe **today** only because (a) the `datappName` Zod regex forbids quote characters and (b) placeholders sit only in XML element-text nodes, never in attribute values. Neither invariant is enforced or asserted where the other lives, so loosening the regex (e.g. to allow more punctuation in display names) or moving a placeholder into an attribute would silently reintroduce an XML-attribute-breakout bug. Worth either escaping quotes too, or adding a comment/assertion coupling the escape set to these assumptions.

src/scripts/build.ts
jarhun88: Fixed in a645d638 — replaced the `zip` CLI call with the `archiver` npm package in `buildTemplateZip.ts`, so it no longer depends on an external binary that's missing on `windows-latest`. Verified the generated zip matches the old `zip -rXq` output byte-for-byte (same entries, sizes, root `Data App Name/` dir entry).

package.json
jarhun88: Removed in 25083e6c — the `pretest` hook was actually redundant: the two test files that need the template zip already build it themselves in their own `beforeAll` hooks (`scaffoldDataApp.test.ts`, `dataAppWorkspaceStore.test.ts`), so this global hook was forcing every `npm test` run to build the zip even for unrelated suites. Removing it also means `npm test` no longer depends on this build step at all — only the specific suites that need it do.

src/tools/web/scaffoldDataApp/dataAppWorkspaceStore.ts
jarhun88: Thought about this — decided to keep it as-is rather than skip the upload when the object already exists. The key here is fixed (no version/hash in it), so an exists-check would treat a stale object from a previous deploy — or a different pod hitting the same shared bucket — as a hit and just serve it forever instead of refreshing. Always writing first is what guarantees the object matches what's actually on disk right now. Added a comment (3e19e1e0) explaining this so it doesn't get "optimized" back into that later.

src/tools/web/scaffoldDataApp/templateIdentity.ts
jarhun88: Good catch — fixed in 3eb11d48, escaping quotes too. Doesn't change anything today since `datappName` can't contain quotes, but it means the function is safe wherever the result ends up, instead of relying on placeholders staying out of attribute values.

PR #931, opened 2026-09-21
refactor(desktop): serve knowledge only from external roots

WillMartin:
## Description

Desktop knowledge is now served solely from external roots named by TABLEAU_KNOWLEDGE_DIR. This is the codebase's standing rule: the desktop build no longer bundles a knowledge corpus, and when no root is configured the corpus is empty rather than an error, so the server degrades gracefully instead of throwing at startup or on the list/search/read tools.

- Delete the 106 bundled files under resources/desktop/knowledge (105 .md + .gitkeep). The corpus is owned and shipped externally (tab-agent-south).
- assets.ts: listKnowledgeSlugs/readKnowledgeBySlug/getConfiguredKnowledgeDir resolve only from external roots; empty roots yield [] / null / ''.
- knowledge/index.ts: drop the two knowledgeCorpusEmptyError throws; an empty corpus lists/searches as empty.
- server.desktop.ts: the empty-corpus startup log names TABLEAU_KNOWLEDGE_DIR when no root is set; the knowledge resource description no longer points at resources/desktop/knowledge.
- Drop the SEA search-knowledge smoke (workflow + buildSea.test.ts): it asserted a bundled 100-module corpus that no longer ships.
- Delete corpus-content tests (corpusIntegrity, resources, search, sectionRead, knowledgeToolReferences); that coverage moves with the corpus to tab-agent-south. index.test.ts is rewired onto a temp external root.

## Motivation and Context

We confusingly had multiple versions of knowledge floating around - some in the MCP, some packaged elsewhere. We're standardizing on the externally vended one.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe): Cleanup - this should largely be defunct knowledge

## Version Bump

- [X] Patch
- [ ] Minor
- [ ] Major

## How Has This Been Tested?

Unittests, locally testing with TAS dev mode and validating it can find relevant knowledge


artbeckTableau (commented):
Review of the knowledge-externalization change. The core logic holds up -- an empty corpus resolving to `[]` / `null` / `''` instead of throwing is correct, and I found no functional bugs.

**Only the first two items below matter for merge. The other three are FYI -- judgment calls or cleanup, not blockers.**

---

### Please fix before merge

**1. Encoding-convention violations (non-ASCII in source)**

Three new lines introduce characters outside our source-encoding rule (Windows-1252 / ASCII); our Windows code-page preflight is sensitive to this:

- `src/desktop/knowledge/index.ts:36` -- right-arrow `U+2192`, which is **not** Windows-1252-representable and can hard-fail preflight.
- `src/desktop/knowledge/index.test.ts:87` -- same `U+2192` right-arrow.
- `src/desktop/assets.ts:256` -- em dash `U+2014` (cp1252-OK but not ASCII).

Swap for `->` and `--`.

**2. Stale references to deleted paths**

The bulk deletion left two comments pointing at files that no longer exist in the repo:

- `src/desktop/derivations.ts:21` -- the AUTHORITY comment for the canonical derivation table still cites `resources/desktop/knowledge/tactics/tree/column-instance-prefixes.md`, now gone with no in-repo pointer to tab-agent-south. This is the one I care most about: wrong derivation strings silently corrupt fields to `None` on load, so the evidentiary basis for that table should remain followable. Please repoint it at the new external source.
- `src/desktop/assets.ts:7` -- the header comment's example SEA asset key uses a deleted knowledge path, which now misleads readers about what the SEA blob actually embeds.

---

### FYI -- not blockers

**3. Lost test coverage (worth a sanity check).** Deleting `sectionRead.test.ts` and `search.test.ts` removed the only coverage for URI-fragment section reads (`extractSection` heading-boundary math, malformed-`%` `decodeURIComponent` fallback) and for search fallbacks (whole-string fallback on empty tokenization, zero-hit `nearestKeywordMatches` broadening). The rewired `index.test.ts` only exercises whole-file reads. The PR says this coverage moves to tab-agent-south -- fine if that repo re-tests these pure functions, but `extractSection` / `searchKnowledgeByWholeString` still live here. Can you confirm the coverage actually moved?

**4. Dead code in `src/scripts/seaSmoke.ts` (~line 65).** The PR removed the CI callers of `--min-knowledge-resources` / `--search-knowledge` but left the flag parsing, smoke logic, and its still-passing test intact and now unreachable. Consider deleting the now-dead path, or note why it's kept.

**5. Silent empty corpus (`src/desktop/knowledge/index.ts:33`).** A misconfigured or unreachable `TABLEAU_KNOWLEDGE_DIR` yields normal-looking empty successes from `search-knowledge` / `list-knowledge-resources`; the only signal is a one-shot startup warning log. This looks intentional per the graceful-degradation goal -- raising only as a design question: is a startup-only warning enough observability when an agent's mandated "FIRST search-knowledge" route silently returns nothing?

WillMartin:
@artbeckTableau fixed the must haves. The "silent corpus" we warn locally in the logs but I'm not particularly worried on it? Probably something to talk about more after merge

PR #933, opened 2026-09-22
fix(desktop): author-action parameter mode — enumerate on miss, resolve caption, reject phantoms

sarahkim-tableau:
## Decision

parameter mode now validates every routing input against the live workbook and, on a miss, enumerates what the workbook actually offers — the same recovery language set mode already speaks ("Available sets: ..."). When that enumeration is **empty**, the message points at the tool that creates the target (`author-parameter` / `author-set`) rather than the dead-end "none" the traces got stuck on. This is the rule the tool now keeps for parameter mode, not a one-off: a first miss becomes a fixable second call instead of the blind retries this tool used to provoke (~426 sampled tool-call errors, W-24198191).

## Why

The tool failed with bare `"X is required"` errors that named no available values, so the agent retried blindly. Worse, several bad inputs were *accepted* and applied as actions that can never fire, then reported as success.

## What changed

- **Descriptions.** Filled the empty `mode` / param `.describe()` text and the tool description so the caller can pick a mode and its required field before the first call.
- **Enumerate on miss.** A missing `sourceField` / `targetParameter` now lists the fields / parameters the workbook has.
- **Empty enumeration → author-first.** When the workbook has no parameters (or the datasource no sets), the recovery message names `author-parameter` / `author-set` and says to retry — the exact stuck case the mined traces documented ("'Available sets: none' meaning the set didn't exist yet and needed to be created first").
- **Reject a blank `sourceField`** (not just an omitted one): it rendered a no-op action with a target but no value to push, which still passed readback.
- **Reject phantom targets.** A `targetParameter` naming no existing parameter, and a `sourceWorksheet` naming no existing worksheet, are rejected — both otherwise persisted as actions that can never fire.
- **Resolve caption → internal token.** A `targetParameter` given by its display caption (e.g. `[Parameters].[Churn Rate]`) resolves to the parameter's internal `[Parameter N]` token, and *that* is what gets emitted — matching `resolveTargetSet`. A parameter's caption is independent of its internal name, so a caption input was both falsely rejected and, if accepted, would have serialized an unresolvable `[Parameters].[Caption]`. The receipt now reports the internal token that landed.
- Adds `findWorkbookParameters` (name + caption) to `authorCalcCore` for the enumeration and resolution.

## Deliberately out of scope

`sourceField` existence is **not** checked here. No enumeration spans its legitimate forms (special pseudo-fields like `[:Measure Names]`, qualified, column-instance), and existence in a datasource doesn't prove the field is on the source worksheet's marks. A correct check needs to enumerate the source worksheet's shelf fields — a larger change with its own fixtures. Filed as a follow-up.

## Reviewer note

The set-mode `sourceWorksheet` check now runs **before** the unknown-set check, so a set-mode call with both a bad worksheet and a bad set now reports the worksheet first.

## Testing

- `authorAction.test.ts`: 69 pass, including caption-resolution, phantom-rejection, and empty-enumeration (author-first) tests for both parameter and set modes.
- Both prose invariant gates (`emittedToolReferences`, `paramOriginDescriptions`) pass — `author-parameter` / `author-set` are registered tool names, so naming them in recovery prose is allowed (existing precedent in `bindTemplate.ts`).
- ESLint + `tsc` clean; desktop bundle rebuilds.
- Validated live against Tableau Desktop via MCP Inspector: `[Parameters].[Churn Rate]` (a caption) resolves and applies as `[Parameters].[Parameter 2]`.


src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -34,13 +35,27 @@ const urlTargetSchema = z.enum(['default-zone-or-browser', 'browser', 'specific-
 // intact. This is the interactivity layer over the key signature.
 const paramsSchema = {
   session: sessionParam(),
-  mode: modeSchema.default('parameter').describe(''),
+  mode: modeSchema
+    .default('parameter')
+    .describe(
+      'Target driven, and the field it needs: parameter (sourceField+targetParameter), ' +
+        'set (targetSet), url (url), filter (targetSheet).',
```
diana-wang, line 42: Did we want to list out all of the required fields here for each mode? There's other fields that are required for all the modes. E.g. Filter actions require a sourceSheet or sourceDashboard, a targetSheet, an activation, and optionally can have filterFields, excludeSourceSheets, and excludeTargetSheets.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -311,8 +311,73 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
             ).toErr();
           }
 
+          // parameter mode needs a source field and an existing target parameter. Both errors
+          // enumerate what the workbook actually offers — the same recovery guidance set mode
+          // gives with "Available sets" — so a first miss becomes a fixable second call rather
+          // than the repeated blind retries this tool used to provoke.
```
diana-wang: nit: simplify all these comments in the PR? e.g. comments like "rather
Comment view than the repeated blind retries this tool used to provoke." probably aren't needed

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: The describe is capped hard by the tools/list byte budget (author-action is pinned at the max), so I kept it to the one distinguishing field per mode. The cross-mode requirements (caption, sourceWorksheet, source-or-dashboard) are enforced with runtime errors that name what's missing + enumerate the valid options, so a bad call gets a precise correction on the first miss rather than a blind retry.
FYI activation isn't required — it defaults to on-select.

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: Done in b733c421 — trimmed the trace-backstory and editorializing tails from the comments this PR added (that one included), kept the substantive why. Comments only, no schema/behavior change.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -311,8 +311,73 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
+          // parameter and set modes drive off a single required source worksheet. A name that
+          // isn't a real worksheet persists as a source the action can never fire from, so reject
+          // it and enumerate the worksheets — the same recovery filter mode already gives.
+          if (mode === 'parameter' || mode === 'set') {
+            if (!worksheetNames.has(effectiveSourceSheet)) {
+              return new ArgsValidationError(
+                `sourceWorksheet "${effectiveSourceSheet}" was not found. Available worksheets: ${worksheetNames.size > 0 ? [...worksheetNames].join(', ') : 'none'}`,
+              ).toErr();
+            }
+          }
```
diana-wang, line 373: similar to a filter actions, a parameter action's source can be a worksheet or a dashboard or a worksheet  (or worksheets) on a dashboard. Will need to verify and cover both (might be able to reuse the code from filters?)

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -311,8 +311,73 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
             ).toErr();
           }
 
+          // parameter mode needs a source field and an existing target parameter. Both errors
+          // enumerate what the workbook actually offers — the same recovery guidance set mode
+          // gives with "Available sets" — so a first miss becomes a fixable second call rather
+          // than the repeated blind retries this tool used to provoke.
+          let resolvedTargetParameter = '';
+          if (mode === 'parameter') {
```
diana-wang, line 317: Doesn't necessarily need to be done as part of this PR, but FYI if we want to cover everything that can go into a parameter action, some other components include an activation (hover/selection/menu), an aggregation for the source field (defaults to None) and behavior on clearing selection

Image

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: You're right about the model — Tableau does allow a dashboard or scoped-worksheet source here. The XSD shares one source group across all action families, and parameter-actions.md has a working `<source dashboard=… worksheet=…>`.

But the param/set renderers have only ever emitted worksheet-only sources, and that predates this PR. This guard doesn't narrow anything — it validates the existing worksheet-only contract, and turns a previously silent breakage (a dashboard name in the worksheet slot serialized a source Tableau errors on when you open the action) into an upfront error.

Real dashboard-source support is more than reusing the filter guards: both renderers have to switch to `renderSourceElement`, and the param/set readback would need to inspect `<source>` or it'd report a dropped source as applied. I'd say this should be a separate follow-up with its own fixtures and live validation, not part of this bugfix.

src/tools/desktop/authoring/datasource/authorAction.ts
```diff
@@ -311,8 +311,66 @@ export const getAuthorActionTool = (server: DesktopMcpServer): DesktopTool<typeo
...
+            // Reject empty/whitespace as well as undefined: renderParameterAction omits the
+            // source-field param when it is blank and readback only checks the target survived,
+            // so a blank sourceField would apply a no-op action and report success.
+            if (sourceField === undefined || sourceField.trim().length === 0) {
+              return new ArgsValidationError(
+                `sourceField is required in parameter mode. Available fields: ${formatAvailableFields(liveXml)}`,
+              ).toErr();
+            }
```
diana-wang, line 325: sourceField is never checked against the real fields - want to validate that it exists in the list of available fields, similar to parameters below.

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: activation's already wired for parameter mode. agg-type's hardcoded to attr and clear-option to do-nothing though, neither's exposed. agree those are a follow-up.

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: yeah this is a whole another issue I wanted to bring up to the team. a plain existence check against the field list false-rejects stuff we actually recommend — [:Measure Names], qualified and column-instance forms aren't in summarizeSchema, so it'd reject valid input and break the happy-path test. Even a clean datasource check doesn't prove the field's on the source sheet's marks, which is the thing that makes it fire. Doing it right means reading the sheet's shelf fields and we've got no helper for that. I'd say it's another separate work.

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Actually it seems the default value changes depending on the parameter type. E.g. numeric values default to sum.

Followup sounds good

src/tools/desktop/authoring/datasource/authorAction.ts
diana-wang: Measure Names/Values in general is kinda complicated so that might need it's own separate effort tbh.

src/tools/desktop/authoring/datasource/authorAction.ts
sarahkim-tableau: followup work created: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002lMIbDYAW/view

diana-wang (approved):

